PR #69, opened 2025-07-26
feat: array_type rule from @typescript-eslint/array-type

zoolsher:
credit: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/rules/array-type.ts


hardfist:
closed since it's already implemented

PR #189, opened 2025-08-07
fix: try fix windows extension

hardfist:
* rename rslint -> rslint.exe in windows
* remove unnecessary chown since already did in move artifacts
rslint language server starts successfully now but diagnostics still not working which will be fixed in following pr


fi3ework (commented):
do we need to add `.exe` for win32 here? https://github.com/web-infra-dev/rslint/blob/yj/try-fix-windows/packages/vscode-extension/src/Rslint.ts#L243

PR #190, opened 2025-08-07
chore: add json-schema

hardfist:
add json-schema for better auto completion in rslint.json
we should auto generated json schema in the future


fi3ework:
and also add IDE configuration in documentation (when we have website, i'll add it later), since we can't add `$schema` field  into the json.

fi3ework (dismissed):

PR #201, opened 2025-08-07
Add TypeScript ESLint rules: dot-notation, explicit-member-accessibility, max-params, member-ordering, no-unnecessary-type-assertion

ScriptedAlchemy:
This PR adds support for five new TypeScript ESLint rules:\n\n- **dot-notation**: Enforce dot notation whenever possible\n- **explicit-member-accessibility**: Require explicit accessibility modifiers on class properties and methods\n- **max-params**: Enforce a maximum number of parameters in function definitions\n- **member-ordering**: Require a consistent member declaration order\n- **no-unnecessary-type-assertion**: Disallow unnecessary type assertions\n\nChanges include:\n- Rule implementations in Go\n- Test configurations and snapshots\n- Updated rstest.config.mts to include new test files\n- API updates to support the new rules


packages/rule-tester/src/index.ts
```diff
@@ -23,46 +23,9 @@ function checkDiagnosticEqual(
   rslintDiagnostic: Diagnostic[],
   tsDiagnostic: TsDiagnostic[],
 ) {
```
hardfist, line 25: I'm afraid that we cant' use snapshot as source of truth now, as we have no way to tell whether the snapshot is right or wrong

PR #250, opened 2025-08-13
feat(cmd): add the init command for rslint

holyfata:
# Summary

Add a new command for rslint to create a default config in current directory.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -1,7 +1,9 @@
 package config
 
 import (
+	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/bmatcuk/doublestar/v4"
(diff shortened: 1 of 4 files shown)


internal/config/config.go
```diff
@@ -129,6 +131,31 @@ type RuleConfig struct {
 	Options map[string]interface{} `json:"options,omitempty"` // Rule-specific options
 }
 
+const defaultJsonc = `
+[
+  {
+    // ignore files and folders for linting
+    "ignores": ["./files-not-want-lint.ts", "./tests/**/fixtures/**.ts"],
```
hardfist, line 138: we can remove the ignore field for initialized project, user can configure it later

holyfata:
@hardfist review please ~

hardfist:
thanks

Issue #280, opened 2025-08-15
[Bug]: An unexpected space after `--fix`


ulivz:
### System Info

  System:
    OS: macOS 15.4.1
    CPU: (10) arm64 Apple M1 Max
    Memory: 186.78 MB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.15.0 - ~/.nvm/versions/node/v22.15.0/bin/node
    npm: 11.4.2 - ~/.nvm/versions/node/v22.15.0/bin/npm
    pnpm: 9.15.9 - ~/Library/pnpm/pnpm
  Browsers:
    Chrome: 139.0.7258.68
    Safari: 18.4

### Details

When I integrate rslint at https://github.com/bytedance/UI-TARS-desktop/pull/1144 ,⚡ It's fast! but encountered an unexpected leftover whitespace that should have been removed.

<img width="724" height="132" alt="Image" src="https://github.com/user-attachments/assets/d5ae15d1-ca3c-4d0d-9271-b809015bf97e" />

#### Source Code:
```ts
  if (!isZodSchema(schema)) {
    return schema as JSONSchema;
  }
```

#### Formatted Result

```ts
  if (!isZodSchema(schema)) {
    return schema  ;
  }
```

#### Expected Result

```ts
  if (!isZodSchema(schema)) {
    return schema;
  }
```

### Reproduce Steps

```bash
git clone https://github.com/bytedance/UI-TARS-desktop
git checkout 3e487dd78f8e37d15fd24f735efe192666825d29
cd multimodal

pnpm install --prefer-offline && pnpm run build
cd tarko/agent
pnpm lint:fix
```

hardfist:
you remind me that I miss the autofix tests in origin test https://github.com/web-infra-dev/rslint/blob/main/packages/rslint-test-tools/tests/typescript-eslint/rules/no-unnecessary-type-assertion.test.ts#L1397

PR #281, opened 2025-08-15
fix: remove unexpected space in output format string

ulivz:
## Summary

Improves the autofix for no-unnecessary-type-assertion rule to properly remove whitespace before 'as' keyword, preventing extra spaces in output.

Close: #280 

- Before: `foo() as number -> foo()  ;`
- After:  `foo() as number -> foo();`


internal/plugins/typescript/rules/no_unnecessary_type_assertion/no_unnecessary_type_assertion.go
```diff
@@ -219,16 +219,24 @@ var NoUnnecessaryTypeAssertionRule = rule.CreateRule(rule.Rule{
 			if node.Kind == ast.KindAsExpression {
 				s := scanner.GetScannerForSourceFile(ctx.SourceFile, expression.End())
 				asKeywordRange := s.TokenRange()
+				
+				sourceText := ctx.SourceFile.Text()
+				startPos := asKeywordRange.Pos()
+				
+				if startPos > expression.End() && sourceText[startPos-1] == ' ' {
```
hardfist, line 226: this seems can't handle following case well
```
const foo = 3               as 3; // multi space
const foo = 3 /* some comment here*/ as 3 // contains comment
```
the origin typescript-eslint implementation use a util to handle these cases https://github.com/typescript-eslint/typescript-eslint/blob/b2ee794265c4c727009e65a4eb5f06fad9686cf8/packages/eslint-plugin/src/rules/no-unnecessary-type-assertion.ts#L304

PR #304, opened 2025-08-26
feat: support rslint-wasm

hardfist:
support run rslint on browser


packages/rslint-wasm/README.md
```diff
@@ -0,0 +1,3 @@
+## eslint-wasm
```
CPunisher, line 1:
```suggestion
## rslint-wasm
```

PR #336, opened 2025-09-04
chore: show lint annotations in ci

hardfist:
## Checklist

- [ ] Tests updated (or not required).


hardfist:
> Ah, it should have been ::warning not ::warn. I'll fix it.

@Thiry1  it works now, thanks!

PR #409, opened 2025-12-01
fix: Support files configuration option

### LINKED ISSUE #408: [Bug]: The files configuration is not take effect
### System Info

  System:
    OS: macOS 15.6.1
    CPU: (14) arm64 Apple M3 Max
    Memory: 391.97 MB / 36.00 GB
    Shell: 5.9 - /bin/zsh
  npmPackages:
    @rslint/core: ^0.1.13 => 0.1.13 

### Details

When I set the `files` configuration item in `rslint.jsonc`, it did not take effect.

### Reproduce link

_No response_

### Reproduce Steps

Set the `files` configuration item in `rslint.jsonc`.

lambda47:
## Summary
Add support for `files` configuration option

- optimize file matching logic
- support negation glob patterns in `files` and `ignores` configurations
## Related Links
Fixed #408 
## Checklist

### DIFF
--- a/website/docs/en/config/index.md
+++ b/website/docs/en/config/index.md
@@ -49,6 +49,25 @@ The configuration file contains an array of configuration entries. Each entry de
 
 ## Configuration Options
 
+### files
+
+- **Type:** `string[]`
+- **Default:** `[]`
+
+An array of glob patterns for files and directories to include during linting. If omitted, all files are included.
+
+```jsonc
+{
+  "files": ["./src/**", "./tests/**"],
+}
+```
+
+Patterns support:
+
+- **Glob patterns**: `*.js`, `**/*.ts`
+- **Directory patterns**: `src/**`, `tests/**`
+- **Negation**: `!important.ts` (when used with other patterns)
+
 ### ignores
 
 - **Type:** `string[]`


website/docs/en/config/index.md
```diff
@@ -49,6 +49,25 @@ The configuration file contains an array of configuration entries. Each entry de
...
+{
+  "files": ["./src/**", "./tests/**"],
+}
+```
+
+Patterns support:
+
+- **Glob patterns**: `*.js`, `**/*.ts`
+- **Directory patterns**: `src/**`, `tests/**`
+- **Negation**: `!important.ts` (when used with other patterns)
```
fansenze: The negation pattern (`!important.ts`) is documented here, but the underlying `doublestar.Match` doesn't support this syntax.
It looks like the logic and test cases are missing here.

internal/config/file.go
```diff
@@ -0,0 +1,143 @@
...
+// Returns (isNeg, actualPattern) where isNeg indicates if the pattern starts with "!".
+func parseNegationPattern(pattern string) (isNeg bool, actualPattern string) {
+	if strings.HasPrefix(pattern, negPatternPrefix) {
+		return true, pattern[1:]
+	}
+	return false, pattern
+}
+
+// classifyPatterns splits a list of glob patterns into its positive and negative patterns.
+func classifyPatterns(patterns []string) (positivePatterns []string, negativePatterns []string) {
```
fansenze: This separates patterns into two groups, which differs from how ignores handles negation (sequential **"last match wins"**). For example, `["src/**", "!**/*.test.ts",
  "**/important.test.ts"]` would exclude `important.test.ts` because all negative patterns are applied after all positive patterns.

fansenze:
Hi @lambda47, thanks again for your contribution and for bringing this issue to our attention!

Since there hasn't been further activity here, and we've since landed #588 which covers the `files` configuration support (along with broader config/file-matching improvements), we'll go ahead and close this PR.

Your original PR and the linked issue were really helpful in shaping our approach. Thanks for your time and effort — we hope to see you around again!

PR #480, opened 2026-02-20
@typescript-eslint/max-params

ScriptedAlchemy:
Summary: Port @typescript-eslint/max-params from ScriptedAlchemy/rslint#13.

Related Links:
- https://github.com/ScriptedAlchemy/rslint/pull/13

Checklist:
- [x] Tests updated
- [x] Documentation updated (rule md)


internal/plugins/typescript/rules/max_params/max_params.go
```diff
@@ -0,0 +1,147 @@
...
+		return false
+	}
+
+	return decl.Type != nil && decl.Type.Kind == ast.KindVoidKeyword
+}
+
+func buildExceedMessage(count int, maxCount int) rule.RuleMessage {
+	return rule.RuleMessage{
+		Id:          "exceed",
+		Description: fmt.Sprintf("Function has too many parameters (%d). Maximum allowed is %d.", count, maxCount),
```
fansenze: The original ESLint core `max-params` rule uses a dynamic name in the message — e.g. `"Function 'foo'..."`, `"Arrow function..."`, `"Method 'bar'..."` depending on the function type. Currently this is hardcoded as `"Function"` for all cases.

It might be worth generating the appropriate name based on the node kind to stay consistent with the upstream behavior. Something like:

- `KindFunctionDeclaration` → `Function 'name'`
- `KindArrowFunction` → `Arrow function`
- `KindMethodDeclaration` → `Method 'name'`
- `KindConstructor` → `Constructor`
- `KindGetAccessor` → `Getter 'name'`
- `KindSetAccessor` → `Setter 'name'`
- `KindFunctionType` → `Function type`

Not a blocker, but would be nice to align!

ScriptedAlchemy:
Closing because this PR is now empty after syncing with current main.

PR #491, opened 2026-03-03
@typescript-eslint/no-deprecated

ScriptedAlchemy:


internal/plugins/typescript/rules/no_deprecated/no_deprecated.go
```diff
@@ -0,0 +1,2315 @@
...
+
+func isIdentifierTaggedTemplateTag(node *ast.Node) bool {
+	if node == nil || node.Parent == nil || node.Parent.Kind != ast.KindTaggedTemplateExpression {
+		return false
+	}
+	taggedTemplate := node.Parent.AsTaggedTemplateExpression()
+	return taggedTemplate != nil && taggedTemplate.Tag == node
+}
+
+func isDeprecatedNodeAssertFailOverload(ctx rule.RuleContext, node *ast.Node, name string) (bool, string) {
```
fansenze: These two helpers (`isDeprecatedNodeAssertFailOverload` here, `isDeprecatedFsExistsImport` at L1880) hard-code the API name, package, and the full deprecation reason string including version numbers.

I ablated both (forced `return false`) and reran the suite: `import assert from 'node:assert'; assert.fail({}, {})` fails immediately — this case is supported *only* by the hard-coding (`fs.exists` is structurally identical). The same root cause appears elsewhere: the skipped `<div aria-grabbed />` test also yields no diagnostic because the deprecation lives in `lib.dom.d.ts`. So `@deprecated` from an external `.d.ts` (lib / node_modules) isn't detected, and instead two specific APIs are patched in by hand — everything else from node_modules goes undetected. The root cause is worth fixing rather than per-API patching.

internal/plugins/typescript/rules/no_deprecated/no_deprecated.go
```diff
@@ -0,0 +1,2315 @@
...
+		found = true
+		if reason == "" {
+			reason = deprecatedReasonFromDeclaration(node)
+		}
+		return reason != ""
+	})
+	return found, reason
+}
+
+func deprecatedVariableInfoByNameInSource(ctx rule.RuleContext, name string) (bool, string) {
```
fansenze, line 1670: This `*ByNameInSource` family walks the whole SourceFile and matches declarations by **name string**, detached from symbol resolution (used as a fallback in `checkIdentifier`, L2085).

This produces false positives when a local variable/parameter shadows a deprecated declaration of the same name — the reported identifier resolves to the shadowing symbol, which is *not* deprecated, yet gets flagged. Verified with two probes:
```ts
/** @deprecated */ const foo = 1;
function bar(foo: number) { return foo; }            // FALSE POSITIVE on `foo`
```
```ts
/** @deprecated */ const value = 1;
function bar() { const value = 2; return value; }    // FALSE POSITIVE on `value`
```
Detection should rely on the resolved symbol, not name matching.

packages/rslint-test-tools/tests/typescript-eslint/rules/no-deprecated.test.ts
```diff
@@ -557,8 +589,10 @@ exists('/foo');
     `,
   ],
   invalid: [
+    // Skipped: React JSX deprecation requires React types (tsgolint parity)
     {
-      code: `
+      skip: true,
```
fansenze: 4 invalid JSX cases are skipped (here + L2956, L2971, L3000) with the note "requires React types". I removed the skips and ran each:
- **AProps (here), `foo-bar:baz-bam` (L2971), `Tab.List` (L3000) all PASS** — the rule reports the deprecated prop with the correct messageId + line/column. Their `@deprecated` comes from **local declarations** (a plain `interface`), so "requires React types" doesn't apply; skipping them just drops working coverage.
- Only `const a = <div aria-grabbed />` (L2956) genuinely **fails** (no diagnostic) — the deprecation is in `lib.dom.d.ts` (external), same root cause as the hard-coded node_modules helpers.

Please un-skip the 3 that pass, and track the external-lib case (aria-grabbed) explicitly rather than skipping silently.

internal/plugins/typescript/rules/no_deprecated/no_deprecated.go
```diff
@@ -0,0 +1,2315 @@
...
+		reportRange := func(diagnosticRange core.TextRange, message rule.RuleMessage) {
+			key := strconv.Itoa(diagnosticRange.Pos()) + ":" + strconv.Itoa(diagnosticRange.End())
+			if reportedRanges[key] {
+				return
+			}
+			reportedRanges[key] = true
+			ctx.ReportRange(diagnosticRange, message)
+		}
+		if sourceFile != nil {
+			diagnostics := ctx.TypeChecker.GetSuggestionDiagnostics(context.Background(), sourceFile)
```
fansenze, line 2112: Two parallel detection paths — `GetSuggestionDiagnostics` (here) + the hand-written AST listeners (L2155+) — deduped by text range via `reportedRanges` (L1996). By inspection (not reproduced) this looks fragile: if the two paths compute slightly different ranges, dedup fails and you get duplicate reports. There are also text-based heuristics: `isInImportStatementRange` (L1364) classifies "inside an import" by parsing the source line as text, and `hasDeprecatedTagInSource` (L205) scans characters backward for `/**`. Upstream is purely symbol/type driven; these heuristics are a smell.

cmd/rslint/api.go
```diff
@@ -243,16 +243,20 @@ func (h *IPCHandler) HandleLint(req api.LintRequest) (*api.LintResponse, error)
 			sourceFiles[filePath] = sourceFile
 			sourceFilesLock.Unlock()
 
+			if len(req.RuleOptions) == 0 {
```
fansenze, line 246: Unrelated infra changes, with an empty PR description:
- This `len(req.RuleOptions) == 0` branch changes global behavior (runs config-enabled rules instead of nothing when no options are passed).
- `RuleCount` (L291) changes from "configured" to "executed".
- `RequiresTypeInfo` (L259): by code reading this appears to be a no-op on the IPC path (no `TypeInfoFiles` is passed and IPC never calls `FilterNonTypeAwareRules`) — if it's actually needed, add a test; if not, drop it.

Please split infra changes into a separate PR and document them.

packages/rslint/src/browser.ts
```diff
@@ -82,7 +82,7 @@ export class BrowserRslintService implements RslintServiceInterface {
         this.expectedSize = dataView.getUint32(0, true); // true for little-endian
 
         // Remove length bytes from buffer
-        this.chunks = [combined.slice(4)];
+        this.chunks = [combined.subarray(4)];
```
fansenze: Unrelated to the rule. On the Node side (`node.ts`) `Buffer.slice`→`subarray` is fine (`slice` is deprecated). But here `combined` is a `Uint8Array`: `.slice()` copies while `.subarray()` returns a shared view — different memory semantics on the IPC message boundary, with no test coverage. I haven't reproduced a concrete bug, but it's an aliasing risk worth isolating and justifying.

fansenze (commented):
Review of the `no-deprecated` port (P0/P1 findings). Several conclusions are backed by ablation/probe experiments, noted inline. Not blocking — happy to discuss.

internal/plugins/typescript/rules/no_deprecated/no_deprecated.go
ScriptedAlchemy: Addressed in `7167fbc9`: removed the hard-coded Node API helpers (`isDeprecatedNodeAssertFailOverload`, `isDeprecatedFsExistsImport`). The two affected import cases (`node:assert.fail` 2-arg overload and `fs.exists`) are now explicitly skipped with comments documenting the root cause: external `.d.ts` deprecations from Node types are not resolved by the current type-checker path.

internal/plugins/typescript/rules/no_deprecated/no_deprecated.go
ScriptedAlchemy: Addressed in `7167fbc9`: added `isLocalNonDeprecatedSymbol`, which resolves the identifier through the type checker before using the name-only fallback. This suppresses the false positives where a local non-deprecated variable/parameter shadows a deprecated declaration of the same name. Applied to the `*ByNameInSource` fallback call sites.

packages/rslint-test-tools/tests/typescript-eslint/rules/no-deprecated.test.ts
ScriptedAlchemy: Addressed in `7167fbc9`: un-skipped the 3 passing JSX cases (`AProps`, `foo-bar:baz-bam`, `Tab.List`). Only `aria-grabbed` remains skipped, with an updated comment explaining the external `lib.dom.d.ts` deprecation limitation.

internal/plugins/typescript/rules/no_deprecated/no_deprecated.go
ScriptedAlchemy: Addressed in `7167fbc9`: added an architectural comment before the `GetSuggestionDiagnostics` loop explaining the two detection paths (type-checker diagnostics + AST listeners), `reportedRanges` deduplication, and the TODO to consolidate once external `.d.ts` deprecations are resolved through the type-checker path.

cmd/rslint/api.go
ScriptedAlchemy: Addressed in `c86294c7`: the IPC infra changes are now included in this PR and PR #1041 is closed as superseded. The combined PR contains the config-driven rules fallback when IPC `RuleOptions` is empty, `RequiresTypeInfo` propagation, and the `RuleCount` semantic change for review in one place.

packages/rslint/src/browser.ts
ScriptedAlchemy: Addressed in `6d873b2a`: reverted `Uint8Array.subarray()` back to `.slice()` in `browser.ts` to preserve copy semantics at the IPC message boundary and avoid shared-view aliasing risk.

internal/plugins/typescript/rules/no_deprecated/no_deprecated.go
```diff
@@ -0,0 +1,2315 @@
...
+				if !ok || propertyName == "" {
+					return
+				}
+				objectType := utils.GetConstrainedTypeAtLocation(ctx.TypeChecker, elementAccess.Expression)
+				var propertySymbol *ast.Symbol
+				if objectType != nil {
+					propertySymbol = checker.Checker_getPropertyOfType(ctx.TypeChecker, objectType, propertyName)
+				}
+				isDeprecated := symbolIsDeprecated(ctx.TypeChecker, propertySymbol)
+				sourceDeprecated, sourceReason := deprecatedStructuralPropertyInfoByNameInSource(ctx, propertyName)
```
fansenze: `isLocalNonDeprecatedSymbol` only guards `checkIdentifier`. Element access (here) and JSX attributes (L2213) still fire on name match alone (`!isDeprecated && sourceDeprecated`), so any same-named deprecated property elsewhere in the file false-positives.

vs typescript-eslint 8.59.0 (upstream → rslint):

```ts
// 1) cross-object collision
const a  = { /** @deprecated */ b: 'string' };
const a2 = { b: 'string' };
a2['b'];                          // 0 → flags `b`

// 2) any-typed object
interface U1 { /** @deprecated */ shared: number }
const obj: any = {};
obj['shared'];                    // 0 → flags `shared`

// 3) JSX attribute
interface U2 { /** @deprecated */ foo: number }
declare function Comp(p: any): any;
const x = <Comp foo={1} />;       // 0 → flags `foo`
```

Dot access (L2272) is symbol-gated and clean; these two need the same gate. The name path is redundant for real cases — `f['shared']` with a real deprecated `shared` is caught by the symbol path.

internal/plugins/typescript/rules/no_deprecated/no_deprecated.go
```diff
@@ -0,0 +1,2315 @@
...
+					continue
+				}
+				symbol := symbolAtLocation(ctx.TypeChecker, node)
+				if isWithinJsxClosingElement(node, diagnostic.Pos(), diagnostic.End()) {
+					continue
+				}
+				if isImportBindingAtRange(node, diagnostic.Pos(), diagnostic.End()) {
+					continue
+				}
+				if isInImportStatementRange(sourceFile, diagnostic.Pos()) {
```
fansenze, line 2134: `isInImportStatementRange` only checks the single line at `pos`, so a specifier on its own line in a multi-line import isn't suppressed and gets double-reported.

vs 8.59.0 (`dep.ts: /** @deprecated */ export const oldValue = 1;`):

```ts
import { oldValue } from './dep'; oldValue;   // 1 → 1 ✓
import {
  oldValue,
} from './dep'; oldValue;                     // 1 → 2 (extra report on the import specifier)
```

Prettier wraps imports by default, so this hits most multi-line deprecated imports. Needs an AST-based import check, not single-line text.

internal/plugins/typescript/rules/no_deprecated/no_deprecated.go
fansenze: Identifier-path guard is fine, but element access (L2250) and JSX attributes (L2213) still hit the name-only fallback. Reproduced remainder + upstream diff: https://github.com/web-infra-dev/rslint/pull/491#discussion_r3345773623 — resolving this in favor of that.

internal/plugins/typescript/rules/no_deprecated/no_deprecated.go
fansenze: The added comment documents the dual paths but doesn't fix them. The one reproducible bug — `isInImportStatementRange` double-reporting multi-line imports — is now tracked with an upstream diff: https://github.com/web-infra-dev/rslint/pull/491#discussion_r3345774251. The dedup-fragility and `hasDeprecatedTagInSource` concerns were never reproduced (smells only). Resolving in favor of the concrete comment.

internal/plugins/typescript/rules/no_deprecated/no_deprecated.go
```diff
@@ -0,0 +1,2315 @@
...
+		return false, ""
+	}
+	for _, decl := range symbol.Declarations {
+		if decl == nil {
+			continue
+		}
+		if decl.Kind == ast.KindBindingElement {
+			continue
+		}
+		if typeChecker.IsDeprecatedDeclaration(decl) || hasDeprecatedTag(decl) || hasDeprecatedTagInSource(decl) {
```
fansenze: Multi-declarator false positive: a statement-level `/** @deprecated */` on `const a = 1, b = 2` is attributed to every binding, so the 2nd+ declarator is flagged.

vs typescript-eslint 8.59.0:

```ts
/** @deprecated */ const a = 1, b = 2;
a;   // upstream 1 → rslint 1
b;   // upstream 0 → rslint flags `b`
```

`const /** @deprecated */ a = 1, b = 2` (per-declarator jsdoc) is correct. Root cause is below the rule: `IsDeprecatedDeclaration`/`hasDeprecatedTag` return true for `b`'s declaration (ablating `hasDeprecatedTagInSource` doesn't change it) — tsgo spreads the statement jsdoc to all declarators while tsc attaches it to the first only. An in-rule fix would gate reporting on the jsdoc actually belonging to the specific binding.

internal/plugins/typescript/rules/no_deprecated/no_deprecated.go
ScriptedAlchemy: Addressed in commit 4c7f2822: element access and JSX fallback are now symbol/context gated. The name-based source fallback only runs when the checker cannot resolve a property symbol and the contextual/object type is not any/unknown, which removes the cross-object and any-typed false positives while preserving real deprecated property reports.

internal/plugins/typescript/rules/no_deprecated/no_deprecated.go
ScriptedAlchemy: Addressed in commit 4c7f2822: multiline import suppression now uses AST/range matching instead of a single-line text check. isImportBindingAtRange recognizes specifier/clause nodes that contain the diagnostic span, so wrapped imports suppress the import-site diagnostic and only the usage site reports.

internal/plugins/typescript/rules/no_deprecated/no_deprecated.go
ScriptedAlchemy: Addressed in commit 4c7f2822: statement-level JSDoc ownership is now bound to the first declarator in a declaration list. '/** @deprecated */ const a = 1, b = 2;' reports 'a' but no longer leaks the statement tag onto 'b', and the regression is covered in both Go tests and the upstream parity fixture.

fansenze (approved):

Issue #476, opened 2026-02-12
[Tracking]: eslint-plugin-jest rules


fansenze:
This is a tracking issue for porting the existing rules from [`eslint-plugin-jest`](https://github.com/jest-community/eslint-plugin-jest?tab=readme-ov-file#rules).

There are **71** rules.

| Category | ✅ Done | 🚫 Won't | Total |
| :--- | :---: | :---: | :---: |
| Recommended | 5 | 0 | 19 |
| Not recommended | 4 | 0 | 52 |
| Deprecated | 0 | 0 | 0 |
| **Total** | **9** | **0** | **71** |

> ✅ = Implemented &nbsp;|&nbsp; 🚫 = Not intending to implement

---

## Recommended rules

<details open>
<summary>✅ 5 &nbsp;|&nbsp; 🚫 0 &nbsp;/&nbsp; Total: 19</summary>

| Status | Name |
| :----: | :--- |
| ✅ | [jest/no-alias-methods](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-alias-methods.md) |
|  | [jest/no-conditional-expect](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-conditional-expect.md) |
|  | [jest/no-deprecated-functions](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-deprecated-functions.md) |
|  | [jest/no-done-callback](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-done-callback.md) |
|  | [jest/no-export](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-export.md) |
| ✅ | [jest/no-focused-tests](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-focused-tests.md) |
|  | [jest/no-identical-title](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-identical-title.md) |
|  | [jest/no-interpolation-in-snapshots](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-interpolation-in-snapshots.md) |
|  | [jest/no-jasmine-globals](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-jasmine-globals.md) |
[...]

PR #593, opened 2026-04-03
feat(eslint-plugin-jest) add `prefer-to-have-length` rule

eryue0220:
## Summary

Port `prefer-to-have-length` from eslint-plugin-jest to rslint.

## Related Links

eslint-plugin-jest/prefer-to-have-length [doc](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/prefer-to-have-length.md) [code](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/prefer-to-have-length.ts)

## Checklist


internal/plugins/jest/rules/prefer_to_have_length/prefer_to_have_length.go
```diff
@@ -0,0 +1,169 @@
...
+					return
+				}
+
+				fileText := sourceFile.Text()
+				innerText := rslintUtils.TrimmedNodeText(sourceFile, inner)
+				expectedText := fileText[argList.Pos():argList.End()]
+				middleText := fileText[expectCall.End():beforeMatcher.End()]
+
+				newText := fmt.Sprintf("expect(%s)%s.toHaveLength(%s)", innerText, middleText, expectedText)
+				fixRange := core.NewTextRange(expectCall.Pos(), node.End())
```
fansenze: `expectCall.Pos()` includes leading trivia (whitespace/comments before the token) — this is how the TypeScript-Go AST works (`fullStartPos`). So for indented code like:

```js
    expect(files.length).toBe(1);
```

the fix range would start *before* the indentation, and the replacement text `expect(files).toHaveLength(1)` doesn't include it — the autofix silently eats the leading whitespace.

Current tests all start at column 1 so they don't catch this.

Other rules in this repo (e.g. `no_focused_tests`) handle this by trimming first:

```go
trimmed := rslintUtils.TrimNodeTextRange(sourceFile, expectCall)
fixRange := core.NewTextRange(trimmed.Pos(), node.End())
```

Might be worth adding a test case with some leading indentation to confirm the fix.

eryue0220: Updated.

fansenze (approved):

PR #614, opened 2026-04-08
feat(eslint-plugin-jest): add `prefer-todo` rule

eryue0220:
## Summary

Port `prefer-todo` from eslint-plugin-jest to rslint.

## Related Links

eslint-plugin-jest/prefer-todo [doc](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/prefer-todo.md) [code](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/prefer-todo.ts)

## Checklist


internal/plugins/jest/rules/prefer_todo/prefer_todo.go
```diff
@@ -0,0 +1,169 @@
...
+
+	prefix := src[callee.Pos():mem.Pos()]
+	var mid string
+	if mem.Kind == ast.KindIdentifier {
+		mid = "todo"
+	} else {
+		mid = "'todo'"
+	}
+	suffix := src[mem.End():callee.End()]
+	return prefix + mid + suffix
```
fansenze, line 108: `callee.Pos()` here includes leading trivia — same issue as #593.

Notice the `Members == 0` branch just above (line 98) already does this correctly:

```go
headTrim := rslintUtils.TrimNodeTextRange(ctx.SourceFile, head)
return src[headTrim.Pos():headTrim.End()] + ".todo"
```

But this `Members > 0` branch doesn't trim `callee`, so for indented code like:

```js
    test.skip("i need to write this test", () => {});
```

`prefix` would pick up the 4 leading spaces → the replacement text becomes `    test.todo(...)` → but the fix range is already trimmed (line 157), so the original indentation is preserved too → **double indentation** in the output.

Fix is straightforward — trim the callee the same way:

```go
calleeTrim := rslintUtils.TrimNodeTextRange(ctx.SourceFile, callee)
prefix := src[calleeTrim.Pos():mem.Pos()]
```

fansenze (dismissed):

fansenze (approved):

PR #623, opened 2026-04-09
feat(eslint-plugin-jest): add `prefer-strict-equal` rule

eryue0220:
## Summary
Port `prefer-strict-equal` from eslint-plugin-jest to rslint.

## Related Links

eslint-plugin-jest/prefer-strict-equal [doc](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/prefer-strict-equal.md) [code](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/prefer-strict-equal.ts)

## Checklist


internal/plugins/jest/rules/prefer_strict_equal/prefer_strict_equal.go
```diff
@@ -0,0 +1,65 @@
...
+						continue
+					}
+
+					ctx.ReportNodeWithSuggestions(
+						memberEntry.Node,
+						buildUseToStrictEqualErrorMessage(),
+						rule.RuleSuggestion{
+							Message: buildSuggestReplaceWithStrictEqualErrorMessage(),
+							FixesArr: []rule.RuleFix{
+								{Range: core.NewTextRange(memberEntry.Node.Pos(), memberEntry.Node.End()), Text: "toStrictEqual"},
```
fansenze: The replacement text here is always bare `toStrictEqual`, but that only works for dot-notation (`.toEqual` → `.toStrictEqual`). For bracket-notation like `["toEqual"]`, this produces `[toStrictEqual]` — which is a **variable reference**, not a string property access. The code would break at runtime.

The original eslint-plugin-jest handles this via `replaceAccessorFixer`, which wraps in quotes for non-identifiers:

```typescript
node.type === AST_NODE_TYPES.Identifier ? text : `'${text}'`
```

Same thing is needed here:

```go
text := "toStrictEqual"
if memberEntry.Node.Kind != ast.KindIdentifier {
    text = "'toStrictEqual'"
}
```

The Go test also has the wrong expectation — it expects `[toStrictEqual]` (no quotes), while the TS test correctly expects `['toStrictEqual']`. You can cross-check:

```
// Go test (wrong):
Output: `expect(something)[toStrictEqual](somethingElse);`

// TS test (correct):
output: "expect(something)['toStrictEqual'](somethingElse);"
```

eryue0220: updated.

internal/plugins/jest/rules/prefer_strict_equal/prefer_strict_equal.go
```diff
@@ -0,0 +1,73 @@
...
+				MemberEntries := jestFnCall.MemberEntries
+				if len(MemberEntries) == 0 {
+					return
+				}
+
+				for _, memberEntry := range MemberEntries {
+					kind := memberEntry.Node.Kind
+					if kind != ast.KindIdentifier && kind != ast.KindStringLiteral {
+						continue
+					}
```
fansenze: `getElementAccessName` in `jest.go` also supports `ast.KindNoSubstitutionTemplateLiteral`, which means `expect(x)[`toEqual`](y)` will produce a `MemberEntry` with Name: `"toEqual"` and `Node.Kind: NoSubstitutionTemplateLiteral`. This filter skips it silently, so the rule won't report on template-literal bracket access.

```go
if kind != ast.KindIdentifier && kind != ast.KindStringLiteral && kind != ast.
KindNoSubstitutionTemplateLiteral {
    continue
}
```

eryue0220: update.

fansenze (approved):

PR #678, opened 2026-04-17
fix(lsp): scope tsConfigPaths per-config to prevent type-aware rule leak (fixes #671)

### LINKED ISSUE #671: [Bug]: LSP reports @typescript-eslint/no-unused-vars, but CLI passes on the same file
### System Info

```text
System:
  OS: macOS 26.2
  CPU: (16) arm64 Apple M3 Max
  Memory: 5.94 GB / 64.00 GB
  Shell: 5.9 - /bin/zsh
Binaries:
  Node: 24.12.0
  npm: 11.6.2
  pnpm: 10.33.0
  bun: 1.3.4
  Deno: 2.5.4
Browsers:
  Chrome: 147.0.7727.56
  Safari: 26.2
npmPackages:
  @rslint/core: ^0.4.2 => 0.4.2
```

Extra editor info:
- VS Code: 1.115.0
- `rstack.rslint`: 0.4.2
- `typescript` in the project: 6.0.2

### Details

I found an inconsistency between the language server and the CLI for `@typescript-eslint/no-unused-vars`.

Reproduction file:
- https://github.com/rstackjs/create-rstack/blob/main/test/skills.test.ts

This callback in `test/skills.test.ts` triggers `rslint` diagnostics in VS Code:

```ts
const xSync = rs.fn((command: string, args: string[], options: unknown) => {
  return {
    stdout: '',
    stderr: '',
    exitCode: 0,
  };
}) as any;
```

In the editor, `rstack.rslint` reports:
- `[@typescript-eslint/no-unused-vars] 'command' is defined but never used.`
- and the same for `args` / `options`

But the CLI passes on the same file with the same config:

```bash
pnpm lint
# Found 0 errors and 0 warnings ...

pnpm exec rslint test/skills.test.ts
# Found 0 errors and 0 warnings ...
```

The config is very small:

```ts
import { defineConfig, ts } from '@rslint/core';

export default defineConfig([
  { ignores: ['**/dist/**'] },
  ts.configs.recommended,
]);
```

I am not sure whether the CLI or the language server is correct here, but I would expect them to be consistent.

This seems different from #555, because here the main problem is that the language server reports the rule while the CLI does not report anything for the same file.

### Reproduce link

https://github.com/rstackjs/create-rstack

### Reproduce Steps

1. Clone `https://github.com/rstackjs/create-rstack`
2. Run `pnpm install`
3. Open `test/skills.test.ts` in VS Code with the `rstack.rslint` extension enabled.
4. Look at the callback parameters around line 70. The language server reports `@typescript-eslint/no-unused-vars`.
5. Run `pnpm lint`.
6. Run `pnpm exec rslint test/skills.test.ts`.
7. Both CLI commands report `Found 0 errors and 0 warnings`.

fansenze:
## Summary

- **Root cause**: `rebuildTsConfigPaths` (`internal/lsp/service.go`) stored a single global `tsConfigPaths`. If any rslint config in the workspace couldn't resolve a tsconfig (no explicit `parserOptions.project` and no auto-detectable `tsconfig.json` under that config's directory), the code set `s.tsConfigPaths = nil` and flipped the type-aware-rule filter into conservative "allow-all" for **every** file in the workspace — including files governed by sibling configs that had perfectly resolvable tsconfigs.
- **Concrete trigger (issue #671)**: [create-rstack](https://github.com/rstackjs/create-rstack) ships a nested `template-rslint/` starter directory that carries its own `rslint.config.ts` but no `tsconfig.json`. Opening any .ts file in VSCode tripped the global fallback, so type-aware rules like `@typescript-eslint/no-unused-vars` ran on `test/skills.test.ts` (outside the root `tsconfig.json`'s `include: ["src"]`) and produced 3 diagnostics that the CLI correctly suppressed.
- **Fix**: track the resolved tsconfig list **per-config-directory** in `tsConfigPathsByConfig`. At lint time pick the nearest config's list via the new `tsConfigPathsForURI`. A config with no resolved tsconfig still disables filtering — but only for files under that config's own scope.
- **Reproduction matrix** (same create-rstack fixture, same probe flow):
  - pre-fix + two configs (root + `template-rslint/`) → **3** `no-unused-vars` diagnostics on `test/skills.test.ts`
  - post-fix + same two configs → **0** diagnostics
- Added regression coverage:
  - `internal/lsp/service_test.go`: updated existing `TestRebuildTsConfigPaths_*` / `TestHandleConfigUpdate_*` cases for per-config shape, plus a dedicated `TestTsConfigPathsForURI_NestedConfigWithoutTsconfigDoesNotLeak`.
  - `packages/vscode-extension/__tests__/fixtures-project-service-scope/template-nested/rslint.config.js`: nested config without tsconfig so the existing E2E asserts the multi-config path end-to-end.

## Related Links

- Fixes https://github.com/web-infra-dev/rslint/issues/671

## Checklist

### DIFF
--- a/packages/vscode-extension/__tests__/suite-project-service-scope/project-service-scope.test.ts
+++ b/packages/vscode-extension/__tests__/suite-project-service-scope/project-service-scope.test.ts
@@ -27,16 +27,15 @@ suite('rslint projectService type-aware scope', function () {
       const current = vscode.languages.getDiagnostics(doc.uri);
       if (predicate(current)) return current;
       await new Promise((resolve) => {
+        let timer: ReturnType<typeof setTimeout>;
         const disposable = vscode.languages.onDidChangeDiagnostics((e) => {
-          for (const uri of e.uris) {
-            if (uri.toString() === doc.uri.toString()) {
-              disposable.dispose();
-              resolve(void 0);
-              return;
-            }
+          if (e.uris.some((uri) => uri.toString() === doc.uri.toString())) {
+            clearTimeout(timer);
+            disposable.dispose();
+            resolve(void 0);
           }
         });
-        setTimeout(() => {
+        timer = setTimeout(() => {
           disposable.dispose();
           resolve(void 0);
         }, 1500);
--- a/packages/vscode-extension/__tests__/suite-project-service-scope/project-service-scope.test.ts
+++ b/packages/vscode-extension/__tests__/suite-project-service-scope/project-service-scope.test.ts
@@ -1,0 +1,93 @@
+import * as assert from 'assert';
+import * as vscode from 'vscode';
+import path from 'node:path';
+
+// Type-aware rule scope when parserOptions uses `projectService: true`
+// (the shape `ts.configs.recommended` exports) without an explicit
+// `project`. The LSP and CLI must agree: only files covered by the
+// fallback tsconfig's `include` get type-aware rules.
+//
+// Fixture: fixtures-project-service-scope
+//   - rslint.config.js  — parserOptions.projectService: true (no explicit project)
+//   - tsconfig.json     — include: ["src"]
+//   - src/covered.ts    — IN tsconfig: no-unused-vars SHOULD fire
+//   - test/skills.test.ts — NOT IN tsconfig: no-unused-vars should NOT fire
+suite('rslint projectService type-aware scope', function () {
+  this.timeout(60000);
+
+  function workspaceRoot(): string {
+    return vscode.workspace.workspaceFolders![0].uri.fsPath;
+  }
+
+  async function waitForDiagnostics(
+    doc: vscode.TextDocument,
+    predicate: (diags: vscode.Diagnostic[]) => boolean,
+  ): Promise<vscode.Diagnostic[]> {
+    for (let i = 0; i < 20; i++) {
+      const current = vscode.languages.getDiagnostics(doc.uri);
+      if (predicate(current)) return current;
+      await new Promise((resolve) => {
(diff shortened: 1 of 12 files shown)


packages/vscode-extension/__tests__/suite-project-service-scope/project-service-scope.test.ts
```diff
@@ -0,0 +1,129 @@
...
+  test('test/skills.test.ts (outside tsconfig include) — no-unused-vars should NOT fire', async () => {
+    const doc = await vscode.workspace.openTextDocument(
+      path.join(workspaceRoot(), 'test/skills.test.ts'),
+    );
+    await vscode.window.showTextDocument(doc);
+
+    // Give the server time to publish diagnostics. Waiting for a specific
+    // predicate is unreliable here because the correct outcome is "no rslint
+    // diagnostics" — we instead wait a bounded amount of time.
+    await new Promise((r) => setTimeout(r, 5000));
```
chenjiahan: This may slow down the test cases

chenjiahan: Can we use `poll` instead?
https://rstest.rs/api/runtime-api/test-api/expect#expectpoll

fansenze: `expect.poll` is incompatible with our vscode test runner (`@vscode/test-electron`). Instead, we replaced the hardcoded 5-second `setTimeout` with `waitForDiagnostics` using a "marker-rule" pattern.

chenjiahan (approved):

PR #776, opened 2026-04-22
feat(eslint-plugin-jest): add `no-mocks-import` rule

eryue0220:
## Summary

Port `no-mocks-import` from eslint-plugin-jest to rslint.

## Related Links

eslint-plugin-jest/no-mocks-import [doc](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-mocks-import.md) [code](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/no-mocks-import.ts)
## Checklist


internal/plugins/jest/rules/no_mocks_import/no_mocks_import.go
```diff
@@ -0,0 +1,50 @@
...
+
+				if len(arguments) == 0 {
+					return
+				}
+
+				firstArg := arguments[0]
+				if firstArg != nil && ast.IsStringLiteral(firstArg) && isMocksImportPath(firstArg.Text()) {
+					ctx.ReportNode(firstArg, buildNoManualImportErrorMessage())
+				}
+			},
```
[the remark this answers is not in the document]
eryue0220: Updated. Add `require` checked, `KindExportNamedDeclaration` and `KindExportAllDeclaration` are no need.

fansenze (commented):
Thanks for the port! The logic and tests look solid overall. Left two comments — one is a crash issue that needs fixing before merge.

internal/plugins/jest/rules/no_mocks_import/no_mocks_import.go
```diff
@@ -0,0 +1,54 @@
...
+	Name: "jest/no-mocks-import",
+	Run: func(ctx rule.RuleContext, options any) rule.RuleListeners {
+		return rule.RuleListeners{
+			ast.KindImportDeclaration: func(node *ast.Node) {
+				if isMocksImportPath(node.AsImportDeclaration().ModuleSpecifier.Text()) {
+					ctx.ReportNode(node, buildNoManualImportErrorMessage())
+				}
+			},
+			ast.KindCallExpression: func(node *ast.Node) {
+				callee := node.AsCallExpression().Expression.AsIdentifier()
```
fansenze: `AsIdentifier()` does a direct type assertion (`n.data.(*Identifier)`) internally — it panics instead of returning nil when the node is not an Identifier. So the `callee == nil` check on the next line never gets a chance to run.

Since the `KindCallExpression` listener fires for *every* call expression in the file, this will crash on things like `console.log()`, `obj.method()`, `require.resolve()`, etc. — basically any call where the callee is not a bare identifier.

You can see the safe pattern used elsewhere in this codebase, e.g. `no_disabled_tests.go:34`:

```go
if callExpr.Expression.Kind != ast.KindIdentifier {
    return
}
callee := node.AsCallExpression().Expression.AsIdentifier()
```

Adding a `Kind` check before calling `AsIdentifier()` should fix it.

internal/plugins/jest/rules/no_mocks_import/no_mocks_import.go
```diff
@@ -0,0 +1,54 @@
...
+				if callee == nil || callee.Text != "require" {
+					return
+				}
+
+				arguments := node.Arguments()
+				if len(arguments) == 0 {
+					return
+				}
+
+				firstArg := arguments[0]
```
fansenze, line 56: Nit: the original ESLint rule uses `isStringNode()` which also matches `NoSubstitutionTemplateLiteral` (template strings with no interpolation, like `` `./__mocks__/index` ``). Here only `IsStringLiteral` is checked, so a case like `` require(`./__mocks__/index`) `` would be missed.

Unlikely in practice, but worth handling for full alignment with the original rule.

eryue0220: Updated.

internal/plugins/jest/rules/no_mocks_import/no_mocks_import.go
eryue0220: Updated.

fansenze (approved):

PR #777, opened 2026-04-22
feat(eslint-plugin-jest): add `no-deprecated-functions` rule

eryue0220:
## Summary

Port `no-deprecated-functions` from eslint-plugin-jest to rslint.

## Related Links

eslint-plugin-jest/no-deprecated-functions [doc](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-deprecated-functions.md) [code](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/no-deprecated-functions.ts)

## Checklist


internal/plugins/jest/rules/no_deprecated_functions/no_deprecated_functions_test.go
```diff
@@ -0,0 +1,88 @@
...
+		[]rule_tester.ValidTestCase{
+			{Code: `jest`, Settings: jestSettings("14")},
+			{Code: `require("fs")`, Settings: jestSettings("14")},
+			{Code: `jest.resetModuleRegistry`, Settings: jestSettings("14")},
+			{Code: `require.requireActual`, Settings: jestSettings("17")},
+			{Code: `jest.genMockFromModule`, Settings: jestSettings("25")},
+			{Code: `jest.genMockFromModule`, Settings: jestSettings("25.1.1")},
+			{Code: `require.requireActual`, Settings: jestSettings("17.2")},
+		},
+		[]rule_tester.InvalidTestCase{
```
fansenze, line 34: The invalid cases only cover 3 out of 6 deprecations (`resetModuleRegistry`, `addMatchers`, `genMockFromModule`). Missing test cases for:

- `require.requireMock` → `jest.requireMock`
- `require.requireActual` → `jest.requireActual`
- `jest.runTimersToTime` → `jest.advanceTimersByTime`

The first two are especially worth testing since they involve a **cross-object replacement** (`require` → `jest`), which exercises a different code path in the fix logic. Would be good to add both dot-style and bracket-style cases for each, same as the other entries.

(Same gap exists in the e2e test file `no-deprecated-functions.test.ts`.)

internal/plugins/jest/utils/jest.go
```diff
@@ -209,3 +216,160 @@ func getElementAccessName(node *ast.Node) string {
...
+	}
+	raw, ok := settings["jest"]
+	if !ok {
+		return "", false
+	}
+	m, ok := raw.(map[string]interface{})
+	if !ok {
+		return "", false
+	}
+	ver, ok := m["version"].(string)
```
fansenze: This only handles `string` typed version values. If a user writes their config with a number (e.g. `{ jest: { version: 27 } }`), the type assertion silently fails and the function falls back to `false` — which eventually means the rule uses the default version 29 instead of the intended 27.

In practice JSON-decoded numbers come in as `float64`, so something like this would cover it:

```go
switch v := m["version"].(type) {
case string:
    ver = v
case float64:
    ver = strconv.Itoa(int(v))
}
```

Not a big deal since most configs use strings, but it would avoid a subtle silent mismatch.

fansenze (commented):
Thanks for the clean implementation! Left a couple of comments.

internal/plugins/jest/rules/no_deprecated_functions/no_deprecated_functions.go
```diff
@@ -0,0 +1,117 @@
...
+		if len(depMap) == 0 {
+			return rule.RuleListeners{}
+		}
+
+		return rule.RuleListeners{
+			ast.KindCallExpression: func(node *ast.Node) {
+				callee := node.AsCallExpression().Expression
+				if callee == nil {
+					return
+				}
```
fansenze, line 89: The `KindCallExpression` listener fires on every call node, and `GetJestFnMemberEntries` recurses into a `KindCallExpression` callee (jest.go:189-190). So `jest.resetModuleRegistry()()` gets reported twice — once for the inner call (callee `jest.resetModuleRegistry`) and once for the outer call (callee `jest.resetModuleRegistry()`, which the helper unwraps into the same chain).

I confirmed locally — rule_tester prints `reported 2 errors` with ranges `[0,24]` and `[0,26]`, and the outer fix swallows a pair of parens (`jest.resetModuleRegistry()()` → `jest.resetModules()()`, losing the call-then-call semantics).

The original ESLint rule guards with `node.callee.type === MemberExpression`, so the outer one is filtered out. A `Kind` check here matches that:

```go
if callee.Kind != ast.KindPropertyAccessExpression && callee.Kind != ast.KindElementAccessExpression {
    return
}
```

Edge case in practice, but worth closing since the fix can produce broken code.

eryue0220: Thanks, updated.

fansenze (approved):

PR #779, opened 2026-04-23
feat(eslint-plugin-jest): add `prefer-to-contain` rule

eryue0220:
## Summary

Port `prefer-to-contain` from eslint-plugin-jest to rslint.

## Related Links

eslint-plugin-jest/prefer-to-contain [doc](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/prefer-to-contain.md) [code](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/prefer-to-contain.ts)

## Checklist


fansenze (commented):
Nice work overall! One minor suggestion:

Lines 176-182 handle preserving `resolves`/`rejects` modifiers in the fix output, but there are no test cases covering this path. Would be great to add a couple of invalid cases like:

```go
{
    Code:   `expect(a.includes(b)).resolves.toBe(true);`,
    Output: []string{`expect(a).resolves.toContain(b);`},
    Errors: []rule_tester.InvalidTestCaseError{
        {MessageId: "useToContain", Line: 1, Column: 32},
    },
},
{
    Code:   `expect(a.includes(b)).resolves.not.toBe(true);`,
    Output: []string{`expect(a).resolves.not.toContain(b);`},
    Errors: []rule_tester.InvalidTestCaseError{
        {MessageId: "useToContain", Line: 1, Column: 36},
    },
},
```

Just to make sure that code path stays correct going forward.

internal/plugins/jest/rules/prefer_to_contain/prefer_to_contain.go
```diff
@@ -0,0 +1,215 @@
...
+				jestFnCall := jestUtils.ParseJestFnCall(node, ctx)
+				if jestFnCall == nil || jestFnCall.Kind != jestUtils.JestFnTypeExpect {
+					return
+				}
+
+				if len(jestFnCall.Members) == 0 || !slices.ContainsFunc(jestFnCall.Members, func(m string) bool {
+					return jestUtils.EQUALITY_METHOD_NAMES[m]
+				}) {
+					return
+				}
```
fansenze: `ContainsFunc` matches if an equality matcher appears *anywhere* in the chain. The original ESLint rule only checks the current call`s tail method, so chains like `expect(a.includes(b)).toBe(true).somethingElse(true)` get reported twice in rslint — once for the inner `.toBe(true)` (correct), once for the outer `.somethingElse(true)` (false positive, because the listener fires on every CallExpression and the outer `Members` still contains `toBe`).

I confirmed locally — rule_tester prints two errors at ranges 22-26 (`toBe`) and 33-46 (`somethingElse`). Worse, if the outer fix wins it rewrites `.toBe(true).somethingElse(true)` to `.toContain(b)`, silently dropping the trailing `.somethingElse(true)` call.

Edge case in practice (the chain isn`t valid jest usage anyway), but it`s a one-line tightening:

```go
last := jestFnCall.Members[len(jestFnCall.Members)-1]
if !jestUtils.EQUALITY_METHOD_NAMES[last] {
    return
}
```

(The empty-slice guard is implicit — accessing `len-1` only after we`ve already established `len(Members) > 0`. If you want to keep the explicit check, just `&&` it in.)

fansenze (approved):

PR #821, opened 2026-04-28
feat(eslint-plugin-jest): add `no-done-callback` rule

eryue0220:
## Summary

Port `no-done-callback` from eslint-plugin-jest to rslint.

## Related Links

eslint-plugin-jest/no-done-callback [doc](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-done-callback.md) [code](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/no-done-callback.ts)

## Checklist


internal/plugins/jest/rules/no_done_callback/no_done_callback.go
```diff
@@ -0,0 +1,178 @@
...
+	argLength := len(nodes)
+	if argLength == 0 {
+		return nil
+	}
+
+	if isJestEach || (jestFnCall.Kind == utils.JestFnTypeTest && argLength >= 2) {
+		return nodes[1]
+	} else if jestFnCall.Kind == utils.JestFnTypeHook && argLength >= 1 {
+		return nodes[0]
+	}
```
fansenze: This panics on `test.each``("only one arg")` — `isJestEach` is true but `argLength` is 1, so `nodes[1]` is out of range. I reproduced it locally:

```
panic: runtime error: index out of range [1] with length 1
    no_done_callback.go:51
```

The `each` chain isn't valid jest usage, but it's valid syntax — the linter shouldn't crash on it.

ESLint side gets away with `node.arguments[1]` returning `undefined` and the later `!callback` check catches it. Easiest fix here is to require `argLength >= 2` on the `isJestEach` branch too:

```go
if (isJestEach || jestFnCall.Kind == utils.JestFnTypeTest) && argLength >= 2 {
    return nodes[1]
}
```

internal/plugins/jest/rules/no_done_callback/no_done_callback.go
```diff
@@ -0,0 +1,178 @@
...
+		Description: "Return a Promise instead of relying on callback parameter",
+	}
+}
+
+func buildErrorSuggestWrappingInPromiseMessage() rule.RuleMessage {
+	return rule.RuleMessage{
+		Id:          "suggestWrappingInPromise",
+		Description: "Wrap in `new Promise({{ callback }} => ...`",
+	}
+}
```
fansenze: `{{ callback }}` is never replaced — `RuleMessage.Description` is a plain string, there's no placeholder rendering in rslint. The original ESLint rule passes `data: { callback: argument.name }` so users see e.g. `Wrap in 'new Promise(done => ...)'`. Here the IDE will show the literal `{{ callback }}`.

rule_tester only compares messageId, so tests pass — but the suggestion text is broken in the editor.

Fix: take the name as a parameter and `fmt.Sprintf` it.

```go
func buildErrorSuggestWrappingInPromiseMessage(callbackName string) rule.RuleMessage {
    return rule.RuleMessage{
        Id:          "suggestWrappingInPromise",
        Description: fmt.Sprintf("Wrap in `new Promise(%s => ...`", callbackName),
    }
}
```

And pass `callbackName` from the call site (line 171).

eryue0220: Updated.

internal/plugins/jest/rules/no_done_callback/no_done_callback.go
eryue0220: Updated.

fansenze (approved):

PR #864, opened 2026-05-07
feat: expose structured SymbolNameKind in tsgo wire format

fansenze:
## Summary

typescript-go uses a `\xFE` sentinel prefix on `Symbol.Name` to mark its
internal symbol names (`__index`, `__call`, `__type`, ...). The previous
wire format escaped the sentinel to ASCII `__` so consumers could parse
the prefix back, which had two problems:

- The "is this an internal symbol" semantic was encoded into the string
  contents — consumers had to do prefix scanning to recover it.
- A user-spelled identifier like `__foo` is indistinguishable from an
  internal name at the prefix level.

This PR replaces the prefix escape with a structured `name_kind` enum
field (`User` / `Internal` / `Private`). The sentinel is stripped from
`name`, and the new field is the single source of truth for symbol-name
origin.

- `cmd/tsgo`: drop `sanitizeSymbolName`; introduce `SymbolNameKind` and
  `classifySymbolName`; `SymbolInfo` gains a `name_kind` field; both
  call sites switched.
- `crates/tsgo-client`: `SymbolData` exposes `name_kind: u8` plus a
  `symbol_name_kind` constant module so consumers can match on it
  without depending on string prefixes.

Wire format change: `name` for internal/private symbols no longer carries
the `__` prefix; consumers should switch from name-prefix detection to
`name_kind`.

## Related Links

N/A

## Checklist

### DIFF
--- a/cmd/tsgo/semantic.go
+++ b/cmd/tsgo/semantic.go
@@ -10,13 +10,35 @@ import (
 	"github.com/microsoft/typescript-go/shim/compiler"
 )
 
-// sanitizeSymbolName replaces the internal symbol name prefix (\xFE) with "__"
-// so that consumers receive valid UTF-8 symbol names.
-func sanitizeSymbolName(name string) []byte {
-	if strings.HasPrefix(name, ast.InternalSymbolNamePrefix) {
-		return []byte("__" + name[len(ast.InternalSymbolNamePrefix):])
+// SymbolNameKind classifies a symbol's name origin so consumers can tell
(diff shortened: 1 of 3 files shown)


fansenze:
Closing — turns out the wire format already exposes this via `CheckFlagsIndexSymbol` (1 << 23) on `SymbolData.check_flags`. The consumer can detect index-signature-backed properties directly without a new field. No rslint change needed.

PR #870, opened 2026-05-08
feat(eslint-plugin-jest): add `prefer-to-be` rule

eryue0220:
## Summary

Port `prefer-to-be` from eslint-plugin-jest to rslint.

## Related Links

eslint-plugin-jest/prefer-to-be [doc](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/prefer-to-be.md) [code](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/prefer-to-be.ts)

## Checklist


internal/plugins/jest/rules/prefer_to_be/prefer_to_be.go
```diff
@@ -0,0 +1,289 @@
...
+		case ast.KindParenthesizedExpression:
+			node = node.AsParenthesizedExpression().Expression
+		case ast.KindAsExpression:
+			node = node.AsAsExpression().Expression
+		case ast.KindTypeAssertionExpression:
+			node = node.AsTypeAssertion().Expression
+		case ast.KindNonNullExpression:
+			node = node.AsNonNullExpression().Expression
+		case ast.KindSatisfiesExpression:
+			node = node.AsSatisfiesExpression().Expression
```
fansenze: Nice port! One thing worth a second look: `unwrapTypeAssertions` strips more node kinds than the original `followTypeAssertionChain` does.

I checked against `eslint-plugin-jest@27.9.0` locally with this file:

```ts
expect(value).toEqual((null));               // 1
expect(value).toEqual(null!);                // 2
expect(value).toEqual(1 satisfies number);   // 3
expect(value).toEqual((NaN));                // 4
expect(value).toEqual((undefined));          // 5
expect(value).toEqual((1));                  // 6
```

Original reports lines 1, 4, 5, 6 — but not 2 and 3. Parens are fine to strip (ESTree doesn't even have a `ParenthesizedExpression` node, so the original sees `(null)` as plain `null` and reports it — your code needs that case to match). But `null!` and `1 satisfies number` get reported here while the original ignores them, so those two are real over-reporting.

Suggest dropping these two cases from the switch and keeping `KindParenthesizedExpression`, `KindAsExpression`, `KindTypeAssertionExpression`.

Small side note: the bracket-notation `not` removal (`expect(x)["not"].toBe(undefined)`) actually works better here than upstream — upstream's `range[0] - 1` trick leaves a stray `]`. Worth a test case to lock it in.

eryue0220: updated.

fansenze (approved):

PR #937, opened 2026-05-14
feat: port rule promise/no-return-wrap

elecmonkey:
## Checklist


fansenze:
Thanks for the port! I ran the rule against `eslint-plugin-promise@7.3.0` as an oracle over a wider edge-case set and found six cases that disagree with upstream, splitting into two root causes.

### 1. Parenthesized callbacks are missed (false negative)

ESTree drops `ParenthesizedExpression` during parsing; the TS AST keeps it. The bind-chain loop in `isInPromise` peels parens with `IsOuterExpression`, but two other spots don't — so once a callback is wrapped in parens, it stops being recognized as a `.then` / `.catch` / `.finally` argument.

Cases that should report but don't:
```js
doThing().then((function() { return Promise.resolve(4) }))
doThing().then(((function() { return Promise.resolve(4) })))
doThing().then((() => Promise.resolve(4)))
doThing().then(((fn).bind(this)))
doThing().then(((fn).bind)(this))
```

Fix: apply the same paren-skip pattern at the trailing return of `isInPromise` and at the `call := parent.Parent` step inside the loop.

### 2. `Promise?.resolve(4)` is wrongly reported (false positive)

ESTree wraps optional chains in a `ChainExpression`, and upstream short-circuits on `ReturnStatement.argument.type !== 'CallExpression'`, so anything inside an optional chain is skipped. The TS AST has no `ChainExpression` wrapper — optional-ness is just a `QuestionDotToken` flag on the inner nodes — so the port walks into `checkCallExpression` and reports.

Case that reports here but not upstream:
```js
doThing().then(function() { return Promise?.resolve(4) })
```

Fix: guard `checkCallExpression` against optional callees — bail when the callee's `PropertyAccessExpression.QuestionDotToken` is set, or, more thoroughly, when any link in the callee chain is optional.

fansenze:
@elecmonkey **Remaining issue: false positive when the return is inside a nested method / accessor / constructor**

I cross-checked the rule against `eslint-plugin-promise@7` (also swept ~20 other JS shapes, those all line up — the cases below are the only remaining divergence I found):

```js
// All of these are flagged by the port, but upstream is silent on them.
doThing().then(function() {
  class Foo { method() { return Promise.resolve(4) } }
  return new Foo()
})

doThing().then(function() {
  return { m() { return Promise.resolve(4) } }
})

doThing().then(function() {
  return { get x() { return Promise.resolve(4) } }
})
// ...same story for setters, static methods, and constructors — 8 shapes in total.
```

**Why it diverges**

In ESTree, every class method / object shorthand / getter / setter / constructor is internally `MethodDefinition` (or `Property`) wrapping a real `FunctionExpression`. Upstream's `getAncestors().filter(FE | ArrowFE).reverse()[0]` lands on that inner `FunctionExpression`, whose parent is `MethodDefinition` / `Property` — not a `CallExpression` — so `isPromise(parent)` returns false and nothing is reported.

In typescript-go's AST these are first-class nodes (`KindMethodDeclaration` / `KindGetAccessor` / `KindSetAccessor` / `KindConstructor`) with the body attached directly — there is no synthetic inner `FunctionExpression`. The port's `nearestFunctionExpressionOrArrow` only matches `KindFunctionExpression` / `KindArrowFunction`, so it walks straight past these nodes and ends up attributing the return to the outer `.then` callback, producing a false positive.

The nested `function inner() {...}` case is fine, by the way — upstream's filter also skips `FunctionDeclaration`, and the port matches that behavior. So this isn't a "handle all function-like kinds" problem; it's specifically the method/accessor/constructor kinds.

**Suggested fix (verified locally)**

Extend the kind set in the helper — but deliberately keep `FunctionDeclaration` out, to preserve the existing upstream-aligned behavior:

```go
func nearestFunctionExpressionOrArrow(node *ast.Node) *ast.Node {
    for cur := node.Parent; cur != nil; cur = cur.Parent {
        switch cur.Kind {
        case ast.KindFunctionExpression,
            ast.KindArrowFunction,
            ast.KindMethodDeclaration,
            ast.KindGetAccessor,
            ast.KindSetAccessor,
            ast.KindConstructor:
            return cur
        }
    }
    return nil
}
```

(Renaming it to something like `nearestFunctionBoundary` would make the new semantics clearer, but that's optional.) I applied this locally and all existing tests still pass, plus the 8 cases above. Worth mirroring those 8 into both the Go and TS test suites as valid cases.

fansenze (approved):

PR #938, opened 2026-05-14
feat: port rule promise/always-return

elecmonkey:
## Checklist


fansenze:
Thanks for the port! I ran the rule against `eslint-plugin-promise@7.3.0` as an oracle over a wider edge-case set and found seven cases that disagree with upstream — all false positives sharing one root cause.

### Cases reported here but valid upstream

(each used as the `.then()` callback)

```js
function(x) { switch (x) { case 1: return 'a'; case 2: return 'b'; default: return 'c'; } }
function() { while (true) { return 1; } }
function() { do { return 1; } while (false); }
function() { outer: { return 1; } }
function() { for (;;) { return 1; } }
function(x) { if (x) { switch (x) { case 1: return 1; default: return 2; } } else { return 0; } }
function() { while (true) { throw new Error(); } }
```

### Root cause

`statementTerminates` handles `Return` / `Throw` / `process.exit|abort` / `Block` / `If` / `Try`, but everything else falls through to `default: return false`. Five missing statement kinds:

- `KindSwitchStatement` — terminates when there's a `default` clause and every case ends in a terminating stmt
- `KindWhileStatement` / `KindForStatement` — when the condition is literal `true` (or empty in `for`), the loop is infinite, so a terminating path in the body terminates the function
- `KindDoStatement` — body runs at least once, so a terminating body terminates the function
- `KindLabeledStatement` — recurse into the inner statement

Upstream uses ESLint's CodePath `finalSegments` so these fall out naturally; the AST-walking approach here has to enumerate them. The `switch`-with-default case in particular is common enough in production code that it'd produce noisy reports.

### Fix

Extend `statementTerminates` to cover the kinds above, and lock the seven cases into the test suite. The `while` / `for` infinite-loop case also needs to account for `break` exiting the loop.

elecmonkey:
ci passed. some functions (IsFunctionLikeContainer, Completion, StatementCompletion, StatementListCompletion) were extracted into `internal/utils/flow.go`

if this pr is merged, maybe I could launch a new pr to update some other things about the functions extracted?

- no_fallthrough
- react/jsx_max_depth
and `react_hooks/react_hooksutil`'s IsFunctionLikeContainer.

@fansenze

internal/utils/flow.go
```diff
@@ -121,3 +121,276 @@ func BlockEndsWithTerminal(block *ast.Node) bool {
...
+}
+
+func isLiteralTrue(expr *ast.Node) bool {
+	expr = ast.SkipOuterExpressions(expr, skipOEKParentheses)
+	return expr != nil && expr.Kind == ast.KindTrueKeyword
+}
+
+func loopBodyTerminates(loop *ast.Node, body *ast.Node) bool {
+	return StatementCompletion(body) == CompletionTerminates && !containsBreakExitingLoop(loop, body)
+}
```
fansenze, line 330: These four are flagged here, valid upstream:

```js
hey.then(function() { while (true) {} })
hey.then(function(x) { while (true) { if (x) return 1; } })
hey.then(function() { while (true) { x++; } })
hey.then(function() { do {} while (true) })
```

For an infinite loop, "body must strictly terminate" is too strict — a body that just falls through and lets the loop iterate again is fine, since the only ways out are `return` / `throw` / a `break` that exits the loop. Replacing the body-terminates check with `!containsBreakExitingLoop(stmt, body)` at the three call sites matches upstream's CodePath model (where infinite loops have no `finalSegments` to complain about):

```go
case ast.KindWhileStatement:
    whileStmt := stmt.AsWhileStatement()
    if whileStmt != nil && isLiteralTrue(whileStmt.Expression) &&
        !containsBreakExitingLoop(stmt, whileStmt.Statement) {
        return CompletionTerminates
    }
    return CompletionFallsThrough
// same shape for ForStatement
case ast.KindDoStatement:
    doStmt := stmt.AsDoStatement()
    if doStmt == nil { return CompletionFallsThrough }
    if isLiteralTrue(doStmt.Expression) {                       // do { ... } while (true)
        if !containsBreakExitingLoop(stmt, doStmt.Statement) {
            return CompletionTerminates
        }
        return CompletionFallsThrough
    }
    if loopBodyTerminates(stmt, doStmt.Statement) {             // non-infinite: body runs once
        return CompletionTerminates
    }
    return CompletionFallsThrough
```

Verified locally — all 62 existing tests still pass.

internal/utils/flow.go
```diff
@@ -121,3 +121,276 @@ func BlockEndsWithTerminal(block *ast.Node) bool {
...
+			return false
+		}
+	}
+	return false
+}
+
+func isLiteralTrue(expr *ast.Node) bool {
+	expr = ast.SkipOuterExpressions(expr, skipOEKParentheses)
+	return expr != nil && expr.Kind == ast.KindTrueKeyword
+}
```
fansenze: Minor: only `KindTrueKeyword` is matched, so this gets flagged:

```js
hey.then(function() { while (1) { return 1; } })
```

ESLint's CodePath also treats non-zero numeric literals and non-empty string literals as definitely-truthy. Easy extension.

internal/utils/flow.go
```diff
@@ -121,3 +121,276 @@ func BlockEndsWithTerminal(block *ast.Node) bool {
...
+		if doStmt != nil && loopBodyTerminates(stmt, doStmt.Statement) {
+			return CompletionTerminates
+		}
+		return CompletionFallsThrough
+	case ast.KindLabeledStatement:
+		labeledStmt := stmt.AsLabeledStatement()
+		if labeledStmt == nil {
+			return CompletionFallsThrough
+		}
+		return StatementCompletion(labeledStmt.Statement)
```
fansenze, line 255: Edge case: silent here, reported upstream:

```js
hey.then(function(x) { outer: { if (x) break outer; return 1; } })
```

If `x` is truthy, `break outer` exits the labeled block and the function falls through. The three-state `Completion` collapses `if (x) break outer;` (no `else`) to `FallsThrough` and the "could-stop" signal is lost by the time the labeled block decides its completion — so it ends up looking like it always terminates via the trailing `return 1`.

Tricky to fix cleanly without enriching the model — flagging for awareness, not blocking.

fansenze (commented):
@elecmonkey Thanks for the port! The 7 cases I flagged earlier are all fixed. Cross-checked against `eslint-plugin-promise@7` and found three remaining divergences, all in `internal/utils/flow.go`:

- **Critical**: 4 false positives from infinite-loop body-completion check (one shared fix)
- **Minor**: `while (1)` not recognized as truthy
- **Edge case**: labeled `break` that escapes a labeled block

Details on each in the inline comments below.

Once these are addressed (or the edge case explicitly deferred with a note), this is good to merge from my side. `flow.go` becoming shared infrastructure is also a nice win — the loop fix here will pay off for any future rule using `StatementCompletion`.

And feel free to open a follow-up PR for the other updates related to the extracted functions.

fansenze (approved):

PR #948, opened 2026-05-15
feat(eslint-plugin-jest): add `valid-title` rule

eryue0220:
## Summary

Port `valid-title` from eslint-plugin-jest to rslint.

## Related Links

eslint-plugin-jest/valid-title [doc](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/valid-title.md) [code](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/valid-title.ts)

## Checklist


fansenze (commented):
Nice work on this — ran the rule against upstream `eslint-plugin-jest@29.15.2` with matching fixtures on both sides and noticed a few alignment gaps. Left inline notes on each.

The first two affect actual rule output users would see; the rest are cosmetic or corner cases.

Side note: the Go and JS tests assert only `messageId` (no line/column/message text), so these gaps don't show up in CI. Might be worth pinning location + message in a couple of representative cases to catch drift going forward.

internal/plugins/jest/rules/valid_title/valid_title.go
```diff
@@ -0,0 +1,471 @@
...
+	}
+	return def
+}
+
+func compileRE2(pat string) *regexp2.Regexp {
+	if pat == "" {
+		return nil
+	}
+	re, err := regexp2.Compile(pat, jsRegexOpts)
+	if err != nil {
```
fansenze, line 75: Silent failure here means an invalid user regex (e.g. `mustMatch: '(unclosed'`) just disables the matcher with no feedback — the lint run looks fine and the rule isn't enforcing anything.

Upstream throws `SyntaxError` and surfaces it as a fatal lint error so users notice immediately. Same concern applies to the `disallowedWords` regex below and `mustNotMatch`. Could we surface the compile error somehow — config diagnostic, or at minimum a log?

internal/plugins/jest/rules/valid_title/valid_title.go
```diff
@@ -0,0 +1,471 @@
...
+func compileDisallowedWords(raw interface{}) *regexp2.Regexp {
+	items, ok := raw.([]interface{})
+	if !ok || len(items) == 0 {
+		return nil
+	}
+	parts := make([]string, 0, len(items))
+	for _, it := range items {
+		w, ok := it.(string)
+		if ok && w != "" {
+			parts = append(parts, regexp.QuoteMeta(w))
```
fansenze: Upstream doesn't escape these — it builds ``new RegExp(`\b(${words.join('|')})\b`, 'iu')`` so users can pass `'foo|bar'` to mean *either* `foo` or `bar`, or use `.` as a wildcard. With `QuoteMeta` each entry is treated as a literal:

| title | options | upstream | this PR |
|---|---|---|---|
| `"just foo"` | `disallowedWords:['foo\|bar']` | flags `"foo"` | (nothing) |
| `"foo_bar in title"` | `disallowedWords:['foo.bar']` | flags `"foo_bar"` (dot matches `_`) | (nothing) |

```suggestion
			parts = append(parts, w)
```

internal/plugins/jest/rules/valid_title/valid_title.go
```diff
@@ -0,0 +1,471 @@
...
+	}
+	return word
+}
+
+func binaryPlusContainsStringLit(n *ast.Node) bool {
+	if n == nil || n.Kind != ast.KindBinaryExpression {
+		return false
+	}
+	be := n.AsBinaryExpression()
+	if be == nil || be.OperatorToken == nil || be.OperatorToken.Kind != ast.KindPlusToken {
```
fansenze: This `KindPlusToken` check is stricter than upstream — their `doesBinaryExpressionContainStringNode` runs on any `BinaryExpression`, regardless of operator. So `it(a - 'b', cb)` stays silent upstream but reports `titleMustBeString` here.

Practical impact is tiny (nobody writes that as a title), but it's a divergence — happy to leave as-is if you prefer the stricter behavior, just flagging.

internal/plugins/jest/rules/valid_title/valid_title.go
```diff
@@ -0,0 +1,471 @@
...
+			continue
+		}
+		gp := parent.Parent
+		return gp != nil && gp.Kind == ast.KindCallExpression
+	}
+	return false
+}
+
+var (
+	reDupPrefix            = regexp.MustCompile(`^([\x60'"]).+?\s+`)
```
fansenze: Upstream uses a literal space in the equivalent regex (`/^([`'"]).+? /u`). With `\s+` here, a tab or newline right after the prefix gets consumed by the autofix:

```js
describe('describe \tfoo', () => {});
// upstream fixed-point: describe('\tfoo', () => {});  (tab preserved, accidentalSpace lingers)
// this PR fixed-point:  describe('foo', () => {});
```

Normal `'describe foo'` and `'describe  foo'` titles converge to the same result on both sides — only mixed whitespace diverges. Edge case, but worth matching:

```suggestion
	reDupPrefix            = regexp.MustCompile(`^([\x60'"]).+? `)
```

internal/plugins/jest/rules/valid_title/valid_title.go
```diff
@@ -0,0 +1,471 @@
...
+func duplicatePrefixReplacement(rawSrc string) string {
+	return reDupPrefix.ReplaceAllString(rawSrc, "$1")
+}
+
+func regexpToMessagePattern(re *regexp2.Regexp) string {
+	if re == nil {
+		return ""
+	}
+	src := re.String()
+	return "/" + strings.ReplaceAll(src, "/", "\\/") + "/"
```
fansenze: Upstream uses `RegExp.toString()` which includes the flag, so error messages read `it should match /good/u`. Here we emit just `/good/` — the `u` flag is always present in upstream (it's hardcoded when compiling) but dropped from the rendered message.

Doesn't change rule behavior, just the error text users see. If full alignment matters you'd want to thread the flag string through alongside the compiled regex.

internal/plugins/jest/rules/valid_title/valid_title.go
```diff
@@ -0,0 +1,471 @@
...
+						ctx.ReportNodeWithFixes(arg, rule.RuleMessage{
+							Id:          "titleMustBeString",
+							Description: "Title must be a string",
+						})
+					}
+					return
+				}
+
+				if title == "" {
+					ctx.ReportNode(arg, rule.RuleMessage{
```
fansenze: Upstream passes `node` (the whole `CallExpression`) here, so `describe('', cb)` highlights the entire call. With `arg` the squiggle shrinks to just `''`:

- upstream range: `L1:1–L1:23`
- this PR range:  `L1:10–L1:12`

Doesn't affect pass/fail, only IDE highlight width.

```suggestion
					ctx.ReportNode(node, rule.RuleMessage{
```

internal/plugins/jest/rules/valid_title/valid_title.go
```diff
@@ -0,0 +1,547 @@
...
+
+func binaryPlusContainsStringLit(n *ast.Node) bool {
+	if n == nil || n.Kind != ast.KindBinaryExpression {
+		return false
+	}
+	be := n.AsBinaryExpression()
+	if be == nil {
+		return false
+	}
+	if utils.IsStringLiteralOrTemplate(be.Left) {
```
fansenze: Subtle divergence from upstream here: `utils.IsStringLiteralOrTemplate` also matches `KindTemplateExpression` (templates with `${...}`), but upstream's `isStringNode` requires `quasis.length === 1` — it doesn't treat templates-with-substitutions as a string node.

So these stay silent here but report `titleMustBeString` upstream:

```js
it(foo + `suffix${x}`, () => {});
it(`prefix${x}` + foo, () => {});
```

PR's behavior is arguably more correct (the expression really is a string). Happy to leave as-is, just flagging — if you'd like exact mirror:

```go
if ast.IsStringLiteralLike(be.Left)  { return true }
if ast.IsStringLiteralLike(be.Right) { return true }
```

Either way, worth a test pinning down whichever side you pick.

internal/plugins/jest/rules/valid_title/valid_title.go
```diff
@@ -0,0 +1,547 @@
...
+		ignoreTypeOfDescribeName: boolFromMap(m, "ignoreTypeOfDescribeName", false),
+		ignoreTypeOfTestName:     boolFromMap(m, "ignoreTypeOfTestName", false),
+	}
+
+	if dw, ok := m["disallowedWords"]; ok && dw != nil {
+		co.disallowedConcat, co.invalidPatterns = compileDisallowedWords(dw, co.invalidPatterns)
+	}
+
+	if mn, ok := m["mustNotMatch"]; ok {
+		co.mustNotMatch, co.invalidPatterns = compileMatcherPatterns(mn, "mustNotMatch")
```
fansenze: Looks like a real bug here — `co.invalidPatterns` gets overwritten instead of appended.

`compileDisallowedWords` takes the existing slice and appends to it, but `compileMatcherPatterns` returns a fresh one, so these later assignments wipe whatever the previous step collected. Reproduced with:

```jsonc
{
  "disallowedWords": ["ok", "(unterminated"],  // invalid regex
  "mustMatch": "^should "                       // valid
}
```

upstream throws `SyntaxError`; this PR reports nothing, and `disallowedConcat` is already nil from the failed compile — so the user's `disallowedWords` ends up silently disabled.

Could we accumulate at the call site:

```go
ms, more := compileMatcherPatterns(mn, "mustNotMatch")
co.mustNotMatch = ms
co.invalidPatterns = append(co.invalidPatterns, more...)
```

A combined-option test would lock it in.

internal/plugins/jest/rules/valid_title/valid_title.go
```diff
@@ -0,0 +1,551 @@
...
+		return false
+	}
+	if ast.IsStringLiteralLike(be.Left) {
+		return true
+	}
+	if ast.IsStringLiteralLike(be.Right) {
+		return true
+	}
+	return binaryPlusContainsStringLit(be.Left) || binaryPlusContainsStringLit(be.Right)
+}
```
fansenze: Subtle follow-on from my earlier "any BinaryExpression, regardless of operator" comment — that was based on ESTree's narrower `BinaryExpression`. In typescript-go's AST, `KindBinaryExpression` also covers `LogicalExpression` (`||`, `&&`, `??`) and `AssignmentExpression` (`=`, `+=`, …), so this helper now matches more than upstream does:

```js
// All three: upstream reports titleMustBeString, port stays silent
it('a' || 'b',  () => {})
it(cond && 'b', () => {})
it(x = 'foo',   () => {})
```

Fix: filter on `OperatorToken.Kind` at the top — bail out for logical / assignment / comma operators. Everything else (arithmetic, comparison, bitwise, `in`/`instanceof`) stays included to match upstream:

```go
func binaryPlusContainsStringLit(n *ast.Node) bool {
    if n == nil || n.Kind != ast.KindBinaryExpression {
        return false
    }
    be := n.AsBinaryExpression()
    if be == nil || be.OperatorToken == nil {
        return false
    }
    switch be.OperatorToken.Kind {
    case ast.KindAmpersandAmpersandToken, ast.KindBarBarToken, ast.KindQuestionQuestionToken,
        ast.KindEqualsToken,
        ast.KindPlusEqualsToken, ast.KindMinusEqualsToken,
        ast.KindAsteriskEqualsToken, ast.KindSlashEqualsToken, ast.KindPercentEqualsToken,
        ast.KindAsteriskAsteriskEqualsToken,
        ast.KindAmpersandEqualsToken, ast.KindBarEqualsToken, ast.KindCaretEqualsToken,
        ast.KindLessThanLessThanEqualsToken, ast.KindGreaterThanGreaterThanEqualsToken, ast.KindGreaterThanGreaterThanGreaterThanEqualsToken,
        ast.KindBarBarEqualsToken, ast.KindAmpersandAmpersandEqualsToken, ast.KindQuestionQuestionEqualsToken,
        ast.KindCommaToken:
        return false
    }
    // ...existing left/right/recurse checks unchanged
}
```

Bonus tightening to match upstream exactly: upstream's `doesBinaryExpressionContainStringNode` only recurses on `left` (when `left.type === BinaryExpression`), not on both sides. Couldn't construct a case where the both-sides recursion alone causes a divergence (operator precedence + TS's explicit `ParenthesizedExpression` swallows them), but it's worth aligning.

Renaming away from `binaryPlusContainsStringLit` (e.g. `binaryExprContainsStringLit`) would also help — the "Plus" in the name is stale after the earlier broadening.

internal/plugins/jest/rules/valid_title/valid_title.go
```diff
@@ -0,0 +1,551 @@
...
+	}
+	switch n.Kind {
+	case ast.KindStringLiteral:
+		return n.AsStringLiteral().Text, true
+	case ast.KindNoSubstitutionTemplateLiteral:
+		return n.AsNoSubstitutionTemplateLiteral().Text, true
+	default:
+		return "", false
+	}
+}
```
fansenze: Upstream's `getStringValue` returns `quasis[0].value.raw` for template literals — the **raw** source between the backticks, with escape sequences preserved. Port returns `.Text` which is the **cooked** value (escapes resolved). For string literals this doesn't matter (`StringLiteral.Text` is cooked, but so is upstream's `node.value`), but template literals diverge on titles containing escapes:

```js
it(`\nfoo\n`, () => {})
// upstream title: "\nfoo\n" (7 chars, literal backslash-n) → trim no-op → no report
// port title:     "<NL>foo<NL>" (5 chars, real newlines)   → trim strips → accidentalSpace REPORTED
```

Practical impact is small (rare to put escape sequences in test titles), but it does cause a real divergence on `accidentalSpace` and on any downstream string-matching check (disallowedWord, mustMatch, …). If `typescript-go/shim/ast` exposes a `RawText` field on `NoSubstitutionTemplateLiteral`, that's the drop-in fix; otherwise scanning the source text between the backticks works.

fansenze (commented):
@eryue0220 Thanks for the port! All 8 of my earlier comments are addressed — really clean follow-through, especially on the `disallowedWords` semantics and the `invalidPatterns` accumulation bug. Cross-checked against `eslint-plugin-jest@main` (which has the latest `invalidEachSpecifier` branch) and found two remaining divergences:

- **Critical**: `binaryPlusContainsStringLit` accepts logical / assignment operators too
- **Minor**: `NoSubstitutionTemplateLiteral` uses cooked text instead of raw

Details in the inline comments below.

Once these are addressed, this is good to merge from my side.

internal/plugins/jest/rules/valid_title/valid_title.go
eryue0220: Updated.

internal/plugins/jest/rules/valid_title/valid_title.go
eryue0220: Updated.

internal/plugins/jest/rules/valid_title/valid_title.go
```diff
@@ -0,0 +1,568 @@
...
+	if ast.IsStringLiteralLike(be.Left) {
+		return true
+	}
+	if ast.IsStringLiteralLike(be.Right) {
+		return true
+	}
+	return binaryExprContainsStringLit(be.Left)
+}
+
+func rawTemplateLiteralText(node *ast.Node, sourceText string) string {
```
fansenze: `node.Pos()` is the token **full start** — it includes leading trivia (whitespace/comments before the `` ` ``). So `sourceText[pos+1 : end-1]` skips the first trivia char instead of the opening backtick, and the returned "title" leaks trivia + the real `` ` `` into every downstream check (`accidentalSpace`, `disallowedWord`, `mustMatch`, `duplicatePrefix`, `invalidEachSpecifier`).

Verified on `c6809a4d` with `rule_tester` — this multiline form (canonical jest formatting, valid upstream) wrongly reports `accidentalSpace`:

```js
it(
  `hello`,
  () => {});
```

Fix by using the trimmed range:

```go
func rawTemplateLiteralText(sf *ast.SourceFile, node *ast.Node) string {
    r := utils.TrimNodeTextRange(sf, node)
    start, end := r.Pos(), r.End()
    if end-start < 2 {
        return ""
    }
    return sf.Text()[start+1 : end-1]
}
```

Worth adding lock-in valid tests for the multiline / leading-space template forms.

fansenze (approved):

PR #950, opened 2026-05-15
feat(eslint-plugin-jest): add `expect-expect` rule

eryue0220:
## Summary

Port `expect-expect` from eslint-plugin-jest to rslint.

## Related Links

eslint-plugin-jest/expect-expect [doc](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/expect-expect.md) [code](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/expect-expect.ts)

## Checklist


internal/plugins/jest/rules/expect_expect/expect_expect.go
```diff
@@ -0,0 +1,272 @@
...
+
+func checkCallExpressionUsed(
+	assertNode *ast.Node,
+	unchecked *[]*ast.Node,
+	assertedFnDecls map[*ast.Node]bool,
+	assertedFnNames map[string]bool,
+	uncheckedByDecl map[*ast.Node][]*ast.Node,
+	uncheckedByName map[string][]*ast.Node,
+) {
+	for n := assertNode.Parent; n != nil; n = n.Parent {
```
fansenze, line 171: This parent walk goes leaf→root, while upstream walks `context.sourceCode.getAncestors(node)` which is root→leaf. For most cases this doesn't matter (only one `it` in the chain), but for nested `it`:

```js
it('outer', () => { it('inner', () => { someAssert(1); }); });
// options: { assertFunctionNames: ['someAssert'] }
```

| | which `it` is credited | which `it` is reported |
|---|---|---|
| upstream | outer (first match root→leaf, splice + break) | **inner** (col 21-23) |
| this PR  | inner (first match leaf→root, remove + return) | **outer** (col 1-3) |

Both flag exactly one error — just on different `it`s. Nested `it` is non-idiomatic so impact is small, but it's a stable divergence.

(For chained `expect(1).toBe(1)` both `it`s end up credited because each of `expect()` and `expect.toBe()` triggers a separate check, which is why this only surfaces for single-call assertions.)

internal/plugins/jest/rules/expect_expect/expect_expect.go
```diff
@@ -0,0 +1,272 @@
...
+		if n.Kind == ast.KindFunctionDeclaration {
+			decl := n.AsFunctionDeclaration()
+			if decl != nil && decl.Name() != nil {
+				declNode := decl.AsNode()
+				fnName := decl.Name().Text()
+				assertedFnDecls[declNode] = true
```
fansenze: There's a hoisting divergence with upstream worth flagging:

```js
// case A: it before func decl
it('t', myTest); function myTest() { expect(1).toBe(1); }

// case B: func decl before it
function myTest() { expect(1).toBe(1); } it('t', myTest);
```

Upstream uses a reactive approach — when the assert call is visited, it calls `getTestCallExpressionsFromDeclaredVariables` to find `it()` references to the enclosing function. This works for case A (the `it` is already in `unchecked`) but **misses case B** — at the time the assert is visited, the `it` hasn't been visited yet so `unchecked` is still empty.

| case | upstream | this PR |
|---|---|---|
| A (it-first) | no errors | no errors |
| B (func-first) | **`noAssertions` on `it`** | no errors |

The stateful `assertedFnDecls` map makes this PR more semantically correct (the test does have an assertion via the hoisted function), but it diverges from upstream. Note that `expect_expect_test.go:25` already codifies this divergence:

```go
{Code: `function myTest() { expect(true).toBeDefined() } it("should pass", myTest);`},
```

Upstream actually reports `noAssertions` on this case. Probably worth a deliberate call: align with upstream (drop the stateful map, accept the upstream FP) or document the intentional deviation.

fansenze (commented):
Nice work — the `fix cr` refactor is clean and addresses gemini's three concerns. Ran the rule against upstream `eslint-plugin-jest@29.15.2` with matching fixtures; most behavior aligns. Two divergences inline:

- F2: hoisting (func-first) — PR is semantically more correct but diverges from upstream, and one existing test (`valid-25`) codifies that divergence
- F3: nested `it` direction — different `it` is reported between the two implementations

Both come from the same design choice: stateful tracking + leaf→root parent walk, vs upstream's reactive lookup + root→leaf ancestor walk.

Same note as #948: tests assert only `messageId`, no line/column, so the F3 divergence doesn't surface in CI.

internal/plugins/jest/rules/expect_expect/expect_expect.go
```diff
@@ -0,0 +1,259 @@
...
+
+func compileAssertPattern(pattern string) *regexp.Regexp {
+	segs := strings.Split(pattern, ".")
+	parts := make([]string, 0, len(segs))
+	for _, s := range segs {
+		if s == "**" {
+			parts = append(parts, `[a-zA-Z0-9_$.]*`)
+		} else {
+			parts = append(parts, strings.ReplaceAll(s, "*", `[a-zA-Z0-9_$]*`))
+		}
```
fansenze: The character class here is more permissive than upstream — `[a-zA-Z0-9_$]` matches `_` and `$`, but upstream's `[a-z\d]` (with `ui` flags) doesn't. This causes a silent divergence:

```js
// assertFunctionNames: ['request.**.expect']
test('x', () => Request.GET_1().foo$Bar().expect(456));
```
- Upstream (eslint-plugin-jest@29.15.2): ❌ reports `noAssertions`
- This PR: ✅ passes (verified by running both upstream and a probe Go test against this rule)

**Suggested fix** — drop `_$` so the classes match upstream:
```go
if s == "**" {
    parts = append(parts, `[a-zA-Z0-9.]*`)
} else {
    parts = append(parts, strings.ReplaceAll(s, "*", `[a-zA-Z0-9]*`))
}
```
The valid test case at `expect_expect_test.go:487` (`Request.GET_1().foo$Bar().expect`) would then need to move to `invalid` or be removed.

internal/plugins/jest/rules/expect_expect/expect_expect.go
```diff
@@ -0,0 +1,259 @@
...
+	assertNames := []string{"expect"}
+	additional := []string{}
+
+	m := internalUtils.GetOptionsMap(options)
+	if m == nil {
+		return assertNames, additional
+	}
+
+	if raw, ok := m["assertFunctionNames"]; ok && raw != nil {
+		if arr, ok := raw.([]interface{}); ok && len(arr) > 0 {
```
fansenze: The `len(arr) > 0` guard means an explicitly empty `assertFunctionNames: []` silently falls back to the default `["expect"]`. Upstream treats `[]` literally:

```js
// assertFunctionNames: []
test('x', () => { expect(true).toBeDefined(); });
```
- Upstream: ❌ reports `noAssertions` (no patterns → nothing matches)
- This PR: ✅ passes (default `expect` quietly added back)

**Suggested fix** — drop the length guard so an empty array is respected:
```go
if arr, ok := raw.([]interface{}); ok {
    out := make([]string, 0, len(arr))
    for _, v := range arr {
        if s, ok := v.(string); ok {
            out = append(out, s)
        }
    }
    assertNames = out
}
```

fansenze (commented):
Nice work — earlier comments are addressed and the test suite is green ✅. Did a differential check against `eslint-plugin-jest@29.15.2` and surfaced three silent divergences worth a look. None are blocking IMO; (1) and (2) are the user-visible ones.

internal/plugins/jest/rules/expect_expect/expect_expect.go
```diff
@@ -0,0 +1,259 @@
...
+					return
+				}
+
+				calleeName := utils.CalleeChainName(callExpr.Expression)
+				jestFn := utils.ParseJestFnCall(node, ctx)
+				isJestTest := jestFn != nil && jestFn.Kind == utils.JestFnTypeTest
+				isExtraBlock := calleeName != "" && slices.Contains(additionalTestBlocks, calleeName)
+
+				if isJestTest || isExtraBlock {
+					if isJestTest && isTodoTestCall(jestFn) {
```
fansenze: The `.todo` skip is gated on `isJestTest`, so it doesn't apply when the call comes in via `additionalTestBlockFunctions`. Upstream applies the todo early-return to both branches:

```js
// additionalTestBlockFunctions: ['foo.todo']
foo.todo("eventual test");
```
- Upstream: ✅ passes (callee.property === 'todo' triggers early return)
- This PR: ❌ reports `noAssertions`

Pretty niche, but mirroring upstream would look like:
```go
if isJestTest || isExtraBlock {
    // skip in both branches when the trailing accessor is `.todo`
    if jestFn != nil && len(jestFn.Members) > 0 && jestFn.Members[len(jestFn.Members)-1] == "todo" {
        return
    }
    // ... rest unchanged
}
```
(Or any equivalent check on the trailing accessor.)

internal/plugins/jest/utils/jest.go
```diff
@@ -217,6 +220,76 @@ func getElementAccessName(node *ast.Node) string {
...
+	switch n.Kind {
+	case ast.KindStringLiteral:
+		return n.AsStringLiteral().Text
+	case ast.KindNumericLiteral:
+		return n.AsNumericLiteral().Text
+	case ast.KindNoSubstitutionTemplateLiteral:
+		return n.AsNoSubstitutionTemplateLiteral().Text
+	default:
+		return ""
+	}
```
fansenze, line 291: Cross-checked against `eslint-plugin-jest@main` and found four ElementAccess cases that diverge — all rooted in this helper:

```js
// (1) Identifier as bracket arg, narrow pattern
test('x', () => foo[bar].expect(1));    // assertFunctionNames: ['foo.expect']
// upstream: chain = "foo.bar.expect" → no match → REPORTS
// port:     chain = "foo.expect"     → match    → silent

// (2) Mirror of (1) with the matching pattern
test('x', () => foo[bar].expect(1));    // assertFunctionNames: ['foo.bar.expect']
// upstream: chain = "foo.bar.expect" → match    → silent
// port:     chain = "foo.expect"     → no match → REPORTS

// (3) Numeric literal should truncate (upstream's isSupportedAccessor doesn't accept numeric)
test('x', () => foo[1].expect(1));      // assertFunctionNames: ['foo.1.expect']
// upstream: chain = null → REPORTS
// port:     chain = "foo.1.expect" → match → silent

// (4) Arbitrary expression should truncate, not silently skip
test('x', () => foo[a+b].expect(1));    // assertFunctionNames: ['foo.expect']
// upstream: chain = null → REPORTS
// port:     chain = "foo.expect" → match → silent
```

**Why it diverges**

Upstream's `getNodeName` uses `joinNames(a, b) = a && b ? a.b : null` and `isSupportedAccessor` only accepts `Identifier` / `StringLiteral` / single-quasi `TemplateLiteral`. Any failed join nullifies the chain.

The port has two parts that drift from this:
1. ElementAccess returns `left` when the key can't be resolved (line 257), which loses the "broken" signal — the outer keeps building a chain that shouldn't exist.
2. `calleeChainLiteralElementKey` misses `KindIdentifier` (cases 1, 2) and incorrectly accepts `KindNumericLiteral` (case 3).

**Suggested fix**

```go
// jest.go:252-259 — propagate "broken" instead of swallowing it
case ast.KindElementAccessExpression:
    ea := expr.AsElementAccessExpression()
    left := CalleeChainName(ea.Expression)
    key := calleeChainLiteralElementKey(ast.SkipParentheses(ea.ArgumentExpression))
    if left == "" || key == "" {
        return ""  // was `return left`
    }
    return left + "." + key

// jest.go:277-290 — match upstream's isSupportedAccessor
func calleeChainLiteralElementKey(n *ast.Node) string {
    if n == nil { return "" }
    switch n.Kind {
    case ast.KindIdentifier:                          // new: upstream treats [bar] as property name "bar"
        return n.AsIdentifier().Text
    case ast.KindStringLiteral:
        return n.AsStringLiteral().Text
    case ast.KindNoSubstitutionTemplateLiteral:
        return n.AsNoSubstitutionTemplateLiteral().Text
    // KindNumericLiteral removed — upstream's isSupportedAccessor doesn't accept numerics
    default:
        return ""
    }
}
```

Verified locally — all four cases above flip to match upstream after the change, and the rule's existing tests still pass. Worth mirroring those four into both the Go and TS test suites as locks.

fansenze (commented):
@eryue0220 Thanks for the port! All 5 of my earlier comments are addressed (including the trickier hoisting + nested-`it` refactors), and the 3 gemini-bot suggestions are folded in cleanly too. Cross-checked against `eslint-plugin-jest@main` and found one remaining divergence cluster — all rooted in `CalleeChainName`'s `ElementAccessExpression` handling.

Details in the inline comment below. Once this is addressed, this is good to merge from my side.

(Worth noting `CalleeChainName` is shared infrastructure for the jest plugin, so fixing here will help future rules too.)

internal/plugins/jest/utils/jest.go
eryue0220: Updated.

fansenze (approved):

PR #963, opened 2026-05-16
fix: handle exhaustive-deps shorthand references

### LINKED ISSUE #961: [Bug]: `react-hooks/exhaustive-deps` false positive for object literal shorthand dependencies
### System Info

  System:
    OS: macOS 26.4.1
    CPU: (18) arm64 Apple M5 Pro
    Memory: 3.46 GB / 48.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.15.0 - /Users/bytedance/.nvm/versions/node/v24.15.0/bin/node
    npm: 11.12.1 - /Users/bytedance/.nvm/versions/node/v24.15.0/bin/npm
    pnpm: 11.1.2 - /Users/bytedance/.nvm/versions/node/v24.15.0/bin/pnpm
    bun: 1.3.14 - /Users/bytedance/.bun/bin/bun
  Browsers:
    Chrome: 147.0.7727.56
    Safari: 26.4

### Details

`react-hooks/exhaustive-deps` false positive for object literal shorthand dependencies

### Reproduce link

https://github.com/elecmonkey/rslint-exhaustive-deps-bug-report

### Reproduce Steps

clone my repro repo and run: 

```bash
pnpm lint:eslint
pnpm lint:rslint
```

minimal reproduction component:
```tsx
import { useMemo } from 'react';

const App = () => {
  const isAdmin = true;

  const value = useMemo(() => ({ isAdmin }), [isAdmin]);

  return <pre>{String(value.isAdmin)}</pre>;
};

export default App;
```

elecmonkey:
## Summary

Fixes #961

### DIFF
--- a/internal/plugins/react_hooks/rules/exhaustive_deps/rule.go
+++ b/internal/plugins/react_hooks/rules/exhaustive_deps/rule.go
@@ -59,9 +59,9 @@ var ExhaustiveDepsRule = rule.Rule{
 		// previous O(N×hooks×file_size) re-scan pattern that surfaced
 		// in PR-808 review feedback for large codebases.
 		caches := &runCaches{
-			extraWrite:     map[*ast.Symbol]bool{},
-			currentAssign:  map[*ast.Symbol]bool{},
-			usedOutside:    map[*ast.Node]bool{},
+			extraWrite:    map[*ast.Symbol]bool{},
+			currentAssign: map[*ast.Symbol]bool{},
+			usedOutside:   map[*ast.Node]bool{},
 		}
 
 		report := func(node *ast.Node, msg string) {
@@ -1717,7 +1717,12 @@ func processIdentifier(
 	// Resolve.
 	var sym *ast.Symbol
 	if tc != nil {
-		sym = tc.GetSymbolAtLocation(id)
+		if isObjectLiteralShorthandReference(id) {
+			sym = tc.GetShorthandAssignmentValueSymbol(id.Parent)
+		}
+		if sym == nil {
+			sym = tc.GetSymbolAtLocation(id)
+		}
 	}
 	// Determine if the symbol's declaration is in component scope (between
 	// callback exclusive and componentFn inclusive).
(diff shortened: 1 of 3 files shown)


internal/plugins/react_hooks/rules/exhaustive_deps/rule.go
```diff
@@ -1736,6 +1736,14 @@ func processIdentifier(
 	if sym != nil {
 		if isInComponentPureScope(sym, callback, componentFn) {
 			// Pure scope hit — proceed.
+		} else if isObjectLiteralShorthandReference(id) {
```
fansenze: Nice catch on #961! One concern though — I think this fix overrides TC even when TC's (admittedly buggy) answer was already classified safely as not-a-dep, which opens a different false-positive.

What's happening: tsgo's `GetSymbolAtLocation` always resolves a shorthand identifier to the `ShorthandPropertyAssignment` node itself — that's the root of #961. The syntactic gate added here triggers for every shorthand, which trips on local bindings that shadow an outer same-named one. e.g. this errors `missing dependency: 'isAdmin'` on this branch but is clean on main:

```tsx
function MyComponent() {
  const isAdmin = true;
  useMemo(() => {
    const isAdmin = false;   // local shadow
    return { isAdmin };
  }, []);
}
```

Same with a nested `function inner(isAdmin) { return { isAdmin } }` or `const { isAdmin } = data` inside the callback.

tsgo actually exposes `GetShorthandAssignmentValueSymbol(shorthandNode)` which does the proper lexical resolve, shadowing included. If we route shorthand through it at the resolve site, this whole branch isn't needed:

```go
var sym *ast.Symbol
if tc != nil {
    if isObjectLiteralShorthandReference(id) {
        sym = tc.GetShorthandAssignmentValueSymbol(id.Parent)
    }
    if sym == nil {
        sym = tc.GetSymbolAtLocation(id)
    }
}
```

I tested this locally — the #961 case still passes and the three shadow shapes above also pass.

fansenze (approved):

PR #969, opened 2026-05-18
feat(eslint-plugin-jest): add `no-jasmine-globals` rule

eryue0220:
## Summary

Port `no-jasmine-globals` from eslint-plugin-jest to rslint.

## Related Links

eslint-plugin-jest/no-jasmine-globals [doc](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-jasmine-globals.md) [code](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/no-jasmine-globals.ts)

## Checklist


internal/plugins/jest/rules/no_jasmine_globals/no_jasmine_globals.go
```diff
@@ -0,0 +1,200 @@
...
+					}
+					return
+				case ast.KindPropertyAccessExpression, ast.KindElementAccessExpression:
+					chain, jasmineObj, tail, ok := jasmineRootCallChain(callee)
+					if !ok {
+						return
+					}
+
+					switch tail {
+					case "any", "anything", "arrayContaining", "objectContaining", "stringMatching":
```
fansenze, line 208: 🐛 Looks like the fix misfires on nested chains.

`tail` here only looks at the *last* segment, so `jasmine.foo.any()` matches `"any"` and walks into this fix branch — replacing `jasmine` with `expect` and producing:

```js
jasmine.foo.any()  →  expect.foo.any()
```

upstream uses the dotted suffix (`getNodeName(callee).replace('jasmine.', '')`), so for `jasmine.foo.any` the `functionName` is `"foo.any"`, doesn't match the list, and falls through to `illegalJasmine` with no fix.

Gating the fix branch on the chain being exactly `jasmine.<name>` should do it:

```go
case "any", "anything", "arrayContaining", "objectContaining", "stringMatching":
    if len(entries) != 2 {
        ctx.ReportNode(node, buildIllegalJasmineMessage())
        return
    }
    ctx.ReportNodeWithFixes(...)
```

(Would need to thread `entries` out of `jasmineRootCallChain`, or do the length check on `strings.Count(chain, ".")`.) A test for `jasmine.foo.any()` would lock it in.

internal/plugins/jest/rules/no_jasmine_globals/no_jasmine_globals.go
```diff
@@ -0,0 +1,200 @@
...
+func isBindingResolved(ident *ast.Node, ctx rule.RuleContext) bool {
+	if ctx.TypeChecker == nil {
+		return false
+	}
+	sym := ctx.TypeChecker.GetSymbolAtLocation(ident)
+	return sym != nil
+}
+
+func jasmineRootCallChain(callee *ast.Node) (chain string, jasmineObj *ast.Node, tail string, ok bool) {
+	entries := jestutils.GetJestFnMemberEntries(callee)
```
fansenze, line 67: 🐛 False positive on chains where `jasmine` isn't actually the root.

`GetJestFnMemberEntries` silently drops `ThisKeyword` / `SuperKeyword` (they hit the `default → return nil` branch in `jest.go`), so `this.jasmine.foo()` flattens to `[jasmine, foo]` and this check happily accepts it as a plain `jasmine.foo()` call.

Reproduced with a unit test:

```js
(this as any).jasmine.foo()  // PR: reports `illegalJasmine`        | upstream: silent
(this as any).jasmine.any()  // PR: fixes to `this.expect.any()`    | upstream: silent
```

upstream's `getNodeName` returns `null` for these (because `this` isn't a named accessor), so the rule skips them entirely.

Could we verify the chain is actually rooted at a bare `jasmine` identifier? A small local guard does the trick without touching the shared `GetJestFnMemberEntries`:

```go
if entries[0].Node.Parent == nil ||
    rslintutils.AccessExpressionObject(entries[0].Node.Parent) != entries[0].Node {
    return "", nil, "", false
}
```

Tests for `this.jasmine.foo()` and `this.jasmine.any()` would lock it in.

internal/plugins/jest/rules/no_jasmine_globals/no_jasmine_globals.go
```diff
@@ -0,0 +1,195 @@
...
+	parent := node.Parent
+	if parent == nil || !ast.IsAssignmentExpression(parent, false) {
+		return
+	}
+	bin := parent.AsBinaryExpression()
+	if bin == nil || bin.Left != node {
+		return
+	}
+
+	if propName == "DEFAULT_TIMEOUT_INTERVAL" {
```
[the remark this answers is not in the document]
fansenze: Quick follow-up — looks like this went a step too far. By `return`ing early on non-`=` operators, the rule now silently skips `+=` etc. entirely instead of just dropping the fix. upstream still reports `illegalJasmine` here (its fix is the buggy part, but the report itself fires).

Verified:

```js
jasmine.DEFAULT_TIMEOUT_INTERVAL += 1000;
// PR:       no error (the test case at L30 even lists this as valid)
// upstream: reports `illegalJasmine`
```

To mirror upstream, the operator check should gate only the fix branch and fall through to the plain `ctx.ReportNode(node, illegalJasmine)`:

```go
if propName == "DEFAULT_TIMEOUT_INTERVAL" &&
    bin.OperatorToken != nil && bin.OperatorToken.Kind == ast.KindEqualsToken {
    right := ast.SkipParentheses(bin.Right)
    if right != nil && isEcmaPrimitiveLiteral(right.Kind) {
        // ... fix ...
        return
    }
}
ctx.ReportNode(node, buildIllegalJasmineMessage())
```

Also worth flipping the valid case on test L30 to invalid.

internal/plugins/jest/rules/no_jasmine_globals/no_jasmine_globals.go
```diff
@@ -0,0 +1,230 @@
...
+						if entryCount != 2 {
+							ctx.ReportNode(node, buildIllegalJasmineMessage())
+							return
+						}
+						ctx.ReportNodeWithFixes(
+							node,
+							buildIllegalMethodMessage(chain, "expect."+tail),
+							rule.RuleFixReplace(ctx.SourceFile, jasmineObj, "expect"),
+						)
+					case "addMatchers":
```
fansenze, line 218: Same gating gap as the matcher case right above — `addMatchers` and `createSpy` only check `tail`, so `jasmine.foo.addMatchers()` walks in and reports `illegalMethod` with chain `"jasmine.foo.addMatchers"`. Upstream's `functionName = calleeName.replace('jasmine.', '')` yields `"foo.addMatchers"`, which doesn't match, so it falls through to plain `illegalJasmine`.

Verified on `9856d814`:

```js
jasmine.foo.addMatchers()
// upstream: illegalJasmine
// PR:       illegalMethod ("`jasmine.foo.addMatchers` → `expect.extend`")

jasmine.foo.createSpy()
// upstream: illegalJasmine
// PR:       illegalMethod
```

Same `entryCount != 2` gate applies — easiest to fold these into the matcher case, or duplicate the check:

```go
case "addMatchers":
    if entryCount != 2 {
        ctx.ReportNode(node, buildIllegalJasmineMessage())
        return
    }
    ctx.ReportNode(node, buildIllegalMethodMessage(chain, "expect.extend"))
case "createSpy":
    if entryCount != 2 {
        ctx.ReportNode(node, buildIllegalJasmineMessage())
        return
    }
    ctx.ReportNode(node, buildIllegalMethodMessage(chain, "jest.fn"))
```

Worth pinning `jasmine.foo.addMatchers()` and `jasmine.foo.createSpy()` as lock-in tests.

internal/plugins/jest/rules/no_jasmine_globals/no_jasmine_globals.go
```diff
@@ -0,0 +1,230 @@
...
+	if len(entries) < 2 || entries[0].Name != "jasmine" || entries[0].Node == nil {
+		return "", nil, "", 0, false
+	}
+
+	rootParent := entries[0].Node.Parent
+	if rootParent == nil || rslintutils.AccessExpressionObject(rootParent) != entries[0].Node {
+		return "", nil, "", 0, false
+	}
+
+	return jestutils.JoinJestFnMemberEntries(entries), entries[0].Node, entries[len(entries)-1].Name, len(entries), true
```
fansenze, line 94: `GetJestFnMemberEntries` silently drops segments that `getElementAccessName` can't name (NumericLiteral, BinaryExpression, …), so `jasmine[1].any()` collapses to `[jasmine, any]` here, slips past the `entryCount == 2` gate, and produces an invalid auto-fix `expect[1].any()`.

Upstream's `isSupportedAccessor` rejects non-string/identifier accessors → `joinNames("jasmine", null)` → `null`, `getNodeName(callee)` → `null`, rule returns early. So upstream is silent here, PR produces a broken fix.

Verified:

```js
jasmine[1].any()
// upstream: silent (no report, no fix)
// PR:       illegalMethod, auto-fix → `expect[1].any()`  (invalid JS at runtime)
```

Same root cause as the `expect-expect` divergence in #950 (the F6 cluster). Local fix in `jasmineRootCallChain` — confirm chain depth matches actual AST depth so silently-dropped segments break the chain instead of being papered over:

```go
expected := len(entries)
actual := 1
for n := callee; ; {
    obj := rslintutils.AccessExpressionObject(n)
    if obj == nil { break }
    actual++
    n = obj
}
if actual != expected {
    return "", nil, "", 0, false
}
```

Longer term it'd be cleaner to have `GetJestFnMemberEntries` propagate "broken" instead of silent-skipping unsupported segments — that'd help every rule using it — but that's a wider audit. A test for `jasmine[1].any()` would lock this in either way.

fansenze (approved):

PR #972, opened 2026-05-19
ci: probe Windows runner disk usage (do not merge)

fansenze:
## Summary
- Temporary diagnostic PR. Adds two PowerShell steps to the `test-go-windows` job that print `Get-PSDrive` and the on-disk size of `GOCACHE`, the relocated `TEMP`, the checkout, and the entire `_work` root, both before and after the Unit Test step.
- Goal: decide whether the batched "delete test-generated cache files" workaround in `test-go-windows` (added in #696/#922 era to avoid disk-full) is still required now that `GOCACHE` is relocated to the work disk via `setup-go`.

## Why
Current Windows Test Go takes 20-39 minutes vs ~7 min on Ubuntu. Most of it is spent in batch 1 (~21 min) link/compile. We want to either:
- persist `GOCACHE` across runs via `lynx-infra/cache`, or
- drop the per-batch GOCACHE cleanup and run a single `go test ./internal/...`,

both of which raise peak disk usage. Need real numbers from the runner before deciding.

## Test plan
- [ ] CI runs the Windows Go job and the two `Disk usage` steps print PSDrive + folder sizes.
- [ ] Capture numbers from the run, then close/revert this PR — not for merge.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,6 +73,28 @@ jobs:
           go-version: ${{ matrix.go-version }}
           cache-name: ci-go-windows
 
+      - name: Disk usage (before tests)
+        shell: pwsh
+        run: |
+          Write-Host "=== PSDrive ==="
+          Get-PSDrive -PSProvider FileSystem | Format-Table -AutoSize
+
+          $paths = @(
+            $env:GOCACHE,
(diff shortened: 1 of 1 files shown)


fansenze:
Diagnostic data collected (job 76695688127). Closing in favor of follow-up PR that drops the batched workaround and switches to single-pass go test.

PR #1017, opened 2026-05-25
feat(eslint-plugin-jest): add`valid-expect` rule

eryue0220:
## Summary

* Port `valid-expect` from eslint-plugin-jest to rslint.

## Related Links

eslint-plugin-jest/valid-expect [doc](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/valid-expect.md) [code](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/valid-expect.ts)

## Checklist


internal/plugins/jest/rules/valid_expect/valid_expect.go
```diff
@@ -0,0 +1,468 @@
...
+	}
+
+	if len(fixes) > 0 {
+		ctx.ReportNodeWithFixes(descriptor.node, msg, fixes...)
+		return
+	}
+	ctx.ReportNode(descriptor.node, msg)
+}
+
+func reportUncalledMatcherMember(ctx rule.RuleContext, node *ast.Node) {
```
fansenze: `reportUncalledMatcherMember` fires for any member chain whose first segment is literally `expect`, even when `expect` itself was never called — upstream only has a `CallExpression` listener, so a bare member access is never visited.

```
expect.extend;    → upstream: no error | here: matcherNotCalled
expect.resolves;  → upstream: no error | here: matcherNotFound
```

Only `assertions`/`hasAssertions` are exempted, so every other bare `expect.x` is a false positive.

internal/plugins/jest/rules/valid_expect/valid_expect.go
```diff
@@ -0,0 +1,468 @@
...
+func isPromiseMethodCall(node *ast.Node) bool {
+	if node == nil || node.Kind != ast.KindCallExpression {
+		return false
+	}
+
+	callee := utils.CalleeChainName(node.AsCallExpression().Expression)
+	return strings.HasPrefix(callee, "Promise.")
+}
+
+func findPromiseCallExpressionNode(node *ast.Node) *ast.Node {
```
fansenze: `findPromiseCallExpressionNode` walks every ancestor up to the function boundary; upstream only inspects `node.parent` (and the grandparent for an `ArrayExpression`), so this latches onto a `Promise.*` call upstream would never reach.

```
Promise.all([ foo(expect(Promise.resolve(2)).resolves.not.toBeDefined()) ])
  upstream: asyncMustBeAwaited on the assertion
  here:     promisesWithAsyncAssertionsMustBeAwaited on Promise.all  (wrong node + messageId)
```

And `await Promise.all([ foo(expect(...)) ])` climbs to the awaited `Promise.all` and reports nothing, while upstream still flags the un-awaited inner assertion (missed report).

internal/plugins/jest/rules/valid_expect/valid_expect.go
```diff
@@ -0,0 +1,468 @@
...
+	Run: func(ctx rule.RuleContext, options any) rule.RuleListeners {
+		opts := parseOptions(options)
+		arrayExceptions := map[string]bool{}
+		asyncInserted := map[*ast.Node]bool{}
+		var descriptors []asyncDescriptor
+
+		return rule.RuleListeners{
+			ast.KindCallExpression: func(node *ast.Node) {
+				entries := utils.GetJestFnMemberEntries(node)
+				if !isExpectChain(entries, ctx.Settings) {
```
fansenze: `isExpectChain` matches the literal name "expect" before `ParseJestFnCall` runs, so aliased imports are never resolved. Other jest rules (prefer-to-contain, no-standalone-expect) support this alias.

```
import { expect as pleaseExpect } from '@jest/globals'; pleaseExpect(p).resolves.toBe(2)
  upstream: asyncMustBeAwaited
  here:     nothing (missed report)
```

internal/plugins/jest/rules/valid_expect/valid_expect.go
```diff
@@ -0,0 +1,468 @@
...
+					return
+				}
+				if len(entries) == 1 && node.Parent != nil && utils.IsMemberAccessNode(node.Parent) {
+					return
+				}
+
+				parsed := utils.ParseJestFnCall(node, ctx)
+				if parsed == nil {
```
fansenze, line 419: Here `parsed == nil` can't distinguish "not jest's expect" from "malformed expect chain": `isExpectChain` passes on the name, `ParseJestFnCall` returns nil for a non-jest `expect`, and this block re-derives a reason and reports anyway.

```
import { expect } from 'chai'; expect(foo).to.equal(bar)
  upstream: no error (resolveToJestFn returns null for a non-jest expect)
  here:     modifierUnknown (false positive)
```

This and the `isExpectChain` gate above would both be fixed by reusing a three-state parse entry (parsed / reason / not-jest-null) like upstream's `parseJestFnCallWithReason`, instead of a name-gate plus reporting-on-nil.

internal/plugins/jest/rules/valid_expect/valid_expect.go
```diff
@@ -0,0 +1,498 @@
...
+						ctx.ReportNode(reportNode, buildErrorModifierUnknownMessage())
+					}
+					return
+				}
+				if parsed.Kind != utils.JestFnTypeExpect {
+					return
+				}
+
+				expectCall := parsed.Head.Local.Node.Parent
+				if expectCall != nil && expectCall.Kind == ast.KindCallExpression {
```
fansenze: Missing upstream's early return: upstream does `if (expect.type !== CallExpression) return;` (valid-expect.ts:292), skipping **both** the arg and the async checks. Here only the arg check is gated, so the async check still runs when `expect` isn't called.

Verified false positives against eslint-plugin-jest@29.15.2:
```
expect.resolves.toBe(2)      -> here: asyncMustBeAwaited | upstream: (none)
expect.rejects.toBe(2)       -> here: asyncMustBeAwaited | upstream: (none)
expect.resolves.not.toBe(2)  -> here: asyncMustBeAwaited | upstream: (none)
```
Fix: `return` when `expectCall` is not a `CallExpression`.

internal/plugins/jest/rules/valid_expect/valid_expect.go
```diff
@@ -0,0 +1,498 @@
...
+							reportNode = topMember
+						}
+					}
+
+					switch reason {
+					case utils.ExpectParseReasonMatcherNotFound:
+						ctx.ReportNode(reportNode, buildErrorMatcherNotFoundMessage())
+					case expectParseReasonMatcherNotCalled:
+						entries := utils.GetJestFnMemberEntries(reportNode)
+						if len(entries) == 2 && (entries[1].Name == "assertions" || entries[1].Name == "hasAssertions") {
```
fansenze: This `assertions`/`hasAssertions` exemption is a false negative. The valid forms (`expect.hasAssertions`, `expect.hasAssertions()`) never reach this `matcher-not-called` branch, so the only thing it suppresses is `expect(x).assertions`, which upstream flags.

Verified against eslint-plugin-jest@29.15.2:
```
expect(x).assertions      -> here: (none) | upstream: matcherNotCalled @1:11-21
expect(x).hasAssertions   -> here: (none) | upstream: matcherNotCalled @1:11-24
```
Deleting these 4 lines keeps the full test suite green and restores the report. Fix: remove the exemption.

internal/plugins/jest/rules/valid_expect/valid_expect.go
```diff
@@ -0,0 +1,498 @@
...
+	return slices.Contains(asyncMatchers, parsed.Matcher)
+}
+
+func isPromiseMethodCall(node *ast.Node) bool {
+	if node == nil || node.Kind != ast.KindCallExpression {
+		return false
+	}
+
+	callee := utils.CalleeChainName(node.AsCallExpression().Expression)
+	return strings.HasPrefix(callee, "Promise.")
```
fansenze: `HasPrefix(callee, "Promise.")` over-matches. Upstream requires the callee object to be exactly `Promise` (valid-expect.ts:39, `isSupportedAccessor(node.callee.object, 'Promise')`), so a deeper chain like `Promise.all.x(...)` is not a promise wrap.

Verified against eslint-plugin-jest@29.15.2:
```
Promise.all.x([expect(...)])
  here:     promisesWithAsyncAssertionsMustBeAwaited on the outer call (+ wrong autofix)
  upstream: asyncMustBeAwaited on the inner assertion
```
Fix: check the callee's member object equals `Promise`, not a string prefix.

internal/plugins/jest/rules/valid_expect/valid_expect.go
```diff
@@ -0,0 +1,498 @@
...
+			return core.NewTextRange(i, i+1)
+		}
+	}
+
+	return internalUtils.TrimNodeTextRange(sourceFile, call).WithEnd(start)
+}
+
+func tooManyArgsRange(sourceFile *ast.SourceFile, args []*ast.Node, maxArgs int) core.TextRange {
+	start := internalUtils.TrimNodeTextRange(sourceFile, args[maxArgs]).Pos()
+	end := internalUtils.TrimNodeTextRange(sourceFile, args[len(args)-1]).End()
```
fansenze, line 321: `tooManyArgs` end column is +1 vs upstream, which subtracts one (`end.column - 1`, valid-expect.ts:325).

Verified against eslint-plugin-jest@29.15.2:
```
expect("a","b").toEqual(1)        -> upstream endCol 26, here 27
3 args, maxArgs:2                -> upstream endCol 38, here 39
```
The Go and TS tests hard-code the +1 values, so they pass but don't match upstream. (Start column is correct, so this is only the missing `-1`, not a UTF-16 issue.) Fix: drop one position from the end.

fansenze (commented):
Reviewed against the real eslint-plugin-jest@29.15.2 output (ran the upstream rule). 4 alignment divergences below — 2 false positive/negative, 1 wrong report+autofix, 1 off-by-one. Your 4 earlier comments are all genuinely fixed (resolving them).

fansenze (approved):

PR #1019, opened 2026-05-25
feat(eslint-plugin-jest): add `prefer-to-have-been-called` rule

eryue0220:
## Summary

* Port `prefer-to-have-been-called` from eslint-plugin-jest to rslint.

## Related Links

eslint-plugin-jest/prefer-to-have-been-called [doc](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/prefer-to-have-been-called.md) [code](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/prefer-to-have-been-called.ts)

## Checklist

### DIFF
--- a/internal/plugins/jest/rules/prefer_to_have_been_called/prefer_to_have_been_called.go
+++ b/internal/plugins/jest/rules/prefer_to_have_been_called/prefer_to_have_been_called.go
@@ -77,12 +77,8 @@ var PreferToHaveBeenCalledRule = rule.Rule{
 					jestFnCall.MatcherEntry.Node,
 					buildPreferMatcherErrorMessage(),
 					rule.RuleFixReplaceRange(
-						core.NewTextRange(matcherCall.Arguments.Pos(), matcherCall.Arguments.End()),
-						"",
-					),
-					rule.RuleFixReplaceRange(
-						core.NewTextRange(replaceStart, matcherParent.End()),
-						replacementMatcher,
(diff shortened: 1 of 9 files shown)


internal/plugins/jest/rules/prefer_to_have_been_called/prefer_to_have_been_called.go
```diff
@@ -0,0 +1,87 @@
...
+
+func buildPreferMatcherErrorMessage() rule.RuleMessage {
+	return rule.RuleMessage{
+		Id:          "preferMatcher",
+		Description: "Use `toHaveBeenCalled`",
+	}
+}
+
+func isZeroLiteral(node *ast.Node) bool {
+	node = jestUtils.UnwrapTypeAssertions(node)
```
fansenze: Nit worth fixing: this should be `UnwrapBasicTypeAssertions`. Upstream's `getFirstMatcherArg` → `followTypeAssertionChain` only unwraps `as` and `<T>` assertions — not `!` or `satisfies` — so these stay unflagged upstream but get reported (and autofixed) here:

```js
expect(method).toBeCalledTimes(0!);
expect(method).toBeCalledTimes(0 satisfies number);
```

`prefer_to_be` uses the same upstream helper and already calls `UnwrapBasicTypeAssertions`, so this keeps the two rules consistent.

fansenze (approved):

PR #1026, opened 2026-05-26
feat(eslint-plugin-jest): add `no-duplicate-hooks` rule

eryue0220:
## Summary

* Port `no-duplicate-hooks` from eslint-plugin-jest to rslint.

## Related Links

eslint-plugin-jest/no-duplicate-hooks [doc](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-duplicate-hooks.md) [code](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/no-duplicate-hooks.ts)

## Checklist


packages/rslint-test-tools/rstest.config.mts
```diff
@@ -439,7 +439,7 @@ export default defineConfig({
     './tests/eslint-plugin-jest/rules/no-mocks-import.test.ts',
     './tests/eslint-plugin-jest/rules/no-standalone-expect.test.ts',
     './tests/eslint-plugin-jest/rules/no-test-prefixes.test.ts',
-    './tests/eslint-plugin-jest/rules/prefer-called-with.test.ts',
+    './tests/eslint-plugin-jest/rules/no-duplicate-hooks.test.ts',
```
fansenze, line 442: This **replaces** `prefer-called-with.test.ts` instead of adding a new line, so `prefer-called-with.test.ts` is dropped from the `include` list and no longer runs in CI — the file still exists and `PreferCalledWithRule` is still registered in `all.go`. Please keep both: add `no-duplicate-hooks.test.ts` on a new line and restore `prefer-called-with.test.ts`.

fansenze (commented):
Rule logic looks correct and faithfully ported (verified against the upstream eslint-plugin-jest rule; the Go test passes and asserts line/column/messageId). One blocking issue inline.

packages/rslint-test-tools/rstest.config.mts
eryue0220: Updated.

fansenze (approved):

PR #1029, opened 2026-05-27
fix(eslint-plugin-runner): await stdio pipe close before terminating workers

fansenze:
## Summary

On `windows-latest`, terminating a worker thread while its stdio pipes are still live lets libuv's concurrent pipe teardown fault **below the JS layer** (a native abort). Because rstest's forked test child intercepts `uncaughtException` / `unhandledRejection` / `process.exit`, the only way that child can "exit unexpectedly" is a native fault — so this surfaced as `Worker exited unexpectedly` in the high-terminate-churn `worker-pool-e2e` suite (e.g. `multiconfig`, `resilience`), a flaky that has been failing `main` CI.

`terminateWorker` already destroyed the pipes first, but `destroy()` only *initiates* the close (it completes on a later tick), so a `terminate()` called right after it still races the in-flight teardown. This PR awaits each pipe's `'close'` — bounded by a 1s grace so a wedged worker can't stall — before terminating. The two are now serialized: by the time we terminate, the pipes are already gone and there is no concurrent teardown left to fault. `terminateWorker` is only ever called from cold paths (init/task timeout, crash respawn, shutdown grace), so normal lint throughput is unaffected.

It also routes the tests' raw `worker.terminate()` calls through `terminateWorker`, so they exercise the same safe teardown as the production timeout / crash / shutdown paths — production never bare-terminates a worker.

**Verification note:** the native abort only reproduces on `windows-latest`. Locally (macOS) all 554 tests pass but cannot exercise the abort, so effectiveness needs to be confirmed across windows CI runs.

## Related Links

- Failing run: https://github.com/web-infra-dev/rslint/actions/runs/26490526197/job/78007746866

## Checklist

### DIFF
--- a/packages/eslint-plugin-runner/src/worker-pool.ts
+++ b/packages/eslint-plugin-runner/src/worker-pool.ts
@@ -170,7 +170,43 @@ const WORKER_FILE = (() => {
 const WORKER_EXIT_GRACE_MS = 5_000;
 
 /**
- * Terminate a worker, closing its piped stdout/stderr FIRST.
+ * Upper bound on how long `terminateWorker` waits for a piped stream to
+ * emit `'close'` after `destroy()` before giving up and terminating
+ * anyway. A normal close lands within a tick; this only guards a wedged
+ * worker whose stream never closes, so teardown can never stall.
+ */
+const PIPE_CLOSE_GRACE_MS = 1_000;
+
+/**
+ * Destroy one worker stdio pipe and resolve once it has ACTUALLY closed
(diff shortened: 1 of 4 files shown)


fansenze:
Superseded by #1031. The pipe-teardown approach here didn't address the root cause: the windows abort comes from tearing down a worker thread that has oxc (a napi addon) loaded (nodejs/node#34567), not from stdio pipe teardown. #1031 skips the windows-flaky worker-pool e2e suites instead — a known worker_threads + napi limitation; a real fix would be process isolation (child_process), tracked separately.

PR #1032, opened 2026-05-28
feat(eslint-plugin-jest): add `prefer-each` rule

eryue0220:
## Summary

* Port `prefer-each` from eslint-plugin-jest to rslint

## Related Links

eslint-plugin-jest/prefer-each [doc](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/prefer-each.md) [code](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/prefer-each.ts)

## Checklist


internal/plugins/jest/rules/prefer_each/prefer_each.go
```diff
@@ -0,0 +1,172 @@
...
+				if jestFnCall == nil {
+					return
+				}
+
+				switch jestFnCall.Kind {
+				case jestUtils.JestFnTypeHook,
+					jestUtils.JestFnTypeDescribe,
+					jestUtils.JestFnTypeTest:
+					if !isInsideTestCase(node) {
+						for i := range loops {
```
fansenze: **Nested/sibling loops diverge from upstream (over-reporting).**

This appends each jest call to *every* enclosing loop on the stack, so each one reports on exit (L121). Upstream keeps a single global `jestFnCalls` and resets it on *both* loop-enter and after reporting, so a spot reports once — the innermost loop only. Verified by running both:

| code | upstream | this PR |
|------|----------|---------|
| `for a { for b { it } }` | 1 `it` | 2 `it` (dup) |
| `for a { it(); for b { nonJest } }` | 0 | 1 |

`for{for{it}}` (e.g. a browser×os matrix) is real code, so users get two warnings for one `.each`.

A 1:1 port of the upstream algorithm is simpler and correct — `ListenerOnExit` is enough, so the loops stack / ranges / `isInsideTestCase` can go:

```go
jestFnCalls := []jestUtils.JestFnType{}
inTestCaseCall := false
enter := func(*ast.Node){ if len(jestFnCalls)==0 || inTestCaseCall {return}; jestFnCalls = jestFnCalls[:0] }
exit  := func(n *ast.Node){ if len(jestFnCalls)==0 || inTestCaseCall {return}; ctx.ReportNode(n, buildPreferEachMessage(recommendFn(jestFnCalls))); jestFnCalls = jestFnCalls[:0] }
// CallExpression:      hook/describe/test → push; test → inTestCaseCall=true
// CallExpression:exit: test → inTestCaseCall=false
```

internal/plugins/jest/rules/prefer_each/prefer_each_test.go
```diff
@@ -0,0 +1,367 @@
...
+				Errors: []rule_tester.InvalidTestCaseError{
+					{
+						MessageId: "preferEach",
+						Message:   "prefer using `it.each` rather than a manual loop",
+					},
+				},
+			},
+			{
+				Code: `
+        for (const suite of suites) {
```
fansenze, line 220: Asserts 2 `it`, but upstream reports only **1** (innermost loop) — the extra one is the outer-loop over-report from `prefer_each.go:151`. `prefer-each.test.ts` has no such case (it mirrors upstream's 13 invalid + 5 valid). After fixing L151, set this back to 1 `it` or drop the case.

internal/plugins/jest/rules/prefer_each/prefer_each_test.go
```diff
@@ -0,0 +1,367 @@
...
```
fansenze: Same divergence: asserts 1 `it`, but upstream reports **0** (it clears the accumulator on inner-loop enter). Also absent from upstream and from `prefer-each.test.ts`. Align with the fix above.

internal/plugins/jest/rules/prefer_each/prefer_each_test.go
```diff
@@ -0,0 +1,367 @@
...
+      it("only returns numbers that are greater than seven", function () {
+        const numbers = getNumbers();
+
+        for (let i = 0; i < numbers.length; i++) {
+          expect(numbers[i]).toBeGreaterThan(7);
+        }
+      });
+    `},
+			{Code: `
+      test("outer", () => {
```
fansenze: This valid case is the opposite of upstream: upstream reports **1 `it`** here (its `inTestCaseCall` boolean is reset to false on exiting `inner`). Contrived (jest doesn't nest test-in-test) and this PR is arguably more correct, but it contradicts upstream silently and isn't in upstream or `prefer-each.test.ts`. If intentional, add a comment.

fansenze (commented):
Port looks solid overall — non-nested cases match upstream exactly. A few alignment issues show up with nested/sibling loops, verified by running both this rule and upstream `eslint-plugin-jest` on the same snippets. Details inline.

fansenze (approved):

PR #1034, opened 2026-05-29
refactor: remove @stylistic plugin and related rules

fansenze:
## Summary

The `@stylistic` plugin ports purely stylistic / formatting rules — spacing, brace style, JSX layout, and so on. Going forward we plan to handle this kind of formatting work in a dedicated formatter rather than shipping it as lint rules. Removing the `@stylistic` plugin now keeps a clean separation between linting and formatting, and avoids maintaining rules that the upcoming formatter will supersede.

This PR removes the `@stylistic` plugin entirely:

- Delete `internal/plugins/stylistic/` (21 rules + `stylisticutil`) and its `@stylistic` registration in `internal/config/config.go`.
- Clean up the react-side reuse leftovers that existed only to produce `@stylistic` variants: drop the `stylisticQuotes` / `stylisticScope` parameters and their now-dead branches in `jsx-curly-brace-presence` / `jsx-curly-spacing`, and the stale `@stylistic` comments in the five shared jsx rules. The react rules keep their own behavior unchanged (verified identical).
- Remove the `@rslint/core` public surface: the `stylisticPlugin` export, `configs/stylistic.ts` preset, and the `'@stylistic'` member of `KnownPlugin`. **This is a breaking change** for any config that referenced `@stylistic`.
- Sync docs (`plugin-registry.ts`, `QUICK_REFERENCE.md`) and remove the stylistic test suite (Go + rstest white-list).

### Verification

- `go build ./...`, targeted `go test` (config + the 5 affected react rule packages), `golangci-lint` (0 issues), `tsgo` typecheck, dogfood `pnpm lint` (0 errors), and rstest e2e for the 5 react rules — all pass.
- Validated the built binary against the rsbuild and rspack repos (normal + `--type-check`): output is identical to baseline `@rslint/core@0.5.3` (0 errors / 0 warnings), no crashes, and no performance regression attributable to this change (before-HEAD ≈ after-removal across all four scenarios).

## Related Links

N/A

## Checklist


chenjiahan (approved):

eryue0220:
Just a question, so will have another tool called `rstylic` or `rsformatter`?

fansenze:
@eryue0220 We're still discussing internally whether it makes sense to add a new formatter. For the Stylistic ESLint plugin, though, our recommendation would be to run it via the upcoming `eslintPlugins` support

chenjiahan:
@eryue0220 Since Rslint already supports linting and type checking, we’re considering evolving it into an all-in-one code quality tool, with built-in formatting support rather than providing ESLint’s legacy stylistic rules.

That said, we’re not looking to build a new formatter from scratch. Instead, we’re exploring integrations with existing community formatters such as Oxfmt, Prettier, Biome, or Dprint.

PR #1047, opened 2026-06-03
feat: support mounting ESLint plugins in rslint config

fansenze:
## Summary

Lets you mount community ESLint plugins in the rslint flat config via the **object form** of `plugins` (`plugins: { example: examplePlugin }`), so their rules run alongside the native ones. The **array form** (`plugins: ['@typescript-eslint']`) stays the built-in native-plugin whitelist; a single entry uses one form — to combine native and community plugins, declare them in separate config entries.

- Plugin JS rules run in a Node worker pool; native rules run in Go.
- Go stays the driver — file discovery, per-file config matching, diagnostic merge and `--fix` — and reverse-dispatches plugin-rule batches (concurrently) to the worker over the CLI IPC channel and the LSP `rslint/pluginLint` request.
- Per-file routing: each dispatched file carries a `configKey` so the worker runs it against the exact config's loaded plugins (monorepo / multi-config supported).
- `--singleThreaded` sizes the plugin worker pool to a single worker, matching the native pass.
- Works end to end: CLI (including `--fix`) and the editor (diagnostics, quick fixes, `source.fixAll`).

See the docs for which ESLint plugin APIs are supported vs. not (type information, custom parser, processors, code-path analysis).

## Checklist


packages/rslint/src/plugin-source.ts
```diff
@@ -0,0 +1,61 @@
...
+ * write `plugins: { uc: unicornPlugin }`). Array-form `plugins` (the
+ * JSON-config name whitelist) is ignored — it carries no live objects.
+ * Returns null when neither yields a plugin object map.
+ */
+export function selectPluginSource(
+  entry: unknown,
+): Record<string, unknown> | null {
+  if (entry == null || typeof entry !== 'object') return null;
+  const e = entry as { eslintPlugins?: unknown; plugins?: unknown };
+  if (
```
fansenze: **An empty `eslintPlugins: {}` shadows an object-form `plugins`, loading zero plugins.**

`selectPluginSource` returns any non-null non-array object, including `{}`. So a config entry with `eslintPlugins: {}` *and* `plugins: { uc: unicornPlugin }` returns `{}` — both the host normalizer (`config-loader.ts:110`) and the worker loader (`plugin-loader.ts:217`) then iterate an empty map and load nothing. The `uc/*` rules are afterwards silently dropped by the plugin gate (`internal/config/rule_registry.go:56`, since the prefix was never registered) — no error, just missing diagnostics.

The documented intent is "explicit `eslintPlugins` wins over object-form `plugins`", but an *empty* `eslintPlugins` shouldn't win. Suggest requiring `Object.keys(e.eslintPlugins).length > 0` before returning it, so an empty value falls through to `plugins`.

packages/rslint/src/eslint-plugin/host.ts
```diff
@@ -0,0 +1,61 @@
...
+ * `configs` empty ⇒ the pool spawns no workers (no-op fast path); a
+ * `lint` call then returns empty per-file results. Init rejects if a
+ * referenced plugin fails to import — the caller decides how loud to be
+ * (CLI fails the run; LSP logs and serves empty).
+ */
+export async function createPluginLintHost(
+  configs: ConfigDescriptor[],
+  onLog?: WorkerPoolOptions['onLog'],
+): Promise<PluginLintHost> {
+  const pool = new WorkerPool({ configs, onLog });
```
fansenze: **`--singleThreaded` is never honored by the plugin worker pool.**

`createPluginLintHost` constructs `new WorkerPool({ configs, onLog })` with no `workerCount`, so the pool always defaults to `min(cpus, 8)` (`worker-pool.ts:273`). Nothing plumbs the flag down: `cli.ts:86` calls `runEngine` with no `runtime` block, and `engine.ts:149` passes no `workerCount`. The `WorkerPoolOptions.workerCount` JSDoc ("1 honors --singleThreaded", `worker-pool.ts:62`) is therefore a dead contract in production — only tests set it.

Thread `runtime.singleThreaded` from the CLI through `runEngine` → `createPluginLintHost` → `WorkerPool({ workerCount: 1 })` so a single-threaded run is actually single-threaded.

internal/lsp/service.go
```diff
@@ -581,6 +576,58 @@ func (s *Server) handleFixAllCodeAction(ctx context.Context, uri lsproto.Documen
...
+		if err != nil {
+			log.Printf("Error running lint for fixAll pass %d: %v", pass, err)
+			break
+		}
+
+		// Fold in eslint-plugin fixes so source.fixAll applies plugin rule fixes
+		// too, not just native. The plugin pass lints the SAME currentContent, so
+		// its fix byte offsets align with ApplyRuleFixes's input; suggestionsMode
+		// is "off" because fixAll applies only autofixes.
+		if pluginDiags := s.lintPluginRulesSync(ctx, uri, currentContent, true, linter.SuggestionsModeOff); len(pluginDiags) > 0 {
```
fansenze: **`source.fixAll` blocks the dispatch loop on a synchronous server→client reverse request, once per fix pass, with no Go-side deadline.**

`computeFixAllContent` calls `lintPluginRulesSync` inside the per-pass loop (this line), which drives an `s.sendRequest(...)` reverse request and blocks the dispatch loop until the client answers. In the normal case this is bounded — the worker's 30s task-timeout (`worker-pool.ts:275`) and cancellation of the originating `codeAction` request both unblock it. But there is no independent timeout on the reverse request itself: if the client never services it (extension wedged, or pool mid-rebuild) and the request context isn't cancelled, the dispatch loop — and thus all editor interaction — stalls.

Consider bounding the reverse request with a context deadline here so a non-responsive client degrades to native-only fixes instead of freezing the LSP.

packages/rslint/src/engine.ts
```diff
@@ -198,6 +257,7 @@ export async function runEngine(opts: EngineRunOptions): Promise<number> {
   const finalExit = await childExit;
   ipc.close();
   removeSignalHandlers();
+  await pluginHost?.shutdown();
```
fansenze: **The init-failure early-exits don't shut the plugin pool down — only this normal path does.**

Once the pool is created (lines 133–158), four exit paths return without `pluginHost.shutdown()`: `return 0` (line 230), `return state.code` (line 232), `return outcome.state.code` (line 236), and `return 2` (line 251). Only this line and the signal handler (line 168) drain it.

This does **not** hang the CLI — `bin/rslint.cjs:45` calls `process.exit()`, which reaps the worker threads. The issue is an asymmetric teardown: those paths skip the graceful drain this one performs. Wrapping the post-creation body in `try { … } finally { await pluginHost?.shutdown() }` would make teardown uniform across every exit.

internal/lsp/eslint_plugin.go
```diff
@@ -0,0 +1,263 @@
...
+// Must be called from the main dispatch loop (it reads jsConfigs + documents
+// to build the input and the generation map).
+//
+// Follow-up (not needed for correctness): when a newer keystroke supersedes an
+// in-flight dispatch we currently rely on generation to drop the stale result,
+// but the Node worker keeps running until it finishes. Sending the client a
+// `$/cancelRequest` for the superseded reverse-request id (s.sendRequest mints
+// "ts%d" ids) would free that worker CPU sooner. It only saves work, never
+// changes the published diagnostics, so it is deferred.
+func (s *Server) dispatchPluginLint(uri lsproto.DocumentUri, generation uint64) {
```
fansenze, line 142: **Plugin quick-fixes can be missing for a `codeAction` that arrives before this async result is merged.**

`pushDiagnostics` stores + publishes native diagnostics synchronously (`service.go:1044`/`1052`), then dispatches the plugin pass here in a goroutine; the result merges back later via `pluginResultCh` (`mergePluginDiagnostics`). A `codeAction` landing in that window reads `s.diagnostics[uri]` (`service.go:449`) with native-only entries, so plugin fixes/suggestions are absent. The window is narrow and self-heals on the next request, and `source.fixAll` is unaffected (it re-lints plugin rules synchronously).

Flagging mostly so it's a conscious known-limitation rather than a surprise; no fix strictly required.

internal/linter/eslint_plugin.go
```diff
@@ -0,0 +1,473 @@
...
+	}
+}
+
+// optionsToArray normalizes rslint's RuleConfig.Options (nil / single value
+// / array) into ESLint's options array (context.options).
+func optionsToArray(opts any) []any {
+	switch v := opts.(type) {
+	case nil:
+		return []any{}
+	case []any:
```
fansenze, line 357: **(latent) A rule whose single option is itself an array loses one nesting level on the plugin-dispatch path.**

`parseArrayRuleConfig` (`internal/config/config.go:248`) unwraps a lone post-severity element: `["error", ["a","b"]]` → `Options = ["a","b"]`. `optionsToArray` (this function) then hits the `[]any` branch and returns it unchanged, and the worker builds `context.options` per-element (`linter/options-defaults.ts:58`) — so the rule sees `options[0] === "a"` instead of `["a","b"]`. (Native rules are unaffected: they consume the raw `Options` through their `Run` closure, `rule_registry.go:71`.)

No fixture or conformance case in this PR exercises an array-valued single option, and such rules are uncommon, so this is **latent** rather than actively breaking — but the conformance harness can't catch it (it would only surface as an ESLint-vs-rslint mismatch if such a case existed). Fixing it cleanly means preserving whether the original option was a single array vs. a multi-element list rather than re-deriving it from the already-folded `Options`.

chenjiahan (approved):

PR #1048, opened 2026-06-03
feat(eslint-plugin-jest): add `prefer-to-have-been-called-times` rule

eryue0220:
## Summary

* Port `prefer-to-have-been-called-times` from `eslint-plugin-jest` to `rslint`

## Related Links

eslint-plugin-jest/prefer-to-have-been-called-times [doc](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/prefer-to-have-been-called-times.md) [code](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/prefer-to-have-been-called-times.ts)

## Checklist

### DIFF
--- a/internal/plugins/jest/rules/prefer_to_have_been_called_times/prefer_to_have_been_called_times.go
+++ b/internal/plugins/jest/rules/prefer_to_have_been_called_times/prefer_to_have_been_called_times.go
@@ -1,0 +1,130 @@
+package prefer_to_have_been_called_times
+
+import (
+	"github.com/microsoft/typescript-go/shim/ast"
+	"github.com/microsoft/typescript-go/shim/core"
+	jestUtils "github.com/web-infra-dev/rslint/internal/plugins/jest/utils"
+	"github.com/web-infra-dev/rslint/internal/rule"
+)
+
+func buildPreferMatcherMessage() rule.RuleMessage {
+	return rule.RuleMessage{
+		Id:          "preferMatcher",
+		Description: "Use `toHaveBeenCalledTimes()`",
+	}
+}
+
+func unwrapMockProperty(mockExpr *ast.Node) *ast.Node {
+	if mockExpr == nil {
+		return nil
+	}
+	mockExpr = ast.SkipParentheses(mockExpr)
+	switch mockExpr.Kind {
+	case ast.KindPropertyAccessExpression:
+		if ast.IsOptionalChain(mockExpr) {
+			return nil
+		}
+		pa := mockExpr.AsPropertyAccessExpression()
+		if !jestUtils.IsNamedMember(pa.Name(), "mock") {
+			return nil
+		}
+		return pa.Expression
+	case ast.KindElementAccessExpression:
+		if ast.IsOptionalChain(mockExpr) {
+			return nil
+		}
+		el := mockExpr.AsElementAccessExpression()
+		if !jestUtils.IsNamedMember(ast.SkipParentheses(el.ArgumentExpression), "mock") {
+			return nil
+		}
+		return el.Expression
+	default:
+		return nil
+	}
+}
+
+// unwrapMockCallsAccessProperty returns the mock function expression when arg is
+// exactly `*.mock.calls` or `*["mock"].calls` (no indexing after `.calls`).
+func unwrapMockCallsAccessProperty(arg *ast.Node) *ast.Node {
+	if arg == nil {
+		return nil
+	}
+	switch arg.Kind {
+	case ast.KindPropertyAccessExpression:
+		if ast.IsOptionalChain(arg) {
+			return nil
+		}
+		pa := arg.AsPropertyAccessExpression()
+		if !jestUtils.IsNamedMember(pa.Name(), "calls") {
+			return nil
+		}
+		return unwrapMockProperty(pa.Expression)
+	case ast.KindElementAccessExpression:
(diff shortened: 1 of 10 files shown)


eryue0220:
Has the gemini reviewer been removed?

packages/rslint-test-tools/rstest.config.mts
```diff
@@ -445,6 +445,7 @@ export default defineConfig({
     './tests/eslint-plugin-jest/rules/prefer-strict-equal.test.ts',
     './tests/eslint-plugin-jest/rules/prefer-to-be.test.ts',
     './tests/eslint-plugin-jest/rules/prefer-to-contain.test.ts',
+    './tests/eslint-plugin-jest/rules/prefer-to-have-been-called-times.test.ts',
```
fansenze, line 448: **Filename mismatch — this test never runs.** The file on disk is `prefer-to-have-been-call-times.test.ts` (`call-times`), but this entry says `called-times`. So the real file is never picked up by `include`, and this line points at a path that does not exist.

`rstest run eslint-plugin-jest/rules` → `testFiles: 25` (should be 26): the phantom entry is silently skipped, so CI stays green.

Fix: rename the test file to `prefer-to-have-been-called-times.test.ts` (matches the rule name everywhere else — Go package, rule `Name`, `.md`, `all.go`). Keep this config line as-is.

internal/plugins/jest/rules/prefer_to_have_been_called_times/prefer_to_have_been_called_times.go
```diff
@@ -0,0 +1,130 @@
...
+				if expectCall == nil || expectCall.Kind != ast.KindCallExpression {
+					return
+				}
+
+				args := expectCall.Arguments()
+				if len(args) == 0 {
+					return
+				}
+
+				inner := unwrapMockCallsAccessProperty(args[0])
```
fansenze: **False negative vs upstream on a parenthesized argument.**

```
input:  expect((fn.mock.calls)).toHaveLength(1);
eslint-plugin-jest@29   reports → expect((fn)).toHaveBeenCalledTimes(1)  (col 25)
this rule               not reported (the test asserts it valid)
```

Root cause: no top-level `SkipParentheses` here, unlike the sibling `unwrapLengthAccessProperty` in `prefer_to_have_length.go`.

Fix (passes Go test + upstream differential):

```diff
- inner := unwrapMockCallsAccessProperty(args[0])
+ unwrapped := ast.SkipParentheses(args[0])
+ inner := unwrapMockCallsAccessProperty(unwrapped)
  ...
- rule.RuleFixRemoveRange(core.NewTextRange(inner.End(), args[0].End())),
+ rule.RuleFixRemoveRange(core.NewTextRange(inner.End(), unwrapped.End())),
```

`args[0].End()` would eat the closing paren (`args[0]` is the `ParenthesizedExpression`); `unwrapped.End()` stops before `)`. Then flip the `expect((fn.mock.calls)).toHaveLength(1)` case from valid to invalid.

fansenze: Detection's fixed, but the fix now drops the outer parens — upstream keeps them:

```
in:       expect((method.mock.calls)).toHaveLength(1);
upstream: expect((method)).toHaveBeenCalledTimes(1);   (eslint-plugin-jest@29)
this PR:  expect(method).toHaveBeenCalledTimes(1);
```

`TrimNodeTextRange(args[0])` spans both parens, so replacing it with `inner` text eats them. Remove-only instead:

```go
rule.RuleFixRemoveRange(core.NewTextRange(inner.End(), ast.SkipParentheses(args[0]).End())),
```

then flip the parenthesized case to `expect((method))...` in both _test.go and .test.ts.

internal/plugins/jest/rules/prefer_to_have_been_called_times/prefer_to_have_been_called_times.go
```diff
@@ -0,0 +1,140 @@
...
+				if inner == nil {
+					return
+				}
+
+				matcherCall := node.AsCallExpression()
```
fansenze: Dead code: `matcherCall` is unused after this guard, and a CallExpression's `.Arguments` is never nil here. Drop these 4 lines.

internal/plugins/jest/rules/prefer_to_have_been_called_times/prefer_to_have_been_called_times.go
```diff
@@ -0,0 +1,140 @@
...
+	"github.com/microsoft/typescript-go/shim/core"
+	jestUtils "github.com/web-infra-dev/rslint/internal/plugins/jest/utils"
+	"github.com/web-infra-dev/rslint/internal/rule"
+	rslintUtils "github.com/web-infra-dev/rslint/internal/utils"
+)
+
+func buildPreferMatcherMessage() rule.RuleMessage {
+	return rule.RuleMessage{
+		Id:          "preferMatcher",
+		Description: "Use `toHaveBeenCalledTimes()`",
```
fansenze: nit: upstream message text is "Prefer `toHaveBeenCalledTimes`", but this uses "Use `toHaveBeenCalledTimes()`". messageId matches so tests pass; only the displayed text diverges.

fansenze (approved):

Issue #474, opened 2026-02-12
[Tracking]: eslint-plugin-promise rules


fansenze:
This is a tracking issue for porting the existing rules from [`eslint-plugin-promise`](https://github.com/eslint-community/eslint-plugin-promise?tab=readme-ov-file#rules).

There are **17** rules.

| Category | ✅ Done | 🚫 Won't | Total |
| :--- | :---: | :---: | :---: |
| Recommended | 5 | 0 | 5 |
| Not recommended | 12 | 0 | 12 |
| Deprecated | 0 | 0 | 0 |
| **Total** | **17** | **0** | **17** |

> ✅ = Implemented &nbsp;|&nbsp; 🚫 = Not intending to implement

---

## Recommended rules

<details open>
<summary>✅ 5 &nbsp;|&nbsp; 🚫 0 &nbsp;/&nbsp; Total: 5</summary>

| Status | Name |
| :----: | :--- |
| ✅ | [promise/always-return](https://github.com/eslint-community/eslint-plugin-promise/blob/main/docs/rules/always-return.md) |
| ✅ | [promise/catch-or-return](https://github.com/eslint-community/eslint-plugin-promise/blob/main/docs/rules/catch-or-return.md) |
| ✅ | [promise/no-new-statics](https://github.com/eslint-community/eslint-plugin-promise/blob/main/docs/rules/no-new-statics.md) |
| ✅ | [promise/no-return-wrap](https://github.com/eslint-community/eslint-plugin-promise/blob/main/docs/rules/no-return-wrap.md) |
| ✅ | [promise/param-names](https://github.com/eslint-community/eslint-plugin-promise/blob/main/docs/rules/param-names.md) |

</details>

---

## Not recommended rules

<details open>
<summary>✅ 12 &nbsp;|&nbsp; 🚫 0 &nbsp;/&nbsp; Total: 12</summary>

| Status | Name |
| :----: | :--- |
| ✅ | [promise/avoid-new](https://github.com/eslint-community/eslint-plugin-promise/blob/main/docs/rules/avoid-new.md) |
| ✅ | [promise/no-callback-in-promise](https://github.com/eslint-community/eslint-plugin-promise/blob/main/docs/rules/no-callback-in-promise.md) |
| ✅ | [promise/no-multiple-resolved](https://github.com/eslint-community/eslint-plugin-promise/blob/main/docs/rules/no-multiple-resolved.md) |
[...]

PR #1049, opened 2026-06-04
feat: add promise/always-return rule

swwind:
## Summary

Ported promise/always-return rule for #474

## Related Links

[original rule link](https://github.com/eslint-community/eslint-plugin-promise/blob/main/docs/rules/always-return.md)

## Checklist

## Notes

This rule assumes node.js environment and considers `process.exit()` and `process.abort()` to be the same as return statements. This might introduce some problems when linting other targets (like dom).


internal/plugins/promise/rules/always_return/always_return.go
```diff
@@ -0,0 +1,294 @@
...
+				if len(opts.IgnoreAssignmentVariable) > 0 && isLastCallback(node) {
+					body := cb.Body()
+					if body != nil && body.Kind == ast.KindBlock {
+						for _, stmt := range body.Statements() {
+							if isIgnoredAssignment(stmt, opts.IgnoreAssignmentVariable) {
+								return
+							}
+						}
+					}
+				}
```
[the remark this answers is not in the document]
swwind: This is not aligned with the original behavior. We only need to port existing old rules right now.

internal/plugins/promise/rules/always_return/always_return.go
```diff
@@ -0,0 +1,294 @@
...
+		}
+	}
+	return false
+}
+
+func stmtTerminates(s *ast.Node) bool {
+	if s == nil {
+		return false
+	}
+	switch s.Kind {
```
fansenze: **False positives.** `stmtTerminates` only recognizes return / throw / process.exit|abort / Block / if-else. A callback that terminates via `switch` / `try` / `while` / `for` / `do` falls through to `return false` and is reported. Upstream uses ESLint code-path analysis and accepts all of these.

Differential (eslint-plugin-promise@7.3.0 vs this rule):

```
switch(x){case 1:return 1;default:return 2}   upstream VALID -> REPORTS
try{return 1}catch(e){return 2}               upstream VALID -> REPORTS
while(true){return 1}                         upstream VALID -> REPORTS
for(;;){return 1}                             upstream VALID -> REPORTS
do{return 1}while(x)                          upstream VALID -> REPORTS
```

Common in real `.then()` callbacks. Needs switch/try/loop termination handling.

internal/plugins/promise/rules/always_return/always_return.go
```diff
@@ -0,0 +1,294 @@
...
+					}
+				}
+
+				body := cb.Body()
+				if body == nil {
+					return
+				}
+				stmts := body.Statements()
+				if !allPathsTerminate(stmts) {
+					ctx.ReportNode(cb, buildThenShouldReturnOrThrowMessage())
```
fansenze: **Report location diverges from upstream on every branching case.** Upstream reports at the bad code-path segment's start node; this reports at the callback node (`cb`). They coincide only for single-segment bodies.

Differential on the migrated invalid cases (column, sometimes line):

```
hey.then(function(){ if(x){return x} })              upstream L1C23 -> rule L1C10
hey.then(function(x){ switch(x){case 1:return 1} })  upstream L1C24 -> rule L1C10
hey.then(({x,y})=>{ if(y){throw e} })                upstream L3    -> rule L2
```

10 of the migrated invalid cases differ. Either match upstream's segment node, or treat callback-node reporting as intentional and document it.

internal/plugins/promise/rules/always_return/always_return_upstream_test.go
```diff
@@ -0,0 +1,258 @@
+// TestAlwaysReturnUpstream migrates the full valid/invalid suite from upstream
+// __tests__/always-return.js 1:1. Position assertions cover line/column for every
```
fansenze, line 2: Inaccurate: only 3 invalid cases assert Line+Column (one more asserts Line only); the rest assert `{ message }` only. Every case with omitted positions is a branching case where the rule's line/column diverges from upstream (see the report-location comment on `always_return.go`). Adding the missing assertions would surface that divergence.

internal/plugins/promise/rules/always_return/always_return.go
```diff
@@ -0,0 +1,294 @@
...
+func isProcessExitOrAbort(node *ast.Node) bool {
+	if node == nil || !ast.IsCallExpression(node) {
+		return false
+	}
+	callee := ast.SkipOuterExpressions(node.AsCallExpression().Expression, skipTransparent)
+	if callee == nil || !ast.IsPropertyAccessExpression(callee) {
+		return false
+	}
+	prop := callee.AsPropertyAccessExpression()
+	if prop.QuestionDotToken != nil {
```
fansenze, line 123: `process?.exit()` is correctly treated as non-terminating here (matches upstream — the optional-member ChainExpression breaks upstream's selector). But `process.exit?.()` (optional *call*) carries no QuestionDotToken on the property access, so it's treated as terminating and accepted, while upstream reports it. Very rare; flagging for completeness.

swwind: eslint also reports error with `process?.exit()` and `process.exit?.()`

internal/plugins/promise/rules/always_return/always_return.go
```diff
@@ -0,0 +1,397 @@
...
+//
+//	(true,  nil)   — every execution path through stmts terminates.
+//	(false, node)  — paths do not all terminate; node is the first branching
+//	                 statement (if/switch/try/loop) to blame, matching upstream's
+//	                 code-path-segment reporting.
+//	(false, nil)   — paths do not terminate, but no specific sub-node can be
+//	                 blamed (flat body); the caller should fall back to cb.
+//
+// Nested function scopes are not entered.
+func stmtsTerminate(stmts []*ast.Node) (bool, *ast.Node) {
```
fansenze: This hand-rolled termination analysis (`stmtsTerminate` / `stmtTerminates`) diverges from upstream (eslint code-path analysis). Verified vs eslint-plugin-promise@7.3.0 + this rule:

```
  while(x){return 1} / for(;c;){return 1}    upstream REPORT → accepts  (loop body may be skipped)
  for(;;){ if(x) break; return 1 }           upstream REPORT → accepts
  while(true){} / for(;;){}                  upstream VALID  → reports  (infinite loop never completes)
  while(true){ if(x){return 1} }             upstream VALID  → reports
  switch(x){case 1:f();default:return 2}     upstream VALID  → reports  (fall-through)
  try{foo()}catch{bar()}finally{return 2}    upstream VALID  → reports  (finally ignored)
```

rslint already exposes the binder control-flow result as `utils.IsFunctionEndReachable` (reads `NodeFlagsHasImplicitReturn` — syntactic, no type info, `RequiresTypeInfo` stays false); `array_callback_return` / `getter_return` / `constructor_super` already use it.

Switching the verdict to `!utils.IsFunctionEndReachable(cb)` fixes all of the above. Keep the `process.exit`/`abort` special-case on top (binder doesn't model `never`), and keep a light branch scan only for the report position (affects where, not whether).

internal/plugins/promise/rules/always_return/always_return_extras_test.go
```diff
@@ -0,0 +1,301 @@
...
+			},
+			// try is empty, catch returns
+			{
+				Code:   `hey.then(function() { try{}catch(e){return 1} })`,
+				Errors: []rule_tester.InvalidTestCaseError{{MessageId: "thenShouldReturnOrThrow", Message: thenMsg}},
+			},
+
+			// ---- loops: body does not terminate ----
+			{
+				Code:   `hey.then(function() { while(true){} })`,
```
fansenze: These lock-ins assert non-upstream behavior (verified vs eslint-plugin-promise@7.3.0). They pass only by matching the rule's current bug, so they'd flag a correct fix as a regression:

```
  while(true){}                                     asserts invalid → upstream VALID
  for(;;){}                                         asserts invalid → upstream VALID
  try{return 1}catch(e){}                            asserts invalid → upstream VALID
  switch(x){case 1:doSomething();default:return 2}   asserts invalid → upstream VALID
  try{return 1}finally{}  (L147)                      asserts valid   → upstream REPORT
```

Upstream never tests switch/try/loop — these are all new lock-ins. Re-baseline against the oracle, or drop them.

swwind:
Should be good now.

fansenze:
implemented in #938

PR #1055, opened 2026-06-05
feat: add promise/no-multiple-resolved rule

swwind:
## Summary

Added `promise/no-multiple-resolved` rule for #474.

Notes: this rule does not have a default value in eslint.

## Related Links

[original docs](https://github.com/eslint-community/eslint-plugin-promise/blob/main/docs/rules/no-multiple-resolved.md)

## Checklist


internal/plugins/promise/rules/no_multiple_resolved/no_multiple_resolved.go
```diff
@@ -0,0 +1,600 @@
...
+// as a parameter are handled differently per mode:
+//   - TS mode: pass the same rCtx — symbol identity distinguishes the inner parameter
+//     from the outer resolver automatically, so no manual shadowing is needed.
+//   - JS mode: build an inner rCtx with parameter-shadowed names removed.
+func walkNestedFunctions(node *ast.Node, rCtx *ruleCtx) {
+	if node == nil {
+		return
+	}
+	switch node.Kind {
+	case ast.KindFunctionExpression, ast.KindArrowFunction:
```
fansenze: False negative: function declarations and class methods are never analyzed — only FunctionExpression/ArrowFunction.

```js
new Promise((resolve, reject) => {
  function onDone(e, v) {
    reject(e)
    resolve(v) // eslint: "already resolved on line 3" — rslint: no report
  }
  fn(onDone)
})
```

internal/plugins/promise/rules/no_multiple_resolved/no_multiple_resolved.go
```diff
@@ -0,0 +1,600 @@
...
+		if rCtx.resolverSymbols != nil {
+			// TS mode: symbol comparison handles shadowing; reuse same rCtx.
```
fansenze, line 142: Divergence: calling an outer resolver inside a nested executor is reported here; eslint doesn't report this.

```js
new Promise((resolve) => {
  new Promise((res) => {
    resolve(1)
    resolve(2) // rslint: "already resolved on line 3" — eslint: no report
  })
})
```

Arguably a correct report — fine to keep, but then it belongs in the .md differences section.

fansenze (commented):
Ran this branch and eslint-plugin-promise 7.3.0 over the same snippets. Beyond the documented correlated-conditions limitation: 12 confirmed false negatives (5 root causes) and 2 false positives — see inline.

internal/plugins/promise/rules/no_multiple_resolved/no_multiple_resolved.go
```diff
@@ -0,0 +1,600 @@
...
+		expr := stmt.AsExpressionStatement().Expression
+		return analyzeExprForResolvers(expr, state, rCtx)
+
+	case ast.KindBlock:
+		return analyzeStmts(stmt.Statements(), state, rCtx)
+
+	case ast.KindIfStatement:
+		return analyzeIfStmt(stmt, state, rCtx)
+
+	case ast.KindReturnStatement:
```
fansenze, line 251: False negative: resolver calls inside a `return` expression are not tracked (the expression is only scanned for nested functions).

```js
new Promise((resolve, reject) => {
  if (error) reject(error)
  return resolve(value) // eslint: "potentially resolved on line 2" — rslint: no report
})
```

Same for `resolve(1); return resolve(2)`.

internal/plugins/promise/rules/no_multiple_resolved/no_multiple_resolved.go
```diff
@@ -0,0 +1,600 @@
...
+		return pathState{dead: true}
+
+	case ast.KindThrowStatement:
+		ts := stmt.AsThrowStatement()
+		if ts.Expression != nil {
+			walkNestedFunctions(ts.Expression, rCtx)
+		}
+		return pathState{dead: true}
+
+	case ast.KindBreakStatement, ast.KindContinueStatement:
```
fansenze: False negative: `break`/`continue` drop `firstCall`, so a resolved path that rejoins after the loop/switch is lost.

```js
new Promise((resolve, reject) => {
  while (foo) {
    if (error) { reject(error); break }
  }
  resolve(value) // eslint: "potentially resolved on line 3" — rslint: no report
})
```

Same with `continue`, and with `case 1: { reject(e); break }` (the trailing-break special case only covers an unwrapped break).

internal/plugins/promise/rules/no_multiple_resolved/no_multiple_resolved.go
```diff
@@ -0,0 +1,600 @@
...
+}
+
+// analyzeExprForResolvers checks if expr is a direct resolver call and updates state.
+// For non-resolver expressions, it walks for nested functions without changing state.
+func analyzeExprForResolvers(expr *ast.Node, state pathState, rCtx *ruleCtx) pathState {
+	if expr == nil {
+		return state
+	}
+	expr = ast.SkipParentheses(expr)
+	if call := isDirectResolverCall(expr, rCtx); call != nil {
```
fansenze, line 435: False negative: resolver calls nested inside an expression are skipped — only a whole-expression direct call is recognized.

```js
new Promise((resolve, reject) => {
  error ? reject(error) : resolve(value)
  resolve(value) // eslint: "already resolved on line 2" — rslint: no report
})
```

Also missed: `error && reject(error)`, `ok = resolve(1)`, `reject(e), resolve(v)`.

internal/plugins/promise/rules/no_multiple_resolved/no_multiple_resolved.go
```diff
@@ -0,0 +1,600 @@
...
+		node.ForEachChild(func(child *ast.Node) bool {
+			walkThrowables(child, rCtx, onThrowable)
+			return false
+		})
+		callee := ast.SkipParentheses(node.AsCallExpression().Expression)
+		isResolver := callee != nil && ast.IsIdentifier(callee) && isResolverIdent(callee, rCtx)
+		onThrowable(isResolver)
+
+	case ast.KindPropertyAccessExpression, ast.KindElementAccessExpression, ast.KindNewExpression,
+		ast.KindYieldExpression, ast.KindAwaitExpression:
```
fansenze: False positive: with `await` counted as throwable, a plain `await` after `resolve()` makes the catch report.

```js
new Promise(async (resolve, reject) => {
  try {
    const r = await foo();
    resolve();
    await r;
  } catch (error) {
    reject(error); // rslint: "potentially resolved on line 4" — eslint: no report
  }
})
```

Dropping `KindAwaitExpression` here aligns it; `await foo()` still counts via the CallExpression case.

internal/plugins/promise/rules/no_multiple_resolved/no_multiple_resolved.go
```diff
@@ -0,0 +1,600 @@
...
+	merged := pathState{dead: true} // start with dead; merge in each clause
+	for _, clause := range cb.Clauses.Nodes {
+		if clause == nil {
+			continue
+		}
+		cc := clause.AsCaseOrDefaultClause()
+		if cc.Expression != nil {
+			walkNestedFunctions(cc.Expression, rCtx)
+		}
+		clauseState := analyzeSwitchCaseStmts(cc.Statements.Nodes, state, rCtx)
```
fansenze: False negative: fallthrough isn't modeled — every clause starts from the pre-switch state.

```js
new Promise((resolve, reject) => {
  switch (x) {
    case 1: reject(e)
    case 2: resolve(v); break // eslint: "potentially resolved on line 3" — rslint: no report
  }
})
```

internal/plugins/promise/rules/no_multiple_resolved/no_multiple_resolved.go
```diff
@@ -0,0 +1,656 @@
...
+		if len(rCtx.breakExitStates) > 0 {
+			s := rCtx.breakExitStates[len(rCtx.breakExitStates)-1]
+			if s != nil {
+				*s = mergeStates(*s, state)
+			}
+		}
+		return pathState{firstCall: state.firstCall, certain: state.certain, dead: true}
+
+	case ast.KindContinueStatement:
+		return pathState{firstCall: state.firstCall, certain: state.certain, dead: true}
```
fansenze, line 316: False negative (left over from my previous round — only `break` was fixed): `continue` still drops the state instead of merging it into the loop exit, but the loop can exit right after a `continue`.

```js
new Promise((resolve, reject) => {
  while (foo) {
    if (error) { reject(error); continue }
  }
  resolve(value) // eslint: "potentially resolved on line 3" — rslint: no report
})
```

Same with a bare `reject(error); continue` body. Fix direction: merge into the enclosing loop's exit state, like `break` does via `breakExitStates`.

internal/plugins/promise/rules/no_multiple_resolved/no_multiple_resolved.go
```diff
@@ -0,0 +1,656 @@
...
+
+	case ast.KindBreakStatement:
+		if len(rCtx.breakExitStates) > 0 {
+			s := rCtx.breakExitStates[len(rCtx.breakExitStates)-1]
+			if s != nil {
+				*s = mergeStates(*s, state)
+			}
+		}
+		return pathState{firstCall: state.firstCall, certain: state.certain, dead: true}
```
fansenze, line 290: False positive: `break` ignores its label and always merges into the innermost loop/switch.

```js
new Promise((resolve, reject) => {
  outer: while (foo) {
    switch (x) {
      case 1: resolve(a); break outer
    }
    reject(b) // rslint: "potentially resolved on line 4" — eslint: no report
  }
})
```

`break outer` exits the whole `while`, so `reject(b)` never runs after `resolve(a)`; the state is wrongly merged into the switch exit.

Related false negative — labeled blocks never push onto `breakExitStates`, so the state is dropped entirely:

```js
new Promise((resolve, reject) => {
  outer: {
    if (error) { reject(error); break outer }
  }
  resolve(value) // eslint: "potentially resolved on line 3" — rslint: no report
})
```

internal/plugins/promise/rules/no_multiple_resolved/no_multiple_resolved.go
```diff
@@ -0,0 +1,656 @@
...
+			return false
+		})
+		return state
+	}
+}
+
+func analyzeIfStmt(stmt *ast.Node, state pathState, rCtx *ruleCtx) pathState {
+	ifStmt := stmt.AsIfStatement()
+	// Condition may contain nested functions but doesn't affect resolver state
+	walkNestedFunctions(ifStmt.Expression, rCtx)
```
fansenze: False negative: resolver calls in statement-head expression positions are only scanned for nested functions, never tracked as settles. Applies to `if`/`while`/`do` conditions, `for` init/condition/incrementor, and `switch` discriminants.

```js
new Promise((resolve) => {
  if (resolve(1)) { foo() }
  resolve(2) // eslint: "already resolved on line 2" — rslint: no report
})
```

Verified the same gap for `while (resolve(1))`, `for (resolve(1); cond; )`, and `switch (resolve(1))`.

internal/plugins/promise/rules/no_multiple_resolved/no_multiple_resolved.go
```diff
@@ -0,0 +1,656 @@
...
+		var msg rule.RuleMessage
+		if state.certain {
+			msg = buildAlreadyResolvedMessage(line)
+		} else {
+			msg = buildPotentiallyResolvedMessage(line)
+		}
+		rCtx.ctx.ReportNode(callExpr, msg)
+		rCtx.reported[callExpr] = true
+	}
+	return pathState{firstCall: callExpr, certain: true}
```
fansenze: Kind mismatch: `resolve?.()` is a conditional settle, but this always returns `certain: true`.

```js
new Promise((resolve) => {
  resolve?.(1)
  resolve(2) // eslint: "potentially resolved on line 2" — rslint: "already resolved on line 2"
})
```

internal/plugins/promise/rules/no_multiple_resolved/no_multiple_resolved.go
```diff
@@ -0,0 +1,656 @@
...
+// walkThrowables walks a node (skipping nested functions) and calls onThrowable
+// in post-order for every throwable expression encountered. The last call to
+// onThrowable corresponds to the last throwable expression in execution order.
+func walkThrowables(node *ast.Node, rCtx *ruleCtx, onThrowable func(isResolver bool)) {
+	if node == nil {
+		return
+	}
+	// Don't descend into nested functions
+	if ast.IsFunctionLike(node) {
+		return
```
fansenze, line 572: Divergence: upstream's `lastThrowableExpression` is a global traversal-order tracker, so a nested function body appearing after the resolve counts as the last throwable; skipping functions here means no report.

```js
new Promise((resolve, reject) => {
  try {
    resolve(foo())
    const f = () => bar()
  } catch (e) {
    reject(e) // eslint: "potentially resolved on line 3" — rslint: no report
  }
})
```

Upstream's behavior is arguably a quirk (`f` never runs inside the try), so documenting this in the .md instead of aligning seems fine.

internal/plugins/promise/rules/no_multiple_resolved/no_multiple_resolved.md
```diff
@@ -0,0 +1,86 @@
...
+  }
+})
+```
+
+## Differences from ESLint
+
+- Correlated conditions across separate `if` statements (e.g. `if (err) { reject(err) }` followed
+  by `if (!err) { resolve(val) }`) are not recognized as mutually exclusive. This rule reports them
+  as potential double-resolution. Full ESLint code-path analysis handles these; our simplified
+  state-based analysis does not.
```
fansenze, line 71: The actual divergence is wider than correlated conditions: any conditionally-guarded second settle after a potential settle is reported, even with completely unrelated conditions.

```js
new Promise((resolve) => {
  if (a) { resolve(1) }
  if (b) { resolve(2) } // rslint: "potentially resolved on line 2" — eslint: no report
})
```

ESLint only propagates a potential state into the sole continuation path (a structural rule, independent of condition correlation). Keeping the behavior is fine — it's semantically defensible — but this section should describe the real boundary.

fansenze (commented):
Re-ran this branch and eslint-plugin-promise 7.3.0 over the same snippets (46 differential probes + a 1:1 check of the migrated upstream suite, 15 valid + 16 invalid all matching v7.3.0 including expected errors).

From my previous round: 5 of 7 are fixed and verified (function declarations/class methods, return expressions, expression-level resolver calls, await-throwable FP, switch fallthrough), and the nested-executor divergence is kept + documented as suggested.

Still open: `continue` is still a false negative (that comment covered both `break` and `continue`), and the new `breakExitStates` mechanism introduces a labeled-`break` false positive. Smaller gaps inline.

internal/plugins/promise/rules/no_multiple_resolved/no_multiple_resolved_extras_test.go
```diff
@@ -0,0 +1,418 @@
...
+						Message: "Promise should not be resolved multiple times. Promise is already resolved on line 2.",
+						Line:    3,
+					},
+				},
+			},
+
+			// ---- Dimension 1: do-while body always runs ----
+			// do-while body runs at least once; resolver call within it counts
+			{
+				Code: `new Promise((resolve, reject) => {
```
fansenze, line 189: Alignment note: for `do-while` / `for-in` / `for-of`, eslint emits a second report — the in-loop call against itself via the loop back edge (`L3: "potentially resolved on line 3"` here) — while this implementation only emits the post-loop one. Plain `for`/`while` don't self-report in eslint, so the asymmetry is upstream behavior.

These three cases lock in the 1-report behavior; worth a note in the .md differences section at least.

packages/rslint-test-tools/tests/eslint-plugin-promise/rules/no-multiple-resolved.test.ts
```diff
@@ -0,0 +1,69 @@
+import { RuleTester } from '../rule-tester';
+
+const ruleTester = new RuleTester();
+
+ruleTester.run('no-multiple-resolved', {} as never, {
```
fansenze, line 5: This JS suite is a 6-case smoke subset, while the sibling promise rules mirror the full upstream suite (always-return 104, catch-or-return 79, no-return-wrap 67, param-names 58 cases). Consider mirroring the upstream cases here too, or confirming the thin version is intentional.

fansenze (approved):

PR #1056, opened 2026-06-05
feat: add promise/no-nesting rule

swwind:
## Summary

Added `promise/no-nesting` rule for #474.

## Related Links

[original docs](https://github.com/eslint-community/eslint-plugin-promise/blob/main/docs/rules/no-nesting.md)

## Checklist

### DIFF
--- a/internal/plugins/promise/rules/no_nesting/no_nesting.go
+++ b/internal/plugins/promise/rules/no_nesting/no_nesting.go
@@ -30,117 +30,55 @@ func isPromiseCallback(node *ast.Node) bool {
 	return isThenOrCatchCall(parent)
 }
 
-// collectScopeBindings collects into names all binding identifiers introduced by fn:
-// parameters and locally declared variables at any nesting depth (stopping at nested
-// function boundaries). This mirrors ESLint's scope.variables for the callback's scope.
-func collectScopeBindings(fn *ast.Node, names map[string]bool) {
-	var params *ast.NodeList
-	switch fn.Kind {
-	case ast.KindFunctionExpression:
-		if fe := fn.AsFunctionExpression(); fe != nil {
-			params = fe.Parameters
-		}
-	case ast.KindArrowFunction:
-		if af := fn.AsArrowFunction(); af != nil {
-			params = af.Parameters
-		}
-	}
-	if params != nil {
-		for _, param := range params.Nodes {
-			if param.Kind == ast.KindParameter {
-				if pd := param.AsParameterDeclaration(); pd != nil {
-					utils.CollectBindingNames(pd.Name(), func(_ *ast.Node, name string) {
-						names[name] = true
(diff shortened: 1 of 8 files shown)


internal/plugins/promise/rules/no_nesting/no_nesting.go
```diff
@@ -0,0 +1,133 @@
...
+	})
+	return found
+}
+
+// argsContainRef reports whether any identifier anywhere in the call's argument
+// list has a name that is bound in fn's parameter list or local variable
+// declarations. Uses utils.HasShadowingParameter for params and
+// utils.HasShadowingDeclaration / utils.HasHoistedVarDeclaration for body
+// declarations, mirroring the upstream's scope-reference check.
+func argsContainRef(callNode *ast.Node, fn *ast.Node) bool {
```
fansenze, line 58: The name-based matching here diverges from ESLint: any identifier appearing anywhere in the argument subtree counts as a "reference" to the enclosing callback's bindings, including positions that aren't variable references at all. Verified by running both implementations:

```js
// ESLint: reports — this rule: no report
// `user` in `p.user` is a property name, not a variable reference
doThing().then(user => getProfile().then(p => p.user))

// ESLint: reports — this rule: no report
// `user` here is an object literal key
doThing().then(user => getProfile().then(p => ({ user: 1 })))

// ESLint: reports — this rule: no report
// the inner `a` is a fresh parameter shadowing the outer one,
// so the inner callback never uses the outer `a`
doThing().then(a => inner.then(a => use(a)))
```

The first two shapes are common in real code — callback param names often collide with data property names (`user`, `data`, …) — so these false negatives would noticeably weaken the rule. Counting only identifiers in actual reference positions (skipping property names, object keys, and binding names), and not attributing a name to the outer callback once an inner function rebinds it, would close the gap.

internal/plugins/promise/rules/no_nesting/no_nesting.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+	body := fn.Body()
+	for _, arg := range args.Nodes {
+		found := false
+		walkIdentifiers(arg, func(name string) bool {
+			if utils.HasShadowingParameter(fn, name) {
+				found = true
+				return true
+			}
+			if body != nil && (utils.HasShadowingDeclaration(body, name) || utils.HasHoistedVarDeclaration(body, name)) {
```
fansenze: One more divergence in the opposite direction, also verified by running both implementations:

```js
// ESLint: no report — this rule: reports
// `arguments` belongs to the enclosing callback, so this nesting can't be flattened
doThing().then(function() { return foo.then(() => use(arguments)) })
```

The param/declaration lookups don't cover the implicit `arguments` binding of a non-arrow callback.

fansenze (approved):

PR #1057, opened 2026-06-05
feat: add promise/no-callback-in-promise rule

swwind:
## Summary

Added `promise/no-callback-in-promise` rule for #474.

This is rule is quite brute force, that only checks callback functions named as `callback`, `cb`, `next` and `done`.

## Related Links

[original docs](https://github.com/eslint-community/eslint-plugin-promise/blob/main/docs/rules/no-callback-in-promise.md)

## Checklist

### DIFF
--- a/internal/plugins/promise/rules/no_callback_in_promise/no_callback_in_promise.go
+++ b/internal/plugins/promise/rules/no_callback_in_promise/no_callback_in_promise.go
@@ -1,0 +1,209 @@
+package no_callback_in_promise
+
+import (
+	"github.com/microsoft/typescript-go/shim/ast"
+	"github.com/web-infra-dev/rslint/internal/rule"
+	"github.com/web-infra-dev/rslint/internal/utils"
+)
+
+const skipTransparent = ast.OEKParentheses
+
+var cbBlacklist = []string{"callback", "cb", "next", "done"}
+var timeoutWhitelist = []string{"setImmediate", "setTimeout", "requestAnimationFrame", "nextTick"}
+
+type Options struct {
+	Exceptions  []string
+	TimeoutsErr bool
+}
+
+func parseOptions(options any) Options {
+	opts := Options{}
+	optsMap := utils.GetOptionsMap(options)
+	if optsMap == nil {
+		return opts
+	}
+	if v, ok := optsMap["timeoutsErr"].(bool); ok {
+		opts.TimeoutsErr = v
+	}
+	if v, ok := optsMap["exceptions"].([]interface{}); ok {
+		for _, e := range v {
+			if s, ok := e.(string); ok {
+				opts.Exceptions = append(opts.Exceptions, s)
+			}
+		}
+	}
+	return opts
+}
+
+func buildCallbackMessage() rule.RuleMessage {
+	return rule.RuleMessage{
+		Id:          "callback",
+		Description: "Avoid calling back inside of a promise.",
+	}
+}
+
+// isCallbackName returns true if name is in cbBlacklist and not in exceptions.
+func isCallbackName(name string, exceptions []string) bool {
+	for _, e := range exceptions {
+		if name == e {
+			return false
+		}
+	}
+	for _, cb := range cbBlacklist {
+		if name == cb {
+			return true
+		}
+	}
+	return false
+}
+
+// isCallbackCall returns true if node is a call expression whose callee (skipping parens) is an
+// identifier with a callback name.
+func isCallbackCall(node *ast.Node, exceptions []string) bool {
+	if !ast.IsCallExpression(node) {
+		return false
+	}
+	callee := ast.SkipOuterExpressions(node.AsCallExpression().Expression, skipTransparent)
+	return callee != nil && ast.IsIdentifier(callee) && isCallbackName(callee.AsIdentifier().Text, exceptions)
+}
+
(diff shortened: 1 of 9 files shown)


internal/plugins/promise/rules/no_callback_in_promise/no_callback_in_promise.go
```diff
@@ -0,0 +1,191 @@
...
+	return callee != nil && ast.IsIdentifier(callee) && isCallbackName(callee.AsIdentifier().Text, exceptions)
+}
+
+// getMemberCallName returns the member-property name (for a.foo()) or the direct identifier name
+// (for foo()) from a call-expression callee. Returns "" for anything else.
+func getMemberCallName(node *ast.Node) string {
+	if !ast.IsCallExpression(node) {
+		return ""
+	}
+	callee := node.AsCallExpression().Expression
```
fansenze: **False positive when the timeout callee is parenthesized.**

`getMemberCallName` reads the callee **without** skipping parens, unlike its siblings `isCallbackCall` and `promiseutil.IsMemberCall` (both call `SkipOuterExpressions`). For a parenthesized timeout callee it returns `""`, so `isInsideTimeout` misses the whitelist and the wrapped callback is reported.

Repro (default `timeoutsErr: false`):
```js
promise.then(() => (setTimeout)(() => cb()))
```
Upstream: **valid** (`setTimeout` is whitelisted; ESTree has no paren node). This port: **reports** `cb()`.

Fix — skip parens like the sibling helpers:
```suggestion
	// Skip parens so a parenthesized callee like `(setTimeout)(cb)` still resolves to
	// "setTimeout" — upstream reads the bare identifier (ESTree has no paren node). Without
	// this, the timeout whitelist is missed and the wrapped call is reported as a false positive.
	callee := ast.SkipOuterExpressions(node.AsCallExpression().Expression, skipTransparent)
```

internal/plugins/promise/rules/no_callback_in_promise/no_callback_in_promise.go
```diff
@@ -0,0 +1,191 @@
...
+						firstArg := ast.SkipOuterExpressions(callExpr.Arguments.Nodes[0], skipTransparent)
+						if firstArg != nil && ast.IsIdentifier(firstArg) {
+							firstArgName = firstArg.AsIdentifier().Text
+						}
+					}
+
+					if isPromiseMemberCall(node) {
+						// Scenario A: a.then(cb) — callback passed directly as argument.
+						if firstArgName != "" && isCallbackName(firstArgName, exceptions) {
+							ctx.ReportNode(callExpr.Arguments.Nodes[0], buildCallbackMessage())
```
fansenze: **Reported range includes the parens for a parenthesized callback arg.**

`firstArgName` is computed from the paren-skipped node, but the report targets the raw `Arguments.Nodes[0]`, so a parenthesized arg reports the wrapping node instead of the identifier.

Repro:
```js
a.then((cb))
```
Upstream reports `cb` (column 9); this port reports `(cb)` (column 8).

Fix — hoist `firstArg` and report it:
```suggestion
					var firstArg *ast.Node
					var firstArgName string
					if callExpr.Arguments != nil && len(callExpr.Arguments.Nodes) > 0 {
						firstArg = ast.SkipOuterExpressions(callExpr.Arguments.Nodes[0], skipTransparent)
						if firstArg != nil && ast.IsIdentifier(firstArg) {
							firstArgName = firstArg.AsIdentifier().Text
						}
					}

					if isPromiseMemberCall(node) {
						// Scenario A: a.then(cb) — callback passed directly as argument. Report the
						// unwrapped identifier (firstArg) to match upstream's node.arguments[0]; the raw
						// Nodes[0] would span the surrounding parens, e.g. `(cb)` instead of `cb`.
						if firstArgName != "" && isCallbackName(firstArgName, exceptions) {
							ctx.ReportNode(firstArg, buildCallbackMessage())
```

fansenze (commented):
Reviewed the new `no-callback-in-promise` rule against upstream `eslint-plugin-promise` node-by-node — logic is faithful. Two paren-handling divergences surfaced; both are low real-user reach but the first is a genuine false positive. I reproduced both and verified the suggested fixes pass `go test ./internal/plugins/promise/...` (all rules green, gofmt/vet clean).

internal/plugins/promise/rules/no_callback_in_promise/no_callback_in_promise.go
swwind: Good catch, fixed in a7720c71 — `getMemberCallName` now skips parens via `ast.SkipOuterExpressions` before reading the callee, consistent with `isCallbackCall`/`promiseutil.IsMemberCall`. Added a regression test (`promise.then(() => (setTimeout)(() => cb()))`) that fails without the fix.

internal/plugins/promise/rules/no_callback_in_promise/no_callback_in_promise.go
swwind: Fixed in a7720c71 — hoisted `firstArg` (paren-unwrapped) and report that instead of the raw `Arguments.Nodes[0]`. Updated the existing `a.then((cb))` test to assert `Column: 9` so it actually locks this in (it previously only checked the messageId and would've passed either way).

fansenze (approved):

PR #1058, opened 2026-06-05
feat: add promise/no-return-in-finally rule

swwind:
## Summary

Ported `promise/no-return-in-finally` rule for #474, clean. good.

## Related Links

[original docs](https://github.com/eslint-community/eslint-plugin-promise/blob/main/docs/rules/no-return-in-finally.md)

## Checklist

### DIFF
--- a/internal/plugins/promise/rules/no_return_in_finally/no_return_in_finally.go
+++ b/internal/plugins/promise/rules/no_return_in_finally/no_return_in_finally.go
@@ -1,0 +1,43 @@
+package no_return_in_finally
+
+import (
+	"github.com/microsoft/typescript-go/shim/ast"
+	"github.com/web-infra-dev/rslint/internal/plugins/promise/promiseutil"
+	"github.com/web-infra-dev/rslint/internal/rule"
+)
+
+const skipTransparent = ast.OEKParentheses
+
+func buildMessage() rule.RuleMessage {
+	return rule.RuleMessage{
+		Id:          "noReturnInFinally",
+		Description: "No return in finally",
+	}
+}
+
+// isFinallyCallback reports whether fn is directly passed as a callback to a
+// .finally() call. Parentheses around fn are skipped.
+func isFinallyCallback(fn *ast.Node) bool {
+	parent := fn.Parent
+	for parent != nil && ast.IsOuterExpression(parent, skipTransparent) {
+		parent = parent.Parent
+	}
+	return parent != nil && promiseutil.IsMemberCall(parent, "finally")
+}
+
+var NoReturnInFinallyRule = rule.Rule{
+	Name: "promise/no-return-in-finally",
+	Run: func(ctx rule.RuleContext, options any) rule.RuleListeners {
+		return rule.RuleListeners{
+			ast.KindReturnStatement: func(node *ast.Node) {
+				fn := promiseutil.NearestFunctionBoundary(node)
+				if fn == nil {
+					return
+				}
+				if isFinallyCallback(fn) {
+					ctx.ReportNode(node, buildMessage())
+				}
+			},
+		}
+	},
+}
--- a/internal/plugins/promise/rules/no_return_in_finally/no_return_in_finally.go
+++ b/internal/plugins/promise/rules/no_return_in_finally/no_return_in_finally.go
@@ -15,60 +15,45 @@ func buildMessage() rule.RuleMessage {
 	}
 }
 
-// isFinallyCallback reports whether fn is directly passed as a callback to a
-// .finally() call. Parentheses around fn are skipped.
-func isFinallyCallback(fn *ast.Node) bool {
-	parent := fn.Parent
-	for parent != nil && ast.IsOuterExpression(parent, skipTransparent) {
-		parent = parent.Parent
+// callbackBody returns the block body of a .finally() callback, or nil if fn
+// isn't a function/arrow expression with a block body. Upstream only ever
+// sees FunctionExpression/ArrowFunctionExpression here, since fn comes
+// straight from the call's first argument.
+func callbackBody(fn *ast.Node) *ast.Node {
+	switch fn.Kind {
+	case ast.KindFunctionExpression:
+		return fn.AsFunctionExpression().Body
+	case ast.KindArrowFunction:
+		return fn.AsArrowFunction().Body
+	default:
+		return nil
 	}
-	if parent == nil || !promiseutil.IsMemberCall(parent, "finally") {
-		return false
-	}
-	callExpr := parent.AsCallExpression()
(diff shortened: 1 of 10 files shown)


internal/plugins/promise/rules/no_return_wrap/no_return_wrap.go
```diff
@@ -74,7 +74,7 @@ func checkCallExpression(ctx rule.RuleContext, opts Options, callNode *ast.Node,
 }
 
 func isInPromise(node *ast.Node) bool {
-	functionNode := nearestFunctionBoundary(node)
+	functionNode := promiseutil.NearestFunctionBoundary(node)
```
fansenze: **`no-return-wrap` regression: the shared `NearestFunctionBoundary` adds `FunctionDeclaration` as a boundary, silently changing this rule's behavior and diverging from upstream.**

The old `no_return_wrap.nearestFunctionBoundary` did **not** include `KindFunctionDeclaration`. The new shared `promiseutil.NearestFunctionBoundary` does, and `isInPromise` now calls it — so a `return Promise.resolve(...)` inside a nested function declaration within a promise callback is no longer reported.

Repro:
```js
Promise.resolve().then(function () {
  function helper() {
    return Promise.resolve(4); // expected: report "resolve"
  }
  return helper();
});
```

|                                  | reports? |
| -------------------------------- | -------- |
| upstream `eslint-plugin-promise` | ✅ yes    |
| rslint `main`                    | ✅ yes    |
| this PR                          | ❌ no     |

Upstream picks the boundary by **filtering** ancestors down to arrow/function expressions only — `FunctionDeclaration` is skipped, so it walks out to the enclosing `FunctionExpression` (the promise callback) and reports:
```js
// eslint-plugin-promise — rules/no-return-wrap.js
let functionNode = getAncestors(context, node)
  .filter((n) => n.type === 'ArrowFunctionExpression' || n.type === 'FunctionExpression')
  .reverse()[0]
```
rslint instead uses a "stop at the first boundary" walk, so adding `FunctionDeclaration` to the set makes it stop at `helper` and miss the promise callback.

**Root cause:** the two rules need *different* boundary sets. `no-return-in-finally` *needs* `FunctionDeclaration` (to mirror upstream only iterating the callback body's direct children), while `no-return-wrap` must *not* have it. Folding both into one shared helper is the wrong abstraction.

**Suggested fix:** don't reuse this exact variant for `no_return_wrap` — either keep its boundary set `FunctionDeclaration`-free, or parameterize the accepted kinds so each rule passes its own set.

internal/plugins/promise/rules/no_return_in_finally/no_return_in_finally.go
```diff
@@ -0,0 +1,77 @@
...
+						body = fn.AsArrowFunction().Body
+					}
+					
+					// ESLint's promise/no-return-in-finally rule only checks top-level statements
+					// in the function's block body. It ignores nested returns.
+					if body != nil && body.Kind == ast.KindBlock {
+						if node.Parent != body {
+							return
+						}
+						ctx.ReportNode(node, buildMessage())
```
fansenze: **Report position and count both diverge from upstream (same root cause).**

Upstream listens on the `.finally` `CallExpression`, scans the callback body's top-level statements, and — if **any** is a return — reports **once** on the `.finally` property. This rule instead listens on `ReturnStatement` and reports each one on the `return` node, which diverges in two ways:

**1. Position.** Upstream reports on the `.finally` identifier (`node: node.callee.property`); this rule reports on the `return`. Every diagnostic lands on a different column. E.g. `myPromise.finally(() => { return 2 })` → this rule reports at **col 27** (the `return`, as asserted in the tests), upstream reports at **col 11** (`finally`).

**2. Count.** Upstream uses `.some(...)`, so multiple top-level returns still produce **one** report; this rule reports per-return. Verified: `myPromise.finally(() => { return 1; return 2; })` → this rule emits **2** diagnostics vs upstream's **1**. (Only reachable via dead code, so low impact.)

**Suggested fix** — mirror upstream's structure (listen on the call, report once on `.finally`). This also removes the dependency on `NearestFunctionBoundary`, which dissolves the `no_return_wrap` regression from the other comment — no need to add `FunctionDeclaration` to the shared helper at all:

```go
ast.KindCallExpression: func(node *ast.Node) {
	if !promiseutil.IsMemberCall(node, "finally") {
		return
	}
	call := node.AsCallExpression()
	if call.Arguments == nil || len(call.Arguments.Nodes) == 0 {
		return
	}
	fn := ast.SkipOuterExpressions(call.Arguments.Nodes[0], skipTransparent)
	var body *ast.Node
	switch fn.Kind {
	case ast.KindFunctionExpression:
		body = fn.AsFunctionExpression().Body
	case ast.KindArrowFunction:
		body = fn.AsArrowFunction().Body
	}
	if body == nil || body.Kind != ast.KindBlock {
		return
	}
	for _, stmt := range body.AsBlock().Statements.Nodes {
		if stmt.Kind == ast.KindReturnStatement {
			callee := ast.SkipOuterExpressions(call.Expression, skipTransparent)
			ctx.ReportNode(callee.AsPropertyAccessExpression().Name(), buildMessage())
			break
		}
	}
},
```

This preserves every existing valid/invalid behavior (computed `['finally']` skipped, `?.finally` matched, parens unwrapped, nested returns ignored, arrow-expression bodies ignored) and drops `isFinallyCallback` + `NearestFunctionBoundary`. The position assertions in both Go test files then need updating to the `.finally` column (**20 / 19 / 11 / 20** for the four upstream invalid cases).

internal/plugins/promise/rules/no_return_wrap/no_return_wrap.go
swwind: Good catch — fixed in 7227b0d2: `no-return-wrap` now uses its own boundary walk (no `FunctionDeclaration`), and `promiseutil.NearestFunctionBoundary` (with `FunctionDeclaration`) is only used by `no-return-in-finally`. Added the nested-`function helper()`-inside-a-promise-callback repro from this comment as a regression test.

internal/plugins/promise/rules/no_return_in_finally/no_return_in_finally.go
swwind: Confirmed against upstream source and fixed in 3d039153: now listens on `CallExpression`, scans only the `.finally()` callback's direct top-level statements, and reports once on the `.finally` property — matching upstream's `body.body.some(...)` for both position and dedup. Dropped `isFinallyCallback` and the now-unused `promiseutil.NearestFunctionBoundary`. Added a regression test for the multiple-top-level-returns dedup case and updated the existing position assertions.

fansenze (approved):

PR #1059, opened 2026-06-05
feat: port rule promise/prefer-await-to-then

swwind:
## Summary

Ported `promise/prefer-await-to-then` rule for #474.

## Related Links

[original docs](https://github.com/eslint-community/eslint-plugin-promise/blob/main/docs/rules/prefer-await-to-then.md)

## Checklist

### DIFF
--- a/internal/plugins/promise/rules/prefer_await_to_then/prefer_await_to_then.go
+++ b/internal/plugins/promise/rules/prefer_await_to_then/prefer_await_to_then.go
@@ -28,11 +28,27 @@ var preferAwaitToCallbackMessage = rule.RuleMessage{
 	Description: "Prefer await to then()/catch()/finally().",
 }
 
-// isTopLevelScoped reports whether node is outside any function-like scope,
+// createsScope reports whether node opens a new eslint-scope scope.
+func createsScope(node *ast.Node) bool {
+	if ast.IsFunctionLike(node) {
+		return true
+	}
+	switch node.Kind {
+	case ast.KindBlock,
+		ast.KindForStatement, ast.KindForInStatement, ast.KindForOfStatement,
+		ast.KindSwitchStatement, ast.KindCatchClause,
+		ast.KindClassDeclaration, ast.KindClassExpression,
+		ast.KindWithStatement, ast.KindClassStaticBlockDeclaration:
+		return true
+	}
+	return false
+}
+
+// isTopLevelScoped reports whether node is outside any scope-creating ancestor,
 // mirroring ESLint's getScope().block.type === 'Program' check.
 func isTopLevelScoped(node *ast.Node) bool {
 	for cur := node.Parent; cur != nil; cur = cur.Parent {
-		if ast.IsFunctionLike(cur) {
+		if createsScope(cur) {
 			return false
 		}
 	}
@@ -67,10 +83,19 @@ func isCypress(node *ast.Node) bool {
 		return false
 	}
 	callee := ast.SkipOuterExpressions(node.AsCallExpression().Expression, skipTransparent)
-	if callee == nil || !ast.IsPropertyAccessExpression(callee) {
+	if callee == nil {
 		return false
 	}
-	obj := ast.SkipOuterExpressions(callee.AsPropertyAccessExpression().Expression, skipTransparent)
+	var rawObj *ast.Node
+	switch {
+	case ast.IsPropertyAccessExpression(callee):
+		rawObj = callee.AsPropertyAccessExpression().Expression
+	case ast.IsElementAccessExpression(callee):
+		rawObj = callee.AsElementAccessExpression().Expression
+	default:
+		return false
+	}
+	obj := ast.SkipOuterExpressions(rawObj, skipTransparent)
 	if obj == nil {
 		return false
 	}
--- a/internal/plugins/promise/rules/prefer_await_to_then/prefer_await_to_then.go
+++ b/internal/plugins/promise/rules/prefer_await_to_then/prefer_await_to_then.go
@@ -127,10 +127,18 @@ var PreferAwaitToThenRule = rule.Rule{
 		return rule.RuleListeners{
 			ast.KindCallExpression: func(node *ast.Node) {
 				callee := ast.SkipOuterExpressions(node.AsCallExpression().Expression, skipTransparent)
-				if callee == nil || !ast.IsPropertyAccessExpression(callee) {
+				if callee == nil {
 					return
 				}
-				nameNode := callee.AsPropertyAccessExpression().Name()
+				var nameNode *ast.Node
+				switch {
+				case ast.IsPropertyAccessExpression(callee):
+					nameNode = callee.AsPropertyAccessExpression().Name()
+				case ast.IsElementAccessExpression(callee):
+					nameNode = ast.SkipOuterExpressions(callee.AsElementAccessExpression().ArgumentExpression, skipTransparent)
+				default:
+					return
+				}
 				if nameNode == nil || !ast.IsIdentifier(nameNode) {
 					return
 				}
--- a/internal/plugins/promise/rules/prefer_await_to_then/prefer_await_to_then.go
+++ b/internal/plugins/promise/rules/prefer_await_to_then/prefer_await_to_then.go
@@ -1,0 +1,111 @@
+package prefer_await_to_then
+
+import (
+	"github.com/microsoft/typescript-go/shim/ast"
+	"github.com/web-infra-dev/rslint/internal/rule"
+	"github.com/web-infra-dev/rslint/internal/utils"
+)
+
+const skipTransparent = ast.OEKParentheses
+
+type Options struct {
+	Strict bool
+}
+
+func parseOptions(options any) Options {
+	opts := Options{}
+	optsMap := utils.GetOptionsMap(options)
+	if optsMap != nil {
+		if v, ok := optsMap["strict"].(bool); ok {
+			opts.Strict = v
+		}
+	}
+	return opts
+}
+
+var preferAwaitToCallbackMessage = rule.RuleMessage{
+	Id:          "preferAwaitToCallback",
+	Description: "Prefer await to then()/catch()/finally().",
+}
+
+// isTopLevelScoped reports whether node is outside any function-like scope,
+// mirroring ESLint's getScope().block.type === 'Program' check.
+func isTopLevelScoped(node *ast.Node) bool {
+	for cur := node.Parent; cur != nil; cur = cur.Parent {
+		if ast.IsFunctionLike(cur) {
+			return false
+		}
+	}
+	return true
+}
+
+// isInsideYieldOrAwait reports whether any ancestor of node is an AwaitExpression
+// or YieldExpression.
+func isInsideYieldOrAwait(node *ast.Node) bool {
+	for cur := node.Parent; cur != nil; cur = cur.Parent {
+		if cur.Kind == ast.KindAwaitExpression || cur.Kind == ast.KindYieldExpression {
+			return true
+		}
+	}
+	return false
+}
+
+// isInsideConstructor reports whether any ancestor of node is a constructor declaration.
+func isInsideConstructor(node *ast.Node) bool {
+	for cur := node.Parent; cur != nil; cur = cur.Parent {
+		if cur.Kind == ast.KindConstructor {
+			return true
+		}
+	}
+	return false
+}
+
+// isCypress reports whether callNode is part of a Cypress cy.* chain.
+// It mirrors eslint-plugin-promise's recursive isMemberCallWithObjectName('cy', node) check.
+func isCypress(node *ast.Node) bool {
+	if node == nil || !ast.IsCallExpression(node) {
+		return false
+	}
+	callee := ast.SkipOuterExpressions(node.AsCallExpression().Expression, skipTransparent)
+	if callee == nil || !ast.IsPropertyAccessExpression(callee) {
+		return false
+	}
+	obj := ast.SkipOuterExpressions(callee.AsPropertyAccessExpression().Expression, skipTransparent)
+	if obj == nil {
+		return false
+	}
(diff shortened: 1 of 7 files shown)


internal/plugins/promise/rules/prefer_await_to_then/prefer_await_to_then.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+var preferAwaitToCallbackMessage = rule.RuleMessage{
+	Id:          "preferAwaitToCallback",
+	Description: "Prefer await to then()/catch()/finally().",
+}
+
+// isTopLevelScoped reports whether node is outside any function-like scope,
+// mirroring ESLint's getScope().block.type === 'Program' check.
+func isTopLevelScoped(node *ast.Node) bool {
```
fansenze, line 61: **`isTopLevelScoped` only stops at function scopes, but ESLint's `getScope().block === Program` also stops at block / loop / `switch` / `catch` / class scopes — so every top-level block-ish position is silently un-reported.**

eslint-scope opens a scope for blocks, loops, `switch`, `catch`, classes and static blocks, not just functions. The original rule reports whenever the innermost scope is *not* `Program`. This walk returns `true` (treat as bare top-level → skip) for all of those, producing false negatives.

Measured — ESLint 7.3.0 (A) vs this PR (B) vs the fix below (C), shown as `startLine:startCol`:

| input (module top level) | A: ESLint | B: this PR | C: with fix |
|---|---|---|---|
| `{ thing.then(cb) }` | `1:9` | **(none)** | `1:9` |
| `if (x) { thing.then(cb) }` | `1:16` | **(none)** | `1:16` |
| `for (const a of xs) { thing.then(cb) }` | `1:29` | **(none)** | `1:29` |
| `for (const a of xs) thing.then(cb)` | `1:27` | **(none)** | `1:27` |
| `try { x() } catch (e) { thing.then(cb) }` | `1:31` | **(none)** | `1:31` |
| `switch (x) { case 1: thing.then(cb) }` | `1:28` | **(none)** | `1:28` |
| `class C { x = p.then() }` | `1:17` | **(none)** | `1:17` |
| `class C { static { p.then() } }` | `1:22` | **(none)** | `1:22` |

Baselines unchanged across A/B/C (no regression): bare `thing.then(cb)` → `(none)`; `function f(){ thing.then(cb) }` → `1:21`.

Two consequences:

1. The divergence is **broader than `prefer_await_to_then.md` documents** — its "Differences from ESLint" section only lists class field initializers and static blocks, but plain blocks, loops, `switch` and `catch` diverge too.
2. The fix below also closes the class-field / static-block rows (last two table rows, C matches A exactly), so the **"Differences from ESLint" section in `prefer_await_to_then.md` should be deleted entirely** — after the fix it documents a difference that no longer exists.

**Fix** (stop at the first scope-creating ancestor and check whether it's the source file; the scope list mirrors eslint-scope's scope-creating nodes — the 8 rows above are the measured evidence):

```go
// createsScope reports whether node opens a new eslint-scope scope.
func createsScope(node *ast.Node) bool {
	if ast.IsFunctionLike(node) {
		return true
	}
	switch node.Kind {
	case ast.KindBlock,
		ast.KindForStatement, ast.KindForInStatement, ast.KindForOfStatement,
		ast.KindSwitchStatement, ast.KindCatchClause,
		ast.KindClassDeclaration, ast.KindClassExpression,
		ast.KindWithStatement, ast.KindClassStaticBlockDeclaration:
		return true
	}
	return false
}

func isTopLevelScoped(node *ast.Node) bool {
	for cur := node.Parent; cur != nil; cur = cur.Parent {
		if createsScope(cur) {
			return false
		}
		if cur.Kind == ast.KindSourceFile {
			return true
		}
	}
	return true
}
```

internal/plugins/promise/rules/prefer_await_to_then/prefer_await_to_then.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+// isCypress reports whether callNode is part of a Cypress cy.* chain.
+// It mirrors eslint-plugin-promise's recursive isMemberCallWithObjectName('cy', node) check.
+func isCypress(node *ast.Node) bool {
+	if node == nil || !ast.IsCallExpression(node) {
+		return false
+	}
+	callee := ast.SkipOuterExpressions(node.AsCallExpression().Expression, skipTransparent)
+	if callee == nil || !ast.IsPropertyAccessExpression(callee) {
```
fansenze: **`isCypress` only walks `PropertyAccessExpression`, but ESLint's `isMemberCallWithObjectName` matches any `MemberExpression` (dot *and* bracket) — so a computed Cypress access like `cy['get']()` is not recognized and gets falsely reported.**

The upstream helper recurses through `node.callee.object` for any `MemberExpression`. Here the `!ast.IsPropertyAccessExpression(callee)` guard bails out the moment it meets an `ElementAccessExpression` (`cy['get']`), so the `cy` chain is missed and the call is reported when ESLint stays silent.

Measured — ESLint 7.3.0 (A) vs this PR (B) vs the fix below (C):

| input | A: ESLint | B: this PR | C: with fix |
|---|---|---|---|
| `function f() { cy['get']('x').then(go) }` | `(none)` | **`1:31`** | `(none)` |
| `function hi() { cy.get(".myClass").then(go) }` (dot baseline) | `(none)` | `(none)` | `(none)` |

(Both cases are wrapped in a function so the result isolates the Cypress check from `isTopLevelScoped`.)

**Fix** (handle `ElementAccessExpression` in the receiver walk — verified to turn the computed case into `(none)` with the dot baseline unchanged):

```go
callee := ast.SkipOuterExpressions(node.AsCallExpression().Expression, skipTransparent)
if callee == nil {
	return false
}
var obj *ast.Node
switch {
case ast.IsPropertyAccessExpression(callee):
	obj = ast.SkipOuterExpressions(callee.AsPropertyAccessExpression().Expression, skipTransparent)
case ast.IsElementAccessExpression(callee):
	obj = ast.SkipOuterExpressions(callee.AsElementAccessExpression().Expression, skipTransparent)
default:
	return false
}
if obj == nil {
	return false
}
```

fansenze (commented):
Two alignment issues vs eslint-plugin-promise `prefer-await-to-then`, both verified by running the original rule (eslint-plugin-promise@7.3.0) against this port and a candidate fix. Details inline.

internal/plugins/promise/rules/prefer_await_to_then/prefer_await_to_then.go
```diff
@@ -0,0 +1,136 @@
...
+func createsScope(node *ast.Node) bool {
+	if ast.IsFunctionLike(node) {
+		return true
+	}
+	switch node.Kind {
+	case ast.KindBlock,
+		ast.KindForStatement, ast.KindForInStatement, ast.KindForOfStatement,
+		ast.KindSwitchStatement, ast.KindCatchClause,
+		ast.KindClassDeclaration, ast.KindClassExpression,
+		ast.KindWithStatement, ast.KindClassStaticBlockDeclaration:
```
fansenze: **`createsScope` doesn't list `TSModuleDeclaration`, so `.then()`/`.catch()`/`.finally()` placed directly at the top level of a `namespace` / `module` body is silently un-reported — a false negative vs ESLint.**

`@typescript-eslint/scope-manager` opens a scope for `TSModuleDeclaration` (both `namespace N {}` and the legacy `module M {}`), the same way it does for blocks/loops/classes. The upstream rule reports whenever `getScope().block.type !== 'Program'`, but `createsScope` walks straight past a module body up to the source file and treats the position as bare top-level → skip.

This is a **narrow, low-priority** gap: only a `.then()` that isn't wrapped by any function/block triggers it (one inside a function in the namespace is already reported). Flagging for ESLint alignment, not blocking.

Measured — ESLint 8 + `@typescript-eslint/parser` 7 + `eslint-plugin-promise` 7 (A) vs this PR (B) vs the fix below (C), as `line:col`; both inputs are type-legal:

| input | A: ESLint | B: this PR | C: with fix |
|---|---|---|---|
| `namespace N { Promise.resolve().then(() => {}) }` | `1:33` | **(none)** | `1:33` |
| `module M { Promise.resolve().then(() => {}) }` | `1:30` | **(none)** | `1:30` |

(`enum` bodies open a scope too, but an enum member initializer can never be a Promise — `TS18033` — so there's no type-legal case to align; deliberately left out.)

All existing Go tests stay green with the fix applied.

**Fix:**

```go
	case ast.KindBlock,
		ast.KindForStatement, ast.KindForInStatement, ast.KindForOfStatement,
		ast.KindSwitchStatement, ast.KindCatchClause,
		ast.KindClassDeclaration, ast.KindClassExpression,
		ast.KindWithStatement, ast.KindClassStaticBlockDeclaration,
		ast.KindModuleDeclaration:
		return true
```

internal/plugins/promise/rules/prefer_await_to_then/prefer_await_to_then.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+// createsScope reports whether node opens a new eslint-scope scope.
+func createsScope(node *ast.Node) bool {
+	if ast.IsFunctionLike(node) {
+		return true
+	}
+	switch node.Kind {
+	case ast.KindBlock,
+		ast.KindForStatement, ast.KindForInStatement, ast.KindForOfStatement,
```
fansenze: **`createsScope` lists `For`/`ForIn`/`ForOf` unconditionally, but eslint-scope only opens a `for` scope when the loop head declares with `let`/`const` — so a `var` / declaration-less loop with a non-braced body is falsely reported at the top level.**

eslint-scope creates a `for` scope only when the head binds the loop variable with `let`/`const` (block scoping). With `var` or no declaration there is no scope, so `getScope().block` stays `Program` and the original rule skips. Here `createsScope` returns `true` for every for-loop, so `isTopLevelScoped` treats the body as scoped and reports. This only surfaces when the body is **not** a block — a braced body is caught by the `Block` case anyway, so those stay aligned.

Measured — ESLint 8 + `@typescript-eslint/parser` 7 + `eslint-plugin-promise` 7 (A) vs this PR (B) vs the fix below (C), as `line:col`, all at module top level:

| input | A: ESLint | B: this PR | C: with fix |
|---|---|---|---|
| `for(;;) thing.then(cb)` | `(none)` | **`1:15`** | `(none)` |
| `for(var i=0;;) thing.then(cb)` | `(none)` | **`1:22`** | `(none)` |
| `for(var a of xs) thing.then(cb)` | `(none)` | **`1:24`** | `(none)` |
| `for(a of xs) thing.then(cb)` | `(none)` | **`1:20`** | `(none)` |
| `for(var k in o) thing.then(cb)` | `(none)` | **`1:23`** | `(none)` |
| `for(let i=0;;) thing.then(cb)` | `1:22` | `1:22` | `1:22` |
| `for(const a of xs) thing.then(cb)` | `1:26` | `1:26` | `1:26` |
| `for(const k in o) thing.then(cb)` | `1:25` | `1:25` | `1:25` |

The `let`/`const` rows (C matches A) and braced/nested bodies stay reported — the fix only removes the `var` / declaration-less false positives. All existing Go tests remain green.

**Fix** (treat the three for-kinds as scope-creating only when the head is a `let`/`const` declaration list):

```go
switch node.Kind {
case ast.KindForStatement, ast.KindForInStatement, ast.KindForOfStatement:
	// eslint-scope only opens a for-scope when the loop head declares the
	// iteration variable with let/const; a var or expression head creates
	// no scope, so the loop body stays in the enclosing scope.
	init := node.Initializer()
	return init != nil && ast.IsVariableDeclarationList(init) &&
		init.Flags&(ast.NodeFlagsLet|ast.NodeFlagsConst) != 0
case ast.KindBlock,
	ast.KindSwitchStatement, ast.KindCatchClause,
	ast.KindClassDeclaration, ast.KindClassExpression,
	ast.KindWithStatement, ast.KindClassStaticBlockDeclaration,
	ast.KindModuleDeclaration:
	return true
}
return false
```

internal/plugins/promise/rules/prefer_await_to_then/prefer_await_to_then.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+var PreferAwaitToThenRule = rule.Rule{
+	Name: "promise/prefer-await-to-then",
+	Run: func(ctx rule.RuleContext, options any) rule.RuleListeners {
+		opts := parseOptions(options)
+		return rule.RuleListeners{
+			ast.KindCallExpression: func(node *ast.Node) {
+				callee := ast.SkipOuterExpressions(node.AsCallExpression().Expression, skipTransparent)
+				if callee == nil || !ast.IsPropertyAccessExpression(callee) {
```
fansenze: Confirmed this is an upstream-alignment miss. `eslint-plugin-promise@7.3.0` reports `function f(){ obj[then]() }` because the upstream rule listens on `CallExpression > MemberExpression.callee` and checks `node.property.name`, so computed identifier member access still matches.

Suggested fix: extract the callee property identifier from both access forms:

- `PropertyAccessExpression`: use `callee.AsPropertyAccessExpression().Name()`
- `ElementAccessExpression`: use `ast.SkipOuterExpressions(callee.AsElementAccessExpression().ArgumentExpression, skipTransparent)` and only accept it when it is an `Identifier`

This keeps the upstream boundary intact: report `obj[then]()` and `obj[(then)]()`, but do not report `obj['then']()` or ``obj[`then`]()`` because those are not identifier properties upstream.

I verified this shape with a temporary overlay test:
- invalid: `function f() { obj[then]() }`
- invalid: `function f() { obj[(then)]() }`
- valid: `function f() { obj['then']() }`
- valid: ``function f() { obj[`then`]() }``

The overlay run passed with the existing `prefer_await_to_then` tests.

swwind: Why `obj[then]()` also considered as promise call, made my mind explode! :exploding_head:

fansenze (approved):

PR #1060, opened 2026-06-05
feat: port rule promise/prefer-catch

swwind:
## Summary

Ported `promise/prefer-catch` rule for #474.

## Related Links

[original docs](https://github.com/eslint-community/eslint-plugin-promise/blob/main/docs/rules/prefer-catch.md)

## Checklist

### DIFF
--- a/internal/plugins/promise/rules/prefer_catch/prefer_catch.go
+++ b/internal/plugins/promise/rules/prefer_catch/prefer_catch.go
@@ -3,6 +3,7 @@ package prefer_catch
 import (
 	"github.com/microsoft/typescript-go/shim/ast"
 	"github.com/microsoft/typescript-go/shim/core"
+	"github.com/microsoft/typescript-go/shim/scanner"
 	"github.com/web-infra-dev/rslint/internal/rule"
 	"github.com/web-infra-dev/rslint/internal/utils"
 )
@@ -58,8 +59,10 @@ var PreferCatchRule = rule.Rule{
 					// Get text of arg1 stripping outer parens to match ESLint's getText behavior
 					innerArg1 := ast.SkipOuterExpressions(args[1], skipTransparent)
 					catcherText := utils.TrimmedNodeText(sf, innerArg1)
-					// Remove from end of arg0 (comma position) to end of arg1
-					removeRange := core.NewTextRange(args[0].End(), args[1].End())
+					// Remove from the comma after arg0 to end of arg1, so trivia
+					// between arg0 and the comma (e.g. a trailing comment) is kept.
+					commaStart := scanner.GetRangeOfTokenAtPosition(sf, args[0].End()).Pos()
+					removeRange := core.NewTextRange(commaStart, args[1].End())
 					ctx.ReportNodeWithFixes(thenName, msg,
 						rule.RuleFixRemoveRange(removeRange),
 						rule.RuleFixInsertBefore(sf, thenName, "catch("+catcherText+")."),
--- a/internal/plugins/promise/rules/prefer_catch/prefer_catch.go
+++ b/internal/plugins/promise/rules/prefer_catch/prefer_catch.go
@@ -1,0 +1,71 @@
+package prefer_catch
+
+import (
+	"github.com/microsoft/typescript-go/shim/ast"
+	"github.com/microsoft/typescript-go/shim/core"
+	"github.com/web-infra-dev/rslint/internal/rule"
+	"github.com/web-infra-dev/rslint/internal/utils"
+)
+
+const skipTransparent = ast.OEKParentheses
+
+func buildPreferCatchToThenMessage() rule.RuleMessage {
+	return rule.RuleMessage{
+		Id:          "preferCatchToThen",
+		Description: "Prefer `catch` to `then(a, b)`/`then(null, b)`.",
+	}
+}
+
+var PreferCatchRule = rule.Rule{
+	Name: "promise/prefer-catch",
+	Run: func(ctx rule.RuleContext, options any) rule.RuleListeners {
+		return rule.RuleListeners{
+			ast.KindCallExpression: func(node *ast.Node) {
+				callExpr := node.AsCallExpression()
+				callee := ast.SkipOuterExpressions(callExpr.Expression, skipTransparent)
+				if callee == nil || !ast.IsPropertyAccessExpression(callee) {
+					return
+				}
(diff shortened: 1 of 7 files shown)


internal/plugins/promise/rules/prefer_catch/prefer_catch_extras_test.go
```diff
@@ -0,0 +1,119 @@
...
+			{Code: `prom.then(fn)`},
+			{Code: `prom.then(null)`},
+			{Code: `prom.then(undefined)`},
+
+			// ---- Dimension 4: parenthesized callee is unwrapped ----
+			// (prom.then) is a ParenthesizedExpression wrapping the PAE; after
+			// SkipOuterExpressions the callee becomes the PAE and matches.
+			// This tests that calling (prom.then)(fn1, fn2) still is fine with no report
+			// since after skip it IS a PAE with name 'then' — actually this SHOULD report.
+			// N/A: there's no case where a parenthesized PAE callee should be excluded.
```
fansenze: This commented-out block is self-contradicting and factually wrong, so it reads as noise rather than documentation:

- It says `(prom.then)(fn1, fn2)` "still is fine with no report" and then immediately "actually this SHOULD report", concluding "N/A".
- Verified against both upstream and this port: `(prom.then)(fn1, fn2)` **does report**. Upstream (eslint-plugin-promise v7.3.0) flags it at `1:7`; this port reports it too and autofixes to `(prom.catch(fn2).then)(fn1)`.

Either drop the block (it documents a test that doesn't exist), or make it a real invalid case:

```go
// Parenthesized PAE callee: SkipOuterExpressions unwraps `(prom.then)` to the PAE, so it matches.
{
    Code:   `(prom.then)(fn1, fn2)`,
    Output: []string{`(prom.catch(fn2).then)(fn1)`},
    Errors: []rule_tester.InvalidTestCaseError{{MessageId: "preferCatchToThen", Line: 1, Column: 7}},
},
```

Also consider dropping the `Dimension 4` / `tsgo AST quirk` references in the file header — that's review-process jargon a future maintainer won't recognize.

internal/plugins/promise/rules/prefer_catch/prefer_catch.go
```diff
@@ -0,0 +1,71 @@
...
+						rule.RuleFixRemoveRange(core.NewTextRange(removeStart, removeEnd)),
+						rule.RuleFixReplace(sf, thenName, "catch"),
+					)
+				} else {
+					// hey.then(fn1, fn2) → hey.catch(fn2).then(fn1)
+					// Get text of arg1 stripping outer parens to match ESLint's getText behavior
+					innerArg1 := ast.SkipOuterExpressions(args[1], skipTransparent)
+					catcherText := utils.TrimmedNodeText(sf, innerArg1)
+					// Remove from end of arg0 (comma position) to end of arg1
+					removeRange := core.NewTextRange(args[0].End(), args[1].End())
```
fansenze: The removal range starts at `args[0].End()`, which swallows any trivia between arg0 and the comma. Upstream's `removeArgument` starts at the comma token (`getTokenBefore(catcher)`), so it preserves that trivia.

Verified divergence (upstream v7.3.0 vs this port):

| input | upstream | this port |
|---|---|---|
| `hey.then(fn1 /*x*/, fn2)` | `hey.catch(fn2).then(fn1 /*x*/)` | `hey.catch(fn2).then(fn1)` — comment dropped |
| `hey.then(fn1 , fn2)` | `hey.catch(fn2).then(fn1 )` | `hey.catch(fn2).then(fn1)` |

Both still produce valid code, so this is low severity, but it's a real diff from ESLint. Fix — start the removal at the comma token, keeping the end untouched so the 3-arg lock-in is unaffected:

```go
// Remove from the comma after arg0 to end of arg1
commaStart := scanner.GetRangeOfTokenAtPosition(sf, args[0].End()).Pos()
removeRange := core.NewTextRange(commaStart, args[1].End())
```

(add import `"github.com/microsoft/typescript-go/shim/scanner"`; reuses the same helper `TrimNodeTextRange` already relies on)

Verified: with this change `hey.then(fn1 /*x*/, fn2)` → `hey.catch(fn2).then(fn1 /*x*/)` (matches upstream), and the full upstream + extras suites still pass — including the 3-arg case `→ hey.catch(fn2).then(fn1, fn3)`.

internal/plugins/promise/rules/prefer_catch/prefer_catch.md
```diff
@@ -0,0 +1,33 @@
...
+
+```javascript
+prom.then(fn)
+prom.catch(handleErr).then(handle)
+prom.catch(handleErr)
+```
+
+## Differences from ESLint
+
+None known.
```
fansenze: `None known` isn't accurate — there's one verified difference on the 3-argument form:

`hey.then(fn1, fn2, fn3)`:
- upstream (v7.3.0) autofixes to the **broken** `hey.catch(fn2).then(fn1fn3)` (its `removeArgument` strips both commas around a middle arg)
- this port autofixes to the valid `hey.catch(fn2).then(fn1, fn3)`

The port's output is better, but it's still a difference worth documenting, e.g.:

> For the (unusual) 3-argument form `x.then(a, b, c)`, upstream's autofix produces invalid output (`x.catch(b).then(ac)`); this port produces valid output (`x.catch(b).then(a, c)`).

internal/plugins/promise/rules/prefer_catch/prefer_catch.go
swwind: Fixed — removal now starts at the comma token via `scanner.GetRangeOfTokenAtPosition`, so trivia like `hey.then(fn1 /*x*/, fn2)` now matches upstream's `hey.catch(fn2).then(fn1 /*x*/)`. Verified with a regression test.

internal/plugins/promise/rules/prefer_catch/prefer_catch_extras_test.go
swwind: Good catch — replaced the broken block with a real invalid case for `(prom.then)(fn1, fn2)` and dropped the review-jargon references throughout the file.

internal/plugins/promise/rules/prefer_catch/prefer_catch.md
swwind: Updated the docs to call out the 3-argument divergence explicitly, as suggested.

fansenze (approved):

PR #1065, opened 2026-06-07
feat: port rule promise/no-promise-in-callback

elecmonkey:
## Summary

promise/no-promise-in-callback

## Related Links

- [Document](https://github.com/eslint-community/eslint-plugin-promise/blob/main/docs/rules/no-callback-in-promise.md)
- [Code](https://github.com/eslint-community/eslint-plugin-promise/blob/main/rules/no-callback-in-promise.js)

## Checklist

### DIFF
--- a/internal/plugins/promise/rules/no_promise_in_callback/no_promise_in_callback.go
+++ b/internal/plugins/promise/rules/no_promise_in_callback/no_promise_in_callback.go
@@ -1,0 +1,127 @@
+package no_promise_in_callback
+
+import (
+	"github.com/microsoft/typescript-go/shim/ast"
+	"github.com/web-infra-dev/rslint/internal/plugins/promise/promiseutil"
+	"github.com/web-infra-dev/rslint/internal/rule"
+	"github.com/web-infra-dev/rslint/internal/utils"
+)
+
+const skipTransparent = ast.OEKParentheses
+
+type Options struct {
+	ExemptDeclarations bool
+}
+
+func parseOptions(options any) Options {
+	opts := Options{}
+	optsMap := utils.GetOptionsMap(options)
+	if optsMap == nil {
+		return opts
+	}
+	if value, ok := optsMap["exemptDeclarations"].(bool); ok {
+		opts.ExemptDeclarations = value
+	}
+	return opts
+}
+
+func messageAvoidPromiseInCallback() rule.RuleMessage {
+	return rule.RuleMessage{
+		Id:          "avoidPromiseInCallback",
+		Description: "Avoid using promises inside of callbacks.",
+	}
+}
+
+func isReturnExpression(node *ast.Node) bool {
+	current := ast.WalkUpParenthesizedExpressions(node.Parent)
+	return current != nil && current.Kind == ast.KindReturnStatement
+}
+
+func isPromiseCallback(node *ast.Node) bool {
+	if node == nil || (node.Kind != ast.KindFunctionExpression && node.Kind != ast.KindArrowFunction) {
+		return false
+	}
+	parent := ast.WalkUpParenthesizedExpressions(node.Parent)
+	if parent == nil || !ast.IsCallExpression(parent) {
+		return false
+	}
+	call := parent.AsCallExpression()
+	if call == nil || call.Arguments == nil || len(call.Arguments.Nodes) == 0 {
+		return false
+	}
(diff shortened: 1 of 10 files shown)


packages/rslint/src/configs/promise.ts
```diff
@@ -11,7 +11,7 @@ const recommended: RslintConfigEntry = {
     // 'promise/catch-or-return': 'error', // not implemented
     // 'promise/no-native': 'off', // not implemented
     // 'promise/no-nesting': 'warn', // not implemented
-    // 'promise/no-promise-in-callback': 'warn', // not implemented
+    'promise/no-promise-in-callback': 'warn',
```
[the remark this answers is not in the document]
elecmonkey: the inconsistent introduced by `catch-or-return`'s porting PR. should not fix in this PR.

internal/plugins/promise/rules/no_promise_in_callback/no_promise_in_callback.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+	return ""
+}
+
+func isCallbackContainer(node *ast.Node, opts Options) bool {
+	if node == nil || isPromiseCallback(node) {
+		return false
+	}
+
+	isFunction := node.Kind == ast.KindFunctionExpression || node.Kind == ast.KindArrowFunction ||
```
fansenze, line 77: **False negative: object/class methods aren't treated as callbacks.**

Diff vs real `eslint-plugin-promise@7.3.0`:

| code | upstream | this PR |
|---|---|---|
| `const o = { onError(err) { Promise.resolve(err) } }` | 1 | 0 |
| `class X { onError(error) { audit().catch(log) } }` | 1 | 0 |

In ESTree a method body is a `FunctionExpression`, so upstream flags methods whose first param is `err`/`error`. tsgo gives `MethodDeclaration`/`GetAccessor`/`SetAccessor`/`Constructor`, which this check skips.

```go
isFunction := node.Kind == ast.KindFunctionExpression || node.Kind == ast.KindArrowFunction ||
	node.Kind == ast.KindMethodDeclaration || node.Kind == ast.KindGetAccessor ||
	node.Kind == ast.KindSetAccessor || node.Kind == ast.KindConstructor ||
	(!opts.ExemptDeclarations && node.Kind == ast.KindFunctionDeclaration)
```

Verified: with this change the cases above report 1 (matching upstream) and `get x() { return p.then() }` stays valid — keeps `Differences from ESLint: None known` accurate.

internal/plugins/promise/rules/no_promise_in_callback/no_promise_in_callback.go
```diff
@@ -0,0 +1,127 @@
...
+		Description: "Avoid using promises inside of callbacks.",
+	}
+}
+
+func isReturnExpression(node *ast.Node) bool {
+	current := ast.WalkUpParenthesizedExpressions(node.Parent)
+	return current != nil && current.Kind == ast.KindReturnStatement
+}
+
+func isPromiseCallback(node *ast.Node) bool {
```
fansenze, line 40: Reuse `promiseutil.IsMemberCall` here:

```go
func isPromiseCallback(node *ast.Node) bool {
	if node == nil || (node.Kind != ast.KindFunctionExpression && node.Kind != ast.KindArrowFunction) {
		return false
	}
	parent := ast.WalkUpParenthesizedExpressions(node.Parent)
	return promiseutil.IsMemberCall(parent, "then") || promiseutil.IsMemberCall(parent, "catch")
}
```

The `len(args) == 0` guard is redundant — if the callee is a `.then`/`.catch` member, the function is necessarily an argument (≥1 arg).

fansenze:
The **Related Links** point to `no-callback-in-promise` (a different rule). Should be `no-promise-in-callback`:
- docs: https://github.com/eslint-community/eslint-plugin-promise/blob/main/docs/rules/no-promise-in-callback.md
- code: https://github.com/eslint-community/eslint-plugin-promise/blob/main/rules/no-promise-in-callback.js

fansenze:
@elecmonkey Thanks for the updates! The code changes look good to me now.

Before we merge, try to rebase this PR onto the latest `main` and resolve the current conflicts. After that, I can take another quick look and we should be good to go.

fansenze (approved):

PR #1066, opened 2026-06-07
feat(promise): port rule no-native

elecmonkey:
## Summary

port rule promise/no-native

## Related Links

- [Document](https://github.com/eslint-community/eslint-plugin-promise/blob/main/docs/rules/no-native.md)
- [Code](https://github.com/eslint-community/eslint-plugin-promise/blob/main/rules/no-native.js)

## Checklist


fansenze (commented):
**`no-native` doesn't separate the value/type namespaces when deciding whether `Promise` is shadowed.** Verdicts below are from running upstream `eslint-plugin-promise@7` (its suite runs espree **and** `@typescript-eslint/parser`), so they hold for both JS and TS:

| code | upstream | this PR |
|---|---|---|
| `interface Promise {} Promise.resolve(1)` | report | **missed** |
| `let x: Promise` · `x as Promise` · `typeof Promise` | report | **missed** |
| `class C extends Promise {}` · `implements Promise<T>` | report | **missed** |
| `var Promise = 1; let y: Promise` | report | **missed** |

Root cause: a `Promise` reference must be matched against a binding **in the same namespace** — value refs against value declarations, type refs against type declarations. The rule currently excludes type positions entirely (`IsPartOfTypeNode || IsPartOfTypeQuery`) and, for value refs, lets a type-only declaration count as a user `Promise`.

Direction (`no_native.go`):

1. Drop the type-position early-return; split on `isTypeRef := IsPartOfTypeNode(node) && !IsPartOfTypeQuery(node)` (the `typeof` operand references the *value*, so it stays a value ref).
2. **value ref**: in `hasNonDefaultLibraryDeclaration`, skip `IsInterfaceDeclaration`/`IsTypeAliasDeclaration` — they merge into the global `Promise` symbol but bind no value.
3. **type ref**: resolve shadowing **syntactically** against type-binding declarations only (`type`/`interface`/`class`/`enum`/`namespace`/`import`/type-param) — ideally a `utils.IsTypeShadowed` mirroring `IsShadowed`. Two traps I verified: (a) don't reuse the value-namespace `IsShadowed`, or `var Promise` wrongly silences a type ref; (b) don't use the checker/symbol here — when a local decl collides with the lib `Promise` (e.g. `type Promise = string`) the merged symbol points back at the lib, so only a syntactic walk stays correct.

Tests: `no_native_extras_test.go` lists `type T = typeof Promise;` as **valid**, but upstream reports it (col 17-24) — that case currently encodes the miss. Move it to invalid.

Docs: trim the `## Differences from ESLint` note ("rslint's native Go rule tester", env/globals) to a short user-facing line — the rest is framework-internal.

I verified a complete fix against 36 differential cases (value/type refs, every shadow namespace, nested scopes, `declare global`, heritage clauses) — all align with upstream. Can share the patch.

swwind:
Would you mind updating this branch with the latest origin/main before we merge?

internal/utils/type_shadowing.go
```diff
@@ -0,0 +1,155 @@
...
+// hasTypeParameterNamed reports whether `node` declares a type parameter named
+// `name`. TypeParameters() is only valid for type-parameterized declarations,
+// so guard the call accordingly.
+func hasTypeParameterNamed(node *ast.Node, name string) bool {
+	switch node.Kind {
+	case ast.KindClassDeclaration, ast.KindClassExpression,
+		ast.KindInterfaceDeclaration, ast.KindTypeAliasDeclaration:
+		// always type-parameterizable
+	default:
+		if !ast.IsFunctionLikeDeclaration(node) {
```
swwind: `ast.IsFunctionLikeDeclaration` excludes the signature-like kinds (`MethodSignature`, `CallSignature`, `ConstructSignature`, `FunctionType`, `ConstructorType`), so type parameters declared on them are not seen here. Verified false positives — each of these reports `"Promise" is not defined.` even though the reference resolves to a local type binding:

```ts
interface I { m<Promise>(x: Promise): void }
type F = <Promise>(x: Promise) => void;
type M<T> = { [Promise in keyof T]: Promise };
type U<T> = T extends Array<infer Promise> ? Promise : never;
const C = class Promise { m(): Promise { return this; } };
```

(For contrast, `class Outer { m<Promise>(x: Promise) {} }` works, because `MethodDeclaration` *is* a FunctionLikeDeclaration — so the guard is just narrower than intended.)

Suggested fix: use `ast.IsFunctionLike` here — the signature kinds embed `FunctionLikeBase`, so `TypeParameters()` is safe to call on them — and add a `KindMappedType` case for the mapped-type parameter. The `infer` and class-expression-name cases are harder to cover with a syntactic walk; leaving them as known limitations documented in `no_native.md` seems fine given how unlikely a type parameter named `Promise` is.

internal/utils/type_shadowing.go
```diff
@@ -0,0 +1,155 @@
...
+			}
+		}
+	}
+	return false
+}
+
+// importBindsName reports whether an import declaration binds `name` (default,
+// namespace, or named import). Imports can introduce a type binding, so they
+// count for type-namespace shadowing.
+func importBindsName(stmt *ast.Node, name string) bool {
```
swwind: `importBindsName` is a near-verbatim copy of the import branch in `HasLocalDeclarationInStatements` (`shadowing.go`), and the `ModuleDeclaration` / `ImportEqualsDeclaration` cases in `hasTypeBindingInStatements` duplicate that function's branches as well. Please extract a shared helper (e.g. `importDeclarationBindsName(stmt, name)`) and call it from both places — otherwise future fixes (e.g. `import type` handling) have to land twice.

internal/plugins/promise/rules/no_native/no_native.go
```diff
@@ -0,0 +1,90 @@
...
+		return !utils.IsTypeShadowed(node, promiseName)
+	}
+
+	// Value reference.
+	if ctx.TypeChecker != nil {
+		symbol := utils.GetReferenceSymbol(node, ctx.TypeChecker)
+		if symbol != nil && hasNonDefaultLibraryValueDeclaration(ctx, symbol) {
+			return false
+		}
+		return !utils.IsShadowed(node, promiseName)
```
swwind: Minor: `return !utils.IsShadowed(node, promiseName)` is duplicated inside and outside the checker branch. Can be flattened to:

```go
if ctx.TypeChecker != nil {
	if symbol := utils.GetReferenceSymbol(node, ctx.TypeChecker); symbol != nil && hasNonDefaultLibraryValueDeclaration(ctx, symbol) {
		return false
	}
}
return !utils.IsShadowed(node, promiseName)
```

Also worth noting: the checker path resolves declarations **across files** — a `declare var Promise` in another file of the project suppresses reports in this file, whereas ESLint analyzes per-file and would still report. Consider adding a line about this to the "Differences from ESLint" section of `no_native.md`.

swwind (commented):
Reviewed with focus on function reuse, edge cases, and 1:1 alignment with ESLint. Upstream's valid/invalid cases are all covered and the Go tests pass locally. Problems found and suggested changes are in the inline comments; two more notes below.

**Behavioral divergence from upstream (please confirm it's intended):** upstream runs this rule under `@typescript-eslint/parser` too, and ts-eslint's scope-manager resolves references by name without separating value/type namespaces — so `interface Promise {} Promise.resolve(1)` and `var Promise = 1; let y: Promise` are *valid* upstream but reported here. The namespace-aware behavior is more TS-correct and `no_native.md` documents it, so this is likely fine — just flagging it as a conscious deviation from strict 1:1.

**Test suggestion:** add an invalid case like `/* global Promise */ Promise.resolve()`. Upstream explicitly asserts that `globals: { Promise: true }` / `env: es6` do **not** exempt the report, and the rule correctly ignores `ctx.Globals` today — a regression test would lock that semantic in before anyone wires globals into this rule later.

internal/utils/type_shadowing.go
elecmonkey: Fixed in 93edf49b. Module, import-equals, and import declaration binding checks now live in shared helpers in `shadowing.go`, and both value-namespace and type-namespace statement walks call those helpers.

internal/utils/type_shadowing.go
elecmonkey: Fixed in 93edf49b. `IsTypeShadowed` now uses `ast.IsFunctionLike`, covers mapped-type parameters, and also handles the two suggested known limitations directly: `infer` bindings in conditional true branches and class-expression self names. The Go regression suite includes all five examples from this thread plus the additional signature-like forms.

internal/plugins/promise/rules/no_native/no_native.go
elecmonkey: Fixed in 93edf49b. The duplicated `IsShadowed` return was flattened as suggested. The Differences from ESLint section now documents project-wide declaration resolution, including the cross-file `declare var Promise` behavior.

elecmonkey:
Follow-up on the review summary: the value/type namespace separation is intentional and remains documented as a TypeScript-specific difference from ESLint. I also added regression coverage proving that both `/* global Promise */` and configured `globals: { Promise: true }` do not exempt native `Promise` references. All focused tests, the full Go suite, RuleTester, golangci-lint, and spell checking pass locally.

swwind:
Sorry for the late reply. We've made a lot of changes to ctx.Globals and ctx.Refs API internally. Would you like to rebase this to use those API?

elecmonkey:
> Sorry for the late reply. We've made a lot of changes to ctx.Globals and ctx.Refs API internally. Would you like to rebase this to use those API?

https://github.com/web-infra-dev/rslint/issues/474#issuecomment-5213217943

if we decide to port it, i'll open a new PR

PR #1076, opened 2026-06-09
feat(eslint-plugin-jest): add `max-expects` rule

eryue0220:
## Summary

* Port `max-expects` from eslint-plugin-jest to rslint
* Fix `max-nested-describe` option parse logic bug

## Related Links

eslint-plugin-jest/max-expects [doc](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/max-expects.md) [code](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/max-expects.ts)

## Checklist


packages/rslint-test-tools/rstest.config.mts
```diff
@@ -432,6 +432,7 @@ export default defineConfig({
 
     // eslint-plugin-jest
     './tests/eslint-plugin-jest/rules/expect-expect.test.ts',
+    './tests/eslint-plugin-jest/rules/max-expect.test.ts',
```
fansenze: The test file is named `max-expect.test.ts` but the rule is `max-expects`. This is not just a naming nit: `scripts/gen-rule-manifest.js` derives the expected test path from the rule directory name (`max_expects` → `max-expects.test.ts`), so with the current name the generated rule manifest marks this rule as `"status": "partial-test"` (no tests) even though tests exist and run. Verified by running `buildManifest()` locally — `max-expects` comes out as `partial-test` while `max-nested-describe` (correctly named) is `full`. All other 28 test files in this group match their rule names.

Please rename the file to `max-expects.test.ts` and update this entry.

internal/plugins/jest/rules/max_expects/max_expects.go
```diff
@@ -0,0 +1,123 @@
...
+		Data: map[string]string{
+			"count": strconv.Itoa(count),
+			"max":   strconv.Itoa(maxAllowed),
+		},
+	}
+}
+
+func isTestCallbackFunction(fn *ast.Node, ctx rule.RuleContext) bool {
+	parent := fn.Parent
+	if parent == nil || parent.Kind != ast.KindCallExpression {
```
fansenze, line 68: This parent check doesn't skip `ParenthesizedExpression`, which causes a false negative vs upstream. ESTree (upstream) has no parenthesized-expression nodes, but the tsgo AST keeps them — so for a parenthesized callback passed to a non-test call, `fn.Parent` is the paren node, this function returns `true`, and the counter gets reset where upstream keeps counting:

```js
test('x', () => {
  expect(1).toBe(1);
  expect(1).toBe(1);
  expect(1).toBe(1);
  run((() => { // note the extra parens
    expect(1).toBe(1);
    expect(1).toBe(1);
    expect(1).toBe(1); // upstream: error (6 > 5) — this port: no diagnostic
  }));
});
```

Verified by running both on this snippet: eslint-plugin-jest reports `Too many assertion calls (6) - maximum allowed is 5`, this port reports nothing. (`it('x', (() => {...}))` — parens directly around a test callback — behaves the same on both sides, so the divergence is limited to non-test calls.)

Suggested fix — walk up through parens before the kind check:

```go
parent := fn.Parent
for parent != nil && parent.Kind == ast.KindParenthesizedExpression {
	parent = parent.Parent
}
if parent == nil || parent.Kind != ast.KindCallExpression {
	return true
}
return utils.IsTypeOfJestFnCall(parent, ctx, utils.JestFnTypeTest)
```

eryue0220: Updated.

internal/plugins/jest/rules/max_expects/max_expects.go
```diff
@@ -0,0 +1,126 @@
...
+			if isTestCallbackFunction(node, ctx) {
+				count = 0
+			}
+		}
+
+		return rule.RuleListeners{
+			ast.KindFunctionExpression:                      maybeResetCount,
+			rule.ListenerOnExit(ast.KindFunctionExpression): maybeResetCount,
+			ast.KindArrowFunction:                           maybeResetCount,
+			rule.ListenerOnExit(ast.KindArrowFunction):      maybeResetCount,
```
fansenze, line 110: The reset listeners only cover `FunctionExpression` / `ArrowFunction`, which misses several function forms and causes false positives vs upstream.

In ESTree, the body of an object/class method, getter/setter, and constructor is **also** a `FunctionExpression` node, so upstream's `FunctionExpression` listener fires for them and resets the counter. In the tsgo AST these are separate kinds (`MethodDeclaration` / `GetAccessor` / `SetAccessor` / `Constructor`) that this rule doesn't listen to — the counter just keeps accumulating across the method boundary:

```js
test('x', () => {
  expect(1).toBe(1);     // upstream: 1   this port: 1
  expect(1).toBe(1);     // upstream: 2   this port: 2
  expect(1).toBe(1);     // upstream: 3   this port: 3
  const obj = {
    m() {                // upstream: reset to 0   this port: no reset
      expect(1).toBe(1); // upstream: 1   this port: 4
      expect(1).toBe(1); // upstream: 2   this port: 5
      expect(1).toBe(1); // upstream: 3   this port: 6 → "Too many assertion calls (6)"
    },
  };
});
```

Verified by running both on this snippet: eslint-plugin-jest (29.15.2) reports nothing, this port reports `Too many assertion calls (6) - maximum allowed is 5`. Same divergence for getters, setters, class methods, and constructors (all five shapes verified both ways).

Suggested fix — register the same reset for these kinds, but only when the node has a body. Signature-only forms (abstract methods, overload signatures) are `TSEmptyBodyFunctionExpression` in ESTree, which upstream's `FunctionExpression` listener does *not* match, so they must not reset:

```go
maybeResetCountForMethod := func(node *ast.Node) {
	if node.Body() != nil {
		maybeResetCount(node)
	}
}
```

```go
ast.KindMethodDeclaration:                       maybeResetCountForMethod,
rule.ListenerOnExit(ast.KindMethodDeclaration):  maybeResetCountForMethod,
ast.KindGetAccessor:                             maybeResetCountForMethod,
rule.ListenerOnExit(ast.KindGetAccessor):        maybeResetCountForMethod,
ast.KindSetAccessor:                             maybeResetCountForMethod,
rule.ListenerOnExit(ast.KindSetAccessor):        maybeResetCountForMethod,
ast.KindConstructor:                             maybeResetCountForMethod,
rule.ListenerOnExit(ast.KindConstructor):        maybeResetCountForMethod,
```

Verified this patch locally: the five false-positive shapes stop reporting, the existing test suite passes unchanged, and an abstract-method case (`abstract m(): void;` followed by `p = expect(1).toBe(1)` as the 6th call — which upstream does report) still reports.

fansenze (approved):

PR #1096, opened 2026-06-15
ci: windows-latest hang diagnostics (debug, do not merge)

fansenze:
## Why

`Test npm packages (windows-latest)` intermittently hangs until the job
timeout (e.g. run 27521879692: a 3-line type-aware lint sat for 30 min, only
ending on a manual cancel). The whole chain is unbounded — the Go lint phase,
the IPC request, the engine's `await childExit`, the test helper, and rstest's
`--testTimeout=0` — so one wedged child eats the entire job.

~6800 faithful local runs on macOS (race build, single- and multi-file
fixtures, GOMAXPROCS 4 and 10) did **not** reproduce it → the deadlock looks
Windows-specific. Docker can't help (no Windows containers on macOS; the runner
is a Windows Server VM, not a container), so the only faithful repro is the
runner itself. The Go child is killed by SIGTERM with no goroutine traceback,
so CI logs never show where it wedges — this PR fixes that.

## What

Opt-in, env-gated diagnostics (a complete no-op unless `RSLINT_HANG_*` is set),
so **one** windows-latest run can capture *where* the Go side is stuck:

| Piece | Role |
|---|---|
| `cmd/rslint/hangdiag.go` | lifecycle phase trace + watchdog that dumps **all** goroutine stacks and exits 99 past a deadline; on-demand dump via SIGQUIT (unix) / SIGBREAK (Windows, no SIGQUIT) |
| `packages/rslint/src/engine.ts` | host-side phase trace + backstop watchdog (Node-side state, forces the child down) |
| `scripts/hang-hunt.mjs` | concurrent CLI-spawn loop over the failing fixture shape, wall-clock budget, fails with the captured dump |
| `.github/workflows/debug-hang.yml` | build `rslint.exe` + loop on windows-latest (CLI-spawn matrix + a real rstest loop); upload goroutine dumps as artifacts |

Verified locally end-to-end: a forced watchdog dumped all 10 goroutines
(including the typescript-go checker-pool ones) with file:line, wrote the dump
file, exited 99, and the host passed it through. Default-off path leaks nothing.

## How to read a run

- **Red `Hunt (...)` job** = reproduced. The failing step prints the Go phase
  trace (last marker = the wedged phase) + the full goroutine dump (the
  deadlock site). Same dump is in the `hang-dumps-*` artifact.
- **Green** = not reproduced this run (bump iterations / re-run).

> Debug-only. **Do not merge** — revert once the root cause is found.

### DIFF
--- a/.github/workflows/debug-hang.yml
+++ b/.github/workflows/debug-hang.yml
@@ -1,0 +1,165 @@
+# Debug-only workflow: hunt the intermittent windows-latest CLI hang and
+# capture WHY in one run. Builds rslint.exe with the (env-gated) hang
+# diagnostics from cmd/rslint/hangdiag.go, then loops many concurrent
+# type-aware lints on windows-latest with the Go watchdog armed. When an
+# invocation wedges, the watchdog dumps every goroutine stack (the deadlock
+# site) to stderr + a file, the loop fails the job, and the dumps are uploaded
+# as an artifact. Delete this workflow + the diagnostics once the root cause is
+# found.
+name: Debug Windows Hang
+
+on:
+  push:
+    branches: ['debug/windows-hang-diagnostics']
+  workflow_dispatch:
+    inputs:
+      rounds:
+        description: max rounds per hunt leg (a 70-min wall-clock budget caps it first)
+        default: '100000'
+
+permissions:
+  contents: read
+
+concurrency:
+  group: debug-hang-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build:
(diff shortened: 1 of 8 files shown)


fansenze:
Root cause found and fixed in #1098 (Windows IPC channel close deadlock — closeWith blocking on an uninterruptible stdin ReadFile). These diagnostics did their job: they captured the goroutine dump that localized it. Fix verified by re-running this same hunt with the patch — ~60k windows-latest invocations, 0 hangs. Closing.

PR #1104, opened 2026-06-16
test: add CLI-driven alignment tests for community ESLint plugins

fansenze:
## Summary

CLI-driven **alignment tests** for three community ESLint plugins that have no native Go port — their upstream rule-test suites are ported verbatim and the **real plugins run through the rslint CLI**.

The `lint()` JS API can't drive them: it spawns `rslint --api`, a pure-Go server that only executes native rules and silently ignores eslint-plugin rules (`cmd/rslint/api.go`). Only the CLI loads the real plugin in a Node worker pool.

## Coverage — 104 rules

| Plugin | Version | Rules |
|---|---|---|
| `@stylistic/eslint-plugin` | v5.10.0 | 93 (4 TS-spacing rules excluded — already covered natively) |
| `eslint-plugin-simple-import-sort` | v13.0.0 | 2 |
| `@eslint-community/eslint-plugin-eslint-comments` | v4.7.2 | 9 |

## How

- **Shared CLI RuleTester** (`tests/src/util/eslint-plugin-rule-tester.ts`): per rule, writes one fixture per case + a generated `rslint.config.mjs` mounting the live plugin, runs the CLI once, and asserts diagnostic count + rendered message (`messageId` resolved via the plugin's `meta.messages` + `data`) + line/column + autofix `output`. Three thin per-plugin wrappers bind `pkg`/`prefix`/`plugin`.
- Cases are ported **verbatim** from upstream tags; assertions are exact (no loose / count-only matches), verified non-vacuous.

## KNOWN GAPS (real rslint↔upstream divergences — isolated per file, never silenced into false greens)

- ts-go strict/module **syntax errors** that upstream parses in sloppy/Babel mode (octal & leading-zero literals, `assert` import attributes, `do`-expressions, Flow syntax).
- **Multi-pass vs single-pass autofix**: rslint fixes to a stable point, upstream's RuleTester is a single pass — diagnostics match, only the fixed \`output\` differs.
- **Babel/Flow-only** parser cases.
- eslint-comments **CSS-language** / **eslint-env** (removed in ESLint v10) / **no-unused-disable** (patches \`Linter#verify\` + \`reportUnusedDisableDirectives\` — mechanism absent in rslint).

## Fix

The eslint-plugin \`SourceCode\` shim was missing \`isESTree\` (ESLint sets \`isESTree = ast.type === \"Program\"\`). \`@stylistic/indent\` — the only rule guarding \`isESTreeSourceCode\` — therefore registered zero listeners and emitted **no diagnostics** under rslint. Added the flag, mirroring ESLint's exact predicate.


fansenze:
Superseded: switching to differential conformance — compare each plugin rule against live ESLint v10 under `eslint-plugin-conformance/` — instead of porting the full upstream RuleTester suites. Split into focused PRs: isESTree fix #1111, then per-plugin conformance. Branch kept as a case source.

PR #1124, opened 2026-06-22
feat(eslint-plugin-jest): add `no-export` rule

eryue0220:
## Summary

Port `no-export` from eslint-plugin-jest to rslint

## Related Links

eslint-plugin-jest/no-export [doc](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-export.md) [code](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/no-export.ts)

## Checklist

### DIFF
--- a/internal/plugins/jest/rules/no_export/no_export.go
+++ b/internal/plugins/jest/rules/no_export/no_export.go
@@ -1,0 +1,114 @@
+package no_export
+
+import (
+	"github.com/microsoft/typescript-go/shim/ast"
+	"github.com/web-infra-dev/rslint/internal/plugins/jest/utils"
+	"github.com/web-infra-dev/rslint/internal/rule"
+	rslintutils "github.com/web-infra-dev/rslint/internal/utils"
+)
+
+func buildUnexpectedExportMessage() rule.RuleMessage {
+	return rule.RuleMessage{
+		Id:          "unexpectedExport",
+		Description: "Do not export from a test file",
+	}
+}
+
+func isModuleExportsMemberExpression(node *ast.Node, ctx rule.RuleContext) bool {
+	if node == nil || !utils.IsMemberAccessNode(node) {
+		return false
+	}
+
+	current := node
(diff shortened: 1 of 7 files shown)


internal/plugins/jest/rules/no_export/no_export.go
```diff
@@ -0,0 +1,114 @@
...
+		property, ok := rslintUtils.AccessExpressionStaticName(current)
+		if !ok {
+			return false
+		}
+		innermostProperty = property
+		current = ast.SkipParentheses(rslintUtils.AccessExpressionObject(current))
+	}
+
+	if current == nil || current.Kind != ast.KindIdentifier ||
+		current.AsIdentifier().Text != "module" || innermostProperty != "exports" {
```
fansenze: **False negative: the bare `exports` alias is never flagged.**

This check requires the root of the member chain to be `module`, so any CommonJS export written through the bare `exports` alias slips through:

```js
module.exports.foo = 1;   // reported ✓

exports.foo = 1;          // NOT reported ✗ — identical CommonJS export
exports["foo"] = 1;       // NOT reported ✗
```

`exports` is the standard alias of `module.exports`, and `exports.foo = …` is one of the most common CommonJS export forms — arguably more common than the `module["exports"]` / deeply-nested cases this rule already goes out of its way to cover, so skipping it is inconsistent.

**Suggestion:** also treat a chain rooted at a bare `exports` identifier as an export (covering `exports.foo`, `exports["foo"]`, `exports.a.b`), and reuse the existing local-shadow guard so `const exports = {}; exports.foo = 1` stays valid. The bare-`exports` branch shouldn't reuse the `innermostProperty == "exports"` condition (that's specific to `module.exports`); note `exports = {}` is a plain identifier assignment — not a member access — so it won't be falsely caught.

eryue0220: Updated.

fansenze (approved):

PR #1136, opened 2026-06-23
feat(rslint): add lintText convenience API for in-memory linting

### LINKED ISSUE #1106: [Feature]: lint in-memory files via the Node API (no disk write)
### What problem does this feature solve?

I run an LLM coding agent (Node/TS) that generates code and lints file changes before *before* storing them in the databse. The point is to catch the dumb render-breaking stuff inside the generation loop especially conditional hooks or const reassignment

Afaik today rslint only supports linting real files on disk

Current linters with in-memory:
- eslint: `lintText(code, { filePath })`, works
- biome: in-process Node API, `lintContent(projectKey, content, { filePath })`, works
- oxlint: disk only
- rslint: disk only

We are using Biome for this because it's the only fast linter with a real in-process Node API right now but I would rather be on rslint to not mix both stacks

### What does the proposed API of configuration look like?

Maybe mirror Biome's shape somehow like this (or even better to support ts-go or other cross file lintings)

```ts
import { Rslint } from "@rslint/core";

const rslint = new Rslint();
const { projectKey } = rslint.openProject("/");
rslint.applyConfiguration(projectKey, config);

const { diagnostics } = rslint.lintContent(projectKey, source, {
  filePath: "virtual/Component.tsx", // virtual, never written to disk
});
```

here's my current Biome setup:

```ts
import { Biome, type Configuration } from "@biomejs/js-api/nodejs";

const biome = new Biome();
const { projectKey } = biome.openProject("/");
biome.applyConfiguration(projectKey, config);

const { diagnostics } = biome.lintContent(projectKey, content, { filePath });
// map diagnostics -> { path, line, column, rule, severity, message }
```

mvanhorn:
## Summary

Adds a `lintText()` convenience API for linting an in-memory source string without writing it to disk, mirroring ESLint's `lintText(code, { filePath })` and Biome's `lintContent`. This is the in-generation-loop entry point requested in #1106 for tools (like LLM coding agents) that want to lint freshly generated source before it ever touches the filesystem.

The change is purely an ergonomic TypeScript surface over the existing `LintOptions.fileContents` VFS map that the Go server already honors (the same plumbing exercised by the existing "virtual file support" API test). No Go/server changes are required.

What's added:

- **`lintText(code, options?)`** standalone convenience function in `src/index.ts`, mirroring the existing `lint()` / `applyFixes()` / `getAstInfo()` wrappers: it spins up a service, lints, and closes it (now in a `finally` so the spawned `rslint --api` subprocess is always released, even on a rejected lint).
- **`RSLintService.lintText(code, options?)`** method in `src/service.ts` for callers holding a long-lived service who want to avoid per-call spin-up. It builds a single-entry VFS request (`files: [filePath]`, `fileContents: { [filePath]: code }`) and delegates to the existing `lint()` path. `filePath` defaults to a virtual `virtual.tsx` when omitted.
- **`LintTextOptions`** public type in `src/types.ts` (re-exported from `index.ts` and `service.ts`).
- Structural test coverage in `tests/lint-text.test.mjs` (happy path, default `filePath`, clean code, and service-vs-standalone parity), following the fixture conventions of the existing `tests/api.test.mjs`.

Example:

```ts
import { lintText } from '@rslint/core';

const { diagnostics } = await lintText('const a = 1; a = 2;', {
  filePath: 'virtual/Component.tsx',
  config: './rslint.json',
});
```

### Known limitations

- When `filePath` is omitted (or set to a path not matched by the configured tsconfig's `include` globs), diagnostics are scoped to that virtual path; the Go program is still built from tsconfig roots, so a virtual file outside those roots may produce zero diagnostics. Passing a `filePath` that the config already includes (the documented usage) lints as expected. Fully decoupling the virtual entry from tsconfig roots would require a server-side change and is intentionally out of scope here.

## Related Links

Closes #1106

## Checklist

### DIFF
--- a/packages/rslint/src/index.ts
+++ b/packages/rslint/src/index.ts
@@ -1,6 +1,7 @@
 import { NodeRslintService } from './node.js';
 import {
   LintOptions,
+  LintTextOptions,
   LintResponse,
   RSLintService,
   ApplyFixesRequest,
@@ -41,6 +42,25 @@ export async function lint(options: LintOptions): Promise<LintResponse> {
   return result;
 }
 
+// Convenience function for in-memory linting (no disk write), mirroring ESLint's lintText
+export async function lintText(
+  code: string,
+  options: LintTextOptions = {},
+): Promise<LintResponse> {
+  const service = new RSLintService(
+    new NodeRslintService({
+      workingDirectory: options.workingDirectory,
+    }),
+  );
+  let result: LintResponse;
+  try {
+    result = await service.lintText(code, options);
+  } finally {
+    await service.close();
+  }
+  return result;
+}
+
 // Convenience function for applying fixes
 export async function applyFixes(
   options: ApplyFixesRequest,
@@ -65,6 +85,7 @@ export async function getAstInfo(
 export {
   type Diagnostic,
   type LintOptions,
+  type LintTextOptions,
   type LintResponse,
   type ApplyFixesRequest,
   type ApplyFixesResponse,
--- a/packages/rslint/src/service.ts
+++ b/packages/rslint/src/service.ts
@@ -1,6 +1,7 @@
 import type {
   RslintServiceInterface as RslintServiceBackend,
   LintOptions,
+  LintTextOptions,
   LintResponse,
   ApplyFixesRequest,
   ApplyFixesResponse,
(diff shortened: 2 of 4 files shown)


fansenze:
@mvanhorn Thanks a lot for the PR and for taking the time to work on this. We already have an existing change that covers the same functionality: https://github.com/web-infra-dev/rslint/pull/1122.
We’re also planning a broader refactor of the `Rslint` Node API afterwards(align with ESLint's Node API), so I’d prefer to keep the implementation consolidated in one place rather than merge another parallel change here.
For that reason, I’m going to close this PR for now. Really appreciate the contribution, and thanks again for helping improve the project.

PR #1147, opened 2026-06-24
feat(eslint-plugin-jest): add `prefer-jest-mocked` rule

eryue0220:
## Summary

Port `prefer-jest-mocked` from eslint-plugin-jest to rslint

## Related Links

eslint-plugin-jest/prefer-jest-mocked [doc](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/prefer-jest-mocked.md) [code](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/prefer-jest-mocked.ts)

## Checklist

### DIFF
--- a/internal/plugins/jest/rules/prefer_jest_mocked/prefer_jest_mocked.go
+++ b/internal/plugins/jest/rules/prefer_jest_mocked/prefer_jest_mocked.go
@@ -1,0 +1,91 @@
+package prefer_jest_mocked
+
+import (
+	"github.com/microsoft/typescript-go/shim/ast"
+	"github.com/web-infra-dev/rslint/internal/rule"
+	"github.com/web-infra-dev/rslint/internal/utils"
+)
+
+func buildUseJestMockedMessage() rule.RuleMessage {
+	return rule.RuleMessage{
+		Id:          "useJestMocked",
+		Description: "Prefer `jest.mocked()` over type assertions",
+	}
+}
+
+func isJestMockType(typeNode *ast.Node) bool {
+	if typeNode == nil || typeNode.Kind != ast.KindTypeReference {
+		return false
+	}
+
+	typeRef := typeNode.AsTypeReferenceNode()
+	if typeRef == nil || typeRef.TypeName == nil || typeRef.TypeName.Kind != ast.KindQualifiedName {
+		return false
+	}
+
+	qualifiedName := typeRef.TypeName.AsQualifiedName()
+	if qualifiedName == nil ||
+		qualifiedName.Left == nil ||
+		qualifiedName.Left.Kind != ast.KindIdentifier ||
+		qualifiedName.Right == nil ||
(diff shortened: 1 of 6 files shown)


internal/plugins/jest/rules/prefer_jest_mocked/prefer_jest_mocked.go
```diff
@@ -0,0 +1,91 @@
...
+	ctx.ReportNodeWithFixes(
+		assertionNode,
+		buildUseJestMockedMessage(),
+		rule.RuleFixReplace(ctx.SourceFile, assertionNode, "jest.mocked("+fnName+")"),
+	)
+}
+
+func checkJestMockAssertion(ctx rule.RuleContext, node *ast.Node, typeNode, expression *ast.Node) {
+	if node.Kind == ast.KindAsExpression {
+		if parent := node.Parent; parent != nil && parent.Kind == ast.KindAsExpression {
```
fansenze: False positive vs eslint-plugin-jest when a chained cast has explicit inner parens:

```ts
((foo as jest.Mock) as unknown).mockReturnValue(1);
// eslint-plugin-jest@29: 0 errors
// here: reports + autofixes to ((jest.mocked(foo)) as unknown)...
```

The parent-skip mirrors upstream's `node.parent.type === TSAsExpression`, but ESTree drops parens while the tsgo AST keeps them — the inner node's parent here is a `ParenthesizedExpression`, so the check misses. Walk up through parens:

```go
if parent := ast.WalkUpParenthesizedExpressions(node.Parent); parent != nil && parent.Kind == ast.KindAsExpression {
    return
}
```

internal/plugins/jest/rules/prefer_jest_mocked/prefer_jest_mocked.go
```diff
@@ -0,0 +1,91 @@
...
+	switch qualifiedName.Right.AsIdentifier().Text {
+	case "Mock", "MockedFunction", "MockedClass", "MockedObject":
+		return true
+	default:
+		return false
+	}
+}
+
+func reportJestMockedAssertion(ctx rule.RuleContext, assertionNode *ast.Node, expression *ast.Node) {
+	innerExpression := ast.SkipOuterExpressions(expression, ast.OEKTypeAssertions)
```
fansenze: Fix output diverges from upstream when the operand is parenthesized:

```ts
((foo) as jest.Mock).mockReturnValue(1);
// eslint-plugin-jest@29: (jest.mocked(foo))...
// here: (jest.mocked((foo)))...
```

`followTypeAssertionChain` lands on `foo` (ESTree has no paren node); `SkipOuterExpressions(..., OEKTypeAssertions)` stops at the `ParenthesizedExpression`. Skip parens too:

```go
innerExpression := ast.SkipOuterExpressions(expression, ast.OEKParentheses|ast.OEKTypeAssertions)
```

fansenze (approved):

PR #1148, opened 2026-06-24
feat(eslint-plugin-jest): add `prefer-spy-on` rule

eryue0220:
## Summary

Port `prefer-spy-on` from eslint-plugin-jest to rslint

## Related Links

eslint-plugin-jest/prefer-spy-on [doc](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/prefer-spy-on.md) [code](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/prefer-spy-on.ts)

## Checklist


internal/plugins/jest/rules/prefer_spy_on/prefer_spy_on.go
```diff
@@ -0,0 +1,150 @@
...
+	return []rule.RuleFix{
+		rule.RuleFixInsertBefore(ctx.SourceFile, left, "jest.spyOn("),
+		rule.RuleFixReplaceRange(
+			core.NewTextRange(obj.End(), prop.Pos()),
+			", "+quote,
+		),
+		rule.RuleFixReplaceRange(
+			core.NewTextRange(prop.End(), jestFnCall.End()),
+			quote+")"+mockImplementation,
+		),
```
fansenze, line 144: **`buildSpyOnFixes`: the autofix can emit unbalanced parentheses → invalid output.**

`getJestFnCall` runs `ast.SkipParentheses` on the RHS, so when the mock is wrapped in parens (`obj.a = (jest.fn())`), `jestFnCall` is the *inner* call and its `End()` sits *before* the closing `)`. The replace range `[prop.End(), jestFnCall.End()]` then deletes the opening `(` but leaves the `)` behind:

```
obj.a = (jest.fn())   ->   jest.spyOn(obj, 'a').mockImplementation())
```

This output fails to parse (`error TS1005: ';' expected`). The upstream rule has the same defect, but only because espree doesn't model parentheses — ts-go *does* expose `ParenthesizedExpression`, so the port shouldn't copy the bug.

Fix: extend the replace end past the parens that directly wrap the call.

```go
// jestFnCallReplaceEnd returns the end of the jest.fn() call including any
// parentheses that directly wrap it. getJestFnCall peels parens off the RHS, so
// its End() can sit before a wrapping ")"; using it raw would delete the "(" and
// leave the ")" dangling.
func jestFnCallReplaceEnd(jestFnCall *ast.Node) int {
	n := jestFnCall
	for n.Parent != nil && n.Parent.Kind == ast.KindParenthesizedExpression {
		n = n.Parent
	}
	return n.End()
}
```

```go
rule.RuleFixReplaceRange(
	core.NewTextRange(prop.End(), jestFnCallReplaceEnd(jestFnCall)),
	quote+")"+mockImplementation,
),
```

Verified against `(jest.fn())`, `((jest.fn()))`, `(jest.fn(() => 10))`, and `(jest.fn()).one.two()` (trailing chain preserved); the existing suite still passes. Please add these as `invalid` cases too.

internal/plugins/jest/rules/prefer_spy_on/prefer_spy_on.go
```diff
@@ -0,0 +1,162 @@
...
+	jestFnEnd := jestFnCallFixEnd(jestFnCall)
+
+	return []rule.RuleFix{
+		rule.RuleFixInsertBefore(ctx.SourceFile, left, "jest.spyOn("),
+		rule.RuleFixReplaceRange(
+			core.NewTextRange(obj.End(), prop.Pos()),
+			", "+quote,
+		),
+		rule.RuleFixReplaceRange(
+			core.NewTextRange(prop.End(), jestFnEnd),
```
fansenze, line 142: For parenthesized mock chains, the fix should not stop at the inner `jest.fn()` call. Please compute the replacement end from the RHS expression: use the unwrapped RHS as the chain body, use the original RHS end as the fix range end when the whole chain is parenthesized, and append the source text between the `jest.fn()` fix end and the unwrapped chain end back into the replacement. This preserves `.mockReturnValue(...)` while removing the outer trailing `)`.

Please add a regression test for `obj.a = (jest.fn().mockReturnValue(1))` expecting `jest.spyOn(obj, 'a').mockImplementation().mockReturnValue(1)`. I verified this shape with the target Go rule test; the previous implementation leaves an extra `)` after applying the fix.

internal/plugins/jest/rules/prefer_spy_on/prefer_spy_on.go
```diff
@@ -0,0 +1,162 @@
...
+				if bin == nil {
+					return
+				}
+
+				left := ast.SkipParentheses(bin.Left)
+				if !jestUtils.IsMemberAccessNode(left) {
+					return
+				}
+
+				jestFnCall := getJestFnCall(bin.Right)
```
fansenze, line 174: Please skip private identifiers before reporting/fixing. `this.#f = jest.fn()` cannot be converted to `jest.spyOn(this, '#f')`, because that targets a public string-named property instead of the private slot. A valid regression test should keep `class A { #f; m() { this.#f = jest.fn(); } }` in the valid cases.

eryue0220: Updated.

internal/plugins/jest/rules/prefer_spy_on/prefer_spy_on.go
eryue0220: Updated.

fansenze (approved):

PR #1153, opened 2026-06-26
fix: unwrap single-element array option for buggy rules

swwind:
This PR fixes the option parsing in several Jest and TypeScript rules to support direct/bare object options (rather than requiring them to be wrapped in a single-element array).

### Example
Previously, passing unwrapped options (e.g., from JSON config or the JS rule-tester) such as `"jest/max-expects": ["error", { "max": 10 }]` failed to parse. The configuration parser unwrapped the single-element option array into a bare map `{ "max": 10 }`, which the Go parser (which expected `[]interface{}`) failed to parse, causing it to fall back to the default options (`{ "max": 5 }`). 

Now, option parsers accept only unwrapped shape.
### Changes
- **Rule Fixes**: Update option parsers for jest/max-expects, jest/max-nested-describe, jest/no-hooks, jest/no-standalone-expect, jest/valid-title, typescript/no-confusing-void-expression, typescript/only-throw-error, and no-restricted-imports to handle unwrapped single-object options.
- **Test Fixes**: Update the corresponding test cases in packages/rslint-test-tools to use the direct option shape.


swwind:
This was fixed in #1122, closing

PR #1163, opened 2026-07-01
ci: lock Rust toolchain to 1.96.1

swwind:
## Summary
- Pin the `toolchain` input passed to `dtolnay/rust-toolchain` in `.github/actions/setup-rust/action.yml` to `1.96.1`, so all CI/release jobs use an exact Rust version instead of whatever the action's `stable` branch currently resolves to.

Temporary fix to our windows runner.


swwind:
It looks like not a rust problem, maybe it is my problem.

swwind:
I'm sure it is our windows runner's problem :window: :-1:

fansenze (approved):

PR #1167, opened 2026-07-02
feat(eslint-plugin-jest): add `no-conditional-expect` rule

eryue0220:
## Summary

* Port `no-conditional-expect` from eslint-plugin-jest to rslint

## Related Links

eslint-plugin-jest/no-conditional-expect [doc](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-conditional-expect.md) [code](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/no-conditional-expect.ts)

## Checklist


internal/plugins/jest/rules/no_conditional_expect/no_conditional_expect.go
```diff
@@ -0,0 +1,271 @@
...
+	Run: func(ctx rule.RuleContext, options any) rule.RuleListeners {
+		_ = options
+
+		testCallbackFunctions := collectTestFunctionCallbacks(ctx)
+		testCaseDepth := 0
+		conditionalDepth := 0
+		promiseCatchDepth := 0
+		// eslint-plugin-jest parity tests expect one diagnostic per test block even
+		// when multiple conditional expects exist (e.g. chained .catch handlers).
+		reportedInCurrentTest := false
```
fansenze: This implementation currently deduplicates diagnostics at the test-block level, which does not match upstream `eslint-plugin-jest` behavior. Once the first conditional `expect` in a test is reported, `reportedInCurrentTest` suppresses every later conditional `expect` in the same test, so users only see the first violation.

I verified this against the current latest versions, `eslint@10.6.0` + `eslint-plugin-jest@29.15.4`:
- two independent conditional `expect`s in the same `it` report 2 diagnostics upstream
- a referenced test callback with two conditional `expect`s reports 2 diagnostics upstream
- two `expect`s inside the same `.catch` callback report 2 diagnostics upstream

Current rslint only reports the first one in those cases.

The fix should not just remove `reportedInCurrentTest`, because upstream's `.catch` behavior is also more specific: it tracks `.catch` with a boolean state, not a depth counter, and reports conditional context and promise-catch context through separate branches. This preserves cases like chained `.catch(...expect...).catch(...expect...)` reporting only 1 diagnostic, while `if (...) expect(...)` inside a `.catch` reports 2 diagnostics, and `.catch(error => expect(...))` outside a test still reports 1.

Suggested shape:

```go
if inTestCase() && conditionalDepth > 0 {
    ctx.ReportNode(node, buildConditionalExpectMessage())
}
if inPromiseCatch {
    ctx.ReportNode(node, buildConditionalExpectMessage())
}
```

Please remove the test-level dedupe, change `promiseCatchDepth` to an `inPromiseCatch` bool, split the report conditions, and add parity tests for the cases above.

eryue0220: Updated.

fansenze (approved):

PR #1212, opened 2026-07-06
fix: make no-tsconfig projects respect global ignores and .gitignore

### LINKED ISSUE #1164: files/ignores/globalIgnores don't restrict file discovery
## Versions
- `@rslint/core`: reproduces on both 0.6.3 and 0.6.4 (latest at time of writing)
- `eslint-plugin-unicorn`: 68.0.0
- Node: v26.3.0
- OS: Linux (Ubuntu 24.04, kernel 6.17)

## Repro
Minimal repro repo: https://github.com/swwind/rslint-fixture-repro

```
git clone git@github.com:swwind/rslint-fixture-repro.git
cd rslint-fixture-repro
pnpm install
pnpm exec rslint --config rslint.config.js
```

Layout:
```
rslint-fixture-repro/
├── rslint.config.js
├── src/
│   └── clean.ts            # valid TS, contains one `null` literal
└── unrelated-fixtures/
    └── broken.js           # deliberately invalid syntax, unrelated to src/
```

`rslint.config.js`:
```js
import { defineConfig } from '@rslint/core';
import unicorn from 'eslint-plugin-unicorn';

export default defineConfig([
  {
    files: ['src/**/*.ts'],
    ignores: ['unrelated-fixtures/**'],
    plugins: { 'unicorn-js': unicorn },
    rules: {
      'unicorn-js/no-null': 'error',
    },
  },
]);
```

`unrelated-fixtures/broken.js` is a file containing deliberately invalid JS syntax (used elsewhere purely as raw test-fixture data, never meant to be parsed as a program).

## Actual output
```
  TypeScript(TS1109)  — [error] Expression expected.
  ╭─┴──────────( unrelated-fixtures/broken.js:23:6 )─────
  │ 22 │  })();
  │ 23 │  ();
  │ 24 │  Employee(name, basepay);
  ╰────────────────────────────────
```
Exit code 1. No `unicorn-js/no-null` diagnostic for `src/clean.ts` is ever printed, and no summary line ("Found N errors...") appears at all.

## Expected output
The `unicorn-js/no-null` violation in `src/clean.ts` should be reported, exactly as it is when `unrelated-fixtures/` is removed (control case):
```
  unicorn-js/no-null  — [error] Use `undefined` instead of `null`.
  ╭─┴──────────( src/clean.ts:2:10 )─────
  │ 1 │  export function example(): unknown {
  │ 2 │    return null; // should trigger unicorn-js/no-null
  │ 3 │  }
  ╰────────────────────────────────

Found 1 error and 0 warnings (linted 2 files with 1 rule in 807ms using 8 threads)
```

## Impact
Any project with even one intentionally-invalid or non-source `.js`/`.ts` file anywhere under the working directory (test fixtures, generated code, scratch files) will have its *entire* lint run silently produce zero rule diagnostics, with no way to exclude the offending file via config.

swwind:
## Summary

Projects without a `tsconfig.json` used to scan files with an ad-hoc directory walk that ignored `{ ignores: [...] }` config entries and `.gitignore` entirely — files users explicitly excluded were silently linted anyway.

**Fix:** that fallback walk is removed. No-tsconfig projects now go through `DiscoverGapFiles`/`walkDir`, the same discovery pipeline already used when a tsconfig only partially covers a project's files — which does honor `ignores` and `.gitignore`, and only walks the tree once instead of twice.

Fixing this surfaced a few related bugs, also fixed here:
- `.gitignore`-derived ignores were sometimes evaluated after config `ignores`, so a config's `!` negation could lose to a `.gitignore` rule depending on discovery order. `.gitignore` patterns are now always evaluated first, so negations always win.
- A nested config with no tsconfig of its own could have its files wrongly excluded just because an *ancestor* config's tsconfig `include` already covered them. File discovery for it now proceeds normally, and existing nearest-config resolution decides ownership.
- `rslint <file>` on an explicitly-passed file that's gitignored/ignored reported "was not found in the project" instead of "is ignored".
- `--allowFiles` (lint-staged style) missed `.gitignore` files in directories between the config root and a nested explicit file, so those files could be wrongly linted.
- `node_modules` is no longer hardcoded-skipped; it's excluded only via `.gitignore`/`ignores`, like any other directory (in practice invisible for most repos, since `node_modules/` is almost always itself gitignored).

**Behavior now:** no-tsconfig projects respect `{ ignores: [...] }` and `.gitignore` the same way tsconfig projects do, and explicitly-passed ignored files are reported as ignored instead of linted or misreported.

## Related Links

Fixes #1164

## Checklist

### DIFF
--- a/cmd/rslint/api.go
+++ b/cmd/rslint/api.go
@@ -154,8 +154,9 @@ func (h *IPCHandler) HandleLint(req api.LintRequest) (*api.LintResponse, error) 
 	// (the JS side resolves any multi-config merge into one entry list).
 	// The --api path never runs the type-check phase (RunLinterOptions.TypeCheck
 	// stays false), so there is no per-program type-check skip mask to build.
-	programs, typeInfoFiles, _ := buildProgramsWithGapFallback(
+	programs, typeInfoFiles, _, rslintConfig, _ := buildProgramsWithGapFallback(
 		programs, nil, rslintConfig, configDirectory, fs, allowedFiles, nil, parseCache, false,
+		len(tsConfigs) > 0, nil,
 	)
 
 	// Collect diagnostics and source files
--- a/cmd/rslint/cmd.go
+++ b/cmd/rslint/cmd.go
@@ -33,7 +33,6 @@ import (
 	"github.com/microsoft/typescript-go/shim/ast"
 	"github.com/microsoft/typescript-go/shim/bundled"
 	"github.com/microsoft/typescript-go/shim/compiler"
-	"github.com/microsoft/typescript-go/shim/core"
 	"github.com/microsoft/typescript-go/shim/scanner"
 	"github.com/microsoft/typescript-go/shim/tspath"
 	"github.com/microsoft/typescript-go/shim/vfs/cachedvfs"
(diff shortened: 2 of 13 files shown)


swwind:
Superseded by #1227, which rebuilt lint target discovery and covers this PR's scope: verified on latest `main` (`a7170f83`) with the #1164 repro (a no-tsconfig project) that both `globalIgnores(['unrelated-fixtures/**'])` and a `.gitignore` entry now correctly exclude the directory from discovery — the unrelated `TS1109` no longer appears and only the real lint diagnostic is reported. Details in https://github.com/web-infra-dev/rslint/issues/1164#issuecomment-4946986162. Closing.

PR #1224, opened 2026-07-07
feat(eslint-plugin-jest): add `no-restricted-jest-methods` rule

eryue0220:
## Summary

* Port `no-restricted-jest-methods` from eslint-plugin-jest to rslint

## Related Links

eslint-plugin-jest/no-restricted-jest-methods [doc](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-restricted-jest-methods.md) [code](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/no-restricted-jest-methods.ts)

## Checklist


internal/plugins/jest/rules/no_restricted_jest_methods/no_restricted_jest_methods.go
```diff
@@ -0,0 +1,119 @@
...
+			return rule.RuleListeners{}
+		}
+
+		return rule.RuleListeners{
+			ast.KindCallExpression: func(node *ast.Node) {
+				if isNestedJestFnCall(node) {
+					return
+				}
+
+				jestFnCall := jestUtils.ParseJestFnCall(node, ctx)
```
fansenze, line 105: This diverges from upstream for parenthesized Jest chains. eslint-plugin-jest still reports restricted methods for `(jest).fn()` and `(jest.fn)()`, but tsgo preserves `ParenthesizedExpression`, so `ParseJestFnCall` currently misses these shapes and returns nil. That makes the restriction easy to bypass by adding parentheses.

Suggested fix: unwrap parenthesized expressions before parsing the Jest member chain, ideally in the shared Jest parser helpers (`GetJestFnMemberEntries` / `ResolveFirstIdentifier`). Also make the nested-call guard skip parenthesis wrappers so `(jest.fn()).mockImplementation(...)` reports `fn` once, matching upstream.

internal/plugins/jest/rules/no_restricted_jest_methods/no_restricted_jest_methods.go
```diff
@@ -0,0 +1,119 @@
...
+				if jestFnCall == nil || jestFnCall.Kind != jestUtils.JestFnTypeJest || len(jestFnCall.Members) == 0 {
```
fansenze, line 106: This also diverges from upstream when `jest` is locally bound. Upstream resolves `const jest = { fn() {} }; jest.fn();` as a local object and does not report it, but this rule currently treats the name `jest` as the Jest global and reports a false positive.

Suggested fix: before accepting a `JestFnTypeJest` call, resolve the head identifier and ignore it when it points to a local declaration rather than the Jest global or an import from `@jest/globals`. Keep existing support for `import { jest } from '@jest/globals'` and `const { jest } = require('@jest/globals')`.

eryue0220:
I've updated.

fansenze (approved):

PR #1229, opened 2026-07-08
feat(config): type languageOptions.globals/sourceType/ecmaFeatures

swwind:
## Summary

- Add `globals`, `sourceType`, and `parserOptions.ecmaFeatures.{jsx,globalReturn,impliedStrict}` to `defineConfig()`'s `LanguageOptions`/`ParserOptions` types — the runtime already forwards/consumes these, only the type was missing them, so writing them in `rslint.config.ts` was a TS error.
- Also add `parserOptions.jsxPragma`/`jsxFragmentName` now that the worker wire forwards them (#1230).
- Leave out `ecmaVersion` — it never reaches the TS scope-manager branch and no rule reads it back, so typing it would silently do nothing.
- Fix a stale comment in `presets/unicorn.ts` claiming `globals` couldn't be expressed at all.

## Related Links

N/A

## Checklist

- [x] Tests updated (or not required) — type-only change, verified via a `tsc` probe and `pnpm typecheck`.
- [x] Documentation updated (or not required) — inline doc comments on the new fields.

### DIFF
--- a/packages/rslint/src/config/define-config.ts
+++ b/packages/rslint/src/config/define-config.ts
@@ -84,7 +84,9 @@ export type RuleEntry =
 export type RulesRecord = Record<string, RuleEntry>;
 
 /**
- * TypeScript parser options. `project` may be a single tsconfig path or a list.
+ * Parser-specific extras, nested under `parserOptions` per ESLint v10's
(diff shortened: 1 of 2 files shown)


swwind:
Most of the options are not stable yet, currently they are not suggested to use. We will add the option types when we have fully tested their functionality.

PR #1232, opened 2026-07-08
fix: error on unknown rule names in config

### LINKED ISSUE #1228: [Bug]: Unknown rules in configuration are silently ignored instead of throwing configuration errors
### Details
When an unknown, misspelled, or unregistered rule (e.g., `'non-existent-rule-name': 'error'`) is specified in `rslint.config.ts`, `rslint` silently ignores the rule. It does not warn the developer or fail the build, which can lead to situations where intended lint checks are never actually executed due to silent typos.

In contrast, **ESLint** performs strict configuration validation at startup and crashes loudly, preventing developers from continuing with invalid configurations.

---

#### **rslint Behavior**
- Running `rslint` with an unknown rule succeeds silently.
- The rule is skipped entirely without any logs or diagnostics.
- Example:
  ```bash
  $ npx rslint
  Found 0 errors and 0 warnings (linted 542 files with 9 rules in 2.158s using 8 threads)
  ```

#### **ESLint Behavior**
ESLint strictly validates rules and plugins at start time and throws errors, unless the rule is set to `"off"` or `0`:

1. **Unknown core rule**:
   ```
   TypeError: Key "rules": Key "non-existent-rule-name": Could not find "non-existent-rule-name" in plugin "@".
   ```
2. **Plugin rule where the plugin is not registered**:
   ```
   A configuration object specifies rule "some-plugin/non-existent-rule", but could not find plugin "some-plugin".
   ```
3. **Plugin rule where the plugin is registered, but the rule does not exist**:
   ```
   TypeError: Key "rules": Key "some-plugin/non-existent-rule": Could not find "some-plugin/non-existent-rule" in plugin "some-plugin".
   ```

> [!NOTE]
> If the unknown/unregistered rule is explicitly set to `"off"` (or `0`), ESLint **bypasses validation** and runs successfully (exit code 0).

---

### Steps to Reproduce
1. In `rslint.config.ts`, add an unknown rule with a non-zero severity (e.g. `'error'` or `'warn'`):
   ```ts
   export default defineConfig([
     {
       files: ['**/*.ts'],
       rules: {
         'non-existent-rule-name': 'error',
         'import-x/this-rule-does-not-exist': 'error',
       }
     }
   ])
   ```
2. Run `npx rslint`.
3. Observe that the linter completes successfully with `exit code 0`, completely ignoring the invalid rule keys.

---

### Expected Behavior
`rslint` should perform config validation at startup:
- Match rule names against natively implemented Go rules.
- Match rule names with plugin namespaces against the loaded plugin metadata returned from the worker initialization.
- If a rule name cannot be resolved in either registry, `rslint` should report a configuration error and fail non-zero (or at least output a warning).
- **Parity with ESLint**: Validation checks should only run on rules that are enabled (severity is not `'off'` or `0`), matches ESLint's behavior of silently ignoring unknown rules when disabled.

swwind:
## Summary

Unknown, misspelled, or unregistered rule names in `rules` (e.g. `rules: { 'non-existent-rule-name': 'error' }`) were silently ignored. The config-time validation step from #1243 now also resolves every enabled rule name, so unknown names are reported alongside rule-options failures. Rules set to `"off"` are exempt, same as ESLint.

- `internal/config`: `ValidateRuleOptions` becomes `ValidateRules(config, registry, eslintPlugins)` — one pass collects unknown-name errors (`unknown rule "x"`, `unknown rule "p/x": plugin "p" has no such rule`, `unknown rule "p/x": plugin "p" is not registered`) and options failures, deduplicated and sorted by rule name. Mounted ESLint-plugin rules resolve through the passed plugin entries, not the registry's placeholders — the placeholders may be unregistered yet at validation time, or stale leftovers from an earlier request in a long-lived process. Declaring a prefix under `plugins` grants no exemption: enabling a plugin's rules without supplying that plugin's rule names is an error.
- **CLI / IPC API**: every failure is fatal — the CLI prints all failures and exits 1 before linting starts; the API rejects the request (and assembles its plugin entries before validating, so a metadata-free request can no longer resolve rule names through another request's stale placeholders).
- **LSP**: failures are downgraded to logged warnings and the offending rules are set to `"off"` in the committed snapshot; every other rule keeps running and the config transaction commits normally. When the plugin host is degraded the catalog carries no plugin entries, so mounted plugin rules surface as harmless warnings instead of rejecting a valid config.

## Breaking Changes

`--rule` with an unknown rule name now also fails with exit code 1 (previously silently ignored), matching ESLint's `--rule` behavior. Use `"off"` if you don't want a rule validated.

## Related Links

Fixes #1228

## Checklist


swwind:
We might implement this alongside with the new schema validation system. Hold short.

PR #1233, opened 2026-07-08
fix: honor languageOptions.parserOptions.jsxPragma / jsxFragmentName

### LINKED ISSUE #1230: [Bug]: languageOptions.parserOptions.jsx related options (jsxPragma, jsxFragmentName) are ignored/broken
### System Info
```text
System:
  OS: Linux (x64)
npmPackages:
  @rslint/core: 0.6.5
```

### Details
When configuring custom JSX pragmas (e.g., `jsxPragma: 'h'` and `jsxFragmentName: 'Fragment'` for Preact or custom JSX frameworks) under `languageOptions.parserOptions`, Rslint fails to respect them. This results in false-positive unused variable/import warnings (e.g. `'h' is defined but never used`) because Rslint incorrectly defaults the JSX pragma name to `'React'`.

We traced this behavior to two distinct issues in the codebase:

1. **Scope Analyzer Option Drop (JS/TS worker)**:
   In `packages/rslint/src/eslint-plugin/linter/ecma-language-plugin.ts` (lines 271–284), the scope manager factory is instantiated without `jsxPragma` or `jsxFragmentName`:
   ```typescript
   const scopeManagerFactory = (() => {
     const inner = makeScopeManagerFactory(ast, {
       filePath: req.filePath,
       sourceType: req.languageOptions?.sourceType ?? 'module',
       ecmaVersion: req.languageOptions?.ecmaVersion,
       globals,
       impliedStrict: ecmaFeatures?.impliedStrict,
       globalReturn: ecmaFeatures?.globalReturn,
       // missing jsxPragma and jsxFragmentName!
     });
   ```
   This causes `ts-scope-manager` in the Node/JS worker to fall back to `'React'`, which impacts any scope/variable checking rules running inside the worker (like `react/jsx-uses-vars`).

2. **Go Config Merge Key Override**:
   In `internal/config/config.go` (lines 569–583), `mergeLanguageOptions` only performs a shallow merge of the raw `parserOptions` map. If a base config configures `parserOptions.project` and a sub-config override configures `parserOptions.jsxPragma`, the `parserOptions` map is replaced entirely, throwing away `project` configurations on the wire:
   ```go
   if len(override.Raw) > 0 {
       mergedRaw := make(map[string]any, len(base.Raw)+len(override.Raw))
       for k, v := range base.Raw {
           mergedRaw[k] = v
       }
       for k, v := range override.Raw {
           mergedRaw[k] = v // completely overwrites nested objects like parserOptions
       }
       merged.Raw = mergedRaw
   }
   ```

3. **Fallback Program compiler options**:
   In `cmd/rslint/programs.go`, the directory-scan/fallback `compiler.Program` options do not map `jsxPragma` or `jsxFragmentName` to `core.CompilerOptions.JsxFactory`. Go's native rules (like `@typescript-eslint/no-unused-vars`) therefore fall back to looking for `React`.

### Reproduce Steps

1. Create a minimal project with a single JSX file `src/index.jsx` (no `tsconfig.json` present to ensure pure-linter config mapping is tested):
   ```jsx
   // src/index.jsx
   import { h } from 'preact';

   export function App() {
     return <div>Hello Preact</div>;
   }
   ```

2. Create identical configurations for ESLint and Rslint targeting this file:
   * **`eslint.config.js`**:
     ```javascript
     import tsParser from '@typescript-eslint/parser';
     import tsPlugin from '@typescript-eslint/eslint-plugin';

     export default [
       {
         files: ['src/**/*.jsx'],
         plugins: {
           '@typescript-eslint': tsPlugin,
         },
         languageOptions: {
           parser: tsParser,
           parserOptions: {
             ecmaFeatures: { jsx: true },
             jsxPragma: 'h',
             jsxFragmentName: 'Fragment',
           }
         },
         rules: {
           '@typescript-eslint/no-unused-vars': 'error'
         }
       }
     ];
     ```
   * **`rslint.config.mjs`**:
     ```javascript
     import { defineConfig } from '@rslint/core';

     export default defineConfig([
       {
         files: ['src/**/*.jsx'],
         plugins: ['@typescript-eslint'],
         languageOptions: {
           parserOptions: {
             ecmaFeatures: { jsx: true },
             jsxPragma: 'h',
             jsxFragmentName: 'Fragment',
           }
         },
         rules: {
           '@typescript-eslint/no-unused-vars': 'error'
         }
       }
     ]);
     ```

3. **Run ESLint**:
   ```bash
   npx eslint src/index.jsx
   ```
   * **Result**: `0` problems (passes cleanly, because `h` is recognized as the JSX pragma).

4. **Run Rslint**:
   ```bash
   npx rslint --config rslint.config.mjs src/index.jsx
   ```
   * **Result**:
     ```text
       @typescript-eslint/no-unused-vars  — [error] 'h' is defined but never used.
       ╭─┴──────────( src/index.jsx:1:10 )─────
       │ 1 │  import { h } from 'preact';
       │ 2 │  
       ╰────────────────────────────────
     ```

swwind:
## Summary

Fixes three separate bugs that all contributed to `languageOptions.parserOptions.jsxPragma` / `jsxFragmentName` being ignored (custom JSX pragmas like Preact's `h` were flagged as unused imports, while ESLint correctly recognized them):

- **JS/TS worker scope-manager** (`ecma-language-plugin.ts`): `jsxPragma`/`jsxFragmentName` are top-level `parserOptions` keys (siblings of `ecmaFeatures`, not nested under it), but the runner only ever read `ecmaFeatures` and never forwarded them to the scope-manager factory, so it silently fell back to the hard-coded `'React'`/`'Fragment'` defaults.
- **Go config merge** (`internal/config/config.go`): `mergeLanguageOptions` shallow-overwrote the whole `parserOptions` object on merge, so a later config entry setting only `jsxPragma` would silently discard an earlier entry's `parserOptions.project` (or any other nested key). Replaced with a recursive deep-merge, and added typed `JsxPragma`/`JsxFragmentName` fields to `ParserOptions`.
- **Go-native rule fallback Programs** (`cmd/rslint/programs.go`): the no-tsconfig directory-scan Program and the gap-file fallback Program never mapped `jsxPragma`/`jsxFragmentName` to `CompilerOptions.JsxFactory`/`JsxFragmentFactory`, so the native `@typescript-eslint/no-unused-vars` rule's implicit-JSX-usage check always looked for `React` regardless of config.

## Related Links

Fixes #1230

## Checklist

### DIFF
--- a/cmd/rslint/cmd.go
+++ b/cmd/rslint/cmd.go
@@ -1233,7 +1233,8 @@ func executeLintPipeline(args lintArgs, ctx context.Context, dispatch linter.Esl
 
 		// Rebuild fallback Program for gap files (content may have changed after fixes).
 		if len(capturedGapFiles) > 0 {
-			fallback, _ := createFallbackProgram(capturedGapFiles, singleThreaded, cwd, fs, parseCache)
+			jsxFactory, jsxFragmentFactory := resolveGapJsxPragma(configMap, rslintConfig)
+			fallback, _ := createFallbackProgram(capturedGapFiles, singleThreaded, cwd, fs, parseCache, jsxFactory, jsxFragmentFactory)
 			if fallback != nil {
 				baseProgs = append(baseProgs, fallback)
 			}
--- a/cmd/rslint/programs.go
+++ b/cmd/rslint/programs.go
@@ -125,7 +125,12 @@ func createProgramsForConfig(
 		includes := []string{"**/*"}
 		rootFiles := vfsmatch.ReadDirectory(fsys, configDir, configDir, sourceExts, excludes, includes, vfsmatch.UnlimitedDepth)
 		if len(rootFiles) > 0 {
-			program, err := utils.CreateProgramFromOptions(singleThreaded, &core.CompilerOptions{AllowJs: core.TSTrue}, rootFiles, host)
+			jsxFactory, jsxFragmentFactory := rslintconfig.ResolveJsxPragmaOptions(entries)
+			program, err := utils.CreateProgramFromOptions(singleThreaded, &core.CompilerOptions{
+				AllowJs:            core.TSTrue,
+				JsxFactory:         jsxFactory,
+				JsxFragmentFactory: jsxFragmentFactory,
+			}, rootFiles, host)
 			if err != nil {
 				w := bufio.NewWriter(os.Stderr)
 				if !reportSyntacticErrors(err, w, tspath.ComparePathsOptions{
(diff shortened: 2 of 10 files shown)


swwind:
There appears to be a bug when having multiple `rslint.config.ts` files for different projects: the current solution may merge these configurations and force all projects to use the same JSX setting.

Given that few projects currently rely on this option to use JSX with ESLint, and linting JS projects are not the primary objective to this project, this issue is being set aside for now; users can configure the relevant settings in `tsconfig.json` if necessary.

PR #1236, opened 2026-07-08
feat(eslint-plugin-jest): add `prefer-comparison-matcher` rule

eryue0220:
## Summary

* Port `prefer-comparison-matcher` from eslint-plugin-jest to rslint

## Related Links

eslint-plugin-jest/prefer-comparison-matcher [doc](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/prefer-comparison-matcher.md) [code](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/prefer-comparison-matcher.ts)

## Checklist

### DIFF
--- a/internal/plugins/jest/rules/prefer_comparison_matcher/prefer_comparison_matcher.go
+++ b/internal/plugins/jest/rules/prefer_comparison_matcher/prefer_comparison_matcher.go
@@ -1,0 +1,126 @@
+package prefer_comparison_matcher
+
+import (
+	"slices"
+
+	"github.com/microsoft/typescript-go/shim/ast"
+	"github.com/microsoft/typescript-go/shim/core"
+	"github.com/microsoft/typescript-go/shim/scanner"
+	"github.com/web-infra-dev/rslint/internal/plugins/jest/utils"
+	"github.com/web-infra-dev/rslint/internal/rule"
+	internalUtils "github.com/web-infra-dev/rslint/internal/utils"
+)
+
+type comparisonMatcher struct {
+	matcher        string
+	negatedMatcher string
+}
+
+var comparisonMatchers = map[ast.Kind]comparisonMatcher{
+	ast.KindGreaterThanToken:       {matcher: "toBeGreaterThan", negatedMatcher: "toBeLessThanOrEqual"},
+	ast.KindLessThanToken:          {matcher: "toBeLessThan", negatedMatcher: "toBeGreaterThanOrEqual"},
+	ast.KindGreaterThanEqualsToken: {matcher: "toBeGreaterThanOrEqual", negatedMatcher: "toBeLessThan"},
+	ast.KindLessThanEqualsToken:    {matcher: "toBeLessThanOrEqual", negatedMatcher: "toBeGreaterThan"},
+}
+
+func buildUseComparisonMatcherMessage(preferredMatcher string) rule.RuleMessage {
+	return rule.RuleMessage{
+		Id:          "useToBeComparison",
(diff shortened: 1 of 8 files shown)


internal/plugins/jest/rules/prefer_comparison_matcher/prefer_comparison_matcher.go
```diff
@@ -0,0 +1,126 @@
...
+				if len(expectArgs) == 0 {
+					return
+				}
+
+				left, right, matchers, ok := parseComparison(expectArgs[0])
+				if !ok {
+					return
+				}
+
+				matcherArgs := node.AsCallExpression().Arguments.Nodes
```
fansenze: This should use the arguments from the parsed matcher call, not from the `CallExpression` currently being visited.

Upstream `parseJestFnCall` stores matcher arguments from the matcher call itself (`member.parent.parent.arguments`). So for `expect(value > 1).toBe(true).foo(false)`, even when ESLint visits the outer `.foo(false)` call, upstream still reads the `toBe(true)` argument and suggests `toBeGreaterThan`. The autofix preserves the outer chain: `expect(value).toBeGreaterThan(1).foo(false)`.

Here, reading `node.AsCallExpression().Arguments` means the outer visit can read `foo(false)` instead of `toBe(true)`, which can compute the opposite matcher and produce an unsafe fix. Please derive the matcher call from `jestFnCall.MatcherEntry` (or expose `MatcherCall` / `MatcherArgs` from `ParseJestFnCall`) and use that call's arguments for `matcherArg`. If duplicate reports are intentionally avoided, the alternative is to only report when `node` is the matcher call, but using parsed matcher args is the closer upstream behavior.

internal/plugins/jest/rules/prefer_comparison_matcher/prefer_comparison_matcher.go
```diff
@@ -0,0 +1,126 @@
...
+
+				matcherEntry := jestFnCall.MatcherEntry
+				if matcherEntry.Node == nil || matcherEntry.Node.Parent == nil {
+					return
+				}
+
+				comparison := ast.SkipParentheses(expectArgs[0])
+				leftText := scanner.GetSourceTextOfNodeFromSourceFile(ctx.SourceFile, left, false)
+				rightText := scanner.GetSourceTextOfNodeFromSourceFile(ctx.SourceFile, right, false)
+				modifierRange := core.NewTextRange(expectCall.End(), matcherEntry.Node.Parent.End())
```
fansenze, line 117: This parser path should fail closed when the expect chain contains an unsupported computed member.

Upstream `getNodeChain` returns `null` if any accessor in the chain is not statically supported. For `expect(value > 1)[foo()].toBe(true)`, the `[foo()]` segment makes the whole chain unparseable, so `jest/prefer-comparison-matcher` reports nothing.

rslint currently appears to recurse into the left side of `ElementAccessExpression` and return the left chain when the computed key is unsupported. That folds `expect(value > 1)[foo()].toBe(true)` into `expect(value > 1).toBe(true)`, and the autofix can delete `[foo()]`, dropping a potentially side-effectful call.

Please update the Jest chain parser used by `ParseJestFnCall` to fail the whole chain when an element access key is not a supported static accessor, instead of returning the left side only. A strict helper returning `(entries, ok)` would make this explicit. Also add a regression test asserting that `expect(value > 1)[foo()].toBe(true)` produces no diagnostic or fix.

eryue0220:
okay, I've updated of all.

fansenze (approved):

PR #1247, opened 2026-07-10
test(vscode): await extension readiness before suites

fansenze:
## Summary

- await the VS Code extension's public activation promise before loading or running Mocha in every suite
- centralize the suite runner so no workspace can omit the readiness barrier
- keep the original plugin `source.fixAll`-on-save regression as the first test in a fresh extension host
- make fix-all coverage fail closed: required diagnostics, code actions, edits, saves, and final content are hard assertions rather than silent early returns
- replace polling and prewarming workarounds with event-driven waits, and close editors before deleting temporary fixtures

## Evidence and scope

The previous runners could load and start Mocha without awaiting `rstack.rslint.activate()`. That is a concrete readiness-contract gap: extension activation initializes the language client, reverse plugin handler, and workspace configuration used by these tests.

The historical CI log alone does not prove that every observed timeout came from that gap. This PR therefore also preserves the cold-start scenario and removes false-green paths, so a future failure reports the missing stage instead of passing or timing out ambiguously.

This is test-harness-only. It contains no production Go/LSP changes. The stale-`didSave` ordering handled separately by #1252 is a real language-client ordering case, but it is not established as the cause of the original #1247 CI failure.

## Validation

- `pnpm --filter rslint exec tsgo -p tsconfig.spec.emit.json`
- `pnpm run build`
- local main VS Code suite: 32 passing
- local final plugin regression: 5/5 independent fresh extension hosts, with the target test first
- CI Linux npm job: main suite 32 passing; plugin suite 2 passing; target test first (1.519s); extension hosts exited cleanly
- CI Windows standalone VS Code job: main suite 32 passing; plugin suite 2 passing; target test first (1.878s); extension hosts exited cleanly


fansenze:
CI evidence for head `7356960a803e272629c7e7706034ef9422862f0a`:

- [Linux npm / VS Code suites](https://github.com/web-infra-dev/rslint/actions/runs/29079571688/job/86319030720) passed: main suite 32 tests; plugin suite 2 tests; the original plugin `source.fixAll`-on-save scenario ran first in its fresh extension host and passed in 1.519s.
- [Windows standalone VS Code suite](https://github.com/web-infra-dev/rslint/actions/runs/29079571688/job/86322142216) passed: main suite 32 tests; plugin suite 2 tests; the same target ran first and passed in 1.878s.

The workflow remains red only in an unrelated Ubuntu Go test. Both the [initial job](https://github.com/web-infra-dev/rslint/actions/runs/29079571688/job/86319030704) and the [failed-job rerun](https://github.com/web-infra-dev/rslint/actions/runs/29079571688/job/86324525550) failed identically:

```text
rslint: eslint-plugin lint error: WorkerPool: closed
TestRunCLI_StdoutIsTTYWireToANSI/stdoutIsTTY=true_colors_output_frames
ipc_cli_test.go:262: no diagnostic in forwarded output
```

#1247 changes only six TypeScript test-harness files, so I stopped rerunning rather than masking that independent, repeatable failure.

PR #1254, opened 2026-07-10
fix(linter): reject malformed rule fix ranges

fansenze:
## Summary

- validate every rule fix range before slicing source text
- leave an entire diagnostic unapplied when one of its fixes is malformed
- distinguish negative, inverted, out-of-bounds, and internally overlapping ranges
- preserve the existing `ApplyRuleFixes` API and add an opt-in reporter for production callers
- log rejected rule name, file, range, and reason in CLI, IPC API, and LSP paths

## Scope

This is defense-in-depth, not the root fix for the stale-`didSave` race. That product-state bug is handled independently in #1252. This PR ensures malformed native or third-party plugin fix data cannot crash any fixer caller and cannot degrade into an unobservable no-op.

Normal conflicts between separate diagnostics retain the existing behavior: one diagnostic is applied and the conflicting one remains unapplied without being reported as malformed.

## Validation

- `go test ./internal/linter -run TestApplyRuleFixes -count=1`
- `go test ./internal/lsp -run TestComputeFixAllContent -count=1`
- `go test ./cmd/rslint -run 'TestHandleLint_FixProducesInBandOutput|TestHandleLint_FixRangeIsUTF16' -count=1`

### DIFF
--- a/cmd/rslint/api.go
+++ b/cmd/rslint/api.go
@@ -357,7 +357,7 @@ func (h *IPCHandler) HandleLint(req api.LintRequest) (*api.LintResponse, error) 
 		output = make(map[string]string)
 		for filePath, fileDiags := range diagnosticsByFile {
 			originalContent := fileDiags[0].SourceFile.Text()
-			fixedContent, _, didFix := linter.ApplyRuleFixes(originalContent, fileDiags)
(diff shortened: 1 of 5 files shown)


fansenze:
Closing for now: the observed CI failure did not involve a malformed fix range, and existing plugin fix paths already normalize or reject the relevant inputs. We can revisit a narrower hardening change when there is an independent native-rule reproduction.

PR #1255, opened 2026-07-10
feat: port rule require-array-join-separator

Ling-yunchi:
## Summary

Port unicorn/require-array-join-separator from eslint-plugin-unicorn to rslint.

The rule requires Array#join() and borrowed Array.prototype.join() calls to provide an explicit separator argument. Autofixes insert ',' while preserving comments, whitespace, trailing commas, and type arguments.

## Related Links

- [Rule documentation](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v64.0.0/docs/rules/require-array-join-separator.md)
- [Upstream implementation](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v64.0.0/rules/require-array-join-separator.js)

## Checklist

- [x] Go tests added.
- [x] Integration tests added.
- [x] Documentation added.
- [x] Unicorn preset updated.


internal/plugins/unicorn/rules/require_array_join_separator/require_array_join_separator.go
```diff
@@ -0,0 +1,159 @@
...
+			depth--
+			if depth == 0 {
+				return opening, s.TokenRange(), previous, previousKind, true
+			}
+		default:
+			if foundOpening {
+				previous = s.TokenRange()
+				previousKind = s.Token()
+			}
+		}
```
swwind: The `KindOpenParenToken`/`KindCloseParenToken` branches never record the token into `previous`, but upstream's `penultimateToken` (`sourceCode.getLastTokens(node, 2)[0]`) is the literal second-to-last token of the call, nested parens included. Two consequences when the single argument's last token is a `)`:

1. **Broken autofix.** For `[].join.call(f(x,))` the last recorded token is the trailing comma *inside* `f(x,)` (recorded at depth 2), so `penultimateKind == KindCommaToken` and `appendSeparatorFix` takes the trailing-comma arm, producing `[].join.call(f(x,) ',',)` — a parse error. ESLint checks `isCommaToken` on the true penultimate token (the inner `)`) and produces `[].join.call(f(x,), ',')`.
2. **Report-range divergence.** For `[].join.call((foo))` or `[].join.call(g())`, `previous` ends up as `foo`/`g` instead of the inner `)`, so the reported range starts earlier than ESLint's (column 18 vs 19 in the `(foo)` case).

Recording `previous` for every token except the final closing paren (i.e. in the paren branches too) fixes both; the depth-1 trailing comma case `[].join.call(f(),)` still works because that comma really is the last token before the closing paren. Could you also add invalid tests for `[].join.call(f(x,))` and `[].join.call((foo))` with exact columns — arguments ending in `)` are the current suite's blind spot.

internal/plugins/unicorn/rules/require_array_join_separator/require_array_join_separator.go
```diff
@@ -0,0 +1,159 @@
...
+	if call == nil || call.Expression == nil {
+		return core.TextRange{}, core.TextRange{}, core.TextRange{}, ast.KindUnknown, false
+	}
+
+	scanStart := call.Expression.End()
+	if call.TypeArguments != nil && len(call.TypeArguments.Nodes) > 0 {
+		scanStart = call.TypeArguments.End()
+	}
+
+	s := scanner.GetScannerForSourceFile(sourceFile, scanStart)
```
swwind, line 67: Lower severity: raw re-scanning mis-lexes regex literals, because the scanner only produces a regex token via `ReScanSlashToken` — a plain `Scan()` lexes the regex body as ordinary tokens. A paren inside a regex in the argument corrupts the depth tracking:

- `[].join.call(s.replace(/\)/g, ''))` — the `)` inside `/\)/` closes depth early, so the returned closing paren is `replace`'s, and the separator is appended to the wrong call.
- `[].join.call(s.split(/\(/))` — the unbalanced `(` means depth never returns to 0, so the rule silently stops reporting (false negative).

A restructure avoids scanning across the argument entirely and fixes the penultimate-token bug above by construction: for the prototype branch scan only from `arguments[0].End()` (that region can contain only trivia, `,`, and `)`), and for the zero-argument direct branch the region between the callee/type-arguments end and `node.End()` contains only `(`, trivia, `)`. If raw-scanner regex hazards are an accepted tradeoff here (`prefer_array_flat_map` scans too, though over a much narrower span), a comment noting the limitation would help.

internal/plugins/unicorn/rules/require_array_join_separator/require_array_join_separator.go
```diff
@@ -0,0 +1,159 @@
...
+					ArgumentsLength:     &zeroArguments,
+					AllowOptionalMember: true,
+				})
+				if !directMethod {
+					call, directMethod = unicornutil.MatchDotMethodCall(node, unicornutil.DotMethodCallOptions{
+						Method:          "call",
+						ArgumentsLength: &oneArgument,
+					})
+					arguments := node.Arguments()
+					if !directMethod || len(arguments) != 1 || ast.IsSpreadElement(arguments[0]) ||
```
swwind: Nit: `len(arguments) != 1` is redundant — `ArgumentsLength: &oneArgument` already guarantees it.

swwind (commented):
Cross-checked against upstream v64.0.0 (`require-array-join-separator.js`, `fix/append-argument.js`, `ast/is-method-call.js`, `ast/is-member-expression.js`, `utils/array-or-object-prototype-property.js`). The matcher logic, `isArrayPrototypeProperty`, fix spacing, and the `MatchDotMethodCall` refactor of `no_thenable` / `prefer_array_flat_map` are all faithful ports — verified option-by-option, including the optional-chain, spread, computed-key, TS wrapper, and type-argument edge cases.

One real bug: `callParenthesesRange` diverges from ESLint's `getLastTokens(node, 2)` when the argument's last token is a `)`, which can make the autofix emit syntactically invalid code (details inline). One lower-severity edge case around raw re-scanning of regex literals, plus a nit.

Ling-yunchi:
Thanks for the detailed review! Addressed all three comments in `41723b58`:

- Restricted scanning to the suffix after the argument, avoiding nested-parenthesis and regex-literal tokenization issues.
- Fixed the autofix and report range for arguments ending in `)`.
- Removed the redundant argument-length check.
- Added regression tests for nested calls, parenthesized arguments, regex literals, exact ranges, and autofix output.

All related Go/JS tests and checks pass.

swwind:
Would you merge the main branch locally? There are some breaking changes in #1248 may cause the build to fail.

Ling-yunchi:
Yep, done — merged the latest `main` and updated the new rule to use `[]any` for `Run`. Everything is green locally now. Thanks for the heads-up!

swwind:
Would you like to merge main branch again? This seems to conflict with your previous pull request.

Ling-yunchi:
Done, merged latest main.

swwind (approved):
LGTM

PR #1284, opened 2026-07-15
feat(eslint-plugin-jest): add `no-interpolation-in-snapshots` rule

eryue0220:
## Summary

Port `no-interpolation-in-snapshots` from eslint-plugin-jest to rslint

## Related Links

eslint-plugin-jest/no-interpolation-in-snapshots [doc](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-interpolation-in-snapshots.md) [code](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/no-interpolation-in-snapshots.ts)

## Checklist

### DIFF
--- a/internal/plugins/jest/rules/no_interpolation_in_snapshots/no_interpolation_in_snapshots.go
+++ b/internal/plugins/jest/rules/no_interpolation_in_snapshots/no_interpolation_in_snapshots.go
@@ -19,16 +19,15 @@ var NoInterpolationInSnapshotsRule = rule.Rule{
 		return rule.RuleListeners{
 			ast.KindCallExpression: func(node *ast.Node) {
 				jestFnCall := utils.ParseJestFnCall(node, ctx)
-				if jestFnCall == nil || jestFnCall.Kind != utils.JestFnTypeExpect {
-					return
-				}
 
-				if !utils.INLINE_SNAPSHOT_MATCHERS[jestFnCall.Matcher] {
+				if jestFnCall == nil ||
+					jestFnCall.Kind != utils.JestFnTypeExpect ||
+					!utils.INLINE_SNAPSHOT_MATCHERS[jestFnCall.Matcher] {
 					return
 				}
 
 				for _, arg := range node.Arguments() {
-					if arg != nil && arg.Kind == ast.KindTemplateExpression {
+					if arg := ast.SkipParentheses(arg); arg != nil && arg.Kind == ast.KindTemplateExpression {
 						ctx.ReportNode(arg, buildNoInterpolationMessage())
 					}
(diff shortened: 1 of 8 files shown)


internal/plugins/jest/rules/no_interpolation_in_snapshots/no_interpolation_in_snapshots.go
```diff
@@ -0,0 +1,38 @@
...
+				if jestFnCall == nil || jestFnCall.Kind != utils.JestFnTypeExpect {
+					return
+				}
+
+				if !utils.INLINE_SNAPSHOT_MATCHERS[jestFnCall.Matcher] {
+					return
+				}
+
+				for _, arg := range node.Arguments() {
+					if arg != nil && arg.Kind == ast.KindTemplateExpression {
```
swwind:
```go
for _, arg := range node.Arguments() {
    if arg != nil && arg.Kind == ast.KindTemplateExpression {
```
`arg.Kind` is checked without unwrapping parentheses first. In typescript-go's AST, parentheses are real nodes (`KindParenthesizedExpression`), unlike ESTree where they're erased — so a parenthesized template argument won't have `Kind == KindTemplateExpression` and will slip through undetected.

False negative, verified locally:

```go
expect(something).toMatchInlineSnapshot((`${interpolated}`));
```
This rule reports 0 errors for it (confirmed via `go test ./internal/plugins/jest/rules/no_interpolation_in_snapshots/...` with this added as an invalid case — it fails with "Expected invalid test case to contain exactly 1 errors (reported 0 errors)").

Upstream `eslint-plugin-jest`'s `no-interpolation-in-snapshots` *does* flag the same snippet (verified locally with `eslint@8` + `eslint-plugin-jest@27`, ecmaVersion latest — reports `Do not use string interpolation inside of snapshots`), because ESTree collapses `(\`${y}\`)` and `` `${y}` `` into the same `TemplateLiteral` node. So this is a real regression introduced by the port, not matching upstream behavior.

Every sibling rule in this plugin that inspects an argument's `.Kind` (`prefer_equality_matcher.go`, `prefer_comparison_matcher.go`, `prefer_to_contain.go`, `no_jasmine_globals.go`, etc.) calls `ast.SkipParentheses(...)` before the comparison. This rule should do the same:

```go
if arg := ast.SkipParentheses(arg); arg != nil && arg.Kind == ast.KindTemplateExpression {
```

eryue0220: Got it, updated.

swwind (approved):

PR #1293, opened 2026-07-15
feat(eslint-plugin-jest): add `no-conditional-in-test` rule

eryue0220:
## Summary

Port `no-conditional-in-test` from eslint-plugin-jest to rslint

## Related Links

eslint-plugin-jest/no-conditional-in-test [doc](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-conditional-in-test.md) [code](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/no-conditional-in-test.ts)

## Checklist


internal/plugins/jest/rules/no_conditional_in_test/rule.go
```diff
@@ -0,0 +1,115 @@
...
+				parsed := jestUtils.ParseJestFnCall(node, ctx)
+				isUnsupportedFitConcurrent := parsed != nil &&
+					parsed.Name == "fit" &&
+					len(parsed.Members) == 1 &&
+					parsed.Members[0] == "concurrent"
+				if !isUnsupportedFitConcurrent && parsed != nil && parsed.Kind == jestUtils.JestFnTypeTest {
+					testCalls[node] = true
+					testCallDepth++
+				}
+				reportOptionalChain(node)
```
swwind: Bug: a test-declaring call using `?.` flags itself as a violation, even though real `eslint-plugin-jest` never reports it.

| code (`allowOptionalChaining: false`) | eslint | rslint |
|---|---|---|
| `it?.('foo', () => {});` | 0 errors | 1 error |
| `test?.('foo', () => {});` | 0 errors | 1 error |

Cause: `reportOptionalChain(node)` runs after `testCallDepth++` for the same node, so the call flags itself.
Fix: call `reportOptionalChain(node)` before incrementing `testCallDepth` for this node (move it above the `if !isUnsupportedFitConcurrent...` block). Please add these as valid test cases.

internal/plugins/jest/rules/no_conditional_in_test/rule.go
```diff
@@ -0,0 +1,115 @@
...
+					reportConditional(node)
+				}
+			},
+
+			ast.KindPropertyAccessExpression: reportOptionalChain,
+			ast.KindElementAccessExpression:  reportOptionalChain,
+
+			ast.KindCallExpression: func(node *ast.Node) {
+				parsed := jestUtils.ParseJestFnCall(node, ctx)
+				isUnsupportedFitConcurrent := parsed != nil &&
```
swwind: Bug: `xit.concurrent` bodies get scanned, but real `eslint-plugin-jest` doesn't recognize `xit.concurrent` as a jest call at all (same as `fit.concurrent`, which this rule already excludes).

| code | eslint | rslint |
|---|---|---|
| `xit.concurrent('foo', () => { if (x) {} });` | 0 errors | 1 error |

Cause: shared `jestUtils.VALID_JEST_FN_CALL_CHAINS` wrongly lists `xit.concurrent` as valid, and this rule only special-cases `fit.concurrent`.
Fix: add the same exclusion for `xit.concurrent` here, or better, remove `fit.concurrent`/`xit.concurrent` from `VALID_JEST_FN_CALL_CHAINS` in `jest.go` (root-cause fix, bigger blast radius). Please add the case above as a valid test case.

internal/plugins/jest/rules/no_conditional_in_test/rule.go
eryue0220: Updated.

internal/plugins/jest/rules/no_conditional_in_test/rule.go
eryue0220: Updated.

swwind:
Found one behavioral divergence from real `eslint-plugin-jest` while comparing edge cases against an actual ESLint + `eslint-plugin-jest` install:

```js
it('outer', () => {
  switch (x) { case 1: break; }
  it('inner', () => { doThing(); });
  if (y) {}
});
```

| | eslint | rslint |
|---|---|---|
| errors | 1 (only the `switch`) | 2 (`switch` **and** `if (y)`) |

Cause: upstream's `inTestCase` is a plain boolean that gets reset to `false` when the inner `it('inner', ...)` call exits, even though the outer test is still active — so any conditional appearing *after* a nested test call goes unflagged upstream. This rule's `testCallDepth`/`testCalls` map tracks nesting depth correctly, which is more accurate but produces an extra error real `eslint-plugin-jest` would never report for this (admittedly rare) case.

Not sure if this edge case matters enough to intentionally replicate upstream's boolean-reset quirk, but flagging since it's a verifiable false positive relative to eslint.

eryue0220:
> Found one behavioral divergence from real `eslint-plugin-jest` while comparing edge cases against an actual ESLint + `eslint-plugin-jest` install:
> 
> ```js
> it('outer', () => {
>   switch (x) { case 1: break; }
>   it('inner', () => { doThing(); });
>   if (y) {}
> });
> ```
> 
> eslint	rslint
> errors	1 (only the `switch`)	2 (`switch` **and** `if (y)`)
> Cause: upstream's `inTestCase` is a plain boolean that gets reset to `false` when the inner `it('inner', ...)` call exits, even though the outer test is still active — so any conditional appearing _after_ a nested test call goes unflagged upstream. This rule's `testCallDepth`/`testCalls` map tracks nesting depth correctly, which is more accurate but produces an extra error real `eslint-plugin-jest` would never report for this (admittedly rare) case.
> 
> Not sure if this edge case matters enough to intentionally replicate upstream's boolean-reset quirk, but flagging since it's a verifiable false positive relative to eslint.

Yes, I noticed this before, and this behavior is intentional. If need to aligned with the `eslint-plugin-jest` strictly then I can change this behavior.

swwind:
> Yes, I noticed this before, and this behavior is intentional. If need to aligned with the eslint-plugin-jest strictly then I can change this behavior.

Our current guideline for migrating existing rules is to strictly preserve all existing behaviors so that users can migrate smoothly. Although this scenario may not be particularly meaningful, we would still like to keep the rule's current behavior here for consistency. Thanks for your contribution!

eryue0220:
> > Yes, I noticed this before, and this behavior is intentional. If need to aligned with the eslint-plugin-jest strictly then I can change this behavior.
> 
> Our current guideline for migrating existing rules is to strictly preserve all existing behaviors so that users can migrate smoothly. Although this scenario may not be particularly meaningful, we would still like to keep the rule's current behavior here for consistency. Thanks for your contribution!

Updated.

swwind (approved):

fansenze (approved):

PR #1294, opened 2026-07-15
fix(website): correct rule manifest identity and test status

elecmonkey:
## Summary

Fix rule manifest generation so implementation status is tracked using the complete rule identity, and close the robustness gaps raised in review.

- Use a single `group:rule` key (via a `ruleKey` helper) for dedup and test-status lookup, so same-named rules from different plugins (e.g. `eslint/no-shadow` and `@typescript-eslint/no-shadow`) are no longer collapsed to one entry.
- Give the `react` plugin a `PLUGIN_NAME = "eslint-plugin-react"` constant (mirroring `react-hooks`) instead of special-casing the directory name in the generator; the registry `group` is updated to match. Route slugs come from `prefix`, so `/rules/react/` URLs and preset mapping are unaffected.
- Match nested / multi-file rule test paths (`tests/{dir}/rules/{rule}/*.test.ts`) and associate every enabled test file with its rule.
- Scan those test files for skip cases, including statement-level `skip: true` objects that have no `name` (line-based placeholder), so surfacing nested files cannot silently hide their skipped cases.
- Fail the build when a group's test directory is missing or the include list parses to zero entries, since the manifest is gitignored and built at site-build time.
- Update the frontend rule-status table row key to `group:rule`.

## Implementation Overview

Measured against `main` by regenerating `rule-manifest.json` before and after:

| Status | Before | After | Change |
| --- | ---: | ---: | ---: |
| Fully Implemented | 402 | 403 | +1 |
| Partial Implemented | 1 | 18 | +17 |
| Partial Tested | 45 | 42 | -3 |
| Total Rules | 448 | 463 | +15 |

The large `partial-impl` increase is expected: nested test files and unnamed `skip: true` cases are now scanned, so rules that were previously reported as `full` (or hidden entirely) now correctly report their skipped cases.

## Rule Changes

Rules restored after replacing bare-name deduplication with complete rule identities (previously dropped because a same-named rule in another plugin won the dedup):

- `eslint/max-params` → `full`
- `eslint/no-dupe-class-members` → `full`
- `eslint/no-empty-function` → `full`
- `eslint/no-implied-eval` → `full`
- `eslint/no-loop-func` → `full`
- `eslint/no-redeclare` → `partial-impl`
- `eslint/no-restricted-imports` → `full`
- `eslint/no-shadow` → `full`
- `eslint/no-unused-expressions` → `partial-impl`
- `eslint/no-unused-private-class-members` → `full`
- `eslint/no-useless-constructor` → `full`
- `eslint/prefer-destructuring` → `full`
- `eslint/prefer-promise-reject-errors` → `full`
- `eslint/require-await` → `full`
- `@typescript-eslint/no-deprecated` → `partial-impl`

Existing rules whose status is corrected once they are keyed and scanned by their own identity:

- `@typescript-eslint/class-methods-use-this`: `partial-test` → `full`
- `@typescript-eslint/prefer-optional-chain`: `partial-test` → `full`
- `@typescript-eslint/naming-convention`: `partial-test` → `partial-impl`
- `@typescript-eslint/no-unused-vars`: `partial-test` → `partial-impl`
- `@typescript-eslint/no-shadow`: `full` → `partial-impl`
- `@typescript-eslint/prefer-promise-reject-errors`: `full` → `partial-test`
- `@typescript-eslint/no-confusing-void-expression`, `no-duplicate-type-constituents`, `no-redeclare`, `no-unnecessary-boolean-literal-compare`, `no-unnecessary-condition`, `prefer-nullish-coalescing`, `strict-boolean-expressions`: `full` → `partial-impl`
- `eslint/no-invalid-regexp`, `no-loss-of-precision`, `prefer-exponentiation-operator`, `prefer-regex-literals`: `full` → `partial-impl`

### DIFF
--- a/scripts/gen-rule-manifest.js
+++ b/scripts/gen-rule-manifest.js
@@ -78,10 +78,19 @@ function groupToTestDir(group) {
   return group.replace(/^@/, '');
 }
 
-function getIncludedRuleTests() {
+// Canonical manifest key. The whole file uses the `group` namespace so a
+// lookup can never accidentally mix `group` and `testDir` spellings.
+function ruleKey(group, rule) {
+  return `${group}:${rule}`;
+}
+
+function getIncludedRuleTests(groups) {
   // Parse rstest.config.mts include list and associate each rule with its
   // enabled test files, including tests nested under a rule directory.
   const config = fs.readFileSync(TEST_CONFIG_PATH, 'utf-8');
+  // Translate the on-disk test directory back to its manifest group once here,
+  // so the returned map lives entirely in the `group` namespace.
+  const testDirToGroup = new Map(groups.map((g) => [groupToTestDir(g), g]));
   // Match uncommented flat and nested paths:
   //   ./tests/{testDir}/rules/{rule}.test.ts
   //   ./tests/{testDir}/rules/{rule}/{test-file}.test.ts
@@ -90,11 +99,11 @@ function getIncludedRuleTests() {
   const included = new Map();
   let match;
   while ((match = includeRegex.exec(config))) {
+    const group = testDirToGroup.get(match[2]);
+    if (!group) continue; // test dir with no matching plugin group
     const testPath = path.resolve(path.dirname(TEST_CONFIG_PATH), match[1]);
-    const testDir = match[2];
     const rule = match[3].replace(/-/g, '_');
-    // Key by test-dir + rule so same-named rules in different plugins stay distinct.
-    const key = `${testDir}:${rule}`;
+    const key = ruleKey(group, rule);
     if (!included.has(key)) included.set(key, new Set());
     included.get(key).add(testPath);
   }
@@ -286,14 +295,15 @@ function getDocPath(rule, pluginDir) {
 }
 
 function buildManifest() {
-  const included = getIncludedRuleTests();
   const ruleEntries = [...getPluginRuleEntries(), ...getCoreRuleEntries()];
   // Deduplicate by group + rule name, keeping first entry.
   const seen = new Map();
   for (const e of ruleEntries) {
-    const key = `${e.group}:${e.rule}`;
+    const key = ruleKey(e.group, e.rule);
     if (!seen.has(key)) seen.set(key, e);
   }
+  const groups = [...new Set(ruleEntries.map((e) => e.group))];
+  const included = getIncludedRuleTests(groups);
   const rules = Array.from(seen.values())
     .sort(
       (a, b) => a.rule.localeCompare(b.rule) || a.group.localeCompare(b.group),
@@ -303,7 +313,7 @@ function buildManifest() {
       let status = 'full';
       let failing_case = [];
       const group = entry.group;
-      const testFiles = included.get(`${groupToTestDir(group)}:${rule}`);
+      const testFiles = included.get(ruleKey(group, rule));
       if (!testFiles) {
         status = 'partial-test';
       } else {
--- a/scripts/gen-rule-manifest.js
+++ b/scripts/gen-rule-manifest.js
@@ -91,6 +91,19 @@ function getIncludedRuleTests(groups) {
   // Translate the on-disk test directory back to its manifest group once here,
   // so the returned map lives entirely in the `group` namespace.
   const testDirToGroup = new Map(groups.map((g) => [groupToTestDir(g), g]));
+  // Fail loudly if a group's expected test directory is missing: otherwise
+  // every rule in that group silently degrades to partial-test, and because
+  // the manifest is gitignored and built at site-build time the flip would
+  // never surface in a diff.
+  for (const [testDir] of testDirToGroup) {
+    const dir = path.join(TESTS_BASE_DIR, testDir);
+    if (!fs.existsSync(dir)) {
+      throw new Error(
+        `Expected test directory not found: ${path.relative(REPO_ROOT, dir)} ` +
+          `(group "${testDirToGroup.get(testDir)}"). Update groupToTestDir or the test layout.`,
+      );
+    }
+  }
   // Match uncommented flat and nested paths:
   //   ./tests/{testDir}/rules/{rule}.test.ts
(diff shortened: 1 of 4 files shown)


scripts/gen-rule-manifest.js
```diff
@@ -85,8 +87,10 @@ function getIncludedRules() {
   const included = new Set();
   let match;
   while ((match = includeRegex.exec(config))) {
+    const testDir = match[1];
     const rule = match[2].replace(/-/g, '_');
-    included.add(rule);
+    // Key by test-dir + rule so same-named rules in different plugins stay distinct.
+    included.add(`${testDir}:${rule}`);
```
swwind: The composite key introduces a status regression here: `includeRegex` (`([\w-]+)\.test\.ts` — `[\w-]` can't span `/`) never matches nested paths like `'./tests/typescript-eslint/rules/no-shadow/no-shadow.test.ts'` (enabled at rstest.config.mts:344). Pre-PR, the name-only key was accidentally satisfied by eslint's flat `no-shadow.test.ts`, so `@typescript-eslint/no-shadow` showed `full`; with the dir-scoped key, `typescript-eslint:no_shadow` is never added and the rule flips to `partial-test` despite having active tests.

`class-methods-use-this`, `naming-convention`, `no-unused-vars`, `prefer-optional-chain` hit the same blind spot (they were already mislabeled pre-PR, so no-shadow is the only new flip). Allowing an optional subdirectory segment after `rules/` in the regex would fix all of them.

scripts/gen-rule-manifest.js
```diff
@@ -270,19 +274,22 @@ function getDocPath(rule, pluginDir) {
...
+    .sort(
+      (a, b) => a.rule.localeCompare(b.rule) || a.group.localeCompare(b.group),
+    )
+    .map((entry) => {
+      const rule = entry.rule;
       let status = 'full';
       let failing_case = [];
       const group = entry.group;
-      if (!included.has(rule)) {
+      if (!included.has(`${groupToTestDir(group)}:${rule}`)) {
```
swwind: The failure mode this PR fixes stays completely silent for the next occurrence: a `groupToTestDir` result that doesn't exist on disk, or a formatter switching rstest.config.mts to double quotes (the regex hard-codes `'`), just makes `included.has()` return false — every affected rule silently becomes `partial-test`, and since `website/generated/rule-manifest.json` is gitignored and built at site-build time, the mass flip never shows up in any diff.

Cheap guards would turn both into build failures: `fs.existsSync(path.join(TESTS_BASE_DIR, groupToTestDir(group)))` per group (the test dirs are enumerable), plus asserting `included.size > 0` after parsing.

scripts/gen-rule-manifest.js
```diff
@@ -73,6 +73,8 @@ function getPluginRuleEntries() {
 
 // Map group name to test directory name: "@typescript-eslint" -> "typescript-eslint", etc.
 function groupToTestDir(group) {
+  // The manifest uses "react", while its tests live under "eslint-plugin-react".
+  if (group === 'react') return 'eslint-plugin-react';
```
swwind: This special case papers over the actual root cause: `internal/plugins/react/plugin.go` is the only plugin.go without a `PLUGIN_NAME` constant (it's a one-line package declaration), so the group falls back to the dir name `react`. The deeper fix is small and safe: add `const PLUGIN_NAME = "eslint-plugin-react"` and update the registry entry's `group` in `website/theme/plugin-registry.ts` — route slugs come from `prefix`, not `group`, so `/rules/react/` URLs and preset mapping are unaffected (this is exactly the existing react-hooks pattern: group `eslint-plugin-react-hooks`, prefix `react-hooks`).

As written, if someone later adds `PLUGIN_NAME` for consistency with the other seven plugins, this branch goes dead and the registry's `group: 'react'` join breaks with no error; and the next plugin missing `PLUGIN_NAME` regresses its whole group to `partial-test` silently.

scripts/gen-rule-manifest.js
```diff
@@ -270,19 +274,22 @@ function getDocPath(rule, pluginDir) {
 function buildManifest() {
   const included = getIncludedRules();
   const ruleEntries = [...getPluginRuleEntries(), ...getCoreRuleEntries()];
-  // Deduplicate by rule name, keeping first entry
+  // Deduplicate by group + rule name, keeping first entry.
   const seen = new Map();
   for (const e of ruleEntries) {
-    if (!seen.has(e.rule)) seen.set(e.rule, e);
+    const key = `${e.group}:${e.rule}`;
```
swwind: Minor, but this file now has two key namespaces with the identical `x:y` shape: this dedup key uses `group` (`@typescript-eslint:no_shadow`) while the `included` set uses `testDir` (`typescript-eslint:no_shadow`), bridged per-lookup at line 292. A maintainer mirroring this key two lines down as `included.has(`${group}:${rule}`)` gets an always-false lookup and a silent all-`partial-test` manifest. A single `ruleKey(group, rule)` helper — or translating testDir→group once at parse time so the file uses one namespace — would remove the trap (the website side already has the `manifestKey` idiom in `plugin-rule-manifest.ts`).

swwind (commented):
Ran a review pass over this PR — the composite-key direction is right, but the keying change surfaces one real status regression plus a few robustness gaps in the same code path.

One finding that can't be anchored inline (untouched lines): **`getSkipCases` (line ~238) only probes the flat `<testsDir>/<rule>.test.ts` path**, so nested/multi-file layouts (`tests/typescript-eslint/rules/no-shadow/`, `no-unused-vars/`, `naming-convention/`, …) are never scanned for skips — and those nested files do contain real `skip: true` cases (no-shadow ×5, no-unused-vars 30+). Today they're shielded by the include-regex gap noted inline (they land on `partial-test` anyway), but the moment that regex learns about nested dirs, these rules would flip to `full` while silently hiding their skipped cases. Worth fixing both together.

elecmonkey:
@codex review

fansenze (approved):

PR #1318, opened 2026-07-20
feat: port rule prefer-object-spread

swwind:
## Summary

Port the `prefer-object-spread` rule from ESLint to rslint.

Requires using object spread syntax instead of `Object.assign` calls whose first argument is an object literal, including the single-argument case where the `Object.assign` call is unnecessary.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/prefer-object-spread
- Source code: https://github.com/eslint/eslint/blob/main/lib/rules/prefer-object-spread.js

## Checklist


internal/rules/prefer_object_spread/prefer_object_spread.go
```diff
@@ -0,0 +1,505 @@
...
+// assertion wrappers, is an unshadowed reference to the global `name` — i.e.
+// no local declaration shadows it and no `/* global name: off */` /
+// languageOptions.globals entry un-declares it. Mirrors the "is this the
+// global variable, not a local rebinding" half of ESLint's ReferenceTracker.
+func isGlobalIdentifier(node *ast.Node, name string, globals map[string]bool) bool {
+	node = utils.SkipAssertionsAndParens(node)
+	if node == nil || !ast.IsIdentifier(node) || node.AsIdentifier().Text != name {
+		return false
+	}
+	if utils.IsShadowed(node, name) {
```
elecmonkey: This still diverges from ESLint's `ReferenceTracker` when the global `Object` binding has any bare write reference in the file. ESLint stops tracking that global and reports no `Object.assign` calls in the file, even calls that appear before the write. I verified with `eslint@10.7.0` and `ecmaVersion: 2026`: `Object = {}; Object.assign({}, foo);`, `Object ||= {}; Object.assign({}, foo);`, and `Object.assign({}, before); Object = {}; Object.assign({}, after);` all produce 0 diagnostics. By contrast, this implementation only checks declaration shadowing / `globals: off`, so the same cases are still reported and autofixed by rslint (confirmed locally with an overlay rule test).\n\nTo match upstream, the global-reference check needs to treat writes to the global `Object` variable as making `Object.assign` untrackable for this rule, rather than only checking local declarations that shadow `Object`.

elecmonkey (commented):
I found one behavior mismatch with ESLint 10.7.0 while reviewing the prefer-object-spread port.

internal/rules/prefer_object_spread/prefer_object_spread.go
swwind: fixed

fansenze (approved):

PR #1319, opened 2026-07-20
feat: port rule dot-notation

swwind:
## Summary

Port the `dot-notation` rule from ESLint to rslint.

Enforces dot notation whenever possible, converting bracket notation (`obj["foo"]`) to dot notation (`obj.foo`) when safe, and vice versa for reserved words when `allowKeywords: false`. Registered as `dot-notation` (core, no type-info requirement), distinct from the existing `@typescript-eslint/dot-notation` which requires a type checker and only runs on typed projects.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/dot-notation
- Source code: https://github.com/eslint/eslint/blob/main/lib/rules/dot-notation.js

## Checklist


elecmonkey (dismissed):

elecmonkey (commented):
I found one behavior mismatch with ESLint 10.7.0 while reviewing the dot-notation port.

internal/rules/dot_notation/dot_notation.go
```diff
@@ -0,0 +1,355 @@
...
+
+		// ECMAScript + Unicode flags mirror ESLint's `new RegExp(pattern, 'u')`
+		// so user patterns using lookaround, backreferences, or `\p{...}` work
+		// identically to the original rule (Go's standard `regexp` / RE2 does
+		// not support those). Invalid regex patterns are silently ignored, as
+		// upstream would throw at rule-construction time in that case, which
+		// is out of scope here.
+		var allowRE *regexp2.Regexp
+		if opts.AllowPattern != "" {
+			if re, err := regexp2.Compile(opts.AllowPattern, regexp2.ECMAScript|regexp2.Unicode); err == nil {
```
elecmonkey, line 168: This diverges from ESLint for invalid `allowPattern` values. ESLint constructs the option with `new RegExp(options.allowPattern, "u")`, so an invalid pattern such as `"("` throws while loading `dot-notation`. I verified this locally with `eslint@10.7.0`: `dot-notation: ["error", { allowPattern: "(" }]` on `a['b'];` throws `Error while loading rule 'dot-notation': Invalid regular expression: /(/u: Unterminated group`.

With this implementation, the compile error is ignored, `allowRE` stays nil, and the same case continues linting and reports/fixes `a['b']` to `a.b`. I also verified that behavior locally against this PR via an overlay Go rule test. To match upstream, invalid `allowPattern` should surface as a rule/configuration error instead of being treated as if no pattern was configured.

swwind: fixed

internal/rules/dot_notation/dot_notation.go
```diff
@@ -0,0 +1,348 @@
...
+			// untouched, so fixes on nested accesses like `a['b']['c']` don't
+			// overlap and can both apply in a single pass.
+			var replacement string
+			if hasQuestionDot {
+				// The untouched `?.` operator already supplies the dot:
+				// `a?.['b']` -> `a?.b`.
+				replacement = value
+			} else {
+				sep := "."
+				if isDecimalIntegerLiteral(sourceFile, elem.Expression) {
```
elecmonkey, line 238: **Minor fixer divergence: parenthesized decimal-integer object doesn't get the disambiguating space.**

```
(5)['prop'];      upstream -> (5) .prop;     rslint -> (5).prop;
(((5)))['prop'];  upstream -> (((5))) .prop; rslint -> (((5))).prop;
```

Upstream checks `isDecimalInteger(node.object)`, and in ESTree `node.object` is the bare numeric `Literal` (parens have no wrapper node), so it inserts `" ."`. Here `elem.Expression` is a tsgo `ParenthesizedExpression`, so `isDecimalIntegerLiteral` returns false and no space is added.

Both outputs are valid and semantically identical (unlike bare `5['prop']`, where the space is *required*), so this is purely a byte-for-byte fixer mismatch — arguably rslint's is cleaner. But if strict upstream parity is the goal, unwrapping parentheses on `elem.Expression` before the decimal-integer check would align them. Verified against ESLint 10.7.

elecmonkey (commented):
Really thorough port — I hammered it with 60+ edge cases against the real ESLint **10.7** rule (running `verify` + `verifyAndFix` on both the default and `@typescript-eslint/parser`, comparing report position, message, and fixed output): unicode/hex escapes in string & template keys, `allowPattern` with `\p{Lu}` and lookbehind, fixer token-adjacency (`instanceof`/`in`/regex-division/tagged-template), numeric-object spacing, computed `null`/`true`/`false` keys, the `let[` destructuring guard, and various TS receiver wrappers. All matched upstream.

elecmonkey (approved):

fansenze (approved):

PR #1325, opened 2026-07-21
perf: prefilter identifiers before checker lookup in prefer-const write scans

swwind:
## Summary

Stacked on #1324 — last of the four brute-force reference-collection rules from the same profile (see #1321).

- `prefer-const`'s `countWriteReferences` and `findWriteInSameBlock` walked the enclosing scope calling `checker.GetSymbolAtLocation` on every identifier before checking anything else. Now the cheap syntax checks come first — name match and `utils.IsWriteReference` (write positions are rare) — so the checker lookup only runs for genuine write candidates. This mirrors the existing `countWritesBySym` in the same file, which already checks name and write position before resolving.
- `isReadBeforeFirstAssign` gets the name pre-filter too (it needs read references, so no write-position condition).

## Benchmark

Linting the TypeScript repo (v6.0.3, 746 files, 152 rules, fully type-aware) on a 22-core machine, measured on top of #1324:

| | before | after |
|---|---|---|
| `GetSymbolAtLocation` CPU attributed to `prefer_const` | 1.63s | **0.02s** |
| wall time | 5.45s | **5.23s** |
| Go CPU profile samples | 18.05s | 17.31s |

Cumulative for the whole stack (#1321 → this PR): wall time 13.7s → 5.23s, Go CPU 29.3s → 17.3s on this benchmark.

## Related Links

- #1321, #1322, #1324

## Checklist

- [x] Tests updated (or not required) — behavior-preserving; existing `prefer_const` suite passes unchanged.

### DIFF
--- a/internal/rules/prefer_const/prefer_const.go
+++ b/internal/rules/prefer_const/prefer_const.go
@@ -267,9 +267,11 @@ func countWriteReferences(sym *ast.Symbol, declName string, declNode *ast.Node, 
 			return
 		}
 
-		if n.Kind == ast.KindIdentifier && !isPartOfDeclaration(n, declNode) {
-			refSym := ctx.TypeChecker.GetSymbolAtLocation(n)
-			if refSym == sym && utils.IsWriteReference(n) {
+		// Cheap syntax checks (name match, write position) come first so the
+		// expensive checker lookup only runs for genuine write candidates.
+		if n.Kind == ast.KindIdentifier && n.Text() == declName &&
+			utils.IsWriteReference(n) && !isPartOfDeclaration(n, declNode) {
+			if ctx.TypeChecker.GetSymbolAtLocation(n) == sym {
 				count++
 			}
 		}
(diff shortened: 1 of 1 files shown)


swwind:
Follow-up data point for "what's next after this stack" — I experimented with the checker-pool size (PERF investigation item 3: pool is hardcoded to 4 in `checkerpool.go`, and the linter's per-file parallelism is bounded by it). Temporarily set `CompilerOptions.Checkers` before program creation and re-ran the same TypeScript-repo benchmark (746 files, 152 rules, 22-core machine, all fixes from this stack applied):

| checkers | lint wall | Go CPU samples | avg parallelism |
|---|---|---|---|
| 1 | 5.23s | — | — |
| 2 | 4.88s | — | — |
| 4 (default) | 4.66–4.82s | 17.2s | ~4.7 cores |
| 8 | 4.98s | 19.2s | ~5.0 cores |
| 22 (= cores) | 5.27s | 21.4s | ~5.3 cores |

Two takeaways:

- **Scaling up is a net loss.** Checker caches are per-instance, so first-touch lazy work (checker init, global symbol merging, import alias resolution) multiplies with the pool size: +4.2s CPU at 22 checkers, while average parallelism barely moves because the `index%N` sharding never splits the long-tail group containing `checker.ts`.
- **Even scaling down barely hurts** — 1 checker costs only ~0.4s wall vs the default. Cross-checker parallelism contributes very little on this benchmark; the wall-time floor is the serial lint of the largest single file (~3s for `checker.ts` with 91 native rules) plus the Node-side plugin pass it overlaps with.

So checker count / shard weighting looks low-priority. The bigger remaining wins on the Go side are a shared per-file reference index (each rule currently re-walks the whole AST for its own collection — `ForEachChild` is ~59% cum CPU after this stack) and parallelizing the no-type-info file group. Not part of this stack; recording here so the numbers aren't lost.

fansenze (dismissed):

fansenze (approved):

fansenze (approved):

PR #1332, opened 2026-07-22
fix(tsgo): preserve merged local value symbols

quininer:
## Summary

A symbol can merge an imported type alias with a local runtime value, as in an imported interface and a same-named local class. Recording every Alias-flagged symbol in alias_symbols redirected runtime references to the imported type and caused the local value implementation to be tree-shaken.

Resolve only pure aliases that do not also carry a runtime Value flag. Merged local values keep their own symbol identity, while normal imports and re-exports continue to use alias resolution.

The semantic regression imports a SymbolLinks interface, declares a same-named local class, and constructs it at runtime. Applied to the parent, it fails with `local value symbol ... was incorrectly aliased`; it passes with this change, as does the cmd/tsgo test suite.

## Checklist

### DIFF
--- a/cmd/tsgo/semantic.go
+++ b/cmd/tsgo/semantic.go
@@ -267,8 +267,9 @@ func CollectSemanticInFile(tc *checker.Checker, file *ast.SourceFile, semantic *
 					semantic.Node2sym[key] = sym_id
 					semantic.Node2type[key] = typeID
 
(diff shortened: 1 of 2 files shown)


fansenze (approved):

quininer:
merge on https://github.com/web-infra-dev/rslint/pull/1333

PR #1335, opened 2026-07-22
perf: add lazy per-file reference index and port no-var to it

swwind:
## Summary

- Add `rule.RefStore`, a lazily built per-file identifier-reference index exposed as `ctx.Refs`. One AST walk buckets reference-position identifiers by name; each queried name is then resolved with the binder's `NameResolver` (the same scope walk the checker uses for identifiers), so lookups never touch the TypeChecker or trigger lazy type computation. Rules only pay for names they actually ask about.
- Port no-var's reference collection to `ctx.Refs`, deleting its per-file AST scans and `GetSymbolAtLocation` calls.
- Add the `shim/binder` module (generated by `gen_shims`) exposing `binder.NameResolver`.

On the microsoft/TypeScript benchmark repo (746 files, 152 rules), no-var drops from ~1.6s CPU to zero profile samples and RefStore itself doesn't register; wall time improves ~5.3s → ~4.1s under the profiler. Lint output is unchanged.

Other reference-collecting rules (no-unused-vars, prefer-const, naming-convention) can migrate to `ctx.Refs` in follow-ups.

Skills updated accordingly.

## Why

ESLint's scope system tracks both symbol definitions and references. Currently, tsgo's symbol system only provides definition locations, forcing rules that need reference information to perform their own AST traversals. As a result, the complexity of the no-var rule can reach O(n × d), where n is the AST size and d is the number of variable declarations.

This PR adds a lazy O(n) preprocessing step to collect and cache symbol references, enabling rules to reuse the information without redundant AST walks.


swwind:
Follow-up candidates: other rules with a manual "collect all references to symbol X" step that `RefStore` could replace.

**High confidence (binder-resolvable, file/scope-local):**
- `no-class-assign` — `internal/rules/no_class_assign/no_class_assign.go` (`checkClassReassignments`)
- `no-func-assign` — `internal/rules/no_func_assign/no_func_assign.go` (`checkReassignments`)
- `no-ex-assign` — `internal/rules/no_ex_assign/no_ex_assign.go` (`checkReassignments`)
- `no-import-assign` — `internal/rules/no_import_assign/no_import_assign.go` (import listener, whole-file name-match + `GetSymbolAtLocation` confirm)
- `no-shadow-restricted-names` — `internal/rules/no_shadow_restricted_names/no_shadow_restricted_names.go` (`collectWrittenUndefinedSymbols`)
- `prefer-const` — `internal/rules/prefer_const/prefer_const.go` (`countWriteReferences`, `isReadBeforeFirstAssign`, `findWriteInSameBlock`, `countWritesBySym`)
- `exhaustive-deps` (react-hooks) — `internal/plugins/react_hooks/rules/exhaustive_deps/exhaustive_deps.go` (`isUsedOutsideHook`, currently bounded to avoid an O(file-size) regression)

**Medium confidence (partial replacement, extra checker-only needs remain):**
- `no-const-assign` — `internal/rules/no_const_assign/no_const_assign.go`
- `no-unused-vars` (core) — `internal/rules/no_unused_vars/no_unused_vars.go` (has its own binder-based resolver already; would consolidate rather than newly benefit)
- `no-unused-vars` (typescript-eslint) — `internal/plugins/typescript/rules/no_unused_vars/no_unused_vars.go` (import/alias/shorthand/JSX-factory paths still need the checker)
- `utils.ReferenceIndex` — `internal/utils/reference_index.go` (`ForEachReferenceByName`/`buildNameIndex` largely duplicates RefStore's candidate walk; shared by `no_loop_func`, `no_unassigned_vars`, `no_unused_vars`)
- `no-loop-func` — `internal/rules/no_loop_func/no_loop_func.go` (`CollectThroughReferences`; `isSafeCore` still needs loop-position reasoning)
- `prefer-arrow-callback` — `internal/rules/prefer_arrow_callback/prefer_arrow_callback.go` (`functionNameReferenced`, already bounded to one function body)

Checked and ruled out: `no_unused_private_class_members` (both variants — keyed on `PrivateIdentifier`, not `Identifier`), `require_atomic_updates` (needs control-flow-ordered reference info, not a flat list), `no_unused_state`/`no_unused_class_component_methods` (property-name string matching, no declared symbol), `no_use_before_define`/`no_undef`/`no_shadow`/`no_redeclare` (single-point lookups or declaration-only scope trees, no aggregate reference step).

fansenze (approved):

swwind:
A data point on which rules are worth migrating to `ctx.Refs`: I tried porting `no-ex-assign` and it regressed, so I dropped it.

Measured in isolation with `--timing all` on a 1200-file synthetic corpus, 5 runs after a warmup — mean rule time went from **69.5 ms to 158.6 ms**, a 2.3× slowdown.

The reason is that `no-ex-assign`'s existing implementation only walks the **catch block**. Its scope is already small and there is no per-declaration re-walk to eliminate, so the port swaps that small scoped walk for `collectCandidates`' full-`SourceFile` walk plus name resolution — strictly more work.

Heuristic for the remaining migration candidates: `ctx.Refs` pays off when a rule's existing reference collection re-walks a large scope once per declaration, and costs more when the rule is already scoped to a small subtree (a catch clause, a single function body). Worth measuring per rule before opening the PR rather than assuming.

PR #1338, opened 2026-07-22
feat(eslint-plugin-jest): add `require-hook` rule

eryue0220:
## Summary

Port `require-hook` from eslint-plugin-jest to rslint

## Related Links

eslint-plugin-jest/require-hook [doc](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/require-hook.md) [code](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/require-hook.ts)

## Checklist

### DIFF
--- a/internal/plugins/jest/rules/require_hook/require_hook.go
+++ b/internal/plugins/jest/rules/require_hook/require_hook.go
@@ -92,16 +92,25 @@ func shouldBeInHook(node *ast.Node, ctx rule.RuleContext, allowedFunctionCalls [
 
 	switch node.Kind {
 	case ast.KindExpressionStatement:
-		return shouldBeInHook(node.AsExpressionStatement().Expression, ctx, allowedFunctionCalls)
+		return shouldBeInHook(ast.SkipParentheses(node.AsExpressionStatement().Expression), ctx, allowedFunctionCalls)
 	case ast.KindCallExpression:
+		call := node.AsCallExpression()
+		if call.Expression.Kind == ast.KindImportKeyword ||
+			call.QuestionDotToken != nil ||
+			ast.IsOptionalChain(node) {
+			return false
+		}
 		if isJestFnCall(node, ctx) {
 			return false
 		}
 		name := utils.CalleeChainName(node)
 		return !containsString(allowedFunctionCalls, name)
 	case ast.KindVariableStatement:
+		if ast.HasSyntacticModifier(node, ast.ModifierFlagsExport) {
+			return false
+		}
 		declList := node.AsVariableStatement().DeclarationList
-		if declList == nil || declList.Flags&ast.NodeFlagsConst != 0 {
+		if declList == nil || declList.Flags&ast.NodeFlagsBlockScoped == ast.NodeFlagsConst {
 			return false
 		}
 		decls := declList.AsVariableDeclarationList().Declarations
@@ -138,9 +147,21 @@ func getFunctionBodyBlock(fn *ast.Node) *ast.Node {
 	return nil
 }
 
+func hasCallExpressionParent(node *ast.Node) bool {
+	if node == nil {
+		return false
+	}
+	parent := node.Parent
+	for parent != nil && parent.Kind == ast.KindParenthesizedExpression {
+		parent = parent.Parent
+	}
+	return parent != nil && parent.Kind == ast.KindCallExpression
+}
+
 func describeCallbackBody(call *ast.Node, ctx rule.RuleContext) *ast.Node {
 	if call == nil ||
 		call.Kind != ast.KindCallExpression ||
+		hasCallExpressionParent(call) ||
 		!utils.IsTypeOfJestFnCall(call, ctx, utils.JestFnTypeDescribe) {
 		return nil
 	}
--- a/internal/plugins/jest/rules/require_hook/require_hook.go
+++ b/internal/plugins/jest/rules/require_hook/require_hook.go
@@ -1,0 +1,183 @@
+package require_hook
+
+import (
+	"strings"
+
+	"github.com/microsoft/typescript-go/shim/ast"
+	"github.com/web-infra-dev/rslint/internal/plugins/jest/utils"
+	"github.com/web-infra-dev/rslint/internal/rule"
+)
+
+func buildUseHookMessage() rule.RuleMessage {
+	return rule.RuleMessage{
+		Id:          "useHook",
+		Description: "This should be done within a hook",
+	}
+}
+
+type Options struct {
+	AllowedFunctionCalls []string
+}
+
+func parseAllowedFunctionCalls(raw any) []string {
+	items, ok := raw.([]interface{})
+	if !ok {
+		return nil
+	}
+	out := make([]string, 0, len(items))
+	for _, item := range items {
+		if s, ok := item.(string); ok {
+			out = append(out, s)
+		}
+	}
+	return out
+}
+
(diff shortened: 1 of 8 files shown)


internal/plugins/jest/rules/require_hook/require_hook.go
```diff
@@ -0,0 +1,183 @@
...
+
+	switch node.Kind {
+	case ast.KindExpressionStatement:
+		return shouldBeInHook(node.AsExpressionStatement().Expression, ctx, allowedFunctionCalls)
+	case ast.KindCallExpression:
+		if isJestFnCall(node, ctx) {
+			return false
+		}
+		name := utils.CalleeChainName(node)
+		return !containsString(allowedFunctionCalls, name)
```
elecmonkey, line 107: Two divergences originate in this `CallExpression` branch:

1. **Optional calls** — `setup?.();` / `foo?.bar();`. Upstream wraps these in an ESTree `ChainExpression` (no case for it → no report), but tsgo produces a plain `CallExpression`, so rslint reports. Suggest skipping calls carrying the optional-chain flag (`ast.NodeFlagsOptionalChain` / `QuestionDotToken`).
2. **Dynamic import** — `import('x');`. Upstream sees `ImportExpression`, not `CallExpression` → no report. Suggest excluding calls whose callee is the `import` keyword (`ast.KindImportKeyword`).

internal/plugins/jest/rules/require_hook/require_hook.go
```diff
@@ -0,0 +1,183 @@
...
+	case ast.KindVariableStatement:
+		declList := node.AsVariableStatement().DeclarationList
+		if declList == nil || declList.Flags&ast.NodeFlagsConst != 0 {
```
elecmonkey: **Exported declarations** — `export let x = foo();`. Upstream sees an `ExportNamedDeclaration` (and `export default` → `ExportDefaultDeclaration`), neither of which the `switch` handles → no report. In tsgo the export is a modifier on the `VariableStatement`, so this branch fires. Suggest bailing out here when the statement carries an `export` modifier.

internal/plugins/jest/rules/require_hook/require_hook_extras_test.go
```diff
@@ -0,0 +1,98 @@
...
+			{
+				Code: `const suite = describe('suite', () => {
+  setup();
+});`,
+				Errors: []rule_tester.InvalidTestCaseError{
+					{MessageId: "useHook", Line: 2, Column: 3},
+				},
+			},
+			{
+				Code: `setup?.();`,
```
elecmonkey: This asserts `setup?.();` as **invalid**, but upstream does not report it (optional call → `ChainExpression`). Once the optional-chain handling above is fixed, this should move to the valid set.

elecmonkey (commented):
Thanks for the port — structure follows upstream closely and the option/`describe`-traversal logic looks correct.

By my Agent: I ran the real eslint-plugin-jest **29.15.5** rule (ESLint `Linter` on both `espree` and `@typescript-eslint/parser`) against this branch and found 4 false positives. They share one root cause: `shouldBeInHook`'s `switch` relies on ESTree wrapper node types, but tsgo's TS AST wraps these constructs differently, so nodes ESTree hides behind an *unhandled* wrapper reach the handled cases directly.

| Code | Upstream | This PR |
|------|----------|---------|
| `setup?.();` | no report | reports |
| `foo?.bar();` | no report | reports |
| `export let x = foo();` | no report | reports |
| `import('\''x'\'');` | no report | reports |

Upstream returns `[]` for all four (both parsers); details inline. Everything else (registration, `allowedFunctionCalls`, const/null/undefined, multi-declarator, `describe`/`describe.only` traversal) matches upstream.

internal/plugins/jest/rules/require_hook/require_hook.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+func shouldBeInHook(node *ast.Node, ctx rule.RuleContext, allowedFunctionCalls []string) bool {
+	if node == nil {
+		return false
+	}
+
+	switch node.Kind {
+	case ast.KindExpressionStatement:
+		return shouldBeInHook(node.AsExpressionStatement().Expression, ctx, allowedFunctionCalls)
```
swwind: **Parenthesized calls** — `(setup());`. TSESTree makes parentheses transparent, so upstream sees a plain `CallExpression` and reports. In tsgo the expression is a `ParenthesizedExpression`, which falls through to `default` → no report. Suggest `ast.SkipParentheses` on the expression before recursing.

internal/plugins/jest/rules/require_hook/require_hook.go
swwind: **`await using` declarations** — `await using x = f();`. In tsgo `NodeFlagsAwaitUsing = NodeFlagsConst | NodeFlagsUsing`, so `Flags&ast.NodeFlagsConst != 0` treats it as `const` and skips it. Upstream only skips `kind === 'const'` and reports this (plain `using` matches on both sides). Suggest an exact match instead: `declList.Flags&ast.NodeFlagsBlockScoped == ast.NodeFlagsConst`.

internal/plugins/jest/rules/require_hook/require_hook_extras_test.go
```diff
@@ -0,0 +1,98 @@
...
+  describe('suite', () => {
+    setup();
+  });
+}`,
+				Errors: []rule_tester.InvalidTestCaseError{
+					{MessageId: "useHook", Line: 3, Column: 5},
+				},
+			},
+			{
+				Code: `foo(describe('suite', () => {
```
swwind, line 48: This case expects the inner `setup();` (2:3) to be reported, but upstream reports only the outer `foo(...)` call (1:1): upstream's `parseJestFnCall` bails when the call's parent is a `CallExpression` (top-of-chain guard), so a `describe` passed as an argument isn't treated as a jest fn call. rslint's `ParseJestFnCall` doesn't have that guard — a pre-existing utils divergence rather than a bug in this PR, but this test codifies the non-upstream behavior as expected.

internal/plugins/jest/rules/require_hook/require_hook_extras_test.go
```diff
@@ -0,0 +1,98 @@
...
+)
+
+func TestRequireHookExtras(t *testing.T) {
+	rule_tester.RunRuleTester(
+		fixtures.GetRootDir(),
+		"tsconfig.json",
+		t,
+		&require_hook.RequireHookRule,
+		[]rule_tester.ValidTestCase{
+			{Code: `(setup());`},
```
swwind: This asserts `(setup());` as **valid**, but upstream reports it — once the parentheses handling above is fixed, this should move to the invalid set (1:1).

swwind (commented):
I ran the real eslint-plugin-jest **29.15.5** rule against a few more edge cases. Beyond the ones already reported above, three more divergences:

| Code | ESLint | This PR |
|------|--------|---------|
| `(setup());` | reports 1:1 | no report |
| `await using x = f();` | reports | no report |
| `foo(describe('suite', () => { setup(); }));` | reports outer call only (1:1) | also reports inner `setup()` (2:3) |

The first two are missed reports in this port, the third is an extra report rooted in a pre-existing utils divergence. Details inline.

eryue0220:
Update for previous CR.

swwind (approved):

fansenze (approved):

Issue #935, opened 2026-05-14
[Feature]: Lint Rstest support


elecmonkey:
## What problem does this feature solve?

Related discussion: https://github.com/web-infra-dev/rstest/issues/85#issuecomment-4435049030

Most of the rules for jest may also be applicable to Rstest. In a typical repo which using Rstest - web-infra-dev/rsbuild - around 2/3 of the `eslint-plugin-jest` rules can pass directly.

So I have a proposal: we could share an abstraction layer under both Jest and Rstest lint rules in our repo. Consider Rslint is now porting rules from `eslint-plugin-jest` actively, we can extract some utils outside first. once the jest rules become stable, we can start building Rstest rules on top of the existing things.

### Rstest-only Recommended (1/1)

| Rule Name | Will implement | Status | Level |
| --- | :---: | :---: | :---: |
| no-async-mock-factory | ✅ | Merged [#2100](https://github.com/web-infra-dev/rslint/pull/2100) | `error` |

### Jest Recommended (13/13)

| Rule Name | Will implement | Status | Level |
| --- | :---: | :---: | :---: |
| expect-expect | ✅ | Merged [#1614](https://github.com/web-infra-dev/rslint/pull/1614) | `warn` |
| no-alias-methods | ✅ | Merged [#1837](https://github.com/web-infra-dev/rslint/pull/1837) | Not recommended[^no-alias-methods] |
| no-conditional-expect | ✅ | Merged [#1548](https://github.com/web-infra-dev/rslint/pull/1548) | `error` |
| no-deprecated-functions | 🚫 | 🚫 | 🚫 |
| no-commented-out-tests | ✅ | Merged [#1339](https://github.com/web-infra-dev/rslint/pull/1339) | `warn` |
| no-disabled-tests | ✅ | Merged [#1475](https://github.com/web-infra-dev/rslint/pull/1475) | `warn` |
| no-done-callback | 🚫 | 🚫 | 🚫 |
| no-export | 🚫 | 🚫 | 🚫 |
| no-focused-tests | ✅ | Merged [#1363](https://github.com/web-infra-dev/rslint/pull/1363) | `error` |
| no-identical-title | ✅ | Merged [#1473](https://github.com/web-infra-dev/rslint/pull/1473) | `error` |
| no-interpolation-in-snapshots | ✅ | Merged [#1604](https://github.com/web-infra-dev/rslint/pull/1604) | `error` |
[...]

PR #1339, opened 2026-07-22
feat(rstest): add no-commented-out-tests rule

elecmonkey:
## Summary

Add `rstest/no-commented-out-tests` to report commented-out Rstest tests and suites.

The rule reconstructs consecutive comments, parses candidate snippets with typescript-go, and analyzes Rstest API chains to distinguish factory calls from actual test registrations. It supports modifiers, conditional APIs, `each` / `for`, tagged templates, and `test.extend`.

## Related Links

## Checklist

### DIFF
--- a/internal/plugins/rstest/rules/no_commented_out_tests/commented_call_analyzer.go
+++ b/internal/plugins/rstest/rules/no_commented_out_tests/commented_call_analyzer.go
@@ -214,20 +214,64 @@ func rootStartsLogicalLine(text string, offset int) bool {
 	return true
 }
 
-func insideMarkdownFence(text string, offset int) bool {
+func opensMarkdownFence(line string) bool {
+	trimmed := strings.TrimSpace(line)
+	return strings.HasPrefix(trimmed, "```") || strings.HasPrefix(trimmed, "~~~")
+}
+
+// markdownFenceTracker reports whether an offset sits inside a fenced code
+// block. Candidates are visited in ascending order, so the tracker advances a
+// cursor across the text once instead of rescanning from the start for each
+// candidate.
+type markdownFenceTracker struct {
+	cursor int
+	inside bool
+}
+
+func (tracker *markdownFenceTracker) insideAt(text string, offset int) bool {
 	if offset < 0 || offset > len(text) {
 		return false
 	}
-	inFence := false
-	for _, line := range strings.Split(text[:offset], "\n") {
-		trimmed := strings.TrimSpace(line)
-		if strings.HasPrefix(trimmed, "```") || strings.HasPrefix(trimmed, "~~~") {
-			inFence = !inFence
-		}
+	if offset < tracker.cursor {
+		tracker.cursor = 0
+		tracker.inside = false
 	}
-	return inFence
+
+	for tracker.cursor < offset {
+		lineEnd := len(text)
+		if newline := strings.IndexByte(text[tracker.cursor:], '\n'); newline >= 0 {
+			lineEnd = tracker.cursor + newline
+		}
+		if lineEnd >= offset {
+			break
+		}
+		if opensMarkdownFence(text[tracker.cursor:lineEnd]) {
+			tracker.inside = !tracker.inside
+		}
+		tracker.cursor = lineEnd + 1
+	}
+
+	// The partial line holding the offset can open a fence too, but the toggle
+	// is not recorded because a later candidate may sit on the same line.
+	if opensMarkdownFence(text[tracker.cursor:offset]) {
+		return !tracker.inside
+	}
+	return tracker.inside
 }
 
+// registrationExpressionStatement returns the ExpressionStatement a registration
+// opens, or nil when the call does not begin a statement of its own.
+//
+// Adjacent line comments are reconstructed into one block before parsing, so a
+// trailing note binds to the call it follows: `test("foo", () => {})` plus a
+// `- flaky` line parses as a BinaryExpression, and a `(fixture)` line as an
+// outer CallExpression. Demanding that the call be the direct child of an
+// ExpressionStatement therefore drops registrations that are complete on their
+// own line, which is the common shape for a disabled test with a note.
+//
+// Walking through parents that start where the call starts accepts those while
+// still rejecting a call nested in a construct opened earlier, such as an array
+// element or an argument, whose parent begins before the call does.
 func registrationExpressionStatement(node *ast.Node) *ast.Node {
 	for node != nil && node.Parent != nil {
 		parent := node.Parent
@@ -241,7 +285,13 @@ func registrationExpressionStatement(node *ast.Node) *ast.Node {
 		case ast.KindExpressionStatement:
 			return parent
 		default:
-			return nil
+			// A node shares its position with its first child, so equal
+			// positions mean the call is the leftmost token of the parent and
+			// whatever follows only trails the completed registration.
+			if parent.Pos() != node.Pos() {
+				return nil
+			}
+			node = parent
 		}
 	}
 	return nil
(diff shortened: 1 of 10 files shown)


elecmonkey:
@codex review

elecmonkey:
@codex review

internal/plugins/rstest/rules/no_commented_out_tests/no_commented_out_tests.go
```diff
@@ -0,0 +1,81 @@
...
+	"regexp"
+
+	"github.com/microsoft/typescript-go/shim/ast"
+	"github.com/microsoft/typescript-go/shim/core"
+	"github.com/web-infra-dev/rslint/internal/rule"
+)
+
+const (
+	testRootPattern            = `(test|it|describe)`
+	testMemberPattern          = `(\s*(\.\s*\w+|\[\s*['"]\w+['"]\s*\]))`
```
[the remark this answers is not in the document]
elecmonkey: > intentionally not support `extend`

internal/plugins/rstest/rules/no_commented_out_tests/no_commented_out_tests.go
```diff
@@ -0,0 +1,235 @@
...
+			segments: []commentBlockSegment{{
+				start:   0,
+				comment: comment,
+			}},
+		}
+		i++
+
+		if comment.Kind == ast.KindSingleLineCommentTrivia {
+			previous := comment
+			for i < len(comments) && lineCommentsAreAdjacent(sourceText, previous, comments[i]) {
```
swwind, line 175: Adjacent `//` lines are merged into one block and parsed as a unit, so a trailing note can absorb the call into a larger expression. `test('foo', () => {})\n- flaky` parses as a single BinaryExpression and `test('foo', () => {})\n(fixture)` as a single CallExpression — both parse without diagnostics, so they reach `registrationExpressionStatement` (`commented_call_analyzer.go:231`), which walks up to a parent that isn't an `ExpressionStatement` and returns nil. Nothing is reported.

A commented-out test followed by an explanatory line is the common shape in real code, and those notes often start with a continuation token (`(`, `[`, `` ` ``, `-`, `+`, `*`, `/`). Accepting the call when its own start begins a logical line — instead of requiring `ExpressionStatement` to be its direct parent — would cover this.

internal/plugins/rstest/rules/no_commented_out_tests/commented_call_analyzer.go
```diff
@@ -0,0 +1,342 @@
...
+		if len(match) < 4 {
+			continue
+		}
+		candidateStart := match[0]
+		rootOffset := match[2]
+		if insideMarkdownFence(text, rootOffset) {
+			continue
+		}
+		relativeRoot := rootOffset - candidateStart
+		foundOffset, found := findRegistrationInCandidate(text[candidateStart:], relativeRoot, scriptKind)
```
swwind, line 424: `findRegistrationInCandidate` receives `text[candidateStart:]` and fully re-parses it for every regex candidate. A file whose tests are all commented out collapses into a single block, so N candidates means N parses of that block's remaining text — quadratic in exactly the input this rule is meant to catch. The rule is also on by default through the rstest recommended preset (`packages/rslint/src/config/presets/rstest.ts:6`), so this is a baseline cost.

Trimming the candidate text to the next candidate/statement boundary, or parsing the block once and reusing the AST, would avoid it.

swwind (commented):
Two things worth fixing before merge. The remaining edge cases I found (type arguments, `*` stripping in block comments, markdown fences, partially commented-out tests) look like acceptable precision-over-recall tradeoffs for this design.

| Code | Expected | This PR |
| --- | --- | --- |
| `// test('foo', () => {})` | report | report |
| `// test('foo', () => {})`<br>`// - flaky` | report | no report |
| `// test('foo', () => {})`<br>`// (fixture)` | report | no report |

internal/plugins/rstest/rules/no_commented_out_tests/no_commented_out_tests.go
elecmonkey: Fixed in 5b1a7cd5 — thanks, this was a real false negative.

Confirmed the mechanism by dumping the AST for the merged block `test("foo", () => {})\n- flaky`:

```
KindExpressionStatement [0,29)
  KindBinaryExpression [0,29)
    KindCallExpression [0,21)      <- the registration
    KindMinusToken [21,23)
    KindIdentifier [23,29)
```

The parse is completely clean (0 diagnostics), so all three positional guards — the `rootStartsLogicalLine` check, `registrationHasCleanLineEnd`, and `hasDiagnosticInRange` — passed. Only the parent-kind check rejected it.

Worth noting *why* that check was contributing nothing: candidates are sliced from the start of their own line (`rstestRootCandidate` is anchored with `^`), so the call is essentially always at the beginning of the parsed text and therefore always the first statement. The only thing that can stop it from being an `ExpressionStatement`'s direct child is something **following** it binding it into a larger expression — which is exactly this bug. So the check was pure false-negative surface.

Took your suggestion. `registrationExpressionStatement` now accepts a call that **opens** its statement instead of one that **is** the whole statement:

```go
default:
    // A node shares its position with its first child, so equal
    // positions mean the call is the leftmost token of the parent and
    // whatever follows only trails the completed registration.
    if parent.Pos() != node.Pos() {
        return nil
    }
    node = parent
```

`test(...) - flaky` (BinaryExpression) and `test(...)(fixture)` (outer CallExpression) both start where the call starts, so they walk up to the `ExpressionStatement`. A call nested in a construct opened earlier — an array element, an argument — has a parent that begins before it, and is still rejected. `hasDiagnosticInRange` was narrowed from the statement range to the call's own range accordingly, so unrelated prose on a neighbouring line neither suppresses nor triggers a report.

Your table now behaves as expected:

| Code | Expected | Before | After |
| --- | --- | --- | --- |
| `// test('foo', () => {})` | report | report | report |
| `// test('foo', () => {})`<br>`// - flaky` | report | no report | **report** |
| `// test('foo', () => {})`<br>`// (fixture)` | report | no report | **report** |

Pinned as test cases, along with two more continuation tokens and the multiline shape:

```go
"// test(\"foo\", () => {})\n// - flaky"
"// test(\"foo\", () => {})\n// (fixture)"
"// test(\"foo\", () => {})\n// + see #123"
"// test(\"foo\", () => {})\n// `helper` is missing"
"// test(\"foo\", () => {\n//   expect(value).toBe(1)\n// })\n// - flaky"
```

All pre-existing cases are untouched (the test diff is +37/−0) and the suite is green.

internal/plugins/rstest/rules/no_commented_out_tests/commented_call_analyzer.go
elecmonkey: Fixed in 5b1a7cd5. You were right, and benchmarking it turned up a **second** quadratic hotspot in the same loop.

**Hotspot A — the one you identified.** Every candidate re-parsed `text[candidateStart:]` all the way to the end of the block. Replaced with a doubling parse window: the candidate is parsed at 1, 2, 4, 8… lines and the search stops as soon as a complete registration is found. A single-line registration costs one parse, a K-line one converges in log(K), so a block costs its own length rather than length × candidate count.

**Hotspot B — not in your comment, found while benchmarking.** `insideMarkdownFence` ran `strings.Split(text[:offset], "\n")` for *every* candidate — quadratic in both time and allocations. Since candidates are already visited in ascending offset order, this is now a `markdownFenceTracker` that advances a cursor across the text once. This alone was worth another ~2.4×, and was why the fix still looked slightly superlinear until it was addressed.

**Benchmark.** One comment block of N commented-out single-line tests, `-benchtime 20x`, Apple M5 Pro, `go test -run '^$' -bench BenchmarkFind`:

| block lines | 166548a6 (reviewed) | 5b1a7cd5 (fixed) | speedup |
| --- | --- | --- | --- |
| 40 | 1.19 ms | 0.181 ms | 6.6× |
| 160 | 15.43 ms | 0.649 ms | 23.8× |
| 640 | 214.69 ms | 1.877 ms | **114×** |

Scaling confirms the complexity change — 4× the input:

- before: 13.0× then 13.9× the time (quadratic)
- after: 3.6× then 2.9× the time (~linear)

Benchmark source, if you want to reproduce (kept local rather than committed — happy to add it if you'd prefer it in the repo):

```go
func buildBlock(lines int) string {
	var builder strings.Builder
	for i := 0; i < lines; i++ {
		fmt.Fprintf(&builder, "test(\"case %d\", () => { expect(value).toBe(%d) })\n", i, i)
	}
	return builder.String()
}

func BenchmarkFindCommentedRstestRegistrations(b *testing.B) {
	for _, lines := range []int{40, 160, 640} {
		text := buildBlock(lines)
		b.Run(fmt.Sprintf("lines=%d", lines), func(b *testing.B) {
			for i := 0; i < b.N; i++ {
				findCommentedRstestRegistrations(text, core.ScriptKindTS)
			}
		})
	}
}
```

**One correctness note on the windowing**, since it is not obvious. A window that stops short can cut a registration in half, and the parser recovers such a slice into a *complete-looking* call whose only diagnostic sits just past the call's end — so `hasDiagnosticInRange` does not catch it and a truncated test gets accepted. Concretely, window 1 of `test("foo", () => {` recovers to `KindCallExpression [0,19)` with its sole diagnostic at `[20,20)`.

An incomplete window is therefore only trusted when it parsed cleanly:

```go
if !complete && len(sourceFile.Diagnostics()) > 0 {
    return 0, false
}
```

Pinned with a valid case that regresses without that guard:

```go
{Code: "// test(\"foo\", () => {\n//   prose that is not code !!!"},
```

I verified each of the three changes is individually load-bearing — reverting any one of them turns a test case red.

fansenze (approved):

PR #1358, opened 2026-07-24
chore: move to pnpm catalog

swwind:
## Summary

Move all dependencies in package.json to use pnpm's catalog feature.

## Checklist

- [ ] Tests updated (or not required).


elecmonkey:
1.  I think `catalog` only provides real value for dependencies that are shared across multiple packages.
2. Having TS 5 and TS 7 coexist is a pattern that needs more careful consideration.

elecmonkey:
ts 5 to 6 has breaking (some checks will fail in this repo). this PR use catalog and update no-breaking deps only. i'll open another PR to update TypeScript. ideally, we should use TS7 only, but may some deps need ts6 as peer, so we'd better install ts7 as typescript and ts6 as a alias.

pnpm-workspace.yaml
```diff
@@ -13,6 +13,102 @@ allowBuilds:
   esbuild: true
   keytar: false
 
+catalogs:
+  build:
```
chenjiahan: I'm wondering what the benefits are of distinguishing between multiple catalog types (build/lint/types)?

swwind: This may potentially solve the different typescript problem, eslint still uses typescript@6, but the remaining can upgrade to typescript@7, so we may write `catalog:lint` for eslint's typescript and `catalog:types` for the remaining packages.

chenjiahan: typescript is a special case and we can add a specific catalog for typescript 6/7, for example: https://github.com/web-infra-dev/rspack/blob/main/pnpm-workspace.yaml#L18-L23

elecmonkey (dismissed):

elecmonkey:
Reverted three dependency bumps from this catalog migration to unblock CI; everything else is kept.

`eslint-plugin-sonarjs`: `^4.2.0` → `4.0.3`: in 4.2.0 the `no-implicit-dependencies` rule throws on load in CI (`"/" is not nested under topDir …`), which fails the entire `sonarjs.test.ts` conformance suite (171 cases).

leaving full TypeScript toolchain for a follow-up PR.

cc @swwind

pnpm-workspace.yaml
elecmonkey: > typescript is a special case and we can add a specific catalog for typescript 6/7, for example: https://github.com/web-infra-dev/rspack/blob/main/pnpm-workspace.yaml#L18-L23

we've flattened them

elecmonkey:
Website build failure was caused by `monaco-editor` being upgraded from `0.55.1` to `0.56.0`. In `0.56.0`, the package `exports` mapping changed, so our existing worker imports like `monaco-editor/esm/vs/language/typescript/ts.worker` resolve incorrectly to a non-existent `esm/vs/esm/vs/...` path. In this PR, I pinned `monaco-editor` back to `0.55.1` to unblock the docs build.

elecmonkey (approved):

chenjiahan (approved):

PR #1361, opened 2026-07-24
perf: port no-class-assign, no-func-assign, no-ex-assign to RefStore

swwind:
## Summary

- Port `no-class-assign`, `no-func-assign`, and `no-ex-assign` to `ctx.Refs`, replacing their manual whole-scope AST walks, name-string matching, and TypeChecker-based shadow detection with `ctx.Refs.References(sym)`, following the pattern the no-var rule already uses.
- `no-ex-assign` no longer requires a TypeChecker (drops `RequiresTypeInfo: true`); reference resolution is now binder-based like the other two.
- Fixes false positives in `no-class-assign` and `no-func-assign` where a shadowing local was still reported when `ctx.TypeChecker` was nil (their old shadow check silently no-opped in that case).
- Re-enables a previously-disabled `no-class-assign` test case for shorthand destructuring assignment (`({A} = 0)`), which now passes correctly.

## Validation

- `go test ./internal/rules/no_class_assign/...`
- `go test ./internal/rules/no_func_assign/...`
- `go test ./internal/rules/no_ex_assign/...`
- `go vet` on the three changed packages
- `internal/config` TypeChecker-nil-safety guard tests (`TestAllRules_SilentOnNilTypeCheckerImpliesRequiresTypeInfo`, `TestAllRules_NilTypeCheckerEarlyReturnImpliesRequiresTypeInfo`, `TestAllRules_DeclaredSchemasCompile`)

## Related Links

N/A

## Checklist

### DIFF
--- a/internal/rules/no_class_assign/no_class_assign.go
+++ b/internal/rules/no_class_assign/no_class_assign.go
@@ -14,317 +14,19 @@ func buildClassReassignmentMessage(className string) rule.RuleMessage {
 	}
 }
 
-// getIdentifierName extracts the name from an identifier node
-func getIdentifierName(node *ast.Node) string {
-	if node == nil || node.Kind != ast.KindIdentifier {
-		return ""
-	}
-	return node.Text()
-}
-
-// isWriteReference checks if a node is a write reference (assignment target)
-func isWriteReference(node *ast.Node) bool {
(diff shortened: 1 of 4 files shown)


swwind:
Closed as we should separate those rules into different pr

PR #1362, opened 2026-07-24
perf: port no-shadow-restricted-names and no-const-assign to ctx.Refs

swwind:
## Summary

- Port `no-shadow-restricted-names` and `no-const-assign` to `ctx.Refs`, replacing manual whole-file/whole-scope AST walks and `TypeChecker.GetSymbolAtLocation` calls with `ctx.Refs.References(sym)`, following the pattern the no-var rule already uses.
- `no-const-assign`'s violations are now collected per declaration and flushed in source order at the end of the file (mirroring `no-unused-vars`'s pattern), since scanning the whole file in one linear pass no longer happens naturally once lookups are grouped by declaration.
- Re-enables a previously-disabled `no-const-assign` test case for shorthand destructuring writes (`({x} = {x: 1})`), which now passes correctly.

The `prefer-const` port and the `RefStore.Resolve` API addition moved to #1415.

## Validation

- `go build ./internal/rule/... ./internal/rules/...`
- `go test ./internal/rules/no_shadow_restricted_names/...`
- `go test ./internal/rules/no_const_assign/...`

## Related Links

N/A

## Checklist

### DIFF
--- a/internal/rules/no_const_assign/no_const_assign.go
+++ b/internal/rules/no_const_assign/no_const_assign.go
@@ -1,11 +1,14 @@
 package no_const_assign
 
 import (
+	"sort"
+
 	"github.com/microsoft/typescript-go/shim/ast"
+
 	"github.com/web-infra-dev/rslint/internal/rule"
+	"github.com/web-infra-dev/rslint/internal/utils"
 )
 
-// Message builder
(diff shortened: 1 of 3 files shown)


swwind:
Closed as we should separate those rules into different pr

PR #1363, opened 2026-07-24
feat(rstest): add no-focused-tests rule

elecmonkey:
## Summary

Add the `rstest/no-focused-tests` rule to report focused Rstest tests and suites using `.only`.

The rule uses the Rstest call parser to recognize final test and `describe` registrations, including modifier chains, conditional and parameterized APIs, `test.extend`, imports, aliases, and namespace APIs.

It provides a suggestion that removes all explicit `.only` modifiers from the registration. When removal would discard a comment, the rule reports the issue without offering an unsafe suggestion.

The rule is included in the recommended preset, with documentation and tests.

`jest/no-focused-tests`, `jest/prefer-to-be` and `jest/prefer-to-have-been-called` are touched by a refactor with no behavior change: the member-accessor removal helper in `internal/plugins/jest/utils/accessor.go` was a second copy of the one in `internal/utils/test_framework/accessor.go`, so it is deleted in favour of the shared helper. The shared version takes the member entry directly, which removes the index lookup its callers only performed to satisfy the old signature.

## Related Links

## Checklist

### DIFF
--- a/internal/plugins/jest/rules/no_focused_tests/no_focused_tests.go
+++ b/internal/plugins/jest/rules/no_focused_tests/no_focused_tests.go
@@ -82,11 +82,7 @@ var NoFocusedTestsRule = rule.Rule{
 					})
 					if idx >= 0 {
 						entry := jestFnCall.MemberEntries[idx]
-						removalFixes, ok := jestUtils.RemoveMemberAccessorFixes(
-							ctx,
-							jestFnCall.MemberEntries,
-							idx,
-						)
+						removalFixes, ok := jestUtils.RemoveMemberAccessorFixes(ctx, &entry)
 						if !ok {
 							ctx.ReportNode(entry.Node, buildErrorFocusedTestMessage())
 							return
--- a/internal/plugins/rstest/rules/no_focused_tests/no_focused_tests.go
+++ b/internal/plugins/rstest/rules/no_focused_tests/no_focused_tests.go
@@ -1,0 +1,133 @@
+package no_focused_tests
+
+import (
+	"sort"
+
+	"github.com/microsoft/typescript-go/shim/ast"
+	"github.com/microsoft/typescript-go/shim/core"
+	rstestUtils "github.com/web-infra-dev/rslint/internal/plugins/rstest/utils"
+	"github.com/web-infra-dev/rslint/internal/rule"
+	rslintUtils "github.com/web-infra-dev/rslint/internal/utils"
+)
+
+func buildErrorFocusedTestMessage() rule.RuleMessage {
+	return rule.RuleMessage{
+		Id:          "focusedTest",
+		Description: "Unexpected focused test",
+	}
+}
+
+func buildErrorSuggestRemoveFocusMessage() rule.RuleMessage {
+	return rule.RuleMessage{
+		Id:          "suggestRemoveFocus",
+		Description: "Suggest removing focus from test",
+	}
+}
+
+var NoFocusedTestsRule = rule.Rule{
+	Name:   "rstest/no-focused-tests",
+	Schema: rule.EmptyArraySchema,
+	Run: func(ctx rule.RuleContext, options []any) rule.RuleListeners {
+		return rule.RuleListeners{
+			ast.KindCallExpression: func(node *ast.Node) {
+				parsed := rstestUtils.ParseRstestFnCall(node, ctx)
+				if parsed == nil ||
+					(parsed.Kind != rstestUtils.RstestFnTypeDescribe &&
+						parsed.Kind != rstestUtils.RstestFnTypeTest) {
+					return
+				}
+
+				onlyEntries := make([]rstestUtils.ParsedRstestFnMemberEntry, 0, 1)
+				for _, entry := range parsed.MemberEntries {
+					if entry.Name == "only" {
+						onlyEntries = append(onlyEntries, entry)
+					}
+				}
+				if len(onlyEntries) == 0 {
+					return
+				}
+
+				reportNode := onlyEntries[0].Node
+				fixes, ok := removalFixes(onlyEntries, ctx)
+				if !ok {
+					ctx.ReportNode(reportNode, buildErrorFocusedTestMessage())
+					return
+				}
+
+				ctx.ReportNodeWithSuggestions(
+					reportNode,
+					buildErrorFocusedTestMessage(),
+					rule.RuleSuggestion{
+						Message:  buildErrorSuggestRemoveFocusMessage(),
+						FixesArr: fixes,
+					},
+				)
+			},
+		}
+	},
+}
+
+func removalFixes(
+	entries []rstestUtils.ParsedRstestFnMemberEntry,
+	ctx rule.RuleContext,
+) ([]rule.RuleFix, bool) {
+	ranges := make([]core.TextRange, 0, len(entries))
+	for _, entry := range entries {
+		textRange, ok := memberRemovalRange(entry.Node)
+		if !ok || rslintUtils.HasCommentInSpan(ctx.Comments.All(), textRange.Pos(), textRange.End()) {
+			return nil, false
+		}
+		ranges = append(ranges, textRange)
+	}
+
+	sort.Slice(ranges, func(i, j int) bool {
+		return ranges[i].Pos() < ranges[j].Pos()
+	})
+
+	fixes := make([]rule.RuleFix, 0, len(ranges))
+	for index, textRange := range ranges {
+		if index > 0 && ranges[index-1].End() > textRange.Pos() {
+			return nil, false
+		}
+		fixes = append(fixes, rule.RuleFixRemoveRange(textRange))
+	}
+	return fixes, true
+}
+
+func memberRemovalRange(nameNode *ast.Node) (core.TextRange, bool) {
+	if nameNode == nil {
+		return core.TextRange{}, false
+	}
+
+	memberNode := nameNode
+	for memberNode.Parent != nil && memberNode.Parent.Kind == ast.KindParenthesizedExpression {
+		memberNode = memberNode.Parent
+	}
+	if memberNode.Parent == nil {
+		return core.TextRange{}, false
+	}
+
+	accessor := memberNode.Parent
+	var receiver *ast.Node
+	switch accessor.Kind {
+	case ast.KindPropertyAccessExpression:
+		property := accessor.AsPropertyAccessExpression()
+		if property.Name() != memberNode {
+			return core.TextRange{}, false
+		}
+		receiver = property.Expression
+	case ast.KindElementAccessExpression:
+		element := accessor.AsElementAccessExpression()
+		if ast.SkipParentheses(element.ArgumentExpression) != nameNode {
+			return core.TextRange{}, false
+		}
+		receiver = element.Expression
+	default:
+		return core.TextRange{}, false
+	}
+
+	if receiver == nil || receiver.End() >= accessor.End() {
+		return core.TextRange{}, false
+	}
+	return core.NewTextRange(receiver.End(), accessor.End()), true
+}
--- a/internal/plugins/rstest/rules/no_focused_tests/no_focused_tests.go
+++ b/internal/plugins/rstest/rules/no_focused_tests/no_focused_tests.go
@@ -1,0 +1,111 @@
+package no_focused_tests
+
+import (
+	"sort"
+
+	"github.com/microsoft/typescript-go/shim/ast"
+	"github.com/microsoft/typescript-go/shim/core"
+	rstestUtils "github.com/web-infra-dev/rslint/internal/plugins/rstest/utils"
+	"github.com/web-infra-dev/rslint/internal/rule"
+	testFramework "github.com/web-infra-dev/rslint/internal/utils/test_framework"
+)
+
+func buildErrorFocusedTestMessage() rule.RuleMessage {
+	return rule.RuleMessage{
+		Id:          "focusedTest",
+		Description: "Unexpected focused test",
+	}
+}
+
+func buildErrorSuggestRemoveFocusMessage() rule.RuleMessage {
+	return rule.RuleMessage{
+		Id:          "suggestRemoveFocus",
+		Description: "Suggest removing focus from test",
+	}
+}
+
+var NoFocusedTestsRule = rule.Rule{
+	Name:   "rstest/no-focused-tests",
+	Schema: rule.EmptyArraySchema,
+	Run: func(ctx rule.RuleContext, options []any) rule.RuleListeners {
+		analysis := rstestUtils.GetRstestCallAnalysis(ctx)
+		reported := map[*ast.Node]bool{}
+		return rule.RuleListeners{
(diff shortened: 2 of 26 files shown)


elecmonkey:
@codex review

internal/plugins/rstest/utils/rstest.go
```diff
@@ -0,0 +1,72 @@
...
+// RstestFnCallParseConfig returns the shared function-call parser config for
+// Rstest.
+func RstestFnCallParseConfig() jestUtils.FnCallParseConfig {
+	return jestUtils.FnCallParseConfig{
+		ImportModule: RstestImportModule,
+		IsValidChain: isValidRstestCall,
+		ParameterizedModifiers: map[string]bool{
+			"each": true,
+			"for":  true,
+		},
```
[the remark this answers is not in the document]
elecmonkey: 👍

internal/plugins/rstest/utils/rstest.go
```diff
@@ -0,0 +1,150 @@
...
+	"for":    true,
+}
+
+// isValidRstestMember reports whether a chained member is legal for the given
+// Rstest root. `fails` is only available on `test`/`it`, not `describe`.
+func isValidRstestMember(root string, member string) bool {
+	if rstestSharedModifiers[member] || rstestFactoryMembers[member] {
+		return true
+	}
+	if member == "fails" && root != "describe" {
```
swwind: `extend` is missing from the allowed members. `TestAPIs.extend(fixtures)` returns `TestAPIs`, which still carries `.only`, so this is a legal (and common) fixture-style focused test:

```ts
import { test } from '@rstest/core';

test.extend({ db: async ({}, use) => use(makeDb()) }).only('focused', () => {});
it.extend({ db }).concurrent.only('focused', () => {});
```

`isValidRstestMember("test", "extend")` returns false, so `isValidRstestCall` rejects the chain and `ParseRstestFnCall` returns nil — nothing is reported. The control case `test.concurrent.only('x', fn)` reports fine, so the difference is exactly this one member.

Note it can't just go into `rstestFactoryMembers`: `extend` is call-once-returns-new-API, not the factory+call shape, so putting it there would make `isFactoryCall` skip the wrong layer. It belongs next to the run-mode members, restricted to `test`/`it` (`Describe` has no `extend`), the same way `fails` is handled.

internal/plugins/jest/rules/no_focused_tests/no_focused_tests.go
```diff
@@ -27,80 +29,85 @@ func buildErrorSuggestRemoveFocusMessage() rule.RuleMessage {
...
-							endRange = entry.Node.End() + 1
+							)
 						}
+					} else {
+						idx := slices.IndexFunc(jestFnCall.MemberEntries, func(entry jestUtils.ParsedJestFnMemberEntry) bool {
+							return entry.Name == "only"
+						})
+						if idx >= 0 {
+							entry := jestFnCall.MemberEntries[idx]
+							startRange := entry.Node.Loc.Pos() - 1
```
swwind: Pre-existing, but this PR moves the block and adds bracket-notation coverage on the Rstest side, so it seems worth fixing here: `endRange = entry.Node.End() + 1` hard-codes "the `]` sits immediately after the literal". With any whitespace the suggestion produces unparseable output:

```ts
describe[ "only" ]('math', () => {});
// suggestion → describe]('math', () => {});

describe[
  'only'
]('math', () => {});
// suggestion → describe\n]('math', () => {});
```

(`startRange = Pos() - 1` happens to be correct because `Pos()` includes leading trivia, so it lands on `[`.) The end should come from the enclosing `ElementAccessExpression`'s end rather than `literal.End() + 1`. The existing tests only use `describe["only"]()` with no spaces, which is exactly the case where the off-by-one cancels out.

internal/plugins/rstest/utils/rstest.go
```diff
@@ -0,0 +1,150 @@
...
+		localName,
+		localNode,
+		ctx,
+		RstestImportModule,
+	)
+	if name == "" || !isValidRstestCall(name, members) {
+		return nil
+	}
+
+	kind := jestUtils.GetJestKind(name)
```
swwind: Non-blocking / latent: kind resolution goes through `jestUtils.GetJestKind`, which knows the Jest-only aliases, so `fdescribe`/`fit`/`xdescribe`/`xit`/`xtest` with no members resolve to a `JestFnTypeDescribe`/`JestFnTypeTest` Rstest call. There is no observable bug in this rule (no `.only` member ⇒ no report, and `fit.only()` is rejected by `isValidRstestCall`), so the `fit()` / `fdescribe()` valid cases pass for the right reason at the diagnostic level but the wrong reason at the parser level.

Since `ParseRstestFnCall` is meant to be reused by the next Rstest rules, it would be worth constraining `name` to `test`/`it`/`describe` (plus whichever hooks you want) here rather than delegating alias knowledge to the Jest helper — otherwise a future `rstest/require-top-level-describe` or `no-identical-title` will treat `fdescribe('a', ...)` as a real suite.

swwind (commented):
Reviewed the parameterization (`NewRule` + `ResolveFunctionReferenceForModule`) — the Jest side looks behavior-identical, and the `runIf`/`skipIf` factory-layer fix in e50ac6f does resolve the earlier double-report (I traced 1/2/3-layer chains, parenthesized callees and tagged templates; each reports exactly once). Three things below.

| Code | Expected | This PR |
| --- | --- | --- |
| `test.extend({ db }).only('x', () => {})` | report `focusedTest` | no diagnostic |
| `describe[ "only" ]('x', () => {})` | suggestion → `describe('x', () => {})` | suggestion → `describe]('x', () => {})` |
| `fdescribe('x', () => {})` | not a Rstest API | parsed as a Rstest suite (no diagnostic today) |

internal/plugins/rstest/rules/no_focused_tests/no_focused_tests.md
```diff
@@ -0,0 +1,50 @@
+# rstest/no-focused-tests
```
swwind: hello?

internal/plugins/rstest/rules/no_focused_tests/no_focused_tests.go
```diff
@@ -0,0 +1,111 @@
...
+				}
+
+				var reportNode *ast.Node
+				for _, entry := range focusEntries {
+					if entry.Node != nil && !reported[entry.Node] {
+						reportNode = entry.Node
+						break
+					}
+				}
+				if reportNode == nil {
```
swwind: `reported` is keyed by the `.only` accessor node, so when every focus entry of a registration has already been reported this returns with no diagnostic at all. With an alias, several call sites share one entry node, and only the first survives:

```ts
const focused = test.only;
focused('a', fn);   // reported — at line 1, col 22
focused('b', fn);   // no diagnostic
```

Two focused tests, one error. `focused.only('a', fn); focused('b', fn);` behaves the same way — the second registration is silent. The dedup itself is needed (without it the shared entry would be reported once per call site with the same fix), but the current shape makes a `no-focused-tests` rule undercount focused tests, and the one report it does emit points at the alias declaration rather than the call site the reader is looking at.

Worth a deliberate choice here: report the call node (or the call's own last entry) and keep `reported` only for suppressing duplicate *suggestions*, so each focused registration still gets its own diagnostic.

internal/utils/test_framework/accessor.go
```diff
@@ -58,6 +58,165 @@ func IsComputedIdentifierAccessor(node *ast.Node) bool {
...
+		}
+		if call.QuestionDotToken != nil {
+			callQuestionDotRange := utils.TrimNodeTextRange(sourceFile, call.QuestionDotToken)
+			if utils.HasCommentInSpan(comments, accessor.End(), callQuestionDotRange.End()) {
+				ranges = append(ranges, callQuestionDotRange)
+			} else {
+				ranges = append(ranges, core.NewTextRange(accessor.End(), callQuestionDotRange.End()))
+			}
+		} else {
+			ranges = append(ranges, utils.TrimNodeTextRange(sourceFile, questionDot))
```
swwind: When the call has no `?.` of its own, this drops the accessor's `questionDot` outright, which changes runtime semantics:

```ts
test?.only('a', fn);      // suggestion → test('a', fn)
test?.["only"]('a', fn);  // suggestion → test('a', fn)
```

The original evaluates to `undefined` when `test` is nullish; the suggestion throws `TypeError`. Emitting `test?.('a', fn)` instead is valid and preserves the guard — which is what the sibling case already does (`test?.only?.()` → `test?.()`).

This also contradicts `no_focused_tests.md:14` ("Suggestions preserve comments and optional-chain boundaries"). It's locked in by `accessor_test.go:211-212` and `no_focused_tests_test.go:236-237`, so if dropping the guard is intentional the doc line should be softened; otherwise those expectations need updating along with this branch.

swwind (commented):
Re-reviewed the fixer and the alias path on 855b7ab2. The accessor range math is solid — I probed ~30 shapes (optional chains, computed keys, parenthesized keys, comments in every gap, alias chains) and every suggestion output parses. Two things below.

| Code | Expected | This PR |
| --- | --- | --- |
| `const focused = test.only;` + `focused('a', fn); focused('b', fn);` | two `focusedTest` reports | one report, anchored at the line-1 alias |
| `test?.only('a', fn)` | suggestion → `test?.('a', fn)` | suggestion → `test('a', fn)` |

internal/plugins/rstest/rules/no_focused_tests/no_focused_tests.md
elecmonkey: My mistake

internal/plugins/rstest/rules/no_focused_tests/no_focused_tests.go
elecmonkey: Fixed in b1ae69f1. Each focused alias call is now reported at its call site, while duplicate suggestions are suppressed. Thanks!

internal/utils/test_framework/accessor.go
elecmonkey: Fixed in cc00653c. Removing `.only` now preserves the optional-chain guard, with regression tests added. Thanks!

internal/plugins/rstest/rules/no_focused_tests/no_focused_tests.go
```diff
@@ -0,0 +1,138 @@
...
+) *ast.Node {
+	for _, focusEntry := range focusEntries {
+		for _, memberEntry := range parsed.MemberEntries {
+			if focusEntry.Node != nil && focusEntry.Node == memberEntry.Node {
+				return focusEntry.Node
+			}
+		}
+	}
+
+	if len(parsed.MemberEntries) > 0 {
```
swwind, line 106: This fallback only fires when the focus came from an alias *and* the call site has members of its own — call-site `.only` is caught by the loop above. In that window it anchors the diagnostic on the last call-site member, which is never the focus marker:

```ts
const focused = test.only;
focused.each(rows)('x', fn);   // "Unexpected focused test" underlines `each`

const t = test.only;
t.skip('a', fn);               // underlines `skip`
```

Underlining `.skip` with "Unexpected focused test" points the reader at the modifier that is the opposite of focus. It is also inconsistent with the member-less alias case, which already reports on the call-site identifier (`focused`, tested at line 132).

Swapping the two fallbacks — `Head.Local.Node` first, last member entry after — anchors both at the call-site identifier and leaves the existing suite green (I ran it). A test for alias focus plus a call-site member would pin this down; the current cases only cover an alias with no members, or call-site `.only`.

internal/utils/test_framework/accessor.go
```diff
@@ -58,6 +58,163 @@ func IsComputedIdentifierAccessor(node *ast.Node) bool {
...
+		}
+	}
+	return ranges
+}
+
+// RemoveAccessorEntryRanges removes entry without requiring the caller to
+// retain the full parsed member slice. It derives the next accessor or call
+// directly from the AST, which is required for semantic entries originating in
+// an alias initializer rather than the final call site's Members.
+func RemoveAccessorEntryRanges(
```
swwind, line 133: Non-blocking: `RemoveAccessorEntryRanges` and the three helpers above it are a near-verbatim second copy of `internal/plugins/jest/utils/accessor.go` — `AccessorQuestionDotToken`, `removeAccessorSyntax` and `RemoveMemberAccessorFixes`, plus `GetAccessorReceiverAndParent` from parse_jest_fn.go:338. Roughly 110 lines duplicated one layer down from the copy they were lifted from.

The new version is the more general of the two: it derives the next accessor from the AST instead of `entries[index+1]`, checks that the entry really is the accessor's name/argument, and adds the `VariableDeclaration` branch. Since `ParsedJestFnMemberEntry = testFramework.MemberEntry` (jest.go:134), the jest side can call this helper directly and map the ranges to `RuleFixRemoveRange`, which would delete its copy outright.

Fine to land as-is and fold in a follow-up — but as long as both exist, a fix to one silently misses the other.

swwind (commented):
Both earlier findings are fixed and hold up under probing: alias call sites now each get their own diagnostic anchored at the call, and the optional-chain guard survives removal — I ran ~20 more shapes (`(test.only)()`, `test.only?.()`, `test?.only?.()`, `test?.[/* c */ 'only']()`, `test.extend({}).only.each(rows)()`, multi-level aliases, `let` reassignment, shadowing, multi-line chains) and every suggestion output parses and preserves semantics. `let focused = test.only; focused = test;` and shadowed aliases correctly report nothing.

One anchoring gap left, plus a non-blocking cleanup.

| Code | Expected | This PR |
| --- | --- | --- |
| `const focused = test.only;` + `focused.each(rows)('x', fn)` | report on `focused` (2:1) | report on `each` (2:9) |
| `const t = test.only;` + `t.skip('a', fn)` | report on `t` (2:1) | report on `skip` (2:3) |

internal/plugins/rstest/rules/no_focused_tests/no_focused_tests.go
elecmonkey: Fixed — the two fallbacks are swapped, so an alias-hidden focus now reports on the
call-site identifier and only falls back to the last member when there is no head
node at all. Added three cases that fail on the previous ordering: `focused.skip`,
`focused.each(rows)`, and a `describe.only` alias with `.sequential`; each reported
at column 9 on the member before the change and reports at column 1 now.

internal/utils/test_framework/accessor.go
elecmonkey: Folded in here rather than a follow-up. `jest/utils/accessor.go` drops
`removeAccessorSyntax` and the body of `RemoveMemberAccessorFixes`, and
`GetAccessorReceiverAndParent` becomes a delegate; all of it now routes through
`RemoveAccessorEntryRanges`. Since the shared version derives the next operation
from the AST, the entry is all it needs, so the signature loses the slice and the
index and the three callers lose the `slices.IndexFunc` they used to build it. The
jest suite passes unchanged.

swwind (approved):
LGTM

fansenze (approved):

PR #1428, opened 2026-07-27
perf: optimize no-unmodified-loop-condition rule

swwind:
## Summary

Use `ctx.Refs.Resolve()` / `ctx.Refs.References()` when possible, falling back to `ctx.TypeChecker` only for symbols `ctx.Refs` can't bind (declared outside this file). Drop `RequiresTypeInfo: true` accordingly.

## Benchmark

Per-rule time from `--timing all`, synthetic corpus (6 loops/file, one condition var each, plus a function-call-modifies-condition case and a never-modified case):

| corpus | main mean/median | rebased PR mean/median | Δ mean |
| --- | --- | --- | --- |
| 1000-stmt bodies, 1 var, 300 files | 2675.9 / 2648.2 ms | 244.3 / 191.3 ms | −90.9% |
| 1000-stmt bodies, 4 vars, 300 files | 2731.3 / 2697.7 ms | 357.3 / 338.3 ms | −86.9% |
| 3000-stmt bodies, 1 var, 100 files | 2717.2 / 2741.2 ms | 236.3 / 195.9 ms | −91.3% |

Byte-identical diagnostics on all corpora.

## Validation

- `go test ./internal/rules/no_unmodified_loop_condition/...` — 83 cases.
- `go vet ./internal/rules/no_unmodified_loop_condition/...`
- `internal/config` `RequiresTypeInfo` guard tests pass with it dropped.

## Related Links

Follow-up to #1335, #1415.

## Checklist

### DIFF
--- a/internal/rules/no_unmodified_loop_condition/no_unmodified_loop_condition.go
+++ b/internal/rules/no_unmodified_loop_condition/no_unmodified_loop_condition.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 
 	"github.com/microsoft/typescript-go/shim/ast"
-	"github.com/microsoft/typescript-go/shim/checker"
 	"github.com/web-infra-dev/rslint/internal/rule"
 	"github.com/web-infra-dev/rslint/internal/utils"
 )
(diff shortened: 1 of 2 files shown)


elecmonkey:
@codex review

elecmonkey (approved):

fansenze (approved):

PR #1437, opened 2026-07-28
perf: optimize no-loop-func rule

swwind:
## Summary

- Resolve through-reference identifiers with `ctx.Refs.Resolve` and look up a symbol's references with `ctx.Refs.References`, replacing the rule's direct `TypeChecker` usage (`utils.GetReferenceSymbol` per identifier and `checker.GetReferencesToSymbolInFile` — an uncached whole-file text scan — per (loop-function, through-reference) pair).
- `ctx.Refs` covers every case the rule needs: it falls back to the checker internally for symbols the binder scope walk can't see (ambient/lib globals, cross-file and merged declarations) and files merged-symbol references under one identity, so the rule keeps no `TypeChecker` fallback of its own. Declaration names that ESLint's scope manager counts as write references (`var`/`let`/`const` initializers, `for-in`/`for-of` bindings, catch parameters) are merged back in.
- Drop `RequiresTypeInfo: true` and the nil-`TypeChecker` early return — the rule now also runs on files without type info.

## Benchmark

Synthetic corpus: 1200 `.ts` files × 16 functions, each function placing closures inside `for-of`/`while` loops that capture outer locals, the loop counter, and an ambient global; weighted toward safe closures (19,200 diagnostics over 230k closures).

Per-rule time from `--timing all` (summed across workers), 5 runs after a warmup:

| | mean | median |
| --- | --- | --- |
| `main` | 185,505.3 ms | 185,825.9 ms |
| this PR | 825.6 ms | 879.0 ms |

Rule time: **−99.5%**.

## Validation

- `go test ./internal/rules/no_loop_func/...`, including `TypeChecker`-fallback coverage for ambient globals (a read-only global is safe; a reassigned global captured by an escaping closure is flagged), verified against real ESLint 10.7.0 behavior.
- Both binaries produce identical diagnostics (19,200 lines, compared line by line) on the benchmark corpus.
- `pnpx cspell lint` and `golangci-lint run --new-from-rev=origin/main` on the changed files.

## Related Links

Follow-up to #1335 (`ctx.Refs`), #1415 (`ctx.Refs.Resolve`), and #1450 (checker fallback inside `ctx.Refs`).

## Checklist

### DIFF
--- a/internal/rules/no_loop_func/no_loop_func.go
+++ b/internal/rules/no_loop_func/no_loop_func.go
@@ -170,6 +170,29 @@ func (r ReferenceEntry) Node() *ast.Node { return r.node }
 // Symbol returns the resolved symbol for the reference, or nil if unresolved.
 func (r ReferenceEntry) Symbol() *ast.Symbol { return r.symbol }
 
+// resolveReference resolves an identifier to its referenced symbol, trying
+// ctx.Refs first: it answers from a per-node binder scope-walk with no
+// TypeChecker round-trip, and correctly resolves same-file locals — the
+// overwhelming majority of through-references a loop-closure can actually
+// capture (loop counters, outer function locals, imported bindings, which are
+// all declared, and therefore locally bound, in this file). It returns nil
+// for identifiers naming a symbol declared outside this file — ambient/lib
+// globals (`console`, `Array`, ...) and checker-merged declarations (a
+// namespace merged with a value, for instance) — which only the TypeChecker
+// can see, so it's used as a fallback there when available. Falling all the
+// way through to nil (no TypeChecker either) simply excludes the reference
+// from the through-reference set — the same outcome as if it had resolved
+// but turned out to be declared outside `funcNode` and never written to.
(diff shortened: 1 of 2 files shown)


elecmonkey (commented):
(The following was written by Codex) I don't think it deserve fixing because it only happens when the checker is missing.

> There is a false negative for TypeScript parameter properties when no TypeChecker is available.
> 
> In `constructor(public a: number) { for (...) { a = i; callbacks.push(() => a); } }`, `ctx.Refs.Resolve(a)` resolves to the local parameter symbol, while `decl.Symbol()` points to the class-property symbol. They are different symbols but share the same `Parameter` declaration, so the current `decl.Symbol() != sym` check rejects the binder references. With `ctx.TypeChecker == nil`, the fallback returns no references and `no-loop-func` misses this case.
> 
> Upstream ESLint reports this case with `@typescript-eslint/parser`, both with and without `parserOptions.project`, so rslint should keep using `ctx.Refs` for this shared-declaration parameter-property case.

elecmonkey (approved):

fansenze (approved):

PR #1450, opened 2026-07-28
feat(rule): fallback to TypeChecker in RefStore.Resolve for cross-file symbols

swwind:
## Summary

`RefStore.Resolve` now falls back to the TypeChecker automatically when the binder scope walk can't place an identifier (cross-file, `.d.ts`, standard-library globals).

## API changes

- `NewRefStore(sourceFile, options, tc)` — third parameter, the file's `*checker.Checker` (nil allowed). `linter.go` passes the file's checker.
- `RefStore.Resolve(node) *ast.Symbol` — tries the binder scope walk first; when that fails and a TypeChecker was supplied, falls back to it. Without a checker, behaves exactly as before (binder-only).
- `RefStore.References(sym) []*ast.Node` — unchanged, but now also finds in-file references to a symbol obtained through `Resolve`'s checker fallback.

## Merged-symbol identity

In a global script file the checker merges the file's top-level declarations with the same-named `lib.d.ts` and cross-file globals into one symbol object, distinct from the file's raw binder symbol, and both identities can reach the index for a single name. `References` normalizes such a symbol onto the checker's identity so a query answers the same either way — one checker call per top-level symbol of a global script file, and none for a module file or a nested scope.

## Locality helpers

`Resolve` can now return a symbol declared outside this file, so its callers need an explicit locality check. Two helpers cover the two questions:

- `utils.IsSymbolDeclaredInFile` — unchanged: any declaration in this file.
- `utils.IsValueSymbolDeclaredInFile` — new: only value-introducing declarations count, skipping interfaces and type aliases. A local `interface Map {}` merges into lib.d.ts's `Map` symbol without shadowing its value. A `namespace` counts regardless of its content, matching typescript-eslint's scope analysis.

Both skip declarations inside a `declare global` block or a module augmentation, which extend the augmented symbol instead of binding a name in the enclosing file.

## Rule fixes

- `no-misleading-character-class`, `unicorn/new-for-builtins`: both pair `Resolve` with `utils.IsValueSymbolDeclaredInFile` at their locality/shadowing checks, instead of treating a non-nil `Resolve` result alone as "declared in this file."
- `unicorn/new-for-builtins`: keeps the syntactic shadow check when `Resolve` finds nothing. A namespace with no value content of its own is outside the binder's value lookup, and a file without type information has no TypeChecker to fall back to, so `namespace Intl { export interface Local {} } Intl.DateTimeFormat()` was reported there.
- `no-use-before-define`: corrected a comment describing `needsCheckerFallback`; no code change.

## Behavior change

`no-use-before-define` now reports a use that precedes a local declaration merged with a lib.d.ts global — `Map; interface Map<K, V> {}` reports `'Map' was used before it was defined`, where it was previously silent. This matches ESLint, whose scope analysis merges the local declaration into the same variable as the predefined global. Locked in by tests.

## Docs

`architecture.md` and the `port-rule` skill (`SKILL.md`, `AST_PATTERNS.md`, `PORT_RULE.md`, `QUICK_REFERENCE.md`, `UTILS_REFERENCE.md`) document the merged `Resolve` method, the pairing of `Resolve` with the two locality helpers, and the global-script exception to "a same-file symbol never reaches the TypeChecker."

## Validation

- `go test` — `./internal/rule/...`, `./internal/utils/...`, `./internal/rules/no_unused_vars/...`, `./internal/rules/no_misleading_character_class/...`, `./internal/rules/no_new_func/...`, `./internal/rules/no_useless_backreference/...`, `./internal/plugins/typescript/rules/no_use_before_define/...`, `./internal/plugins/unicorn/rules/new_for_builtins/...` — all pass.
- `golangci-lint run --new-from-rev=origin/main` on changed packages — clean.

## Checklist


internal/utils/ts_eslint.go
```diff
@@ -1344,6 +1344,39 @@ func IsSymbolDeclaredInFile(symbol *ast.Symbol, sf *ast.SourceFile) bool {
...
+func IsValueSymbolDeclaredInFile(symbol *ast.Symbol, sf *ast.SourceFile) bool {
+	if symbol == nil {
+		return false
+	}
+	for _, decl := range symbol.Declarations {
+		switch decl.Kind {
+		case ast.KindInterfaceDeclaration, ast.KindTypeAliasDeclaration:
+			continue
+		case ast.KindModuleDeclaration:
+			if !ast.IsInstantiatedModule(decl, false /*preserveConstEnums*/) {
```
elecmonkey: When `preserveConstEnums` is enabled, or implied by `isolatedModules`, a namespace containing only exported const enums is instantiated at runtime, but hard-coding `false` classifies it as type-only. Consequently consumers such as `unicorn/new-for-builtins` treat a local `namespace Map { export const enum E { A } }` as the ambient built-in and incorrectly report `Map()`; use the active compiler option’s `ShouldPreserveConstEnums()` value when determining whether the module introduces a value.

swwind: Good catch on the hardcoded `false`. Digging into it, though, I found that upstream (typescript-eslint's scope analysis) doesn't consult `preserveConstEnums` — or the module's instantiation state — at all here. Verified with real ESLint + `eslint-plugin-unicorn`: a `namespace Map { export const enum E { A } }` and a `namespace Map { export interface Local {} }` both shadow `Map()` identically (no report in either case), regardless of `preserveConstEnums`. Only a bare `interface`/`type` (not wrapped in a namespace) fails to shadow.

So `IsInstantiatedModule` was the wrong check to use here in the first place, not just missing the right config value — a namespace declaration always counts as value-introducing upstream, no matter what it contains. I dropped that branch entirely instead of threading `preserveConstEnums` through, and fixed the one existing test (`namespace Intl { export interface Local {} }`) that had encoded the old, incorrect expectation.

d3045a03

fansenze (approved):

PR #1456, opened 2026-07-28
feat(eslint-plugin-jest): add `require-to-throw-message` rule

eryue0220:
## Summary

* Port `require-to-throw-message` from eslint to rslint

## Related Links

eslint-plugin-jest/require-to-throw-message [doc](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/require-to-throw-message.md) [code](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/require-to-throw-message.ts)

## Checklist


swwind (commented):
One edge case where this diverges from `eslint-plugin-jest`:

```js
(expect(() => { throw new Error('a'); }).toThrow)();
```

Real ESLint reports `addErrorMessage` here; this rule does not (false negative).

Cause: `FindExpectModifiersAndMatcher` in `internal/plugins/jest/utils/parse_jest_fn.go` checks `member.Node.Parent.Parent.Kind == ast.KindCallExpression` to detect the matcher being invoked. When the matcher access itself is wrapped in parens, the grandparent is a `ParenthesizedExpression` instead, so the chain isn't recognized as an expect call at all. ESTree (what real ESLint parses) has no such wrapper node, so this only shows up on the TS AST rslint uses.

This is in shared utils, not new code from this PR, so it's not blocking — but worth a follow-up since any rule built on `ParseJestFnCall`/`FindExpectModifiersAndMatcher` likely has the same gap whenever a matcher call is parenthesized.

eryue0220:
> One edge case where this diverges from `eslint-plugin-jest`:
> 
> ```js
> (expect(() => { throw new Error('a'); }).toThrow)();
> ```
> 
> Real ESLint reports `addErrorMessage` here; this rule does not (false negative).
> 
> Cause: `FindExpectModifiersAndMatcher` in `internal/plugins/jest/utils/parse_jest_fn.go` checks `member.Node.Parent.Parent.Kind == ast.KindCallExpression` to detect the matcher being invoked. When the matcher access itself is wrapped in parens, the grandparent is a `ParenthesizedExpression` instead, so the chain isn't recognized as an expect call at all. ESTree (what real ESLint parses) has no such wrapper node, so this only shows up on the TS AST rslint uses.
> 
> This is in shared utils, not new code from this PR, so it's not blocking — but worth a follow-up since any rule built on `ParseJestFnCall`/`FindExpectModifiersAndMatcher` likely has the same gap whenever a matcher call is parenthesized.

update.

swwind (approved):

PR #1459, opened 2026-07-29
feat(rule): resolve local named export specifiers in RefStore

swwind:
## Summary

- `ctx.Refs.Resolve`/`References` now resolve a local named export's real target (`export { foo }` / `export { foo as bar }`) through the ordinary binder scope walk, matching ESLint scope-manager semantics that treat it as a read. Re-exports (`export { foo } from 'mod'`) stay fully excluded either way, since neither side references anything in this file. `referenceMeaning` widens to `Value|Type|Namespace|Alias` for export specifiers. Type-only specifiers (`export type { foo }` / `export { type foo }`) resolve the same way, mirroring the type-flagged reference ESLint's scope-manager records for them; consumers that only follow runtime reads classify the reference site (`utils.IsReadReference` / `utils.IsNonReferenceIdentifier`) — `prefer-const`'s `isReadBeforeFirstAssign` now does exactly that, so a type-only export is not a read before assignment while `typeof x` still is.

- Drops two hand-rolled workarounds that existed only because of this gap:
  - `no_unassigned_vars`'s `collectLocalExportReads`/`binderLocalExportTarget`/`isLocalExportRead` — a standalone binder walk duplicating what `ctx.Refs.References` now does directly.
  - `no_use_before_define`'s `findLocalBindingName` and the `SkipAlias`/declarations-concat step in `checkNamedExport` — `ctx.Refs.Resolve` now lands directly on the local symbol (e.g. an import specifier's own symbol for a re-exported import), so there's no alias chain left to walk by hand.

- `no_unassigned_vars` also drops its now-dead `RequiresTypeInfo`/`ctx.TypeChecker`/`ctx.Refs` nil guards — it never calls `ctx.TypeChecker`, and `ctx.Refs` is non-nil whenever a `Program` exists, same reasoning as #1454.

- Extracts `IsReadReference` into `internal/utils/write_reference.go`, next to `IsWriteReference`, out of `no_unassigned_vars`'s private copy — a symmetric primitive other `ctx.Refs` consumers can reuse.

## Related Links

- #1454

## Validation

- `go test ./internal/rule/...` — 8 new tests covering local export reads, alias-only labels, re-exports with/without alias, and type-only local exports (indexed like scope-manager type references, still resolving actual type symbols, checker-identity stability).
- `go test ./internal/rules/prefer_const/...` — new cases locking the read classification: `export { x }` counts as a read before assignment, `export type { x }` / `export { type x }` does not, and `ignoreReadBeforeAssign` no longer swallows the type-only case.
- `go test ./internal/plugins/typescript/rules/no_use_before_define/...` — new valid/invalid cases for type-only named exports.
- `go test ./internal/rules/no_unassigned_vars/...`, `./internal/utils/...`.
- Report positions and type-only semantics verified against ESLint / typescript-eslint (`prefer-const`, `no-use-before-define`).
- Spot-checked every other `ctx.Refs.References`/`Resolve` consumer for regressions (`no_unused_vars` TS plugin, `new_for_builtins`, `no_class_assign`, `no_const_assign`, `no_func_assign`, `no_import_assign`, `no_misleading_character_class`, `no_obj_calls`, `no_shadow_restricted_names`, `no_var`, `prefer_const`) — all pass unmodified.
- `pnpx cspell lint` + `golangci-lint run --new-from-rev=origin/main` on changed files — clean.

## Checklist

### DIFF
--- a/internal/plugins/typescript/rules/no_use_before_define/edge_cases_test.go
+++ b/internal/plugins/typescript/rules/no_use_before_define/edge_cases_test.go
@@ -128,6 +128,14 @@ func TestEdgeCases(t *testing.T) {
 				Options: map[string]interface{}{"allowNamedExports": true},
 			},
 
+			// ----- Type-only named exports -----
+			{Code: `let x = 1; export type { x };`},
+			{Code: `type T = number; export type { T };`},
+			{
+				Code:    `export type { x }; let x = 1;`,
+				Options: map[string]interface{}{"allowNamedExports": true},
+			},
+
 			// ----- Decorators with classes:false -----
 			{
 				Code: `
@@ -386,6 +394,33 @@ declare global {
 				Errors: []rule_tester.InvalidTestCaseError{
 					{MessageId: "noUseBeforeDefine"},
 					{MessageId: "noUseBeforeDefine"},
+				},
+			},
+
+			// ----- Type-only named exports before define (typescript-eslint -----
+			// ----- reports these even with the default ignoreTypeReferences) -----
+			{
+				Code: `export type { x }; let x = 1;`,
+				Errors: []rule_tester.InvalidTestCaseError{
+					{MessageId: "noUseBeforeDefine", Line: 1, Column: 15},
+				},
+			},
+			{
+				Code: `export { type x }; let x = 1;`,
+				Errors: []rule_tester.InvalidTestCaseError{
+					{MessageId: "noUseBeforeDefine", Line: 1, Column: 15},
+				},
+			},
+			{
+				Code: `export type { x as y }; let x = 1;`,
+				Errors: []rule_tester.InvalidTestCaseError{
+					{MessageId: "noUseBeforeDefine", Line: 1, Column: 15},
+				},
+			},
+			{
+				Code: `export type { T }; type T = number;`,
+				Errors: []rule_tester.InvalidTestCaseError{
+					{MessageId: "noUseBeforeDefine", Line: 1, Column: 15},
 				},
 			},
 
--- a/internal/plugins/typescript/rules/no_use_before_define/no_use_before_define.go
+++ b/internal/plugins/typescript/rules/no_use_before_define/no_use_before_define.go
@@ -445,9 +445,10 @@ func checkIdentifier(ctx rule.RuleContext, opts options, node *ast.Node, refScop
 		return
 	}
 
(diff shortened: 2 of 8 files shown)


elecmonkey (commented):
I found a regression from making local export specifiers visible to every `RefStore` consumer.

`RefStore.isReferencePosition` now treats every non-re-export `ExportSpecifier` as a reference, but this also includes explicit type-only exports such as `export type { value }` and `export { type value }`. `@typescript-eslint/parser`/scope-manager does not create a read reference for those forms; it only reports the later assignment as a write reference.

This changes behavior for consumers that treat any non-write `ctx.Refs.References` entry as a runtime read. For example, with:

```ts
let value;
export type { value };
value = 1;
```

ESLint's `prefer-const` with `{ ignoreReadBeforeAssign: true }` still reports at the assignment (`value = 1`). On this branch, a temporary overlay test against `internal/rules/prefer_const` reports 0 diagnostics because the type-only export is incorrectly seen as a read before assignment, so `prefer-const` suppresses the warning.

The `ExportSpecifier` branch should exclude `ast.IsTypeOnlyImportOrExportDeclaration(p)` in addition to source-bearing re-exports, or otherwise keep type-only export specifiers out of the shared reference index. Please also add regression coverage for both `export type { value }` and `export { type value }`, ideally through a shared consumer such as `prefer-const` so this does not regress outside the rules that motivated the refactor.

swwind:
Fixed in 0a7c8b56 — thanks for catching this.

Rather than dropping the value space in `referenceMeaning` (which would also drop `export type { x }; let x = 1` out of `no-use-before-define`'s named-export check — typescript-eslint reports that even with the default `ignoreTypeReferences: true`), the index keeps type-only specifiers, mirroring the type-flagged reference the scope-manager records for them. The fix is on the consumer side: `prefer-const`'s `isReadBeforeFirstAssign` now classifies the reference site with `utils.IsNonReferenceIdentifier`, so a type-only export is not a read before assignment while `typeof x` still counts as one (both verified against ESLint).

Regression coverage: `prefer-const` cases for `export type { value }`, `export { type value }`, and the aliased form — the report stays at the assignment and `ignoreReadBeforeAssign: true` no longer swallows it; RefStore tests for the index semantics; `no-use-before-define` valid/invalid cases for type-only named exports.

elecmonkey (approved):

PR #1461, opened 2026-07-29
refactor(jest,rstest): share no-mocks-import rule

elecmonkey:
## Summary

Move the framework-neutral `no-mocks-import` implementation to `internal/utils/test_framework/rules/no_mocks_import`.

Jest and Rstest keep their own rule wrappers, while both delegate to the shared implementation. This removes the direct Rstest dependency on the Jest rule.

The intended dependency direction is:

```text
Jest rules   -> Jest utils   \
                              -> internal/utils/test_framework
Rstest rules -> Rstest utils /
```


elecmonkey:
@codex review

fansenze (approved):

PR #1463, opened 2026-07-29
fix(jest): preserve semantics in accessor edits

elecmonkey:
## Summary

Fix `no-focused-tests` suggestions for bracket notation containing whitespace or line breaks, such as `describe[ "only" ]()`.

The previous implementation inferred the surrounding brackets from the string literal coordinates. This change uses the complete accessor AST range instead, ensuring the suggestion correctly produces `describe()`.


elecmonkey:
@codex review

internal/plugins/jest/rules/no_focused_tests/no_focused_tests.go
```diff
@@ -27,6 +27,39 @@ func buildErrorSuggestRemoveFocusMessage() rule.RuleMessage {
...
+
+	switch accessor.Kind {
+	case ast.KindPropertyAccessExpression:
+		// Removing an optional accessor would also remove its `?.` token and
+		// turn a guarded call into an unconditional one.
+		if accessor.AsPropertyAccessExpression().QuestionDotToken != nil {
+			return core.TextRange{}, false
+		}
+	case ast.KindElementAccessExpression:
+		if accessor.AsElementAccessExpression().QuestionDotToken != nil {
```
swwind: This guard regresses `describe?.["only"]()`: before this PR, the old code (raw positions off the string-literal node) computed a removal range of `(10, 18)`, which preserves the leading `?.` and yields the correct, valid suggestion `describe?.()`. This PR's guard sees `QuestionDotToken != nil` on the accessor and unconditionally drops the suggestion, so a previously-working case now gets no fix at all.

A more precise fix: compute the removal range starting after the accessor's `QuestionDotToken.End()` rather than at `receiver.End()`, so `?.` is preserved and the suggestion is still offered. Confirmed against real `eslint-plugin-jest`, whose fix range for this input is `[10,18)` — same effect.

internal/plugins/jest/rules/no_focused_tests/no_focused_tests.go
```diff
@@ -27,6 +27,39 @@ func buildErrorSuggestRemoveFocusMessage() rule.RuleMessage {
 	}
 }
 
+func focusedMemberRemovalRange(
+	ctx rule.RuleContext,
+	entry *jestUtils.ParsedJestFnMemberEntry,
+) (core.TextRange, bool) {
+	receiver, accessor := jestUtils.GetAccessorReceiverAndParent(entry)
```
swwind: Separately (not a regression from this PR — old code also mishandled it, emitting a syntax-breaking `describe[]()` suggestion): `describe[("only")]()` still gets no suggestion today, because `GetJestFnMemberEntries` stores the paren-unwrapped node as `entry.Node` (jest.go:225) while its real `.Parent` is the `ParenthesizedExpression`, not the `ElementAccessExpression`. `GetAccessorReceiverAndParent` hits the `default` branch and returns `(nil, nil)`. Not blocking this PR, just flagging since bracket-notation suggestions are exactly its stated scope — worth a follow-up.

swwind (commented):
Two findings on the removal-range logic in `focusedMemberRemovalRange`.

internal/plugins/jest/rules/no_focused_tests/no_focused_tests.go
elecmonkey: Fixed in 45a11af9. Suggestions now preserve the existing `?.` boundary across calls and chained accessors, with edge-case coverage for whitespace and comments.

internal/plugins/jest/rules/no_focused_tests/no_focused_tests.go
elecmonkey: Fixed in 4ce2ac96. Parenthesized computed keys are now traced back to their element accessor, including nested parentheses and optional access.

internal/plugins/jest/rules/no_focused_tests/no_focused_tests.go
```diff
@@ -27,6 +27,154 @@ func buildErrorSuggestRemoveFocusMessage() rule.RuleMessage {
 	}
 }
 
+func focusedMemberAccessor(
```
swwind: `focusedMemberAccessor` is `jestUtils.GetAccessorReceiverAndParent` (`internal/plugins/jest/utils/parse_jest_fn.go:479`) plus the paren-walk, and the `?.`-aware start offset lives here too. The shared helper still has both blind spots, and `prefer-to-be` calls it in `appendRemoveNotModifierFix` — where the failure is an autofix, not a suggestion.

Confirmed by running the rule tester on:

```
expect("a string")[("not")].toBeUndefined();
```

`FindExpectModifiersAndMatcher` still recognizes `not` (its `member.Node.Parent` check tolerates the parens), so `reportPreferToBe(..., stripNot=true)` renames the matcher — but `GetAccessorReceiverAndParent` returns `(nil, nil)` for the parenthesized key, so the `.not` removal fix is silently dropped by the `continue`. Result:

```
Expected to have 1 outputs but had 10:
  expect("a string")[("not")].toBeDefined();
  expect("a string")[("not")].toBeUndefined();
  expect("a string")[("not")].toBeDefined();
  ...
```

So the first pass inverts the assertion (`not.toBeUndefined()` → `not.toBeDefined()`), and the fixer never converges — it flip-flops for the full 10 passes. `expect("a string")["not"].toBeUndefined()` (no parens) fixes correctly, confirming the parens are the trigger.

The second blind spot reproduces there too: `expect("a string")?.["not"].toBeUndefined()` fixes to `expect("a string").toBeDefined()`, i.e. the `?.` is swallowed by the `receiver.End()`-based range — the same thing this PR fixes here with `questionDot.End()`.

Suggestion: move the paren-walk and the optional-chain boundary into `GetAccessorReceiverAndParent` and have this rule call it, so `prefer-to-be` / `prefer-to-have-been-called` get the fix too.

swwind (commented):
Re-checked both follow-ups (`45a11af9`, `4ce2ac96`). Both earlier comments are addressed, and the approach holds up. `go test ./internal/plugins/jest/rules/no_focused_tests/` is green, and I compared the new expectations against real `eslint-plugin-jest`:

| Code | ESLint | This PR |
| --- | --- | --- |
| `describe[ "only" ]()` | `describe[]()` | `describe()` |
| `describe[("only")]()` | `describe[]()` | `describe()` |
| `describe?.only()` | `describe?()` | `describe?.()` |
| `describe?.["only"]?.()` | `describe?.?.()` | `describe?.()` |
| `describe?.only?.each()()` | `describe??.each()()` | `describe?.each()()` |
| `describe?.only["each"]()()` | `describe?["each"]()()` | `describe?.["each"]()()` |
| `describe[/* keep */ "only"]()` | `describe[/* keep */()` | `describe/* keep */ ()` |
| `(describe?.["only"])()` | not reported | reported, no suggestion |

Upstream produces syntactically invalid code in every bracket/optional case; this PR produces valid code in all of them. Two things worth calling out as deliberate and correct:

- The `callExpr.Expression != accessor` bail is load-bearing. I patched it out and re-ran: the suggestion for `(describe?.["only"])()` then produces `(describe?.)()`, which is a hard syntax error (`Unexpected token ')'`) — the preserved `?.` needs a following property / `[` / `(` and there is none inside the parens. `(describe.only)()` → `(describe)()` either way, so the guard costs nothing.
- Switching to per-token removal when a comment is in range (rather than dropping the suggestion) keeps every comment permutation I tried valid, including line comments (`describe[// c\n "only"]()` → `describe// c\n ()`).

One new finding inline. It is pre-existing, not a regression from this PR, so not blocking — but it is the same root cause this PR is fixing, so it seems worth doing here.

Two nits:

- The expectations added to `packages/rslint-test-tools/.../no-focused-tests.test.ts` do not actually assert anything new: that `RuleTester` only compares diagnostic counts and ignores `messageId`, line/column and `suggestions` entirely (`tests/eslint-plugin-jest/rule-tester.ts`). Real coverage is the Go test; just flagging so the parity file is not mistaken for verification.
- `AGENTS.md` asks to keep tests minimal and behavior-focused. ~30 comment/whitespace permutations could probably collapse to one representative per class (no comment / comment inside brackets / comment around `?.` / comment around the following connector).

elecmonkey:
@codex review

swwind (commented):
This PR's parenthesis-unwrapping fix in `GetAccessorReceiverAndParent` (used by `FindExpectModifiersAndMatcher`) makes `expect(...)[("matcher")](...)` parse successfully where it previously didn't. Two rules that build their fix range directly from `matcherEntry.Node.Parent.End()` instead of going through that same unwrapping accessor now fire on this newly-parseable input but emit invalid JS/TS, since `.Parent` is the `ParenthesizedExpression`, not the `ElementAccessExpression`. Neither file is touched by this PR's diff, so noting them here rather than as inline comments.

| Code | ESLint | This PR |
| --- | --- | --- |
| `expect(a === b)[("toBe")](false);` | `expect(a).not.toBe(b);` | `expect(a).not.toBe](b);` |
| `expect(value > 1)[("toBe")](true);` | `expect(value).toBeGreaterThan(1);` | `expect(value).toBeGreaterThan](1);` |

Both confirmed with `go test` against the exact inputs above, and cross-checked against real `eslint-plugin-jest` (which handles both correctly).

**`internal/plugins/jest/rules/prefer_equality_matcher/prefer_equality_matcher.go:110`**

`modifierRange` is built from `matcherEntry.Node.Parent.End()`. For a parenthesized bracket key (`[("toBe")]`), `.Parent` is the `ParenthesizedExpression`, so the range ends before the closing `]`. `expect(a === b)[("toBe")](false);` now fixes to `expect(a).not.toBe](b);` — invalid syntax. Route this through the same paren-unwrapping accessor `GetAccessorReceiverAndParent` uses (or call it directly) so the range ends at the real `ElementAccessExpression`.

**`internal/plugins/jest/rules/prefer_comparison_matcher/prefer_comparison_matcher.go:118`**

Same issue as `prefer_equality_matcher.go:110` — `matcherEntry.Node.Parent.End()` stops at the `)` instead of the `]` for a parenthesized bracket key. `expect(value > 1)[("toBe")](true);` now fixes to `expect(value).toBeGreaterThan](1);` — invalid syntax.

elecmonkey:
> This PR's parenthesis-unwrapping fix in `GetAccessorReceiverAndParent` (used by `FindExpectModifiersAndMatcher`) makes `expect(...)[("matcher")](...)` parse successfully where it previously didn't. Two rules that build their fix range directly from `matcherEntry.Node.Parent.End()` instead of going through that same unwrapping accessor now fire on this newly-parseable input but emit invalid JS/TS, since `.Parent` is the `ParenthesizedExpression`, not the `ElementAccessExpression`. Neither file is touched by this PR's diff, so noting them here rather than as inline comments.
> 
> Code	ESLint	This PR
> `expect(a === b)[("toBe")](false);`	`expect(a).not.toBe(b);`	`expect(a).not.toBe](b);`
> `expect(value > 1)[("toBe")](true);`	`expect(value).toBeGreaterThan(1);`	`expect(value).toBeGreaterThan](1);`
> Both confirmed with `go test` against the exact inputs above, and cross-checked against real `eslint-plugin-jest` (which handles both correctly).
> 
> **`internal/plugins/jest/rules/prefer_equality_matcher/prefer_equality_matcher.go:110`**
> 
> `modifierRange` is built from `matcherEntry.Node.Parent.End()`. For a parenthesized bracket key (`[("toBe")]`), `.Parent` is the `ParenthesizedExpression`, so the range ends before the closing `]`. `expect(a === b)[("toBe")](false);` now fixes to `expect(a).not.toBe](b);` — invalid syntax. Route this through the same paren-unwrapping accessor `GetAccessorReceiverAndParent` uses (or call it directly) so the range ends at the real `ElementAccessExpression`.
> 
> **`internal/plugins/jest/rules/prefer_comparison_matcher/prefer_comparison_matcher.go:118`**
> 
> Same issue as `prefer_equality_matcher.go:110` — `matcherEntry.Node.Parent.End()` stops at the `)` instead of the `]` for a parenthesized bracket key. `expect(value > 1)[("toBe")](true);` now fixes to `expect(value).toBeGreaterThan](1);` — invalid syntax.

I’m starting to lose hope over the endless edge cases in autofix 😅 But thanks for the review—I’ll keep working through them and iterating on the fixes.

swwind:
> I’m starting to lose hope over the endless edge cases in autofix 😅 But thanks for the review—I’ll keep working through them and iterating on the fixes.

Jest & rstest users will thank you for your contribution to fix those edge cases! :smile:

internal/plugins/jest/utils/accessor.go
```diff
@@ -0,0 +1,209 @@
...
+	callNode *ast.Node,
+	callee *ast.Node,
+	replacement string,
+) (rule.RuleFix, bool) {
+	callExpr := callNode.AsCallExpression()
+	if callExpr == nil || callee == nil {
+		return rule.RuleFix{}, false
+	}
+
+	start := callee.End()
```
swwind: `start := callee.End()` assumes `callee` is the call's direct callee, but `prefer-to-have-been-called` passes the paren-unwrapped accessor. For a parenthesized callee the range starts inside the parens, so the replacement swallows the closing `)`:

```js
(expect(method).toHaveBeenCalledTimes)(0);
```

autofixes to `(expect(method).not.toHaveBeenCalled();` — invalid syntax (confirmed with the rule tester; upstream produces `(expect(method).not.toHaveBeenCalled)();`). Same behavior exists on main, so not a regression — but since this PR rewrites this fix path, it seems worth handling here: `start := callExpr.Expression.End()` covers it and is identical for the existing callers' non-parenthesized cases (the optional-call branch already handles `)?.(`). Verified that one-line change against both caller packages' tests, including the case above.

swwind (commented):
Re-checked the squashed tip (`933fade1`). Both findings from my previous round (`prefer-equality-matcher` / `prefer-comparison-matcher` building the modifier range from `matcherEntry.Node.Parent.End()`) are fixed via `GetAccessorReceiverAndParent` and covered by tests, and the shared edit helpers in `utils/accessor.go` now apply the `?.`/bracket/comment handling consistently across the five touched rules. CI is green, and I cross-checked against real `eslint-plugin-jest` v29.15.5:

| Code | ESLint | This PR |
| --- | --- | --- |
| `expect("a string")[("not")].toBeUndefined();` | `expect("a string")[)].toBeDefined();` (invalid) | `expect("a string").toBeDefined();` |
| `expect("a string")?.["not"].toBeUndefined?.();` | `expect("a string")?.].toBeDefined?.();` (invalid) | `expect("a string")?.toBeDefined?.();` |
| `expect(method)?.["toHaveBeenCalledTimes"](0);` | `expect(method).not.toHaveBeenCalled();` | `expect(method)?.not['toHaveBeenCalled']();` |
| `expect(method.mock.calls)?.["toHaveLength"](1);` | `expect(method).toHaveBeenCalledTimes(1);` | `expect(method)?.['toHaveBeenCalledTimes'](1);` |

Upstream emits syntactically invalid code in the bracket-`not` cases; this PR stays valid in every case I tried except one remaining spot, noted inline. It is pre-existing behavior on main, not a regression — flagging it because the helper carrying it is new in this PR.

internal/plugins/jest/utils/accessor.go
elecmonkey: Fixed in https://github.com/web-infra-dev/rslint/pull/1463/commits/236ff8739baa0464e9960740277e616c0a16a913

internal/plugins/jest/utils/accessor.go
```diff
@@ -0,0 +1,209 @@
...
+		core.NewTextRange(questionDot.End(), questionDot.End()),
+		suffix,
+	), true
+}
+
+// ReplaceCallSuffixFix replaces a call's type arguments and arguments while
+// preserving an optional-call token owned by the call.
+func ReplaceCallSuffixFix(
+	callNode *ast.Node,
+	callee *ast.Node,
```
swwind: Nit: `callee` is now dead — it is only nil-checked, and both call sites already guarantee non-nil (`prefer_to_have_been_called.go:61` checks `matcherAccessor == nil`, `prefer_to_be.go:179` checks `matcherExpr == nil`). Dropping the parameter would make it obvious that the range is derived from the call node alone.

internal/plugins/jest/utils/accessor.go
```diff
@@ -0,0 +1,209 @@
...
+	replacement string,
+) (rule.RuleFix, bool) {
+	callExpr := callNode.AsCallExpression()
+	if callExpr == nil || callee == nil {
+		return rule.RuleFix{}, false
+	}
+
+	start := callExpr.Expression.End()
```
swwind, line 207: Nit, pre-existing on main and not a regression: a comment between the callee and the argument list is swallowed, because the replaced range starts at the callee's end rather than at the `(`.

`expect(value).toEqual /* keep */ (undefined);` → this PR `expect(value).toBeUndefined();`, upstream `expect(value).toBeUndefined /* keep */ ();`.

Cosmetic only (output stays valid), but it is the one comment case in this area the PR does not preserve, so it may be worth folding in while the helper is being written.

swwind (commented):
Two nits inline.

internal/plugins/jest/utils/accessor.go
elecmonkey: Both nits addressed in 9bc4ab55.

1. Dropped the dead `callee` parameter — `ReplaceCallSuffixFix` is now `(sourceFile, callNode, replacement)`, and the two call sites (`prefer_to_have_been_called.go`, `prefer_to_be.go`) were updated. The range is derived from the call node alone, which reads much clearer.

2. Folded in the comment case as well. The start of the replaced range is now advanced with `scanner.GetRangeOfTokenAtPosition` to the `<` or `(` that actually opens the suffix, instead of starting at the callee end. So `expect(value).toEqual /* keep */ (undefined);` now fixes to `expect(value).toBeUndefined /* keep */ ();`, matching upstream.

`go vet` is clean and the `prefer_to_be` / `prefer_to_have_been_called` Go tests pass.

swwind (approved):
LGTM

PR #1468, opened 2026-07-30
feat: port rule no-useless-assignment

swwind:
## Summary

Port the `no-useless-assignment` rule from ESLint to rslint.

### Differences from ESLint

- A variable read from inside a parameter decorator's arguments is treated as used, so assignments to it are never reported: `const pipe = {}; class C { handler(@Body(pipe) body) {} }`. ESLint reports `const pipe = {}` here — a known false positive ([eslint/eslint#20947](https://github.com/eslint/eslint/issues/20947)).
- For deeply nested `try`/`catch`/`switch` combinations, this rule's control flow modeling can occasionally diverge from ESLint's own, which has open, accepted-but-unfixed bugs in the same area ([eslint/eslint#17579](https://github.com/eslint/eslint/issues/17579)) and its own false negatives on some such shapes. Matching it exactly there is hard to pin down, so it is left as a known gap. These shapes are rare in practice; in the large majority of real code this rule reports the same findings as ESLint.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/no-useless-assignment
- Source code: https://github.com/eslint/eslint/blob/main/lib/rules/no-useless-assignment.js

## Checklist

### DIFF
--- a/internal/rules/all.go
+++ b/internal/rules/all.go
@@ -118,6 +118,7 @@ import (
 	"github.com/web-infra-dev/rslint/internal/rules/no_unused_labels"
 	"github.com/web-infra-dev/rslint/internal/rules/no_unused_private_class_members"
 	"github.com/web-infra-dev/rslint/internal/rules/no_unused_vars"
+	"github.com/web-infra-dev/rslint/internal/rules/no_useless_assignment"
 	"github.com/web-infra-dev/rslint/internal/rules/no_useless_backreference"
 	"github.com/web-infra-dev/rslint/internal/rules/no_useless_call"
 	"github.com/web-infra-dev/rslint/internal/rules/no_useless_catch"
@@ -289,6 +290,7 @@ func GetAllRules() []rule.Rule {
 		prefer_arrow_callback.PreferArrowCallbackRule,
 		no_dupe_else_if.NoDupeElseIfRule,
 		no_throw_literal.NoThrowLiteralRule,
+		no_useless_assignment.NoUselessAssignmentRule,
 		no_useless_backreference.NoUselessBackreferenceRule,
 		no_useless_call.NoUselessCallRule,
 		no_useless_catch.NoUselessCatchRule,
--- a/internal/rules/no_useless_assignment/cfg.go
+++ b/internal/rules/no_useless_assignment/cfg.go
@@ -1,0 +1,1404 @@
+package no_useless_assignment
+
+import (
+	"github.com/microsoft/typescript-go/shim/ast"
+
+	"github.com/web-infra-dev/rslint/internal/utils"
+)
+
+// This file builds a control-flow graph for one code path root (a source file,
+// a function, a class static block, or a class field initializer) and answers
+// "can the value written here still be read?" over it.
+//
+// The graph mirrors the shape ESLint's code path analysis produces, because the
+// rule's observable behaviour depends on it:
+//
+//   - the end of a `try` block flows into its `catch` block as well as past the
+//     statement;
+//   - the first node inside a `try` block (or inside the `catch` block of a
+//     `try` that has a `finally`) that could throw forks to the handler, so an
+//     assignment later in the block is bypassed on the throwing path;
+//   - a `finally` block is laid out twice, once for normal completion and once
+//     for the path that leaves the `try` statement through `return`, `throw`,
+//     or a suspended `yield`; `break` and `continue` do not take that path;
+//   - `&&`, `||`, `??`, `?:`, optional chains, and destructuring defaults fork
+//     around the operand they may skip.
+//
+// Reads and writes are recorded as events in evaluation order, so liveness is a
+// plain backward dataflow over the graph.
+
+type eventKind uint8
+
+const (
+	eventRead eventKind = iota
+	eventWrite
+)
+
+type event struct {
+	kind   eventKind
+	sym    *ast.Symbol
+	assign *assignment
+}
+
+type block struct {
+	events      []event
+	successors  []*block
+	hasIncoming bool
+
+	// Per-symbol dataflow scratch, reset by liveWrites for each symbol.
+	use     bool
+	def     bool
+	liveIn  bool
+	liveOut bool
+}
+
+// writeSite is one place in the graph where an assignment appears. An
+// assignment inside a `finally` block appears twice — once per copy of the
+// block — and only counts as unused when every site is dead.
+type writeSite struct {
+	blk   *block
+	index int
+}
+
+// assignment is one reportable write: the identifier ESLint reports on plus the
+// variable it targets.
+type assignment struct {
+	sym        *ast.Symbol
+	identifier *ast.Node
+	sites      []writeSite
+	// silent marks an assignment that still overwrites the variable but is
+	// never reported, because it sits in a `try` block.
+	silent bool
+}
+
+type tryPosition uint8
+
+const (
+	posTry tryPosition = iota
+	posCatch
+)
+
+// tryFrame tracks one enclosing `try` statement while its blocks are built.
+type tryFrame struct {
+	position     tryPosition
+	hasFinally   bool
+	catchEntry   *block
+	finallyEntry *block
+	thrownForked bool
+	thrownAny    bool
+	returnedAny  bool
+}
+
+// jumpTarget is one `break`/`continue` destination. continueTo is nil for
+// targets that only `break` can reach (switch statements, labelled blocks).
+type jumpTarget struct {
+	label      string
+	breakTo    *block
+	continueTo *block
+}
+
+type builder struct {
+	// assignByIdent maps a write target identifier to the assignment it
+	// belongs to; identifiers absent from the map target untracked variables.
+	assignByIdent map[*ast.Node]*assignment
+	// readNodes maps a reference identifier that reads a tracked variable to
+	// that variable's symbol.
+	readNodes map[*ast.Node]*ast.Symbol
+
+	blocks []*block
+	cur    *block
+
+	jumps      []jumpTarget
+	tryStack   []*tryFrame
+	chainJoins []*block
+}
+
+func (b *builder) newBlock() *block {
+	blk := &block{}
+	b.blocks = append(b.blocks, blk)
+	return blk
+}
+
+func (b *builder) link(from, to *block) {
+	if from == nil || to == nil {
+		return
+	}
+	from.successors = append(from.successors, to)
+	to.hasIncoming = true
+}
+
+// enter makes blk the current block when something can reach it, and marks the
+// path unreachable otherwise.
+func (b *builder) enter(blk *block) {
+	if blk.hasIncoming {
+		b.cur = blk
+	} else {
+		b.cur = nil
+	}
+}
+
+func (b *builder) emitRead(sym *ast.Symbol) {
+	if b.cur == nil {
+		return
+	}
+	b.cur.events = append(b.cur.events, event{kind: eventRead, sym: sym})
+}
+
+func (b *builder) emitWrite(a *assignment) {
+	if b.cur == nil {
+		return
+	}
+	a.sites = append(a.sites, writeSite{blk: b.cur, index: len(b.cur.events)})
+	b.cur.events = append(b.cur.events, event{kind: eventWrite, sym: a.sym, assign: a})
+}
+
+// ---------------------------------------------------------------------------
+// throw / return routing
+// ---------------------------------------------------------------------------
+
+// throwFrame returns the innermost `try` frame an exception raised at the
+// current position lands in, or -1 when it leaves the code path entirely.
+func (b *builder) throwFrame() int {
+	for i := len(b.tryStack) - 1; i >= 0; i-- {
+		f := b.tryStack[i]
+		if f.position == posTry || (f.hasFinally && f.position == posCatch) {
+			return i
+		}
+	}
+	return -1
+}
+
+func throwTarget(f *tryFrame) *block {
+	if f.position == posTry && f.catchEntry != nil {
+		return f.catchEntry
+	}
+	return f.finallyEntry
+}
+
+// returnFrame returns the innermost `try` frame whose `finally` block runs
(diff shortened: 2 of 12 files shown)


elecmonkey:
@codex review

internal/rules/no_useless_assignment/no_useless_assignment.go
```diff
@@ -0,0 +1,541 @@
...
+	if node == nil {
+		return
+	}
+	switch node.Kind {
+	case ast.KindIdentifier:
+		cb(node)
+
+	case ast.KindParenthesizedExpression:
+		forEachTargetIdentifier(node.AsParenthesizedExpression().Expression, cb)
+	case ast.KindNonNullExpression:
```
elecmonkey: This unwraps TypeScript-only assignment target wrappers (`as`, non-null, `satisfies`, and type assertions), but upstream `extractIdentifiersFromPattern()` does not treat these nodes as assignment patterns under `@typescript-eslint/parser`.

Repro with ESLint 10.8.0 + @typescript-eslint/parser 8.65.0 + TypeScript 6.0.3:

```ts
function f() {
  let x = 1;
  g(x);
  (x as any) = 2;
}
```

ESLint reports 0 diagnostics, while this PR reports `x`. The same mismatch happens for `x! = 2`, `(x satisfies any) = 2`, and `(<any>x) = 2`.

internal/rules/no_useless_assignment/no_useless_assignment.go
```diff
@@ -0,0 +1,541 @@
...
+
+// isExported reports whether a module-level binding leaves the file, in which
+// case a later assignment to it can still be observed elsewhere. Both an
+// `export` modifier and an `export { ... }` re-export only apply to the
+// declaration actually named at the top level of the file — a block-scoped
+// variable that merely shares that name is a different binding.
+func isExported(sym *ast.Symbol, exportedNames map[string]bool, root *ast.Node) bool {
+	for _, decl := range sym.Declarations {
+		target := decl
+		if decl.Kind == ast.KindVariableDeclaration {
```
elecmonkey: This export check misses destructured top-level exports because destructured bindings are declared on `BindingElement`, not directly on `VariableDeclaration`. As a result, `target.Parent != root` filters them out before checking the exported declaration.

Repro:

```js
export let { a } = obj;
console.log(a);
a = 2;
```

ESLint 10.8.0 reports 0 diagnostics, but this PR reports the final `a = 2`. The same issue occurs for `let { a } = obj; export { a };` and `export { a as b };`.

internal/rules/no_useless_assignment/cfg.go
```diff
@@ -0,0 +1,1462 @@
...
+	// A `typeof x` inside the annotation references the variable, so the
+	// annotation is walked even though nothing in it runs.
+	b.expr(decl.Type)
+	if decl.Initializer == nil {
+		// `let x;` writes nothing this rule can report on, but a binding
+		// pattern can still contain evaluated expressions.
+		b.patternReads(decl.Name())
+		return
+	}
+	b.patternReads(decl.Name())
```
elecmonkey: The CFG currently evaluates the whole destructuring target's reads before the RHS, then emits all writes afterwards. That loses the per-element evaluation order that ESLint's source-order segment tracking observes.

Repro:

```js
function f() {
  let { a: x, [x]: y } = obj;
  console.log(y);
}
```

ESLint 10.8.0 reports 0 diagnostics because the computed key `[x]` reads the value written by the previous element. This PR reports the `x` binding as an unused assignment. The same mismatch appears for assignment patterns such as `({ a: x, [x]: y } = obj)`.

elecmonkey: There is also a reverse mismatch for destructuring defaults after RHS writes. ESLint reports the RHS assignment here, but this CFG treats the default read as if it can use the RHS-assigned value:

```js
function f() {
  let x = 0;
  console.log(x);
  let { a = x } = (x = 1, obj);
  console.log(a);
}
```

ESLint 10.8.0 reports `x = 1`; this PR reports 0 diagnostics. Since this rule is explicitly trying to match upstream behavior, this ordering needs to be modeled the same way ESLint's code-path segments see it.

internal/rules/no_useless_assignment/cfg.go
```diff
@@ -0,0 +1,1462 @@
...
+	b.pushJump(node, after, head)
+	b.enter(body)
+	if stmt.Initializer != nil {
+		if stmt.Initializer.Kind == ast.KindVariableDeclarationList {
+			// The loop variable is bound, not assigned: ESLint only tracks
+			// declarators that carry an initializer.
+			declList := stmt.Initializer.AsVariableDeclarationList()
+			if declList != nil && declList.Declarations != nil {
+				for _, decl := range declList.Declarations.Nodes {
+					b.patternReads(decl.Name())
```
elecmonkey: For `for-in` / `for-of` destructuring, this only calls `patternReads()` on the loop target. `patternReads()` intentionally skips destructuring default expressions, so reads from defaults are dropped.

Repro:

```js
function f(xs) {
  let y = 1;
  console.log(y);
  y = 2;
  for (let [x = y] of xs) console.log(x);
}
```

ESLint 10.8.0 reports 0 diagnostics, but this PR reports `y = 2`. The same issue reproduces with `for (let [x = y] in xs)`.

internal/rules/no_useless_assignment/cfg.go
```diff
@@ -0,0 +1,1462 @@
...
+			// well as past the statement.
+			b.link(b.cur, frame.catchEntry)
+			normalEnds = append(normalEnds, b.cur)
+		}
+		frame.position = posCatch
+		frame.thrownForked = false
+		b.enter(frame.catchEntry)
+		catchClause := stmt.CatchClause.AsCatchClause()
+		if catchClause.VariableDeclaration != nil {
+			b.patternReads(catchClause.VariableDeclaration.Name())
```
elecmonkey: Catch binding destructuring has the same default-read gap: only `patternReads()` is used, so default expressions are not modeled as reads.

Repro:

```js
function f() {
  let y = 1;
  console.log(y);
  y = 2;
  try {
    throw {};
  } catch ({ x = y }) {
    console.log(x);
  }
}
```

ESLint 10.8.0 reports 0 diagnostics, but this PR reports `y = 2`.

elecmonkey (commented):
I double-checked the findings against the current PR head (`13bb8a694d7c4ce94d0086b4d53f0d806b0891f5`) and a fresh temporary oracle project using:

- eslint 10.8.0
- @typescript-eslint/parser 8.65.0
- TypeScript 6.0.3

Two non-inline issues also remain:

1. `no-useless-assignment` is still missing from Rslint's recommended presets. Upstream marks the rule as recommended and enables it in `eslint:recommended`, but this PR leaves it commented out as `not implemented` in both `packages/rslint/src/config/presets/javascript.ts` and `packages/rslint/src/config/presets/typescript.ts`.

2. The upstream `/* exported foo */` script directive case is skipped in the migrated tests, but the rule docs do not mention it as a difference from ESLint. Repro:

```js
/* exported foo */
let foo = 1;
console.log(foo);
foo = 2;
```

ESLint 10.8.0 reports 0 diagnostics; this PR reports the final assignment. If this directive remains unsupported, it should be documented in `Differences from ESLint`.

internal/rules/no_useless_assignment/no_useless_assignment.go
swwind: Fixed in 2aee0100 — `forEachTargetIdentifier` no longer unwraps `as` / `!` / `satisfies` / `<T>` wrappers, matching upstream's `extractIdentifiersFromPattern`. Such a write is now fully invisible to the rule: it is not reported, and it neither reads nor overwrites the variable (`x = 2; (x as any) = 3;` still reports `x = 2`; with a later `g(x)` nothing is reported). All four wrapper forms plus the pattern-nested variant are covered in the extras tests.

internal/rules/no_useless_assignment/no_useless_assignment.go
swwind: Fixed in 2aee0100 — `isExported` now climbs from the binding element through the pattern to the enclosing variable statement before checking the export modifier / `export { ... }` names, so `export let { a } = obj`, `let { a } = obj; export { a }`, and `export { a as b }` are all recognized. A block-scoped shadow of an exported name is still reported (test added).

internal/rules/no_useless_assignment/cfg.go
swwind: Fixed in 2aee0100 — destructuring is now laid out in run-time evaluation order: the initializer/right-hand side first, then each element's computed key, forked default, and write in source order. `let { a: x, [x]: y } = obj` and the assignment-pattern variant are both clean now (tests added).

swwind: Resolved in 2aee0100, but deliberately not by matching ESLint here. The rule now models the run-time evaluation order, and at run time the initializer's `x = 1` executes before the default `a = x` may read it — the store is genuinely reachable by a read, so reporting it (as ESLint does, based on source positions) would be a false positive. We chose to keep the run-time model rather than reproduce that report; the divergence is called out in the rule docs under “Differences from ESLint”, and both this case and the computed-key analogue are locked in as valid tests.

internal/rules/no_useless_assignment/cfg.go
swwind: Fixed in 2aee0100 — for-in/of heads now lay out the full pattern including forked default reads, so `for (let [x = y] of xs)` and the `for...in` variant no longer report `y = 2`. Tests added for both.

internal/rules/no_useless_assignment/cfg.go
swwind: Fixed in 2aee0100 — catch bindings go through the same pattern layout now, so `catch ({ x = y })` reads its default. Test added.

internal/rules/no_useless_assignment/cfg.go
```diff
@@ -0,0 +1,1462 @@
...
+		b.enter(right)
+		b.expr(binary.Right)
+		b.patternWrites(binary.Left)
+		b.link(b.cur, join)
+		b.enter(join)
+
+	case operator == ast.KindEqualsToken:
+		b.patternReads(binary.Left)
+		b.expr(binary.Right)
+		b.patternWrites(binary.Left)
```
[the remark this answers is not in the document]
swwind: Resolved in 2aee0100 — destructuring now follows the run-time evaluation order (right-hand side first, then per-element computed keys/defaults/writes). For this example the rule now reports `a = 2` — the store the computed key immediately overwrites — and keeps `a = 1`, whose value `console.log(a)` observes. For the record, actual ESLint 10.8 reports `a = 1` and misses `a = 2` (its model is position-based); we intentionally follow the run-time order instead, documented in the rule docs.

internal/rules/no_useless_assignment/cfg.go
```diff
@@ -0,0 +1,1462 @@
...
+func labelOf(node *ast.Node) string {
+	parent := node.Parent
+	if parent == nil || parent.Kind != ast.KindLabeledStatement {
+		return ""
+	}
+	labeled := parent.AsLabeledStatement()
+	if labeled.Statement != node {
+		return ""
+	}
+	return labeled.Label.Text()
```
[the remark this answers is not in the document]
swwind: Fixed in 2aee0100 — `labelsOf` now collects every label wrapped around a loop or switch, so `continue outer` resolves and keeps its back edge; the example no longer reports `v = 1`. One note: ESLint 10.8 itself *does* report `v = 1` here (its continue context carries only the innermost label), so this is a spot where the port deliberately does not reproduce ESLint's false positive; documented in the rule docs.

internal/rules/all.go
```diff
@@ -289,6 +290,7 @@ func GetAllRules() []rule.Rule {
 		prefer_arrow_callback.PreferArrowCallbackRule,
 		no_dupe_else_if.NoDupeElseIfRule,
 		no_throw_literal.NoThrowLiteralRule,
+		no_useless_assignment.NoUselessAssignmentRule,
```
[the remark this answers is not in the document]
swwind: Fixed in 13811435 — the rule is now enabled in both recommended presets.

swwind:
Review feedback is addressed as of 13811435. Honest status:

**Fixed**
- TS assertion wrappers around assignment targets are invisible to the rule, matching upstream.
- Destructured top-level exports (`export let { a } = obj`, `export { a }`, `export { a as b }`) are recognized.
- Destructuring is modeled in run-time evaluation order (initializer first, then per-element computed keys, forked defaults, writes); the `{ a: x, [x]: y }` family and the for-in/of / catch default-read gaps are gone.
- Multi-labelled loops resolve `break`/`continue` through every wrapping label.
- The rule is enabled in both recommended presets.

**Deliberate divergences from ESLint** — I decided to follow run-time semantics here instead of reproducing ESLint's false positives; all documented in the rule docs under “Differences from ESLint”:
- `let { a = x } = (x = 1, obj)` and `let { [a]: q } = (a = 1, obj)` are clean here — the default/key can read the initializer's store; ESLint reports them.
- `({ [a = 1]: x } = (a = 2, obj)); console.log(a)` reports `a = 2` (genuinely dead) instead of ESLint's `a = 1`.
- `outer: inner: while (…) { … continue outer; }` keeps its back edge; ESLint 10.8 reports the assignment before it.

**Left as skipped**
- `/* exported */` stays unsupported (same scope decision as the no-unused-vars port); the migrated upstream test remains skipped and the gap is documented.
- Differential fuzzing against ESLint 10.8 (~6k cases) leaves exactly three divergent shapes, all deeply nested try/catch/finally exception paths; they are recorded in the extras tests as `Skip: true` cases with ESLint's output as the expectation and a note on what rslint reports instead (rslint routes the throw a destructuring/assignment target can raise into the enclosing `finally`, which is defensible at run time).

elecmonkey (dismissed):

elecmonkey (approved):

Issue #1216, opened 2026-07-07
[Tracking]: Schema-driven rule options frameworks with auto TS type generation


swwind:
Add JSON Schema-based validation, default values, and TypeScript type generation for rule options.

Reference ESLint document: https://eslint.org/docs/latest/extend/custom-rules#options-schemas

## Steps

1. Add JSON Schema support
   - Vendor a JSON Schema validation library
   - Add a small wrapper for compiling and validating schemas
   - Add unit tests
2. Implement new schema framework
   - Make rules accept a plain options array (aligned with ESLint) instead of the legacy rslint-specific format
   - Remove support for the legacy object-style rule configuration
   - Update all existing rules (using scripts for now)
   - Ensure the full test suite passes
   - Validate user configs against each rule's schema with clear error messages
   - Populate unspecified options using schema defaults
   - Verify the behavior end-to-end with a couple of real rules
   - Write a guide describing how to author rule schemas consistently for future rules
3. Generate TypeScript types for rule options
   - Generate precise TypeScript types for `rslint.config.ts` from each rule's schema
   - Integrate type generation into the build process
4. Migrate existing rule packages
   - Migrate rules package by package, remove legacy code in rules one by one.
   - Packages:
     - `eslint`
     - `typescript-eslint`
     - `import`
     - `jest`
     - `jsx_a11y`
     - `promise`
     - `react`
     - `react_hooks`
     - `unicorn`
5. Remove legacy compatibility code
   - Remove the legacy rule configuration format
   - Remove compatibility shims and unused code
   - Ensure the new implementation is the only supported path

## Checklist

- [x] Step 1 — Add JSON Schema support (#1222)
- [x] Step 2 — Implement new schema framework
- [x] Step 3 — Generate TypeScript types for rule options
- [x] Step 4 — Migrate `eslint`
- [x] Step 4 — Migrate `typescript-eslint`
- [x] Step 4 — Migrate `import`
- [x] Step 4 — Migrate `jest`
- [x] Step 4 — Migrate `jsx_a11y`
- [x] Step 4 — Migrate `promise`
- [x] Step 4 — Migrate `react`
- [x] Step 4 — Migrate `react_hooks`
- [x] Step 4 — Migrate `unicorn`
- [x] Step 5 — Remove legacy compatibility code

PR #1469, opened 2026-07-30
feat(typescript-eslint): declare options JSON Schema for @typescript-eslint/ rules

swwind:
## Summary

Step 4 of #1216 for the `typescript-eslint` package: all 130 `@typescript-eslint/` rules declare their options JSON Schema, so configs are validated the way ESLint validates them. 55 rules that take no options share `rule.EmptyArraySchema`; 75 embed a `<rule>.schema.json` taken from upstream's `meta.schema`, wrapped the way ESLint's `getRuleOptionsSchema` wraps the tuple form. `rule.LegacyUnwrapOptions`, `utils.GetOptionsMap` and `utils.GetOptionsString` are gone from this package, so the removed bare-object config form is no longer accepted.

That part is mechanical. The rest is not:

**Eight rules could not read their options from a real config.** `no-base-to-string`, `no-duplicate-type-constituents`, `no-meaningless-void-operator`, `no-misused-spread`, `prefer-promise-reject-errors`, `promise-function-async`, `require-array-sort-compare` and `restrict-plus-operands` only accepted a Go options struct via a type assertion. A config's `[{ "ignoreStringArrays": false }]` failed that assertion and fell through to defaults, silently; their Go tests passed the struct, so nothing caught it. They parse decoded JSON now, and several structs gained the `json` tags they were missing.

**Regex-valued options are validated as regexes.** Nine options across seven rules are compiled as regular expressions by the rule, but upstream declares them as bare strings, so an unparsable pattern was dropped and the rule linted as if nothing were configured. They are marked `format: "regex"` (37 schema nodes), as ESLint core's own `dot-notation` schema already does:

| rule | options |
| --- | --- |
| `ban-ts-comment` | `descriptionFormat` |
| `dot-notation` | `allowPattern` |
| `naming-convention` | `filter`, in both its string and object form (the object form is shared with `custom`) |
| `no-empty-object-type` | `allowWithName` |
| `no-require-imports` | `allow` |
| `no-unused-vars` | `vars` / `args` / `caughtErrors` / `destructuredArray` ignore patterns |
| `no-var-requires` | `allow` |

Patterns are checked with the ECMAScript engine, so JS-only syntax such as lookbehind still validates. This diverges from upstream, so each rule has a test pinning it. `ban-ts-comment`, `naming-convention`, `no-empty-object-type`, `no-require-imports`, `no-unused-vars` and `no-var-requires` also compile and match the pattern at runtime with the same ECMAScript engine (regexp2), matching `dot-notation`'s existing approach — otherwise a pattern that passes schema validation could still fail to compile with Go's RE2 and silently do nothing.

**`no-unused-vars` ignored the bare `vars` string.** Upstream's schema allows `"all"` / `"local"` at slot 0 and the rule implements `vars == "local"`, but parsing only handled the object form, so `["local"]` did nothing. It is honored now.

**Two stale rslint-only options are removed.** `no-floating-promises`' `allowForKnownSafeCallsInline` / `allowForKnownSafePromisesInline` and `no-misused-spread`'s `allowInline` predate #502, which taught `utils.TypeOrValueSpecifier` to accept a bare string; post-#502 rules pass `nil` for that argument and put bare strings in `allow`. Upstream declares no such options and its schemas are `additionalProperties: false`, so they were unreachable from a config. The test cases using them moved to the upstream option names and still pass.

**`$defs` resolves under the draft-4 dialect.** Eight rules' schemas reach into upstream's `$defs` via pointers like `#/items/0/$defs/banConfig`. `$defs` is a 2019-09 keyword, so draft-4 treats it as unknown, but the pointer resolves against the raw document regardless and the constraints still apply. Had that been wrong, validation would have silently loosened for all eight; `TestNoRestrictedTypesTypesSchema` pins it.

**Dead branches from the removed bare-object config form are gone.** `consistent-indexed-object-style` and `consistent-generic-constructors` also read a `{ "style": ... }` object, but upstream's schema is a bare string enum at slot 0. Same for `return-await`'s options wrapper type and `class-literal-property-style`'s duplicate string arm.

**Test cases that pinned unreachable configs are deleted**, because config validation now rejects the shapes they described: a bare integer for `max-params`, a non-object for `class-methods-use-this`, `null` / `false` / numeric ban configs for `no-restricted-types`, out-of-enum values for `no-type-alias`, unknown keys for rules that take no options, and a two-element options array under a `maxItems: 1` schema.

**Deliberately unchanged.** Options the schema declares because upstream declares them, but this port does not implement: `no-unsafe-member-access.allowOptionalChaining`, `naming-convention.failureMessage`, `no-unused-expressions.ignoreDirectives`, `no-base-to-string.checkUnknown`, `prefer-promise-reject-errors.allow`, and `prefer-readonly-parameter-types.allow`. A config using them validates and is ignored, as before.

**Follow-up for step 5.** Four rules here delegate to shared implementations that also serve the un-migrated ESLint core rules, and those still call the legacy helpers: `internal/rules/no_shadow`, `internal/rules/no_redeclare`, `internal/rules/no_restricted_imports`, `internal/utils/no_unused_expressions.go`. Behavior is unchanged today, but deleting the helpers needs the `eslint` package migrated in the same step.

## Related Links

- Tracking issue: #1216
- Test-case options validation this builds on: #1464
- Bare-string `TypeOrValueSpecifier` support that made the `*Inline` options redundant: #502

## Checklist

### DIFF
--- a/internal/plugins/typescript/rules/adjacent_overload_signatures/adjacent_overload_signatures.go
+++ b/internal/plugins/typescript/rules/adjacent_overload_signatures/adjacent_overload_signatures.go
@@ -181,7 +181,8 @@ func checkBodyForOverloadMethods(ctx rule.RuleContext, node *ast.Node) {
 }
 
 var AdjacentOverloadSignaturesRule = rule.CreateRule(rule.Rule{
-	Name: "adjacent-overload-signatures",
+	Name:   "adjacent-overload-signatures",
+	Schema: rule.EmptyArraySchema,
 	Run: func(ctx rule.RuleContext, options []any) rule.RuleListeners {
 		// Check the source file at the beginning
 		checkBodyForOverloadMethods(ctx, &ctx.SourceFile.Node)
--- a/internal/plugins/typescript/rules/array_type/array_type.go
+++ b/internal/plugins/typescript/rules/array_type/array_type.go
@@ -1,6 +1,7 @@
 package array_type
 
 import (
+	_ "embed"
 	"fmt"
 
 	"github.com/microsoft/typescript-go/shim/ast"
@@ -8,9 +9,27 @@ import (
 	"github.com/web-infra-dev/rslint/internal/utils"
 )
 
+//go:embed array_type.schema.json
+var schemaJSON []byte
+
 type ArrayTypeOptions struct {
 	Default  string `json:"default"`
 	Readonly string `json:"readonly,omitempty"`
+}
+
+func parseOptions(options []any) ArrayTypeOptions {
+	opts := ArrayTypeOptions{Default: "array"}
+	if len(options) == 0 {
+		return opts
+	}
+	optsMap, _ := options[0].(map[string]interface{})
+	if defaultVal, ok := optsMap["default"].(string); ok {
+		opts.Default = defaultVal
+	}
+	if readonlyVal, ok := optsMap["readonly"].(string); ok {
+		opts.Readonly = readonlyVal
+	}
+	return opts
 }
 
 // Check whatever node can be considered as simple
@@ -257,34 +276,10 @@ func buildArrayFixes(
 }
 
 var ArrayTypeRule = rule.CreateRule(rule.Rule{
-	Name: "array-type",
-	Run: func(ctx rule.RuleContext, _options []any) rule.RuleListeners {
-		options := rule.LegacyUnwrapOptions(_options)
-		opts := ArrayTypeOptions{
-			Default: "array",
-		}
-		// Parse options with dual-format support (handles both array and object formats)
-		if options != nil {
-			var optsMap map[string]interface{}
-			var ok bool
-
-			// Handle array format: [{ option: value }]
-			if optArray, isArray := options.([]interface{}); isArray && len(optArray) > 0 {
-				optsMap, ok = optArray[0].(map[string]interface{})
-			} else {
-				// Handle direct object format: { option: value }
-				optsMap, ok = options.(map[string]interface{})
-			}
-
-			if ok {
-				if defaultVal, ok := optsMap["default"].(string); ok {
-					opts.Default = defaultVal
-				}
-				if readonlyVal, ok := optsMap["readonly"].(string); ok {
-					opts.Readonly = readonlyVal
-				}
-			}
-		}
+	Name:   "array-type",
+	Schema: rule.NewSchema(schemaJSON),
+	Run: func(ctx rule.RuleContext, options []any) rule.RuleListeners {
+		opts := parseOptions(options)
 
 		defaultOption := opts.Default
 		readonlyOption := opts.Readonly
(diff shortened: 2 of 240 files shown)


elecmonkey (commented):
I found two schema compatibility issues that should be addressed before merging.

First, regex-valued options are validated with `regexp2`, whose accepted syntax is not equivalent to ECMAScript Unicode RegExp syntax. For example, `new RegExp("\\p{Script=Greek}+", "u")` is valid in Node, but `regexp2` rejects `Script=Greek` because its `\p{...}` parser only accepts names from its internal Unicode category table. Consequently, a previously valid `@typescript-eslint/no-unused-vars` configuration now fails during rule-option validation. The rule runtime also compiles these patterns with `regexp2`, so validation and runtime compilation need a solution that supports the full Unicode ECMAScript syntax accepted upstream.

Second, `no_restricted_imports.schema.json` omits several constraints from ESLint's current `no-restricted-imports` schema. Pattern objects must contain exactly one of `group` or `regex`, and the restricted/allowed import-name options are mutually exclusive in several combinations. Path objects also disallow specifying both `importNames` and `allowImportNames`. The current schema accepts all of these contradictory shapes. In particular, the existing test that supplies both `group` and `regex` treats the invalid configuration as valid and silently gives `regex` precedence, while ESLint rejects it at configuration time.

Please restore the upstream `oneOf` and `not` constraints and add regression coverage for both invalid restricted-import configurations and valid Unicode ECMAScript property escapes.

[by Codex]

swwind:
> First, regex-valued options are validated with `regexp2`, whose accepted syntax is not equivalent to ECMAScript Unicode RegExp syntax. For example, `new RegExp("\\p{Script=Greek}+", "u")` is valid in Node, but `regexp2` rejects `Script=Greek` because its `\p{...}` parser only accepts names from its internal Unicode category table. Consequently, a previously valid `@typescript-eslint/no-unused-vars` configuration now fails during rule-option validation. The rule runtime also compiles these patterns with `regexp2`, so validation and runtime compilation need a solution that supports the full Unicode ECMAScript syntax accepted upstream.

This was an interesting opinion, I may investigate for better go regex engines later, but not in this PR.


> Second, `no_restricted_imports.schema.json` omits several constraints from ESLint's current `no-restricted-imports` schema. Pattern objects must contain exactly one of `group` or `regex`, and the restricted/allowed import-name options are mutually exclusive in several combinations. Path objects also disallow specifying both `importNames` and `allowImportNames`. The current schema accepts all of these contradictory shapes. In particular, the existing test that supplies both `group` and `regex` treats the invalid configuration as valid and silently gives `regex` precedence, while ESLint rejects it at configuration time.


We are fixing typescripts-eslint's no_restricted_imports rule, those are core rule problems, not related to this PR.

elecmonkey (dismissed):

fansenze (approved):

PR #1471, opened 2026-07-30
feat: Add basePath support and improve linting for config entries

### LINKED ISSUE #1201: Align config basePath with ESLint flat config
## Summary

Rslint should consider supporting the ESLint flat config `basePath` option on config entries.

In ESLint, `basePath` sets a subdirectory base for a config object. Relative `basePath` values are resolved from the config file directory, or from cwd when using an alternate config via `--config`; `files` and `ignores` in that object are then resolved relative to that base path.

## Why

Rslint recently aligned explicit `--config` handling with ESLint by resolving `files`, `ignores`, and `parserOptions.project` from cwd. Supporting `basePath` would close the next compatibility gap for users sharing or migrating ESLint flat configs.

Without support, configs like this are silently interpreted with the wrong match base:

```ts
export default [
  {
    basePath: 'packages/foo',
    files: ['src/**/*.ts'],
    ignores: ['fixtures/**'],
  },
];
```

## Suggested behavior

- Add `basePath?: string` to config entries.
- Resolve relative `basePath` from the config directory for auto-discovered configs.
- Resolve relative `basePath` from cwd for explicit `--config` configs.
- Resolve that entry's `files`, `ignores`, and `languageOptions.parserOptions.project` relative to the effective base path.
- Treat `{ basePath, ignores }` as a global ignore entry, matching ESLint's intent for scoped global ignores.

## References

- ESLint docs: https://eslint.org/docs/latest/use/configure/configuration-files#specify-files-and-ignores
- ESLint issue: https://github.com/eslint/eslint/issues/15683

Penty-d:
## Summary

Adding basePath support to flat config entries. Each entry can now specify an optional basePath that scopes its files / ignores / parserOptions.project patterns to a subdirectory, instead of always resolving relative to the config file's directory or cwd — matching ESLint's flat-config basePath semantics.

## Related Links
Fixes #1201 

## Checklist

### DIFF
--- a/internal/config/base_path.go
+++ b/internal/config/base_path.go
@@ -1,10 +1,12 @@
 package config
 
 import (
+	"fmt"
 	"path"
 	"strings"
 
 	"github.com/microsoft/typescript-go/shim/tspath"
+	"github.com/microsoft/typescript-go/shim/vfs"
 )
 
 // ResolveBasePaths desugars per-entry BasePath into ordinary relative
@@ -13,27 +15,41 @@ import (
 // configDirectory is the match root the rest of the engine already uses:
 // config-file directory for auto-discovered configs, cwd for --config.
 // Relative BasePath values resolve against it. Absolute BasePath values are
-// rebased via ConvertToRelativePath.
+// rebased via ConvertToRelativePath. fsys supplies filesystem case sensitivity
+// (as the rest of config matching does) so an absolute basePath whose casing
+// differs from the match root still resolves on case-insensitive filesystems.
+//
+// A BasePath that resolves outside configDirectory is rejected: its desugared
+// patterns would be ../-prefixed and could never match under the engine's
+// single-match-root model, indistinguishable from a typo. ESLint reports such
+// files as "ignored because outside of base path"; rslint fails fast instead
+// so the misconfiguration is actionable.
 //
 // JS/TS configs normally desugar basePath in Node before the payload reaches
 // Go; this path covers JSON/JSONC configs and any raw ConfigEntry constructed
 // with BasePath still set.
-func ResolveBasePaths(config RslintConfig, configDirectory string) RslintConfig {
+func ResolveBasePaths(config RslintConfig, configDirectory string, fsys vfs.FS) (RslintConfig, error) {
 	if len(config) == 0 {
-		return config
+		return config, nil
 	}
+	useCaseSensitive := fsys == nil || fsys.UseCaseSensitiveFileNames()
 	for i := range config {
-		resolveBasePathForEntry(&config[i], configDirectory)
+		if err := resolveBasePathForEntry(&config[i], configDirectory, useCaseSensitive); err != nil {
+			return nil, fmt.Errorf("config entry at index %d: %w", i, err)
+		}
 	}
-	return config
+	return config, nil
 }
 
-func resolveBasePathForEntry(entry *ConfigEntry, configDirectory string) {
+func resolveBasePathForEntry(entry *ConfigEntry, configDirectory string, useCaseSensitive bool) error {
 	if entry.BasePath == "" {
-		return
+		return nil
 	}
 
-	relativeBase := resolveRelativeBasePath(entry.BasePath, configDirectory)
+	relativeBase, err := resolveRelativeBasePath(entry.BasePath, configDirectory, useCaseSensitive)
+	if err != nil {
+		return err
+	}
 	entry.BasePath = ""
 
 	if len(entry.Files) > 0 {
@@ -81,6 +97,7 @@ func resolveBasePathForEntry(entry *ConfigEntry, configDirectory string) {
 		}
 		entry.Files = []string{catchAll}
 	}
+	return nil
 }
 
 func entryScopesToSubtree(entry ConfigEntry) bool {
@@ -90,26 +107,64 @@ func entryScopesToSubtree(entry ConfigEntry) bool {
 		entry.LanguageOptions != nil
 }
 
-// resolveRelativeBasePath returns a POSIX-ish path relative to configDirectory.
-// Empty means the config root itself (no prefix).
-func resolveRelativeBasePath(basePath string, configDirectory string) string {
+// resolveRelativeBasePath returns a glob-escaped POSIX-ish path relative to
+// configDirectory. Empty means the config root itself (no prefix). It rejects
+// a basePath that resolves outside configDirectory, since the resulting
+// ../-prefixed patterns could never match under the engine's single-match-root
+// model.
+func resolveRelativeBasePath(basePath string, configDirectory string, useCaseSensitive bool) (string, error) {
 	if basePath == "" {
-		return ""
+		return "", nil
+	}
+	// tspath.ResolvePath normalizes to forward slashes, but callers (and test
+	// fixtures) may hand in a configDirectory with native separators. Normalize
+	// once so the containment check and relative rebase compare like for like.
+	if configDirectory != "" {
+		configDirectory = tspath.NormalizePath(configDirectory)
 	}
 	absolute := tspath.ResolvePath(configDirectory, basePath)
+	if _, ok := RelativePathWithinConfigRoot(absolute, configDirectory, useCaseSensitive); !ok {
+		return "", fmt.Errorf(
+			"key \"basePath\": value %q resolves outside the config match root %q; basePath must be a subdirectory of the config root",
+			basePath,
+			configDirectory,
+		)
+	}
 	relative := tspath.ConvertToRelativePath(absolute, tspath.ComparePathsOptions{
-		UseCaseSensitiveFileNames: true,
+		UseCaseSensitiveFileNames: useCaseSensitive,
 		CurrentDirectory:          configDirectory,
 	})
 	relative = strings.ReplaceAll(tspath.NormalizePath(relative), "\\", "/")
 	relative = strings.TrimPrefix(relative, "./")
 	if relative == "." || relative == "" {
-		return ""
+		return "", nil
 	}
-	return relative
+	return escapeGlobBasePath(relative), nil
 }
 
-// rebasePattern prefixes a relative glob/path with relativeBase.
+// escapeGlobBasePath escapes glob metacharacters in a resolved basePath so the
+// directory is treated literally when spliced into glob patterns. ESLint treats
+// basePath as a literal path; without escaping, a basePath such as
+// "packages/[locale]" would be interpreted as a character class and never match.
+//
+// Escaping uses character classes ([*], [?], [[] ...), which the doublestar and
+// picomatch matchers understand on every platform and which survive the
+// tspath.NormalizePath applied to patterns before matching. A bare ']' or '}'
+// outside a class/brace pair is already literal to the matchers, so only the
+// characters that open a class/brace or introduce a wildcard need escaping. A
+// leading '!' (basePath "!vendor") has no class form the matchers accept as
+// literal, so it remains a known limitation.
+func escapeGlobBasePath(s string) string {
+	return strings.NewReplacer(
+		"[", "[[]",
+		"*", "[*]",
+		"?", "[?]",
+		"{", "[{]",
+	).Replace(s)
+}
+
+// rebasePattern prefixes a relative glob/path with the glob-escaped
+// relativeBase produced by resolveRelativeBasePath.
 // Leading ! negation is preserved; absolute patterns are left unchanged;
 // empty relativeBase only strips a leading ./.
 func rebasePattern(pattern string, relativeBase string) string {
--- a/internal/config/base_path.go
+++ b/internal/config/base_path.go
@@ -142,21 +142,8 @@ func rebasePattern(pattern string, relativeBase string) string {
 	return joined
 }
 
+// isAbsolutePattern reports whether pattern is an absolute path in any form
+// tspath recognizes (POSIX, UNC, Windows drive-letter).
 func isAbsolutePattern(pattern string) bool {
-	if pattern == "" {
-		return false
-	}
-	// POSIX absolute, Windows drive, or UNC.
-	if strings.HasPrefix(pattern, "/") || strings.HasPrefix(pattern, "\\") {
-		return true
-	}
-	if len(pattern) >= 3 {
-		drive := pattern[0]
-		if (drive >= 'A' && drive <= 'Z') || (drive >= 'a' && drive <= 'z') {
-			if pattern[1] == ':' && (pattern[2] == '/' || pattern[2] == '\\') {
-				return true
-			}
-		}
-	}
 	return tspath.GetEncodedRootLength(pattern) > 0
 }
--- a/internal/config/base_path.go
+++ b/internal/config/base_path.go
@@ -1,0 +1,162 @@
+package config
+
+import (
+	"path"
+	"strings"
+
+	"github.com/microsoft/typescript-go/shim/tspath"
+)
+
+// ResolveBasePaths desugars per-entry BasePath into ordinary relative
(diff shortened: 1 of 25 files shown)


internal/config/config.go
```diff
@@ -186,7 +194,7 @@ func (config *RslintConfig) UnmarshalJSON(data []byte) error {
 		// neutral but remains non-nil for isGlobalIgnoreEntry.
 		hasNonGlobalKey := false
 		for key := range raw {
-			if key != "ignores" && key != "name" {
+			if key != "ignores" && key != "name" && key != "basePath" {
```
swwind, line 207: `ValidateConfig` never checks `BasePath`, so `"basePath": ""` is silently treated as "not set" — `ResolveBasePaths`'s `entry.BasePath == ""` early-return skips it with no error, while the TS-side `normalizeConfig` throws `"basePath" must be a non-empty string` for the same input. Go's zero-value `string` can't distinguish "key absent" from "key present but empty" once decoded, so the fix has to live here, using the `raw` map that's still in scope at this point (e.g. `if rawBasePath, ok := raw["basePath"]; ok { ... }`), not in `ValidateConfig` after the fact.

This also matters beyond the deprecated `rslint.config.json` file path: the low-level `@rslint/core/internal` `lint()` API (`packages/rslint/src/internal/node.ts`) forwards caller-supplied config straight to Go over IPC without going through `normalizeConfig`, so it can hit this same silent no-op.

internal/config/base_path.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+	return joined
+}
+
+func isAbsolutePattern(pattern string) bool {
+	if pattern == "" {
+		return false
+	}
+	// POSIX absolute, Windows drive, or UNC.
+	if strings.HasPrefix(pattern, "/") || strings.HasPrefix(pattern, "\\") {
```
swwind: These manual POSIX/UNC/drive-letter checks duplicate `tspath.GetEncodedRootLength`, which already covers all of them (see the fallback on the last line of this function). Consider simplifying `isAbsolutePattern` to just `return tspath.GetEncodedRootLength(pattern) > 0`.

packages/rslint/src/config/base-path.ts
```diff
@@ -0,0 +1,192 @@
...
+function isRecord(value: unknown): value is Record<string, unknown> {
+  return value !== null && typeof value === 'object' && !Array.isArray(value);
+}
+
+/** True for absolute POSIX or Windows paths (including drive-letter forms). */
+export function isAbsolutePattern(pattern: string): boolean {
+  if (path.posix.isAbsolute(pattern) || path.win32.isAbsolute(pattern)) {
+    return true;
+  }
+  // path.win32.isAbsolute requires a platform-aware check; also catch
```
swwind: The comment here is incorrect — `path.win32.isAbsolute` is not gated on the host OS; `path.win32.isAbsolute('C:/foo')` returns `true` regardless of platform. The regex fallback on the next line is therefore dead code and can be dropped along with this comment.

packages/rslint/src/config/config-file-loader.ts
```diff
@@ -127,6 +147,8 @@ export function normalizeConfig(config: unknown): Record<string, unknown>[] {
     }
   }
 
+  const configDirectory = options.configDirectory ?? process.cwd();
```
swwind, line 152: This `process.cwd()` fallback is what the API's `overrideConfig` path lands on: `rslint.ts` calls `normalizeConfig(override)` without `{ configDirectory: this.#cwd }`, so an `overrideConfig` basePath desugars against `process.cwd()` while Go matches the resulting patterns relative to the instance `cwd`. With `cwd: '/repo'` and `basePath: '/repo/packages/foo'` run from `/`, the override silently matches nothing — and the `ResolveBasePaths` fallback in `cmd/rslint/api.go` can't repair it because `basePath` is already stripped here. Same omission in `packages/rule-tester/src/index.ts`, which then passes a different `configDirectory` to `lint()`. Relative basePaths are root-invariant, which is why the current tests don't catch this.

internal/config/config.go
```diff
@@ -29,6 +29,14 @@ type RslintConfig []ConfigEntry
...
 type ConfigEntry struct {
 	Name string `json:"name,omitempty"`
+	// BasePath is the optional ESLint flat-config subdirectory root for this
+	// entry. Relative values resolve from the config match root (config-file
+	// directory for auto-discovered configs, cwd for --config). JS/TS configs
+	// desugar basePath in Node before the payload reaches Go; JSON/JSONC
+	// configs are desugared in LoadRslintConfig via ResolveBasePaths. The field
+	// is cleared after desugaring so matching always sees ordinary relative
+	// files/ignores/project patterns.
+	BasePath string `json:"basePath,omitempty"`
```
swwind, line 40: The `rslint --init` migration (`config_init.go`) decodes this field into `ConfigEntry`, but `generateEntryCode`/`buildOverrideFields`/`formatIgnoreEntry` never emit it, so every basePath-scoped entry is silently widened to the repo root — e.g. `{"basePath": "packages/foo", "ignores": ["fixtures/**"]}` becomes a global ignore for every `fixtures/**` in the repo. Since the source JSON is removed after migration, the scope loss is unrecoverable and unannounced.

internal/config/base_path.go
```diff
@@ -0,0 +1,149 @@
...
+
+// resolveRelativeBasePath returns a POSIX-ish path relative to configDirectory.
+// Empty means the config root itself (no prefix).
+func resolveRelativeBasePath(basePath string, configDirectory string) string {
+	if basePath == "" {
+		return ""
+	}
+	absolute := tspath.ResolvePath(configDirectory, basePath)
+	relative := tspath.ConvertToRelativePath(absolute, tspath.ComparePathsOptions{
+		UseCaseSensitiveFileNames: true,
```
swwind: `UseCaseSensitiveFileNames` is hardcoded to `true`, while sibling config code derives it from `fs.UseCaseSensitiveFileNames()` (e.g. `ignore_pattern.go`). On a case-insensitive filesystem, an absolute basePath with different casing (`C:\Repo\packages\foo` vs config root `c:/repo`) yields `../Repo/...` patterns that match nothing — and the TS mirror uses `path.relative` (case-insensitive on win32), so Go and TS disagree for the same config.

internal/config/base_path.go
```diff
@@ -0,0 +1,149 @@
...
+		}
+		return body
+	}
+	if relativeBase == "" {
+		if negated {
+			return "!" + body
+		}
+		return body
+	}
+	joined := path.Join(relativeBase, body)
```
swwind, line 193: `rebasePattern` splices the basePath directory into patterns without glob-escaping (same in `base-path.ts`). `basePath: "packages/[locale]"` turns into a character class and never matches; `"!vendor"` turns into a negated pattern. ESLint treats basePath as a literal path, so both sides diverge from it here — the prefix needs glob-escaping, or matching by path containment.

internal/config/base_path.go
```diff
@@ -0,0 +1,149 @@
...
+func entryScopesToSubtree(entry ConfigEntry) bool {
+	return entry.Rules != nil ||
+		entry.Plugins != nil ||
+		entry.Settings != nil ||
+		entry.LanguageOptions != nil
+}
+
+// resolveRelativeBasePath returns a POSIX-ish path relative to configDirectory.
+// Empty means the config root itself (no prefix).
+func resolveRelativeBasePath(basePath string, configDirectory string) string {
```
swwind: A basePath resolving outside the match root (`"../shared"`, or an absolute path elsewhere) desugars into `../`-prefixed patterns that discovery can never match — the entry loads and validates cleanly but contributes nothing, indistinguishable from a typo'd glob. ESLint reports "File ignored because outside of base path" per file; worth rejecting or diagnosing this, e.g. via the existing ok-flag from `RelativePathWithinConfigRoot`.

packages/rslint/src/config/base-path.ts
```diff
@@ -0,0 +1,187 @@
...
+  if (body === '' || isAbsolutePattern(body)) {
+    return (negated ? '!' : '') + body;
+  }
+
+  if (!relativeBase) {
+    return (negated ? '!' : '') + body;
+  }
+
+  // path.posix.join collapses `a` + `../b` and keeps `**` segments intact.
+  const joined = path.posix.join(relativeBase, body);
```
swwind: The desugaring exists twice (here and `base_path.go`) with no shared fixtures, and there's already observable drift for the same entry: `path.posix.join` keeps a trailing slash where Go's `path.Join` strips it (trailing-slash ignores then behave differently); `packages\foo` is slash-normalized by Go but kept literal here on POSIX; both sides collapse `!!pat` to a positive pattern while ESLint treats `!!` as un-ignore. A shared fixture corpus — or making desugaring Go-only — would pin this down. Relatedly, `applyBasePathToConfig`/`ApplyBasePathOptions` are exported with no consumers and default to `process.cwd()`, contradicting the match-root rule used everywhere else.

swwind (commented):
Verified the earlier review comments against the current head. Most are fully addressed; the table below summarizes, and the remaining gaps are left as inline comments.

| Earlier comment | Status |
|---|---|
| BasePath doc comment misleading (config.go) | ✅ Resolved |
| Reject empty/null basePath via raw map | ✅ Resolved (both rejected, with tests) |
| `--init` migration drops basePath | ✅ Resolved (overrides, ignore entries, global-ignore extraction; 3 tests) |
| Simplify `isAbsolutePattern` via `GetEncodedRootLength` | ✅ Resolved |
| Hardcoded `UseCaseSensitiveFileNames` | ⚠️ Resolved on the Go side (derived from vfs); the TS mirror is still case-sensitive — see inline |
| Glob-escape basePath prefix | ⚠️ Partial — `[ * ? {` escaped on both sides with tests, but leading `!` and picomatch extglobs remain — see inline |
| Outside-root basePath silent no-op | ✅ Resolved (now rejected with a clear error at CLI/API/LSP) |
| Dead win32 regex fallback in base-path.ts | ✅ Resolved |
| `overrideConfig`/rule-tester desugar against `process.cwd()` | ✅ Resolved (instance cwd / config dir passed; no production caller hits the fallback) |
| TS/Go desugar duplication and drift | ⚠️ Partial — trailing-slash, backslash, `!!` drift fixed symmetrically, but no shared fixtures and some edge divergences remain — see inline |

packages/rslint/src/config/base-path.ts
```diff
@@ -0,0 +1,198 @@
...
+ * engine's single-match-root model, so it fails fast instead of silently
+ * contributing nothing (mirrors the Go `ResolveBasePaths` rejection).
+ */
+export function resolveRelativeBasePath(
+  basePath: string,
+  configDirectory: string,
+): string {
+  const root = configDirectory || process.cwd();
+  const absoluteBase = path.resolve(root, basePath);
+  let relative = path.relative(root, absoluteBase).replace(/\\/g, '/');
```
swwind, line 44: The outside-root rejection here uses case-sensitive `path.relative`, while the Go side derives case sensitivity from the filesystem (`base_path.go` uses `fsys.UseCaseSensitiveFileNames()`).

On macOS (case-insensitive APFS), with cwd `/users/me/proj` and a config computing basePath from `__dirname` `/Users/me/proj/packages/app`, `path.relative` returns a `../`-prefixed path, so this throws "resolves outside the config match root" and the whole lint fails — while the Go JSON path accepts the identical config. Windows only agrees by accident because `path.win32.relative` is case-insensitive.

packages/rslint/src/config/base-path.ts
```diff
@@ -0,0 +1,198 @@
...
+
+/**
+ * Escape glob metacharacters in a resolved basePath so the directory is treated
+ * literally when spliced into glob patterns. ESLint treats basePath as a literal
+ * path; without escaping, basePath "packages/[locale]" would be interpreted as a
+ * character class and never match. Character-class escapes ([*], [?], [[] ...)
+ * are understood by both the Go (doublestar) and TS (picomatch) matchers on
+ * every platform.
+ */
+function escapeGlobBasePath(s: string): string {
```
swwind, line 20: `escapeGlobBasePath` (and its Go twin) doesn't neutralize picomatch extglob openers: `@(`, `+(`, `!(`, `(a|b)` are all unhandled (`?(` and `*(` only happen to be neutralized as a side effect of escaping `?`/`*`).

For a literal directory `packages/@(app)`, the escaped basePath is still an extglob to picomatch (which runs with extglobs enabled in `config-hierarchy.ts` and `rslint.ts`), matching files under `packages/app/`, while Go's doublestar treats it literally and matches only `packages/@(app)/`. So TS and Go disagree on which files the entry governs, contradicting the "understood by both matchers" contract in this function's comment.

internal/config/base_path.go
```diff
@@ -0,0 +1,204 @@
...
+// escapeGlobBasePath escapes glob metacharacters in a resolved basePath so the
+// directory is treated literally when spliced into glob patterns. ESLint treats
+// basePath as a literal path; without escaping, a basePath such as
+// "packages/[locale]" would be interpreted as a character class and never match.
+//
+// Escaping uses character classes ([*], [?], [[] ...), which the doublestar and
+// picomatch matchers understand on every platform and which survive the
+// tspath.NormalizePath applied to patterns before matching. A bare ']' or '}'
+// outside a class/brace pair is already literal to the matchers, so only the
+// characters that open a class/brace or introduce a wildcard need escaping. A
```
swwind, line 154: A basePath with a leading `!` is neither escaped nor rejected: `{"basePath": "!vendor", "files": ["src/**"]}` desugars to `files: ["!vendor/src/**"]`, which both matchers parse as a negated pattern — the entry silently inverts instead of scoping to the directory literally named `!vendor`.

The comment here documents it as a known limitation, but the TS mirror neither handles nor documents it. Since no character-class form works, explicitly rejecting a leading `!` in `basePath` on both sides would close the gap.

packages/rslint/tests/config-loader.test.ts
```diff
@@ -353,6 +353,106 @@ describe('normalizeConfig', () => {
...
+
+  test('rejects a basePath resolving outside the config match root', () => {
+    expect(() =>
+      normalizeConfig([{ basePath: '../shared', rules: {} }], {
+        configDirectory: '/repo',
+      }),
+    ).toThrow(/resolves outside the config match root/);
+  });
+
+  test('strips trailing slashes like the Go path.Join', () => {
```
swwind, line 447: The TS and Go desugar test corpora only partially overlap: trailing-slash stripping is tested only here, while `!!` collapse and case-insensitive/drive-letter rebase are tested only in `base_path_test.go`.

There's already unpinned drift observable: for `C:foo`, Go's `GetEncodedRootLength` treats it as rooted and leaves it unprefixed, while `path.win32.isAbsolute('C:foo')` is false on the TS side so it gets rebased. A shared fixture table consumed by both suites would pin these edges against future divergence.

fansenze:
Hi @Penty-d, just checking in – are you still planning to keep working on this?

I apologise for the inconvenience, but the main branch has gone through some significant architectural changes recently, so this PR will now have quite a few conflicts.

If you’re still keen to see it through, I’m happy to help with the merge/rebase, or we can pair on it. But if you don’t have the time right now, I’d be glad to take over and get it across the finish line. Just let me know what works best for you.

Thanks again for your work on this!

Penty-d:
@fansenze Happy to hand this off. Since GitHub doesn’t transfer an existing PR to another contributor’s fork, could you open a replacement PR from your branch? I’ll close this one and link to the new PR.

fansenze:
@Penty-d Thanks! That works for me. Given the significant architectural changes on main, I’ll reimplement basePath against the current architecture in a new PR. I’ll close this PR for now, link the replacement PR here once it’s ready, and add you as a co-author in recognition of your original work. Thanks again for getting this started!

PR #1473, opened 2026-07-30
feat(rstest): add no-identical-title rule

elecmonkey:
## Summary

Add `rstest/no-identical-title` with support for Rstest modifiers, parameterized APIs, imports, and same-file `test.extend()` aliases.

Share the title-tracking implementation with Jest while preserving existing Jest behavior. Register the rule in `rstest.configs.recommended`.

## Related Links

## Validation

- `go test ./internal/plugins/rstest/... ./internal/plugins/jest/rules/no-identical_title ./internal/utils/test_framework/...`
- `go test ./internal/plugins/jest/...`
- `go test ./internal/config/...`
- `go vet ./internal/plugins/rstest/... ./internal/utils/test_framework/... ./internal/plugins/jest/rules/no_identical_title`
- `pnpm --filter @rslint/test-tools exec rstest run tests/cli/js-config/presets.test.ts`
- `pnpm exec prettier --check packages/rslint/src/config/presets/rstest.ts packages/rslint-test-tools/tests/cli/js-config/presets.test.ts internal/plugins/rstest/rules/no_identical_title/no_identical_title.md`
- `git diff --check`

## Checklist

### DIFF
--- a/internal/plugins/rstest/utils/parse_rstest_fn.go
+++ b/internal/plugins/rstest/utils/parse_rstest_fn.go
@@ -1,8 +1,6 @@
 package utils
 
 import (
-	"slices"
-
 	"github.com/microsoft/typescript-go/shim/ast"
 	"github.com/web-infra-dev/rslint/internal/rule"
 	internalUtils "github.com/web-infra-dev/rslint/internal/utils"
@@ -40,11 +38,6 @@ type rstestResolvedAPI struct {
 	state        rstestAPIState
 	originalNode *ast.Node
 	mode         RstestImportMode
-}
-
-func IsTypeOfRstestFnCall(node *ast.Node, ctx rule.RuleContext, kinds ...RstestFnType) bool {
-	parsed := ParseRstestFnCall(node, ctx)
-	return parsed != nil && len(kinds) > 0 && slices.Contains(kinds, parsed.Kind)
 }
 
 // ParseRstestFnCall parses a final Rstest test/describe registration call.
--- a/internal/plugins/rstest/utils/parse_rstest_fn.go
+++ b/internal/plugins/rstest/utils/parse_rstest_fn.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/microsoft/typescript-go/shim/ast"
 	"github.com/web-infra-dev/rslint/internal/rule"
+	internalUtils "github.com/web-infra-dev/rslint/internal/utils"
 	testFramework "github.com/web-infra-dev/rslint/internal/utils/test_framework"
 )
 
@@ -153,8 +154,8 @@ func parseRstestChain(node *ast.Node) (*ast.Node, []rstestChainPart, bool, bool)
 			return nil, nil, false, false
 		}
 		nameNode := ast.SkipParentheses(element.ArgumentExpression)
-		name := staticRstestMemberName(nameNode)
-		if name == "" {
+		name, ok := internalUtils.GetStaticStringLiteralValue(nameNode)
+		if !ok || name == "" {
 			return nil, nil, false, false
 		}
 		return root, append(parts, rstestChainPart{
@@ -188,20 +189,6 @@ func parseRstestChain(node *ast.Node) (*ast.Node, []rstestChainPart, bool, bool)
 		return root, parts, rootInvoked, true
 	default:
 		return nil, nil, false, false
-	}
-}
-
-func staticRstestMemberName(node *ast.Node) string {
-	if node == nil {
-		return ""
-	}
-	switch node.Kind {
-	case ast.KindStringLiteral:
-		return node.AsStringLiteral().Text
-	case ast.KindNoSubstitutionTemplateLiteral:
-		return node.AsNoSubstitutionTemplateLiteral().Text
-	default:
-		return ""
 	}
 }
 
--- a/internal/plugins/rstest/utils/parse_rstest_fn.go
+++ b/internal/plugins/rstest/utils/parse_rstest_fn.go
@@ -1,0 +1,358 @@
+package utils
+
+import (
+	"slices"
+
+	"github.com/microsoft/typescript-go/shim/ast"
+	"github.com/web-infra-dev/rslint/internal/rule"
+	testFramework "github.com/web-infra-dev/rslint/internal/utils/test_framework"
+)
+
+type rstestAPIState uint8
+
+const (
+	rstestAPIInvalid rstestAPIState = iota
+	rstestAPITest
+	rstestAPITestWithExtend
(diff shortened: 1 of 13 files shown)


elecmonkey:
@codex review

elecmonkey:
@codex review

internal/plugins/rstest/utils/parse_rstest_fn.go
```diff
@@ -0,0 +1,358 @@
...
+		case "extend":
+			if state == rstestAPITestWithExtend && part.invocation == rstestCallInvoked {
+				return rstestAPITestWithExtend
+			}
+		}
+	case rstestAPIDescribe:
+		switch part.name {
+		case "only", "skip", "todo", "concurrent", "sequential":
+			if part.invocation == rstestNotInvoked {
+				return rstestAPIDescribe
```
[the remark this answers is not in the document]
elecmonkey: This chain is not supported by Rstest: `describe.each(rows)` returns `DescribeCall`, which has no `.only` (TS2339), and the runtime returns a plain function without modifier properties. The supported order is `describe.only.each(rows)(...)`, which the parser already handles.

internal/plugins/rstest/utils/parse_rstest_fn.go
```diff
@@ -0,0 +1,337 @@
...
+			if part.invocation == rstestCallInvoked {
+				return rstestAPIDescribe
+			}
+		case "each", "for":
+			if part.invocation == rstestCallInvoked || part.invocation == rstestTagInvoked {
+				return rstestAPIParameterizedDescribe
+			}
+		}
+	}
+	return rstestAPIInvalid
```
swwind, line 336: **`applyRstestChainPart` has no case for already-parameterized states**

The switch only handles `rstestAPITest`/`rstestAPITestWithExtend`/`rstestAPIDescribe`. Once a chain reaches `rstestAPIParameterizedTest` or `rstestAPIParameterizedDescribe` (via `.each`/`.for`), any further chain part (e.g. a following `.only`) has no matching case and falls through to `return rstestAPIInvalid`, discarding the whole call instead of keeping it parameterized/ignored.

Concretely: `describe.each([1]).only("g", () => { test("dup") })` never gets a title layer pushed (since `ParseRstestFnCall` returns nil), so its nested `test("dup")` title is compared against the outer scope instead of being scoped independently — a false positive. The test suite also encodes `test.only.extend({})("same"); test.only.extend({}).only("same");` as `valid`, which is a false negative caused by the same gap, not actually correct behavior.

Suggest adding cases for `rstestAPIParameterizedTest`/`rstestAPIParameterizedDescribe` that pass through modifiers like `.only`/`.skip`/etc. without invalidating the parameterized state.

internal/plugins/rstest/utils/parse_rstest_fn.go
```diff
@@ -0,0 +1,337 @@
...
+			if state == rstestAPIInvalid {
+				break
+			}
+		}
+		if state == rstestAPIInvalid {
+			continue
+		}
+		resolved.state = state
+		return resolved, 0, true
+	}
```
swwind, line 274: **Const-alias tracking only exists for rstest, not jest**

This alias-following (`const appTest = test.extend({})`) is implemented only in rstest's parser. Since the PR shares the title-tracking rule with Jest, a Jest file doing the same `const appTest = test.extend({}); appTest("same"); appTest("same");` won't be flagged, while the rstest equivalent will — worth a note (or a follow-up) so the two consumers of the shared rule don't silently diverge in capability.

swwind (commented):
Found 1 correctness bug and 2 lower-severity design notes. Details inline.

| Case | Expected | This PR |
|---|---|---|
| `describe.each([1]).only("g", () => { test("dup") })` with a sibling top-level `test("dup")` | independent describe scope, no report | falsely reports duplicate — chain parse returns nil, so no title layer is pushed for the describe body and the nested `test` is compared against the outer file-level layer |
| `test.only.extend({})("same", ...); test.only.extend({}).only("same", ...);` | duplicate-title report | not reported — chain parse returns nil after `.only` following `.extend` (this exact case is encoded as `valid` in the PR's own test fixtures) |

internal/plugins/rstest/utils/parse_rstest_fn.go
```diff
@@ -0,0 +1,337 @@
...
+		root, parts, rootInvoked, ok := parseRstestChain(property.Expression)
+		if !ok {
+			return nil, nil, false, false
+		}
+		nameNode := property.Name()
+		if nameNode == nil || nameNode.Kind != ast.KindIdentifier {
+			return nil, nil, false, false
+		}
+		return root, append(parts, rstestChainPart{
+			name: nameNode.AsIdentifier().Text,
```
swwind, line 140: **`parseRstestChain` duplicates `member_chain.go`'s tree walk**

This re-walks the same Identifier/PropertyAccess/ElementAccess/CallExpression/TaggedTemplateExpression node kinds that `internal/utils/test_framework/member_chain.go`'s `GetMemberEntries` already covers. Since jest uses the shared helper, a future fix to chain-walking (e.g. optional chaining) applied only there would leave rstest out of sync. Might be worth extending the shared helper with the extra invocation-kind tracking rstest needs instead of maintaining a parallel walker.

internal/plugins/rstest/utils/parse_rstest_fn.go
elecmonkey: not problem

internal/plugins/rstest/utils/parse_rstest_fn.go
elecmonkey: repeat with the former review.

elecmonkey:
> Found 1 correctness bug and 2 lower-severity design notes. Details inline.
> 
> Case	Expected	This PR
> `describe.each([1]).only("g", () => { test("dup") })` with a sibling top-level `test("dup")`	independent describe scope, no report	falsely reports duplicate — chain parse returns nil, so no title layer is pushed for the describe body and the nested `test` is compared against the outer file-level layer
> `test.only.extend({})("same", ...); test.only.extend({}).only("same", ...);`	duplicate-title report	not reported — chain parse returns nil after `.only` following `.extend` (this exact case is encoded as `valid` in the PR's own test fixtures)

Thanks for your review!

The first point has already been disproven earlier in the discussion. The second point is not an issue. Regarding the third point, although the two code blocks look similar, they aren’t easy to abstract into shared logic without adding unnecessary complexity.

swwind:
> Thanks for your review!

You are welcome.

swwind (approved):

fansenze (approved):

PR #1475, opened 2026-07-30
feat(rstest): add no-disabled-tests rule

elecmonkey:
## Summary

Add `rstest/no-disabled-tests` with support for skipped, incomplete, parameterized, imported, and same-file custom test APIs.

Share the framework-neutral checks with Jest while preserving existing Jest behavior. Register the rule as a warning in `rstest.configs.recommended`.

## Related Links

## Validation

- `go test ./internal/plugins/rstest/... ./internal/utils/test_framework/... ./internal/plugins/jest/rules/no_disabled_tests ./internal/config`
- `go test ./internal/plugins/jest/...`
- `go vet ./internal/plugins/rstest/... ./internal/utils/test_framework/... ./internal/plugins/jest/rules/no_disabled_tests`
- `pnpm -C packages/rslint build:js`
- `pnpm --filter @rslint/test-tools exec rstest run tests/cli/js-config/presets.test.ts`
- `git diff --check`

## Checklist

- [x] Tests updated.
- [x] Documentation updated.

### DIFF
--- a/internal/plugins/rstest/rules/no_disabled_tests/no_disabled_tests_test.go
+++ b/internal/plugins/rstest/rules/no_disabled_tests/no_disabled_tests_test.go
@@ -27,7 +27,10 @@ func TestNoDisabledTestsRule(t *testing.T) {
 			{Code: `test("case", context => { context.skip(); });`},
 			{Code: `test("case", ctx => { ctx.skip(); }); request.skip();`},
 			{Code: `describe("empty suite"); describe.concurrent("empty suite");`},
+			// an identifier is not necessarily an options object
 			{Code: `test("case", options);`},
+			{Code: `test("case", { timeout: 100 }, () => {});`},
+			{Code: `test.todo("case", { timeout: 100 });`},
 			{Code: `fit("case", () => {}); xit("case", () => {}); xtest("case", () => {}); fdescribe("suite", () => {}); xdescribe("suite", () => {});`},
 			{Code: `pending();`},
 			{Code: `import { pending } from "actions"; pending();`},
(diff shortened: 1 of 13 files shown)


elecmonkey:
@codex review

internal/plugins/rstest/rules/no_disabled_tests/no_disabled_tests_test.go
```diff
@@ -0,0 +1,218 @@
...
+			{Code: `test.todo("case"); it.todo("case"); describe.todo("suite");`},
+			{Code: `test.todo.each([1])("case"); test.todo.for([1])("case");`},
+			{Code: `const todoTest = test.todo; todoTest("case");`},
+			{Code: `test.skipIf(condition)("case", () => {}); test.runIf(condition)("case", () => {});`},
+			{Code: `describe.skipIf(condition)("suite", () => {}); describe.runIf(condition)("suite", () => {});`},
+			{Code: `test.skipIf(true)("case", () => {}); test.runIf(false)("case", () => {});`},
+			{Code: `test("case", context => { context.skip(); });`},
+			{Code: `test("case", ctx => { ctx.skip(); }); request.skip();`},
+			{Code: `describe("empty suite"); describe.concurrent("empty suite");`},
+			{Code: `test("case", options);`},
```
swwind, line 31: `TestCall` in `@rstest/core` has a `(description, options: TestOptions, fn?)` overload, so `test("case", { timeout: 100 })` is a two-argument registration with no callback — precisely the incomplete test this rule exists to catch. Jest has no such overload, which is why inheriting its `len(args) < 2` check is not sufficient here.

Please handle at least the object-literal form: report `missingFunction` when the second argument is an object literal and there is no third argument. The identifier form on this line can stay valid.

internal/plugins/rstest/utils/rstest.go
```diff
@@ -16,6 +16,8 @@ type RstestImportMode = testFramework.ReferenceMode
 type ParsedRstestFnCall struct {
 	testFramework.ParsedCall
 	Parameterized bool
+	HasSkip       bool
```
swwind: These flags are needed only because members consumed while resolving a const alias (`const skipped = test.skip`) never reach `Members`. `no-focused-tests` is next on the Rstest list and needs the same for `.only`, so this will grow a third bool, and each new bool has to be threaded through `rstestResolvedAPI`, `applyResolvedRstestChainPart`, and every rule's adapter.

Please append the alias-chain members to `ParsedCall.Members` (or add a `ResolvedMembers []string`) instead. Rules then keep using `slices.Contains(Members, "skip")`, which is how the Jest side already reads, and future modifiers need no struct changes.

swwind (commented):
Two things I'd want addressed before this merges. Everything else I looked at is fine.

1. `test("case", { timeout: 100 })` is not reported — a false negative specific to Rstest, details inline.
2. `HasSkip` / `HasTodo` on `ParsedRstestFnCall` won't scale to the next rules on the Rstest list, and it's cheaper to fix here than after consumers exist.

internal/plugins/rstest/rules/no_disabled_tests/no_disabled_tests_test.go
elecmonkey: Fixed in a3a2a985. The shared rule now takes a `HasOptionsOverload` flag; with it on, `missingFunction` is reported when the second argument is an object literal and there is no third argument. `test("case", options)` stays valid as suggested. Added Go + JS cases for the reported and valid forms and updated the doc's limitations section.

internal/plugins/rstest/utils/rstest.go
elecmonkey: Fixed in f129ccaf. Both bools are gone: the resolver now accumulates every applied chain member (including ones consumed through a const alias) and they land in `ParsedCall.Members`, so the rule reads `slices.Contains(parsed.Members, "skip")` like the Jest adapter. `MemberEntries` still holds only call-site members, since alias members have no node in the call expression — noted in a comment.

swwind (approved):
LGTM

fansenze (approved):

PR #1479, opened 2026-07-31
feat(unicorn): port rule prefer-set-has

elecmonkey:
## Summary

Port the `prefer-set-has` rule from eslint-plugin-unicorn to rslint.

Prefer `Set#has()` over `Array#includes()` when a `const` array binding is used
only for existence checks. When the binding has a rewritable TypeScript type
annotation (`T[]`, `readonly T[]`, `Array<T>`, `ReadonlyArray<T>`), the fix is
applied automatically as `new Set(...)` + `Set<T>` / `ReadonlySet<T>`;
otherwise the same edit is offered as a manual suggestion.

## Related Links

- ESLint rule: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-set-has.md
- Source code: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/rules/prefer-set-has.js

## Checklist

### DIFF
--- a/internal/plugins/unicorn/all.go
+++ b/internal/plugins/unicorn/all.go
@@ -11,6 +11,7 @@ import (
 	"github.com/web-infra-dev/rslint/internal/plugins/unicorn/rules/prefer_array_flat_map"
 	"github.com/web-infra-dev/rslint/internal/plugins/unicorn/rules/prefer_node_protocol"
 	"github.com/web-infra-dev/rslint/internal/plugins/unicorn/rules/prefer_number_properties"
+	"github.com/web-infra-dev/rslint/internal/plugins/unicorn/rules/prefer_set_has"
 	"github.com/web-infra-dev/rslint/internal/plugins/unicorn/rules/require_array_join_separator"
 	"github.com/web-infra-dev/rslint/internal/plugins/unicorn/rules/require_number_to_fixed_digits_argument"
 	"github.com/web-infra-dev/rslint/internal/rule"
@@ -28,6 +29,7 @@ func GetAllRules() []rule.Rule {
 		prefer_array_flat_map.PreferArrayFlatMapRule,
 		prefer_node_protocol.PreferNodeProtocolRule,
 		prefer_number_properties.PreferNumberPropertiesRule,
+		prefer_set_has.PreferSetHasRule,
 		require_array_join_separator.RequireArrayJoinSeparatorRule,
 		require_number_to_fixed_digits_argument.RequireNumberToFixedDigitsArgumentRule,
 	}
--- a/internal/plugins/unicorn/rules/prefer_set_has/prefer_set_has.go
+++ b/internal/plugins/unicorn/rules/prefer_set_has/prefer_set_has.go
@@ -1,0 +1,543 @@
+// Package prefer_set_has ports eslint-plugin-unicorn's `prefer-set-has` rule.
+//
+// It flags a `const` array binding that is used only for existence checks via
+// `Array#includes()` and recommends declaring it as a `Set` so `Set#has()` can
+// be used instead. When the binding has a rewritable TypeScript type annotation
+// the fix is applied automatically; otherwise the same edit is offered as a
+// manual suggestion.
+package prefer_set_has
+
+import (
+	_ "embed"
+	"fmt"
+
+	"github.com/microsoft/typescript-go/shim/ast"
+	"github.com/web-infra-dev/rslint/internal/plugins/unicorn/unicornutil"
+	"github.com/web-infra-dev/rslint/internal/rule"
+	"github.com/web-infra-dev/rslint/internal/utils"
+)
+
+const (
+	messageIDError      = "error"
+	messageIDSuggestion = "suggestion"
+)
+
+// maxArrayLength mirrors upstream's MAX_ARRAY_LENGTH = 2 ** 32 - 1.
+const maxArrayLength = float64(1<<32) - 1
+
+//go:embed prefer_set_has.schema.json
+var schemaJSON []byte
+
+// methodsReturnsArray lists methods whose result is always an array (never a
+// string), mirroring upstream's `methodsReturnsArray`.
+var methodsReturnsArray = map[string]bool{
+	// `Array`
+	"copyWithin": true,
+	"fill":       true,
+	"filter":     true,
+	"flat":       true,
+	"flatMap":    true,
+	"map":        true,
+	"reverse":    true,
+	"sort":       true,
+	"splice":     true,
+	"toReversed": true,
+	"toSorted":   true,
+	"toSpliced":  true,
+	"with":       true,
+	// `String`
+	"split": true,
+	// `Iterator`
+	"toArray": true,
+}
+
+// methodsReturnsArrayAndString lists methods that can return an array OR a
+// string, mirroring upstream's `methodsReturnsArrayAndString`.
+var methodsReturnsArrayAndString = map[string]bool{
+	"slice":  true,
+	"concat": true,
+}
+
+type options struct {
+	minimumItems int
+}
+
+func parseOptions(rawOptions []any) options {
+	opts := options{minimumItems: 0}
+	optsMap := utils.GetOptionsMap(rawOptions)
+	if optsMap == nil {
+		return opts
+	}
+	if value, ok := optsMap["minimumItems"]; ok {
+		if number, ok := utils.CoerceIntegral(value); ok {
+			opts.minimumItems = number
+		}
+	}
+	return opts
+}
+
+func messageError(name string) rule.RuleMessage {
+	return rule.RuleMessage{
+		Id:          messageIDError,
+		Description: fmt.Sprintf("`%s` should be a `Set`, and use `%s.has()` to check existence or non-existence.", name, name),
+		Data:        map[string]string{"name": name},
+	}
+}
+
+func messageSuggestion(name string) rule.RuleMessage {
+	return rule.RuleMessage{
+		Id:          messageIDSuggestion,
+		Description: fmt.Sprintf("Switch `%s` to `Set`.", name),
+		Data:        map[string]string{"name": name},
+	}
+}
+
+var PreferSetHasRule = rule.Rule{
+	Name:   "unicorn/prefer-set-has",
+	Schema: rule.NewSchema(schemaJSON),
+	Run: func(ctx rule.RuleContext, rawOptions []any) rule.RuleListeners {
+		opts := parseOptions(rawOptions)
+
+		return rule.RuleListeners{
+			ast.KindIdentifier: func(node *ast.Node) {
+				checkIdentifier(ctx, opts, node)
+			},
+		}
+	},
+}
+
+// checkIdentifier mirrors the body of upstream's `context.on('Identifier', …)`.
+func checkIdentifier(ctx rule.RuleContext, opts options, node *ast.Node) {
+	// ctx.Refs is the stand-in for ESLint's scope manager; without it we
+	// cannot collect references, so degrade gracefully (JS-only runs).
+	if ctx.Refs == nil {
+		return
+	}
+
+	declaration := arrayVariableDeclarator(ctx, node)
+	if declaration == nil {
+		return
+	}
+	initializer := declaration.Initializer
+
+	if opts.minimumItems > 0 {
+		arraySize, ok := getKnownArraySize(ctx, initializer)
+		if !ok || arraySize < opts.minimumItems {
+			return
+		}
+	}
+
+	// The binder attaches the symbol to the VariableDeclaration node (node's
+	// parent), which is the currency of ctx.Refs.
+	symbol := node.Parent.Symbol()
+	if symbol == nil {
+		return
+	}
+
+	// Upstream collects references via getVariableIdentifiers, which includes
+	// every declaration name of the variable. When a `var` is declared more
+	// than once (`var foo = ...; var foo = ...`), the other declaration's name
+	// is an identifier that classifies as neither includes/length/extra, so
+	// getReferenceGroups bails. ctx.Refs.References excludes declaration names,
+	// so reproduce that bail explicitly: a variable with more than one
+	// declaration is not a single `const` array binding.
+	if len(symbol.Declarations) != 1 {
+		return
+	}
+
+	references := ctx.Refs.References(symbol)
+	identifiers := make([]*ast.Node, 0, len(references))
+	for _, reference := range references {
+		if reference != node {
+			identifiers = append(identifiers, reference)
+		}
+	}
+
+	groups, ok := getReferenceGroups(identifiers)
+	if !ok || len(groups.includes) == 0 {
+		return
+	}
+
+	if len(groups.extra) > 0 && !isKnownUniqueArrayExpression(ctx, initializer) {
+		return
+	}
+
+	if len(groups.includes) == 1 && !isMultipleCall(groups.includes[0], node) {
+		return
+	}
+
+	typeText, hasRewritableType := setTypeAnnotationText(ctx, declaration)
+
+	buildFixes := func() []rule.RuleFix {
+		return buildSetFixes(ctx, node, declaration, groups, typeText)
+	}
+
+	// A type annotation that can't be safely rewritten downgrades the fix to a
+	// manual suggestion, matching upstream.
+	if declaration.Type != nil && !hasRewritableType {
+		ctx.ReportNodeWithDeferredSuggestions(node, messageError(node.AsIdentifier().Text), func() []rule.RuleSuggestion {
+			return []rule.RuleSuggestion{{
+				Message:  messageSuggestion(node.AsIdentifier().Text),
+				FixesArr: buildFixes(),
+			}}
+		})
+		return
+	}
+
+	ctx.ReportNodeWithDeferredFixes(node, messageError(node.AsIdentifier().Text), buildFixes)
+}
+
+// buildSetFixes assembles the edits shared by the autofix and suggestion paths:
+// rewrite the type annotation (when possible), wrap the initializer in
+// `new Set(...)`, and rename `.includes`/`.length` to `.has`/`.size`.
+func buildSetFixes(
+	ctx rule.RuleContext,
+	node *ast.Node,
+	declaration *ast.VariableDeclaration,
+	groups referenceGroups,
+	typeAnnotationText string,
+) []rule.RuleFix {
+	fixes := make([]rule.RuleFix, 0, 3+len(groups.includes)+len(groups.lengths))
+
+	if typeAnnotationText != "" && declaration.Type != nil {
+		fixes = append(fixes, rule.RuleFixReplace(ctx.SourceFile, declaration.Type, typeAnnotationText))
+	}
+
+	initializer := declaration.Initializer
+	fixes = append(fixes, rule.RuleFixInsertBefore(ctx.SourceFile, initializer, "new Set("))
+	fixes = append(fixes, rule.RuleFixInsertAfter(initializer, ")"))
+
+	for _, identifier := range groups.includes {
+		property := propertyOfMemberAccess(outerParen(identifier).Parent)
+		if property != nil {
+			fixes = append(fixes, rule.RuleFixReplace(ctx.SourceFile, property, "has"))
+		}
+	}
+
+	for _, identifier := range groups.lengths {
+		property := propertyOfMemberAccess(outerParen(identifier).Parent)
+		if property != nil {
+			fixes = append(fixes, rule.RuleFixReplace(ctx.SourceFile, property, "size"))
+		}
+	}
+
+	return fixes
+}
+
+// propertyOfMemberAccess returns the `.name` node of a non-computed property
+// access, or nil.
+func propertyOfMemberAccess(node *ast.Node) *ast.Node {
+	if node == nil || !ast.IsPropertyAccessExpression(node) {
+		return nil
+	}
+	return node.AsPropertyAccessExpression().Name()
+}
+
+// referenceGroups classifies the references to the candidate binding.
+type referenceGroups struct {
+	includes []*ast.Node
+	lengths  []*ast.Node
+	extra    []*ast.Node
+}
+
+// getReferenceGroups mirrors upstream's `getReferenceGroups`. It returns
+// ok=false when any reference cannot be classified (upstream returns
+// `undefined`, which makes the caller bail).
+func getReferenceGroups(identifiers []*ast.Node) (referenceGroups, bool) {
+	var groups referenceGroups
+	for _, identifier := range identifiers {
+		switch {
+		case isIncludesCall(identifier):
+			groups.includes = append(groups.includes, identifier)
+		case isLengthRead(identifier):
+			groups.lengths = append(groups.lengths, identifier)
+			groups.extra = append(groups.extra, identifier)
+		case isAllowedExtraReference(identifier):
+			groups.extra = append(groups.extra, identifier)
+		default:
+			return referenceGroups{}, false
+		}
+	}
+	return groups, true
+}
+
+// isIncludesCall reports whether `identifier` is the object of a
+// `identifier.includes(x)` call with exactly one non-spread argument, mirroring
+// upstream's `isIncludesCall` (isMethodCall defaults: non-optional, no spread).
+func isIncludesCall(node *ast.Node) bool {
+	member := outerParen(node).Parent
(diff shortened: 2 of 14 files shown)


elecmonkey:
@codex review

swwind (commented):
Ported `prefer-set-has` closely tracks upstream, but two edge cases diverge from real `eslint-plugin-unicorn` (v72) behavior — verified in eslint-playground — where the rule silently produces no report/fix instead of the one upstream gives:

| Code | ESLint | This PR |
|---|---|---|
| `const foo=[1,2,3]; foo.includes(1); const n=(foo).length;` | reports, autofixes 3rd line to `.size` | no report — `(foo).length`'s parenthesized identifier is unclassifiable, so the whole rule bails |
| `const foo=[NaN,1,2]; foo.forEach(x=>log(x)); for(let i=0;i<3;i++) foo.includes(1);` | reports, autofixes to `new Set([NaN,1,2])`/`.has` | no report — `NaN` fails static evaluation, so the array is conservatively treated as "not known unique" |

Also a few spots reimplementing logic that already exists elsewhere in the codebase — worth reusing instead, so a future fix doesn't need to be applied twice. Details inline.

internal/plugins/unicorn/rules/prefer_set_has/prefer_set_has.go
```diff
@@ -0,0 +1,543 @@
...
+	current := node
+	for i := 0; i < depth && current != nil; i++ {
+		current = current.Parent
+	}
+	return current
+}
+
+// isLengthRead mirrors upstream's `isLengthRead`: a non-computed `.length`
+// member read that is neither an assignment target nor a call callee.
+func isLengthRead(identifier *ast.Node) bool {
```
swwind, line 367: **`isLengthRead` doesn't unwrap parentheses.** It reads `identifier.Parent` directly instead of unwrapping via `outerParen` first, unlike `isIncludesCall` a few lines up which does call `ast.SkipParentheses`. Same gap in `isIterableUse`/`isArrayOrArgumentSpread`/`isAllowedForEachCall`.

Fix: unwrap through `outerParen(identifier)` before checking `.Parent`, consistent with the other classifiers in this file.

internal/plugins/unicorn/rules/prefer_set_has/static_value.go
```diff
@@ -0,0 +1,581 @@
...
+		return staticValue{kind: svNumber, number: left.number / right.number}, true
+	case ast.KindPercentToken:
+		return staticValue{kind: svNumber, number: math.Mod(left.number, right.number)}, true
+	case ast.KindAsteriskAsteriskToken:
+		return staticValue{kind: svNumber, number: math.Pow(left.number, right.number)}, true
+	}
+	return staticValue{}, false
+}
+
+func evaluateStaticIdentifier(ctx rule.RuleContext, node *ast.Node, visited map[*ast.Symbol]bool) (staticValue, bool) {
```
swwind, line 766: **`evaluateStaticIdentifier` only special-cases `Infinity`.** `NaN` and `undefined` aren't modeled, so any array containing them fails static evaluation and the rule conservatively skips the autofix.

Fix: add `NaN`/`undefined` alongside the existing `Infinity` check (same shadow-check pattern via `utils.IsShadowed`).

internal/plugins/unicorn/rules/prefer_set_has/prefer_set_has.go
```diff
@@ -0,0 +1,543 @@
...
+	}
+	// `node.parent.parent.callee.object === node` — the receiver must be the
+	// identifier itself (parentheses are transparent in ESTree).
+	return ast.SkipParentheses(match.Object) == node
+}
+
+// outerParen returns node walked up through any enclosing
+// ParenthesizedExpression nodes — the ESTree view in which parentheses are
+// invisible.
+func outerParen(node *ast.Node) *ast.Node {
```
swwind: **`outerParen` duplicates an existing helper.** `typescript-go/internal/ast/utilities.go:845` already exports `ast.WalkUpParenthesizedExpressions` for this. Using it here (and at the call sites from the `isLengthRead` comment) would fix that gap and drop the local duplicate.

internal/plugins/unicorn/rules/prefer_set_has/static_value.go
```diff
@@ -0,0 +1,581 @@
...
+			continue
+		}
+		if comment.Pos() < childRange.Pos() || comment.End() > childRange.End() {
+			return true
+		}
+	}
+	return false
+}
+
+// --- static value model -------------------------------------------------
```
swwind, line 405: **Second, independently-maintained constant-folding evaluator.** `internal/utils/static_string_evaluator.go`'s `StaticStringEvaluator` already does literal folding + const-identifier resolution via `ctx.Refs` and is used by `no_unnecessary_condition`/`prefer_find`. Worth checking whether extending that one with numeric/bigint support is feasible instead of maintaining two evaluators with diverging semantics (see the `NaN`/`Infinity` comment below).

internal/plugins/unicorn/rules/prefer_set_has/prefer_set_has.go
```diff
@@ -0,0 +1,543 @@
...
+		if parent.AsDeleteExpression().Expression == target {
+			return true
+		}
+	}
+	return isForInOrForOfTarget(target)
+}
+
+// isTypeScriptExpressionWrapper reports whether `parent` is a TS `as` / `<T>` /
+// `!` wrapper whose expression is `child`.
+func isTypeScriptExpressionWrapper(parent *ast.Node, child *ast.Node) bool {
```
swwind, line 423: **`isTypeScriptExpressionWrapper` is narrower than the shared helper.** It hand-rolls a switch over `AsExpression`/`TypeAssertionExpression`/`NonNullExpression`; `ast.SkipOuterExpressions(node, ast.OEKAssertions)` — already called elsewhere in this file (`static_value.go:429`) — also covers `SatisfiesExpression`. As written, `evaluateStaticValueInner` treats `x satisfies T` as transparent but `isAssignmentTarget` (via this function) doesn't, so a `satisfies`-wrapped write target is classified inconsistently with an `as`-wrapped one in the same file.

internal/plugins/unicorn/rules/prefer_set_has/prefer_set_has.go
```diff
@@ -0,0 +1,543 @@
...
+// invisible.
+func outerParen(node *ast.Node) *ast.Node {
+	current := node
+	for current.Parent != nil && current.Parent.Kind == ast.KindParenthesizedExpression {
+		current = current.Parent
+	}
+	return current
+}
+
+var multipleCallNodeKinds = map[ast.Kind]bool{
```
swwind: **`multipleCallNodeKinds` duplicates `IsFunctionLikeContainer`.** The function-boundary kinds (`FunctionDeclaration`/`Expression`/`ArrowFunction`/`MethodDeclaration`/`GetAccessor`/`SetAccessor`/`Constructor`), with the comment explaining tsgo's non-wrapping of methods, duplicate `utils.IsFunctionLikeContainer` (`internal/utils/flow.go:128`) exactly.

internal/plugins/unicorn/rules/prefer_set_has/static_value.go
```diff
@@ -0,0 +1,581 @@
...
+func elementTypeText(ctx rule.RuleContext, elementType *ast.Node) string {
+	for elementType != nil && elementType.Kind == ast.KindParenthesizedType {
+		elementType = elementType.AsParenthesizedTypeNode().Type
+	}
+	return utils.TrimmedNodeText(ctx.SourceFile, elementType)
+}
+
+// typeArgumentsText returns the source text of a type-argument list including
+// its angle brackets (`<T>`), mirroring upstream's `sourceCode.getText(typeArguments)`.
+func typeArgumentsText(ctx rule.RuleContext, typeArguments *ast.NodeList) string {
```
swwind: **`typeArgumentsText` duplicates an existing bracket-recovery scan.** The backward/forward `<`/`>` scan is the identical pattern already used in `internal/plugins/typescript/rules/method_signature_style/method_signature_style.go:138` (for type-parameter lists). Could be factored into a shared `internal/utils` helper.

internal/plugins/unicorn/all.go
```diff
@@ -26,6 +27,7 @@ func GetAllRules() []rule.Rule {
 		prefer_array_flat.PreferArrayFlatRule,
 		prefer_array_flat_map.PreferArrayFlatMapRule,
 		prefer_number_properties.PreferNumberPropertiesRule,
+		prefer_set_has.PreferSetHasRule,
```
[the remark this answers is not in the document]
elecmonkey: Fixed in commit 9194f541 — the preset entry is now enabled as `'error'`.

internal/plugins/unicorn/rules/prefer_set_has/prefer_set_has.go
elecmonkey: Fixed in 8ed80b99. All four classifiers (`isLengthRead`, `isIterableUse`, `isArrayOrArgumentSpread`, `isAllowedForEachCall`) now unwrap parentheses via `ast.WalkUpParenthesizedExpressions`, consistent with `isIncludesCall`. Added lock-in tests for `(foo).length`, `for-of (foo)`, `[...(foo)]`, and `(foo).forEach`.

internal/plugins/unicorn/rules/prefer_set_has/static_value.go
elecmonkey: Fixed in 8ed80b99. `evaluateStaticIdentifier` now models `NaN` (as `svNumber` + `math.NaN()`) and `undefined` (new `svUndefined` kind), both with `IsShadowed` guards. `isComparable`, `sameValueZero`, and `truthy` all handle the new kind. Tests confirm arrays containing NaN/undefined are now recognized as known-unique and report+autofix correctly.

internal/plugins/unicorn/rules/prefer_set_has/prefer_set_has.go
elecmonkey: Fixed in 8ed80b99. Deleted the local `outerParen` helper; all 8 call sites now use `ast.WalkUpParenthesizedExpressions` from the tsgo shim.

internal/plugins/unicorn/rules/prefer_set_has/static_value.go
elecmonkey: Acknowledged. I've opted not to merge the two evaluators in this PR — the static-value model here encodes rule-specific semantics (NaN/-0/SameValueZero/array-size) that don't map cleanly onto the string evaluator's scope. Leaving this as a potential follow-up.

internal/plugins/unicorn/rules/prefer_set_has/prefer_set_has.go
elecmonkey: Fixed in 8ed80b99. Replaced the hand-rolled switch with `ast.IsOuterExpression(parent, ast.OEKAssertions)` + `parent.Expression() == child`, which automatically covers `SatisfiesExpression` and stays in sync with the downward `SkipOuterExpressions` call in the same file.

internal/plugins/unicorn/rules/prefer_set_has/prefer_set_has.go
elecmonkey: Fixed in 8ed80b99. Deleted the `multipleCallNodeKinds` map; `isMultipleCall` now uses `utils.IsFunctionLikeContainer` for the function-boundary check (loop kinds stay as a local switch).

internal/plugins/unicorn/rules/prefer_set_has/static_value.go
elecmonkey: Fixed in 8ed80b99. Extracted the bracket-recovery scan into `utils.SliceEnclosingDelimiters(text, start, end, open, close)` with a unit test. Both `typeArgumentsText` in this rule and `method_signature_style.go` now use the shared helper; the old `safeSlice` method was removed.

internal/plugins/unicorn/rules/prefer_set_has/static_value.go
```diff
@@ -0,0 +1,588 @@
...
+func evaluateStaticBinary(ctx rule.RuleContext, binary *ast.BinaryExpression, visited map[*ast.Symbol]bool) (staticValue, bool) {
+	if binary.OperatorToken == nil {
+		return staticValue{}, false
+	}
+	left, lok := evaluateStaticValueInner(ctx, binary.Left, visited)
+	right, rok := evaluateStaticValueInner(ctx, binary.Right, visited)
+	if !lok || !rok {
+		return staticValue{}, false
+	}
+	if left.kind != svNumber || right.kind != svNumber {
```
swwind, line 644: **Static evaluation is number-only, so foldable non-numeric elements bail.** `evaluateStaticBinary` requires both operands to be `svNumber`, and the `evaluateStaticValueInner` switch has no `TemplateExpression` / `PropertyAccessExpression` arm. Upstream uses eslint-utils' `getStaticValue`, which folds string concatenation, template literals with static substitutions, and well-known member constants.

With an extra reference present (so `isKnownUniqueArrayExpression` is consulted), all three of these report and autofix upstream but silently produce nothing here:

```js
const foo = ['a' + 'b', 'c'];  // 'ab'
const foo = [`a${1}`, 'b'];    // 'a1'
const foo = [Number.NaN, 1];   // NaN
```

Fix: handle string `+` in `evaluateStaticBinary`, add a `KindTemplateExpression` arm, and resolve the `Number.*` / `Math.*` constants the evaluator already models as bare identifiers.

internal/plugins/unicorn/rules/prefer_set_has/static_value.go
```diff
@@ -0,0 +1,588 @@
...
+	if !isNumber || !isArrayLength(number) {
+		return 0, false
+	}
+	return int(number), true
+}
+
+// isLengthProperty mirrors upstream's `isLengthProperty`: a non-computed
+// `length: value` property assignment.
+func isLengthProperty(property *ast.Node) bool {
+	if property == nil || property.Kind != ast.KindPropertyAssignment {
```
swwind: **`{length}` shorthand isn't recognized.** `isLengthProperty` only accepts `KindPropertyAssignment`. ESTree models `{length}` as a `Property` with `computed: false` and an Identifier key, so upstream takes the `getStaticArrayLength(property.value, scope)` path and resolves `length` through scope.

```js
const length = 3;
const foo = Array.from({length});
foo.includes(1);
foo.includes(2);
```

With `{minimumItems: 2}` upstream reports 2:7 and autofixes; here `getObjectLength` returns not-ok, the size stays unknown, and the rule bails.

Fix: also accept `ast.KindShorthandPropertyAssignment` and evaluate its `Name()` identifier.

internal/plugins/unicorn/rules/prefer_set_has/prefer_set_has.go
```diff
@@ -0,0 +1,534 @@
...
+		return nil
+	}
+	// Upstream only checks `VariableDeclaration` → `VariableDeclaration` (the
+	// declarator) → `VariableDeclaration` (statement); it does NOT require
+	// `const` here (the const check happens for referenced initializers). But
+	// the identifier must resolve through `findVariable`, and non-const shapes
+	// are excluded downstream via reference analysis. Match upstream's exact
+	// structural gate: parent list's parent is the statement.
+	statement := declarationList.Parent
+	if statement == nil || statement.Kind != ast.KindVariableStatement {
```
swwind: **The `VariableStatement` gate is stricter than upstream.** Upstream's structural check is `parent.parent.type === 'VariableDeclaration'` — the declaration node, not a statement — so a `for`-initializer declaration qualifies:

```js
for (const foo = [1, 2]; ;) {
  foo.includes(1);
  foo.includes(2);
}
```

Upstream reports 1:12 and autofixes to `for (const foo = new Set([1, 2]); ;)`. Here `declarationList.Parent` is the `ForStatement`, so the rule bails. The comment above claims this gate matches upstream exactly.

Fix: keep `VariableStatement` only inside the export check (`HasSyntacticModifier` needs the statement; nothing else does). Note `isMultipleCall`'s `ancestorAt(node, 4)` has to move with it — upstream's `root` is `node.parent.parent.parent`, which for a `for` initializer is the `ForStatement` itself, so the enclosing `for` is deliberately *not* a multiple-call boundary. I verified that upstream gives no report for `for (const foo = [1, 2]; ;) { foo.includes(1); }` but does report when the single `includes` sits in a nested loop or a callback. A fixed depth of 4 would over-report the first case; deriving the root from `declarationList.Parent` (skipping a `VariableStatement` when present) covers both shapes.

internal/plugins/unicorn/rules/prefer_set_has/static_value.go
```diff
@@ -0,0 +1,588 @@
...
+	}
+	typeReference := typeAnnotation.AsTypeReferenceNode()
+	if typeReference.TypeName == nil || !ast.IsIdentifier(typeReference.TypeName) {
+		return "", false
+	}
+	typeArguments := typeReference.TypeArguments
+	if typeArguments == nil || len(typeArguments.Nodes) != 1 {
+		return "", false
+	}
+	if hasCommentsOutsideNode(ctx, typeAnnotation, typeArguments.Nodes[0]) {
```
swwind: **Comment check for `Array<T>` uses the wrong child node.** Upstream passes the whole type-argument list (the `<…>` span) to `hasCommentsOutsideNode`; this passes `typeArguments.Nodes[0]`, so any comment between the angle brackets and the type argument counts as "outside" and blocks the rewrite.

```ts
const foo: Array</* c */ string> = ['a', 'b'];
foo.includes('a');
foo.includes('b');
```

Upstream autofixes the annotation to `Set</* c */ string>`. Here `setTypeAnnotationText` returns not-rewritable, so the report downgrades to a suggestion — and since the suggestion skips the annotation edit entirely, applying it yields `const foo: Array<string> = new Set(['a', 'b'])`, which doesn't type-check. `Array<string /* c */>` hits the same path; `/* c */ string[]` and plain `Array<string>` are unaffected.

Fix: compare against the widened `<…>` range — the same span `typeArgumentsText` already recovers via `SliceEnclosingDelimiters` — rather than the first type argument.

swwind (commented):
Re-reviewed after the fixes — all seven earlier findings check out, thanks.

Six new divergences from real `eslint-plugin-unicorn` (v72), verified in eslint-playground and reproduced against this branch:

| Code | ESLint | This PR |
|---|---|---|
| `const foo = ['a' + 'b', 'c'];` + `forEach` + `includes` in a function | reports 1:7, autofixes | no report |
| ``const foo = [`a${1}`, 'b'];`` + same refs | reports 1:7, autofixes | no report |
| `const foo = [Number.NaN, 1];` + same refs | reports 1:7, autofixes | no report |
| `const length = 3; const foo = Array.from({length});` + 2 `includes` (`minimumItems: 2`) | reports 2:7, autofixes | no report |
| `for (const foo = [1, 2]; ;) { foo.includes(1); foo.includes(2); }` | reports 1:12, autofixes | no report |
| `const foo: Array</* c */ string> = ['a', 'b'];` + 2 `includes` | autofixes to `Set</* c */ string>` | reports, suggestion only — and it leaves `Array<…>`, so the suggested code is `const foo: Array<string> = new Set([…])` |

The first five are missed reports; the last is an autofix downgraded to a suggestion whose output doesn't type-check. Details inline.

Two smaller things, not worth their own threads:

- `SliceEnclosingDelimiters` silently falls back to offset `0` / `len(text)` when the delimiter isn't found (the `no open` and `start past len` unit tests codify this). Both call sites always have brackets so it's harmless today, but a `(string, bool)` return would stop a future caller from emitting silently-wrong fixer text.
- The docs and the package comment say "a `const` array binding", but the rule fires on `let`/`var` too (matching upstream) — verified in both.

Cases I checked that already match, for the record: `let`/`var` bindings, `export {foo}`, `type T = typeof foo`, class static blocks, `(foo.includes)(1)`, `foo?.includes(…)`, `(foo).length`, `NaN`/`undefined`/`Infinity` elements, and a `satisfies`-wrapped left-hand side.

elecmonkey:
Thanks — all six reproduce, and all six are fixed in 162f4029 + e3994fe6. Verified each against eslint-plugin-unicorn v72 before and after.

| Case | Fix |
|---|---|
| `'a' + 'b'` / `` `a${1}` `` / `Number.NaN` | string `+`, `TemplateExpression` and `Number.*`/`Math.*` constants now fold |
| `{length}` shorthand | `isLengthProperty` → `lengthPropertyValue`, accepts `ShorthandPropertyAssignment` |
| `for` initializer | gate no longer requires `VariableStatement`; `ancestorAt(node, 4)` → `multipleCallRoot` |
| `Array</* c */ string>` | comment check compares against the `<…>` span |

You were right that the `for` fix needed the root to move with it — a fixed depth of 4 overshoots the `ForStatement`, so `for (const foo = [1, 2]; ;) { foo.includes(1); }` would have over-reported. Both directions are locked in by tests.

Also took both smaller suggestions: `SliceEnclosingDelimiters` is now `RangeEnclosingDelimiters` returning `(start, end, ok)` with a thin wrapper — the range form is what the `Array<T>` fix needed, so it stopped being cosmetic — and the docs/package comment no longer say `const`.

One deliberate limit worth flagging. ToString feeds the SameValueZero uniqueness check, so a number that stringifies differently from JS would make two equal elements look distinct and the emitted `new Set(...)` would drop one. Go's shortest-decimal `'f'` matches `Number::toString` exactly for |x| in [1e-6, 1e21), but past that JS switches to exponential notation in a format `'f'` won't produce, so `toStringValue` bails there. `const foo = [1e21 + '', 'x']` reports upstream and stays silent here — a missed report rather than a bad fix. Locked in as a valid case with that note.

internal/plugins/unicorn/rules/prefer_set_has/static_value.go
elecmonkey: Fixed in e3994fe6. String `+` folds when either operand is a string, `TemplateExpression` has an arm, and `Number.*` / `Math.*` constants resolve through `PropertyAccessExpression` (with `IsShadowed` on the object identifier, so a local `Number` binding is not mistaken for the global). Concat / template / ToString semantics mirror `StaticStringEvaluator` rather than being reinvented. Numbers only fold inside the plain-decimal band — see the note in my top-level comment.

internal/plugins/unicorn/rules/prefer_set_has/static_value.go
elecmonkey: Fixed in e3994fe6. `isLengthProperty` became `lengthPropertyValue`, returning the value node and accepting `ShorthandPropertyAssignment` (whose value is the key Identifier, as in ESTree). Computed keys still bail, matching upstream's `!computed`.

internal/plugins/unicorn/rules/prefer_set_has/prefer_set_has.go
elecmonkey: Fixed in 162f4029, including the `ancestorAt` move you flagged. `multipleCallRoot` derives the root from the declaration list and skips a `VariableStatement` when there is one, so the `ForStatement` itself is the root at initializer position. `VariableStatement` is now only consulted for the export check, which is the only thing that needs it. Tests cover both directions: the two-`includes` report and the single-`includes` non-report.

internal/plugins/unicorn/rules/prefer_set_has/static_value.go
elecmonkey: Fixed in 162f4029. The check now compares against the `<…>` span (via the new `RangeEnclosingDelimiters`) instead of `typeArguments.Nodes[0]`, so the comment stays inside and the rewrite survives as an autofix, carrying the comment into `Set<…>` verbatim. `T[]` and `readonly T[]` keep passing `elementType`, matching upstream. Covered for both `Array</* c */ string>` and `Array<string /* c */>`.

elecmonkey:
Rebased onto main and fixed CI (349ca4c8).

The Go job was failing on every platform, but only one package was involved: `createPreferSetHasProgram` in the extras test still used the old `fixtures.GetRootDir() string` API, which became `embedfs.Root` in #1507 — so it stopped compiling once merged with main, and one package failing to build fails the whole run. Adapted the same way as your #1532: `rootDir.Dir` for the path arguments, and the overlay built on `rootDir.FS` rather than `NewOverlayVFSForFile`, which hardcodes the OS FS and would bypass the embedded fixtures.

The rebase changed the SHAs I referenced earlier — the two review-fix commits are now bd87d44b and 13799de7.

fansenze (approved):

PR #1482, opened 2026-07-31
feat: port rule preserve-caught-error

swwind:
## Summary

Port the `preserve-caught-error` rule from ESLint to rslint.

The rule requires errors constructed and thrown inside a `catch` block to carry the caught error as their `cause`, and reports a `cause` set to anything else. It covers the global error constructors, plus custom classes listed in `errorClassNames`, and offers a suggestion that attaches the caught error.

Enabled in the `javascript` and `typescript` recommended presets (upstream marks it `recommended`). Dogfooding surfaced one violation in `packages/vscode-extension/src/main.ts`, fixed in this PR.

`utils.ToInt` is extracted from `max_lines_per_function` so both rules share one JSON-number coercion helper.

### Differences from ESLint

- A parenthesized computed `cause` key — `{ [('cause')]: error }` — is reported as a missing `cause`; every other spelling of the key is accepted by both.
- For a constructor with type arguments and no argument list — `new AppError<T>;` — the suggestion inserts the arguments after the type arguments rather than before them.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/preserve-caught-error
- Source code: https://github.com/eslint/eslint/blob/main/lib/rules/preserve-caught-error.js

## Checklist

### DIFF
--- a/internal/rules/preserve_caught_error/preserve_caught_error.go
+++ b/internal/rules/preserve_caught_error/preserve_caught_error.go
@@ -53,7 +53,7 @@ func parseOptions(options []any) Options {
 			if !ok {
 				continue
 			}
-			position, ok := utils.ToInt(entry["argumentPosition"])
+			position, ok := utils.CoerceIntegral(entry["argumentPosition"])
 			if !ok {
 				continue
 			}
@@ -329,7 +329,7 @@ func insertAt(pos int, text string) rule.RuleFix {
 // insertion lands outside any parentheses wrapping that argument.
 func appendArguments(info thrownError, text string) []rule.RuleFix {
 	args := info.args()
-	return []rule.RuleFix{insertAt(args[len(args)-1].End(), text)}
+	return []rule.RuleFix{rule.RuleFixInsertAfter(args[len(args)-1], text)}
 }
 
 // addArgumentsToEmptyCall inserts text as the whole argument list, whether the
@@ -338,7 +338,7 @@ func addArgumentsToEmptyCall(info thrownError, text string) []rule.RuleFix {
 	if info.arguments != nil {
 		return []rule.RuleFix{insertAt(info.arguments.Pos(), text)}
 	}
-	return []rule.RuleFix{insertAt(info.expression.End(), "("+text+")")}
+	return []rule.RuleFix{rule.RuleFixInsertAfter(info.expression, "("+text+")")}
 }
 
 // insertCauseIntoOptions adds `cause` to an existing inline options object.
@@ -355,7 +355,7 @@ func insertCauseIntoOptions(optionsArg *ast.Node, caughtName string) []rule.Rule
 		return []rule.RuleFix{insertAt(properties.Pos(), "cause: "+caughtName)}
 	}
 	lastProperty := properties.Nodes[len(properties.Nodes)-1]
-	return []rule.RuleFix{insertAt(lastProperty.End(), ", cause: "+caughtName)}
+	return []rule.RuleFix{rule.RuleFixInsertAfter(lastProperty, ", cause: "+caughtName)}
 }
 
 // buildMissingCauseFixes attaches the caught error as the `cause` of an error
--- a/internal/rules/preserve_caught_error/preserve_caught_error.go
+++ b/internal/rules/preserve_caught_error/preserve_caught_error.go
@@ -189,6 +189,12 @@ func classifyThrownError(ctx rule.RuleContext, expression *ast.Node, opts Option
 	info.expression = expression
 
 	callee := ast.SkipParentheses(info.callee)
+	// Parentheses end an optional chain, so a chain used as the callee stays a
+	// `ChainExpression` in ESLint's AST — `(ns?.AppError)('m')` matches neither
+	// a global error name nor a configured class.
+	if ast.IsOptionalChain(callee) {
+		return info, false
+	}
 	if isBuiltInGlobalError(ctx, callee) {
 		info.builtIn = true
 		info.className = callee.AsIdentifier().Text
@@ -419,6 +425,19 @@ func buildIncorrectCauseFixes(ctx rule.RuleContext, cause causeInfo, caughtName 
 	return []rule.RuleFix{rule.RuleFixReplace(ctx.SourceFile, cause.value, caughtName)}
 }
 
+// isCaughtErrorShadowed reports whether the identifier attached as `cause`
+// resolves to a binding other than the catch parameter it is spelled like. A
+// `var` sharing the name is hoisted out of the catch clause and still resolves
+// to the parameter, so only a closer block-scoped binding shadows it.
+func isCaughtErrorShadowed(ctx rule.RuleContext, causeValue *ast.Node, param *ast.Node) bool {
+	caught := param.Symbol()
+	if caught == nil {
+		return false
+	}
+	target := ctx.Refs.Resolve(causeValue)
+	return target != nil && target != caught
+}
+
 func suggestIncludeCause(fixes []rule.RuleFix) []rule.RuleSuggestion {
 	if len(fixes) == 0 {
 		return nil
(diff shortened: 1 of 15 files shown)


internal/rules/preserve_caught_error/preserve_caught_error.go
```diff
@@ -0,0 +1,496 @@
...
+		return info, false
+	}
+	// ESLint sees an optional chain as a `ChainExpression`, which is neither a
+	// `CallExpression` nor a `NewExpression`, so those never reach the rule.
+	if info.callee == nil || ast.IsOptionalChain(expression) {
+		return info, false
+	}
+	info.expression = expression
+
+	callee := ast.SkipParentheses(info.callee)
```
elecmonkey, line 191: **[P1] Preserve `ChainExpression` semantics for parenthesized callees**

Checking only `ast.IsOptionalChain(expression)` before stripping callee parentheses misses cases where the call/new expression itself is not optional, but its callee is a parenthesized optional chain.

With `{ errorClassNames: ['AppError'] }`:

```js
try {} catch (err) {
  throw (ns?.AppError)('m');
}
```

ESLint 10.8.0 reports nothing because the ESTree callee is a `ChainExpression`, so `isThrowingNewError` does not match it. Rslint strips the parentheses, matches `AppError`, and reports `missingCause`. `new (ns?.AppError)('m')` has the same mismatch.

Please reject a callee whose parenthesized inner expression is an optional chain before matching configured error classes. The control case `(ns?.errors).AppError('m')` should remain reportable, matching ESLint.

internal/rules/preserve_caught_error/preserve_caught_error.go
```diff
@@ -0,0 +1,496 @@
...
+	for _, member := range properties.Nodes {
+		if member.Kind == ast.KindSpreadAssignment {
+			// A spread can contribute a `cause` that is impossible to verify.
+			return causeInfo{status: causeUnknown}
+		}
+		name := member.Name()
+		if name == nil {
+			continue
+		}
+		if propertyName, ok := utils.GetStaticPropertyName(name); ok && propertyName == "cause" {
```
elecmonkey, line 283: **[P1] Recognize parenthesized static `cause` keys**

`GetStaticPropertyName` does not unwrap the expression inside a computed property name, so `{ [('cause')]: err }` is treated as missing a cause.

ESLint recognizes this as the static key `"cause"`. If the value is `err`, ESLint reports nothing; if it is `other`, ESLint reports `incorrectCause` and replaces only the value. Rslint currently reports `missingCause` in both cases and suggests appending a duplicate `cause` property.

This is listed under “Differences from ESLint”, but the port should align with upstream rather than preserve the divergence. Please unwrap parentheses before evaluating computed static property names and update the lock-in test/documentation.

internal/rules/preserve_caught_error/preserve_caught_error.go
```diff
@@ -0,0 +1,496 @@
...
+	return []rule.RuleFix{rule.RuleFixInsertAfter(args[len(args)-1], text)}
+}
+
+// addArgumentsToEmptyCall inserts text as the whole argument list, whether the
+// call already has empty parentheses (`new Error()`) or none (`new Error`).
+func addArgumentsToEmptyCall(info thrownError, text string) []rule.RuleFix {
+	if info.arguments != nil {
+		return []rule.RuleFix{insertAt(info.arguments.Pos(), text)}
+	}
+	return []rule.RuleFix{rule.RuleFixInsertAfter(info.expression, "("+text+")")}
```
elecmonkey, line 347: **[P2] Match ESLint's insertion point for type arguments without an argument list**

For:

```ts
try {} catch (err) {
  throw new AppError<string>;
}
```

ESLint 10.8.0 inserts the new argument list after the callee and before the type arguments:

```ts
new AppError({ cause: err })<string>
```

Rslint inserts after the complete `NewExpression`:

```ts
new AppError<string>({ cause: err })
```

The same mismatch occurs with `new (AppError)<string>`. Cases that already have `()` or arguments are aligned. Since suggestions are part of rule behavior, please use the upstream token position for the no-argument-list case and remove this documented divergence.

internal/rules/preserve_caught_error/preserve_caught_error.go
```diff
@@ -0,0 +1,496 @@
...
+							// With several `cause` definitions a suggestion
+							// would be more confusing than helpful.
+							return nil
+						}
+						return suggestIncludeCause(buildIncorrectCauseFixes(ctx, cause, caughtName))
+					})
+					return
+				}
+
+				if utils.IsNameShadowedBetween(cause.value, parentCatch, caughtName) {
```
elecmonkey: **[P1] Do not treat `var` as a closer catch binding**

`IsNameShadowedBetween` treats all declarations found in a block or loop initializer as block-scoped. This makes the rule report `caughtErrorShadowed` for `var` declarations that do not introduce a closer binding than the catch parameter:

```js
try {} catch (err) {
  var err;
  throw new Error('m', { cause: err });
}
```

ESLint 10.8.0 reports nothing here, while Rslint reports `caughtErrorShadowed`. The same false positive occurs for a nested block, `for (var err of values)`, a classic `for (var err = 0; ...)`, and `var` in another `switch` case. Equivalent `let`/`const` cases should continue to report.

Please distinguish `var` from block-scoped declaration lists when checking shadowing, or resolve the `cause` identifier to the actual catch binding instead of relying on this name-only helper.

elecmonkey (changes requested):
I compared this port against ESLint 10.8.0 at its current `main` implementation and ran the full migrated suite plus a 154-case differential matrix. The matrix produced 19 mismatches, grouped into the issues below. Two are already documented as differences, but this port is expected to preserve upstream behavior, so they should be fixed rather than locked in.

The existing rule tests pass, and the other 135 generated cases match ESLint.

internal/rules/preserve_caught_error/preserve_caught_error.go
swwind: Confirmed and fixed. `classifyThrownError` now rejects a callee that is still an optional chain after parentheses are stripped, so `(ns?.AppError)('m')` and `new (ns?.AppError)('m')` are no longer matched. Added valid cases for both, plus `(ns?.errors.AppError)('m')`, and kept `(ns?.errors).AppError('m')` as a reportable control case.

internal/rules/preserve_caught_error/preserve_caught_error.go
swwind: Confirmed and fixed in the shared helper: `utils.GetStaticPropertyName` now skips parentheses inside a computed property name. ESLint's AST has no parenthesized-expression node, so this is not specific to this rule — `no-dupe-keys` and `accessor-pairs` treat `[('a')]` as the static key `a` as well, and the shared fix aligns all of them. `{ [('cause')]: err }` is now valid, `{ [('cause')]: other }` reports `incorrectCause` and the suggestion replaces only the value. The lock-in test and the "Differences from ESLint" bullet are gone.

internal/rules/preserve_caught_error/preserve_caught_error.go
swwind: The mismatch is real, but I'd rather keep this one as a documented difference, because the upstream token position produces broken output and the problem is not limited to type arguments. When there is no argument list, `getFirstTokenBetween(lastCalleeToken, lastToken, isOpeningParen)` finds no `(`, so the fixer falls back to inserting directly after the callee:

- `new AppError<string>` → `new AppError({ cause: err })<string>` — does not parse.
- `new (AppError)` → `new (AppError({ cause: err }))` — calls `AppError` and constructs the result, so the suggestion changes semantics.

This port appends the argument list after the whole expression instead, so the suggestion always applies cleanly. I extended the difference note to cover the parenthesized-callee case as well and added a lock-in test for it. If the project prefers byte-identical suggestions over applicable ones, I'll switch both cases to the upstream position.

internal/rules/preserve_caught_error/preserve_caught_error.go
swwind: Confirmed and fixed. The rule no longer uses the name-only helper: it resolves the `cause` identifier with `ctx.Refs.Resolve` and compares the result with the catch parameter's symbol, which mirrors ESLint's scope walk. A `var` hoists out of the catch clause and still resolves to the parameter, so `var err` in the same block, in a nested block, in `for (var err of list)`, in a classic `for`, and in another `switch` case are all valid now; `let`/`const`/`class` in a nested block still report.

elecmonkey (approved):

fansenze (approved):

Issue #1336, opened 2026-07-22
[Feature]: Skip fix computation for `no-var` and similar rules when `--fix` is not requested


elecmonkey:
### What problem does this feature solve?

Reporting a `no-var` violation is a single syntax check. But deciding whether the `var→let` autofix is safe, `canFix` does full-scope reference analysis and calls `GetSymbolAtLocation` on every candidate identifier (for a top-level `var` this scans the whole file).

However, `canFix`'s result does not affect whether we report — it's purely for the fix. Yet this expensive analysis runs regardless of `--fix`, so in common lint-only CI runs it's computed and immediately discarded — pure waste.

### What does the proposed API of configuration look like?

Thread the "is autofix needed" intent from the CLI layer all the way down to rule execution, so a rule can know whether the current run will actually apply fixes. On that basis, decouple the "expensive analysis done solely to compute the fix" from the "analysis required for reporting itself" — when a run doesn't need fixes, the rule skips the former entirely.

`no-var`: its report decision is a pure syntax check, while `canFix`'s full-scope reference analysis serves only the fix and does not affect whether we report. So it can be skipped wholesale when no fix is needed, with identical reporting results.

Default behavior is unchanged (fixes are still computed as usual under LSP, the API, and `--fix`); only the lint-only CLI path turns this analysis off, saving full-file symbol resolution on the common CI route.

fansenze:
Make sense! I will follow up on this.

swwind:
Found a couple of interesting ESLint projects that maybe related:

* `eslint-plugin-autofix` provides enhanced autofixers for a number of existing ESLint rules. Most of them are reimplementations of core rules with improved fixing behavior.
* `eslint-plugin-no-autofix` wraps existing rules and strips the autofix information from their reports. It doesn't improve performance or prevent the original rules from generating fixes internally—it simply hides the fix metadata, which is useful if you don't want ESLint to report that a rule is auto-fixable.

fansenze:
Closing this issue for now, since this has already been integrated at the framework level.

PR #1484, opened 2026-07-31
feat(unicorn): port rule prefer-array-some

elecmonkey:
## Summary

Port the `unicorn/prefer-array-some` rule from eslint-plugin-unicorn to rslint, at upstream `9c2b0bb1` ("Harden rules", #3585).

Suggests replacing `.find()`/`.findLast()` boolean checks, `.findIndex()`/`.findLastIndex()` comparisons against `-1`/`0`, and `.filter(…).length > 0` with `.some(…)`.

`RequiresTypeInfo: false` — syntactic-first, type info only refines (skips keyed-collection receivers, confirms array for the variable-usage branch).

### Differential validation

22 (rslint) vs 21 (ESLint) on a controlled fixture. rslint covers all 21 ESLint reports. The extra one is `Int8Array.filter(fn).length > 0`, which the ESLint side does not report because it runs the last published release; upstream's `isKnownNonIndexedCollection` change makes typed arrays targets, and its test suite has this case as invalid. Behavior here follows that.

## Related Links

- Rule doc: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-array-some.md
- Source: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/9c2b0bb1/rules/prefer-array-some.js

## Checklist

### DIFF
--- a/internal/plugins/unicorn/rules/prefer_array_some/prefer_array_some.go
+++ b/internal/plugins/unicorn/rules/prefer_array_some/prefer_array_some.go
@@ -375,7 +375,7 @@ func checkFilterLength(ctx rule.RuleContext, node *ast.Node) {
 
 	filterProperty := filterCall.Property
 	ctx.ReportNodeWithDeferredFixes(filterProperty, messageArrayFilter(), func() []rule.RuleFix {
-		lengthObjectEnd := parenthesizedEndOf(ctx, filterCall.Call)
+		lengthObjectEnd := parenthesizedEnd(ctx, filterCall.Call)
 		end := node.End()
 		// Removing `.length > 0` would drop comments in the removed range.
 		if utils.HasCommentInSpan(ctx.Comments.All(), lengthObjectEnd, end) {
@@ -391,7 +391,7 @@ func checkFilterLength(ctx rule.RuleContext, node *ast.Node) {
 			// parentheses wrapping the member survive the removal.
 			rule.RuleFixRemoveRange(core.NewTextRange(lengthObjectEnd, utils.TrimNodeTextRange(ctx.SourceFile, lengthMember).End())),
 			// Remove `> 0`, starting after any parentheses around the member.
-			rule.RuleFixRemoveRange(core.NewTextRange(parenthesizedEndOf(ctx, lengthMember), end)),
+			rule.RuleFixRemoveRange(core.NewTextRange(parenthesizedEnd(ctx, lengthMember), end)),
 		}
 	})
 }
@@ -604,10 +604,6 @@ func parenthesizedEnd(ctx rule.RuleContext, node *ast.Node) int {
 	return parenthesizedRange(ctx, node).End()
 }
 
-func parenthesizedEndOf(ctx rule.RuleContext, node *ast.Node) int {
-	return parenthesizedEnd(ctx, node)
-}
-
 // insertBeforeRange returns a zero-width fix that inserts text at the start of
 // the given range.
 func insertBeforeRange(textRange core.TextRange, text string) rule.RuleFix {
@@ -728,10 +724,7 @@ func classifyReceiver(ctx rule.RuleContext, node *ast.Node, targetNames, nonTarg
 // a plain identifier name are resolved (matching upstream's getTypeFromVariable
 // preconditions).
 func constInitializer(ctx rule.RuleContext, idNode *ast.Node) *ast.Node {
-	if ctx.TypeChecker == nil {
-		return nil
-	}
-	sym := ctx.TypeChecker.GetSymbolAtLocation(idNode)
+	sym := ctx.Refs.Resolve(idNode)
 	if sym == nil || len(sym.Declarations) != 1 {
 		return nil
 	}
--- a/internal/plugins/unicorn/rules/prefer_array_some/prefer_array_some.go
+++ b/internal/plugins/unicorn/rules/prefer_array_some/prefer_array_some.go
@@ -384,9 +384,11 @@ func checkFilterLength(ctx rule.RuleContext, node *ast.Node) {
 		return []rule.RuleFix{
 			rule.RuleFixReplace(ctx.SourceFile, filterProperty, "some"),
 			// Remove `.length`: from the end of the filter call up to the end
-			// of the `.length` member expression.
-			rule.RuleFixRemoveRange(core.NewTextRange(lengthObjectEnd, parenthesizedEndOf(ctx, lengthMember))),
-			// Remove `> 0`.
+			// of the member expression itself. Upstream's
+			// removeMemberExpressionProperty stops at the member's own end, so
+			// parentheses wrapping the member survive the removal.
+			rule.RuleFixRemoveRange(core.NewTextRange(lengthObjectEnd, utils.TrimNodeTextRange(ctx.SourceFile, lengthMember).End())),
+			// Remove `> 0`, starting after any parentheses around the member.
 			rule.RuleFixRemoveRange(core.NewTextRange(parenthesizedEndOf(ctx, lengthMember), end)),
 		}
 	})
--- a/internal/plugins/unicorn/rules/prefer_array_some/prefer_array_some.go
+++ b/internal/plugins/unicorn/rules/prefer_array_some/prefer_array_some.go
@@ -862,22 +862,24 @@ func classifyType(ctx rule.RuleContext, t *checker.Type, targetNames, nonTargetN
 	if targetNames.Has(name) {
 		return classTarget
 	}
-	// Interface / class heritage: `interface Items extends Array<string>` (and
-	// `class X extends Array`) resolve to the array target through their base
-	// types. Upstream reaches the same conclusion via its interface-heritage
-	// walk. Only consult heritage when the receiver could be an array target;
-	// keyed-collection classification does not widen through heritage.
+	// Interface heritage: `interface Items extends Array<string>` resolves to
+	// the array target through its base types. Upstream reaches the same
+	// conclusion via its interface-heritage walk, which runs with
+	// `checkClassHeritage: false` — `class MyArray extends Array {}` therefore
+	// stays a non-target. Only consult heritage when the receiver could be an
+	// array target; keyed-collection classification does not widen through it.
 	if targetNames.Has("Array") && classifyHeritage(ctx, t, targetNames, nonTargetNames) == classTarget {
 		return classTarget
 	}
 	return classNonTarget
 }
 
-// classifyHeritage inspects the base types of a class / interface symbol,
-// returning classTarget when any base resolves to the target classification.
+// classifyHeritage inspects the base types of an interface symbol, returning
+// classTarget when any base resolves to the target classification. Classes are
+// excluded to match upstream's `checkClassHeritage: false`.
 func classifyHeritage(ctx rule.RuleContext, t *checker.Type, targetNames, nonTargetNames *utils.Set[string]) typeClass {
 	symbol := checker.Type_symbol(t)
-	if symbol == nil || symbol.Flags&(ast.SymbolFlagsClass|ast.SymbolFlagsInterface) == 0 {
+	if symbol == nil || symbol.Flags&ast.SymbolFlagsInterface == 0 {
 		return classUnknown
 	}
 	declared := checker.Checker_getDeclaredTypeOfSymbol(ctx.TypeChecker, symbol)
--- a/internal/plugins/unicorn/rules/prefer_array_some/prefer_array_some.go
+++ b/internal/plugins/unicorn/rules/prefer_array_some/prefer_array_some.go
@@ -1,0 +1,948 @@
+package prefer_array_some
+
+import (
+	"github.com/microsoft/typescript-go/shim/ast"
+	"github.com/microsoft/typescript-go/shim/checker"
+	"github.com/microsoft/typescript-go/shim/core"
+	"github.com/microsoft/typescript-go/shim/scanner"
+	"github.com/web-infra-dev/rslint/internal/plugins/unicorn/unicornutil"
+	"github.com/web-infra-dev/rslint/internal/rule"
+	"github.com/web-infra-dev/rslint/internal/utils"
+)
+
+const (
+	messageIDArraySome           = "some"
+	messageIDArraySomeSuggestion = "some-suggestion"
+	messageIDArrayFilter         = "filter"
+)
+
+func messageArraySome(method string) rule.RuleMessage {
+	return rule.RuleMessage{
+		Id:          messageIDArraySome,
+		Description: "Prefer `.some(…)` over `." + method + "(…)`.",
+		Data:        map[string]string{"method": method},
+	}
+}
+
+func messageArraySomeSuggestion(method string) rule.RuleMessage {
+	return rule.RuleMessage{
+		Id:          messageIDArraySomeSuggestion,
+		Description: "Replace `." + method + "(…)` with `.some(…)`.",
+		Data:        map[string]string{"method": method},
+	}
+}
+
+func messageArrayFilter() rule.RuleMessage {
+	return rule.RuleMessage{
+		Id:          messageIDArrayFilter,
+		Description: "Prefer `.some(…)` over non-zero length check from `.filter(…)`.",
+	}
+}
+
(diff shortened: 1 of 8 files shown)


swwind (commented):
Verified against `eslint-plugin-unicorn@72` — plain-JS cases via `Linter`, type-aware cases with `parserOptions.project`.

| Code | ESLint | This PR |
|---|---|---|
| `((array.filter(fn).length)) > 0` | fixes to `((array.some(fn)))` | fixes to `((array.some(fn)` — unbalanced parens |
| `(( array.filter(fn).length )) > 0` | fixes to `(( array.some(fn) ))` | fixes to `(( array.some(fn)` — unbalanced parens |
| `(( (( array.filter(fn) )).length )) > (( 0 ))` | reports 1:13 | no report |
| `array.filter(typeof foo).length > 0` | no report | reports 1:7 |
| `array.filter(void 0).length > 0` | no report | reports 1:7 |
| `array.filter(delete a.b).length > 0` | no report | reports 1:7 |
| `class MyArray extends Array {}` receiver, `if (a.find(fn)) {}` | no report | reports |
| `function f(a: Int8Array) { if (a.find(fn)) {} }` | no report | reports |
| `function f(a: Int8Array) { a.findIndex(fn) !== -1 }` | no report | reports |

One missed report (parenthesized `0`), six extra reports, and two autofixes that produce syntactically invalid output. Details inline.

Nits:
- `parenthesizedEndOf` is an exact alias of `parenthesizedEnd`; one of them can go.
- `constInitializer` resolves the identifier with `ctx.TypeChecker.GetSymbolAtLocation`; `ctx.Refs` is the repo's symbol-lookup path.

internal/plugins/unicorn/rules/prefer_array_some/prefer_array_some.go
```diff
@@ -0,0 +1,948 @@
...
+		if utils.HasCommentInSpan(ctx.Comments.All(), lengthObjectEnd, end) {
+			return nil
+		}
+
+		// `.filter` → `.some`; remove `.length`; remove `> 0`.
+		return []rule.RuleFix{
+			rule.RuleFixReplace(ctx.SourceFile, filterProperty, "some"),
+			// Remove `.length`: from the end of the filter call up to the end
+			// of the `.length` member expression.
+			rule.RuleFixRemoveRange(core.NewTextRange(lengthObjectEnd, parenthesizedEndOf(ctx, lengthMember))),
```
swwind: **Autofix emits unbalanced parentheses when `.length` is parenthesized.**

`((array.filter(fn).length)) > 0` is fixed to `((array.some(fn)` — the `.length` removal ends at the *parenthesized* end of the member expression, so the closing `))` is deleted along with it. Upstream's `removeMemberExpressionProperty` ends this removal at `sourceCode.getRange(memberExpression)[1]`, the member's own end; only the second removal (`> 0`) uses the parenthesized end.

```go
rule.RuleFixRemoveRange(core.NewTextRange(lengthObjectEnd, utils.TrimNodeTextRange(ctx.SourceFile, lengthMember).End())),
```

internal/plugins/unicorn/rules/prefer_array_some/prefer_array_some.go
```diff
@@ -0,0 +1,948 @@
...
+	}
+	// We assume the user already follows `unicorn/explicit-length-check`; only
+	// `> 0` and `!== 0` are allowed there.
+	if bin.OperatorToken.Kind != ast.KindGreaterThanToken &&
+		bin.OperatorToken.Kind != ast.KindExclamationEqualsEqualsToken {
+		return
+	}
+	// The right side must be the RAW literal `0` — `0.` / `0x00` / `.0` are
+	// distinct source forms upstream deliberately does not match.
+	if !isRawZeroLiteral(ctx.SourceFile, bin.Right) {
```
swwind: **A parenthesized `0` is not matched.**

`(( (( array.filter(fn) )).length )) > (( 0 ))` reports in ESLint but not here. ESTree drops parentheses, so `binaryExpression.right` is the `Literal` and `right.raw === '0'` holds; in tsgo `bin.Right` is a `ParenthesizedExpression` and `isRawZeroLiteral` bails on the kind check. Skipping parens first keeps the `0.` / `.0` / `0x00` valid cases valid, since the inner literal's source text is still `0`.

```go
if !isRawZeroLiteral(ctx.SourceFile, ast.SkipParentheses(bin.Right)) {
```

internal/plugins/unicorn/rules/prefer_array_some/prefer_array_some.go
```diff
@@ -0,0 +1,948 @@
...
+// It rejects `.filter` arguments that cannot be a callback (literals, objects,
+// arrays, etc.), while treating `.bind()` calls as possible functions.
+func isNodeValueNotFunction(node *ast.Node) bool {
+	switch node.Kind {
+	case ast.KindArrayLiteralExpression,
+		ast.KindObjectLiteralExpression,
+		ast.KindClassExpression,
+		ast.KindTemplateExpression,
+		ast.KindNoSubstitutionTemplateLiteral,
+		ast.KindPrefixUnaryExpression,
```
swwind, line 423: **`typeof` / `void` / `delete` arguments are treated as possible callbacks.**

Upstream's `impossibleNodeTypes` contains ESTree `UnaryExpression`, which covers `!x`, `-x`, `typeof x`, `void x` and `delete x`. tsgo splits the last three into `TypeOfExpression` / `VoidExpression` / `DeleteExpression`, so they fall through here and `array.filter(typeof foo).length > 0`, `array.filter(void 0).length > 0`, `array.filter(delete a.b).length > 0` all report. Adding those three kinds to the list closes the gap.

internal/plugins/unicorn/rules/prefer_array_some/prefer_array_some.go
```diff
@@ -0,0 +1,948 @@
...
+		return classTarget
+	}
+	return classNonTarget
+}
+
+// classifyHeritage inspects the base types of a class / interface symbol,
+// returning classTarget when any base resolves to the target classification.
+func classifyHeritage(ctx rule.RuleContext, t *checker.Type, targetNames, nonTargetNames *utils.Set[string]) typeClass {
+	symbol := checker.Type_symbol(t)
+	if symbol == nil || symbol.Flags&(ast.SymbolFlagsClass|ast.SymbolFlagsInterface) == 0 {
```
swwind: **Class heritage should not widen to `Array`.**

`is-array.js` passes `checkClassHeritage: false`, so upstream walks *interface* heritage only — that is what makes `interface Items extends Array<string>` report while `class MyArray extends Array {}` does not. Including `ast.SymbolFlagsClass` here reports the class case where ESLint is silent; dropping it keeps the interface behavior intact.

internal/plugins/unicorn/rules/prefer_array_some/prefer_array_some.go
```diff
@@ -0,0 +1,948 @@
...
+
+const (
+	classUnknown typeClass = iota
+	classTarget
+	classNonTarget
+)
+
+var arrayTargets = utils.NewSetFromItems("Array", "ReadonlyArray")
+
+var indexedCollectionTargets = func() *utils.Set[string] {
```
swwind: **Typed arrays are an intentional divergence that needs documenting.**

Upstream's `knownNonArrayTypeNames` includes every typed array, so `Int8Array#find`, `#findIndex` and `#filter` are all skipped — not just `.filter`. Treating them as targets is defensible (typed arrays do have `.some()`), but the PR description attributes the difference to `projectService` vs `project`; re-running ESLint with `parserOptions.project` still produces no report, so the cause is this target-set change rather than scan scope. Please document the divergence in `prefer_array_some.md`, or align with upstream.

internal/plugins/unicorn/rules/prefer_array_some/prefer_array_some.go
```diff
@@ -0,0 +1,948 @@
...
+	}
+
+	methodNode := call.Property
+	method := methodNode.AsIdentifier().Text
+
+	comparison := comparisonParent(node) // non-nil only when isCompare
+
+	// Removing the comparison would drop comments between the call and the end
+	// of the comparison; upstream withholds the suggestion in that case.
+	wouldDropComments := isCompare && utils.HasCommentInSpan(
```
swwind: **Comment scan runs even when no suggestion is requested.**

`ctx.Comments.All()` materializes every comment in the file on first use, but `wouldDropComments` only decides whether a suggestion is attached — it never affects whether we report. `checkFindIndexComparison` and `checkFilterLength` already run the same check inside their deferred builders; doing that here too and returning `nil` when comments would be dropped keeps lint-only runs off the comment scan (empty results attach no artifact, so the report is identical to today's `ReportNode` path). See #1336 and the `perf: defer …` series.

internal/plugins/unicorn/rules/prefer_array_some/prefer_array_some_upstream_test.go
```diff
@@ -0,0 +1,317 @@
...
+		{Code: `type Items = string[] | {filter(predicate: Function): {length: number}}; function foo(items: Items) { items.filter(fn).length > 0; }`, Tsx: false, Output: []string{`type Items = string[] | {filter(predicate: Function): {length: number}}; function foo(items: Items) { items.some(fn); }`}, Errors: []rule_tester.InvalidTestCaseError{{MessageId: messageFilter}}},
+		{Code: `declare function getItems(): string[]; getItems().filter(fn).length > 0;`, Tsx: false, Output: []string{`declare function getItems(): string[]; getItems().some(fn);`}, Errors: []rule_tester.InvalidTestCaseError{{MessageId: messageFilter}}},
+		// A typed array shares `Array#filter()` and `Array#some()`.
+		{Code: `function f(array: Int8Array) { return array.filter(fn).length > 0; }`, Tsx: false, Output: []string{`function f(array: Int8Array) { return array.some(fn); }`}, Errors: []rule_tester.InvalidTestCaseError{{MessageId: messageFilter}}},
```
swwind: **Codifies the typed-array divergence.**

ESLint does not report this. Keeping the case is fine, but it belongs with the documented divergence rather than in the file described as a 1:1 upstream migration.

internal/plugins/unicorn/rules/prefer_array_some/prefer_array_some.go
elecmonkey: Fixed in 3afd96c4. The `.length` removal now ends at the member's own end (`utils.TrimNodeTextRange`), only the `> 0` removal uses the parenthesized end. `((array.filter(fn).length)) > 0` → `((array.some(fn)))`; both paren forms added as regression cases.

internal/plugins/unicorn/rules/prefer_array_some/prefer_array_some.go
elecmonkey: Fixed in c8aaa32e — `ast.SkipParentheses(bin.Right)` before the raw-text check. `(( (( array.filter(fn) )).length )) > (( 0 ))` now reports and fixes to `(( (( array.some(fn) )) ))`, matching your upstream snapshot; the `0.` / `.0` / `0x00` valid cases are unaffected.

internal/plugins/unicorn/rules/prefer_array_some/prefer_array_some.go
elecmonkey: Fixed in 63e10fac — added `KindTypeOfExpression` / `KindVoidExpression` / `KindDeleteExpression` alongside the prefix/postfix kinds, with a comment on why ESTree's single `UnaryExpression` needs five kinds here. All four forms added as valid cases.

internal/plugins/unicorn/rules/prefer_array_some/prefer_array_some.go
elecmonkey: Fixed in 7847c642 — `classifyHeritage` now requires `ast.SymbolFlagsInterface`, so `class MyArray extends Array {}` no longer reaches the array target while the `interface Items extends Array<string>` case keeps reporting. Negative cases added for both the find and filter branches.

internal/plugins/unicorn/rules/prefer_array_some/prefer_array_some.go
elecmonkey: Fixed in a1a18009 — the comment check moved inside the deferred suggestion builder and returns `nil` on a hit. Confirmed the equivalence you describe at rule/context.go:199: an empty build result falls through to `emitRange(…, nil, nil)`, which is exactly what `ReportNode` emits, so lint-only runs never touch `ctx.Comments.All()` now.

internal/plugins/unicorn/rules/prefer_array_some/prefer_array_some.go
elecmonkey: Aligned with v72 in cafa38ee: `isKnownNonIndexedCollection` is gone, all three branches use `isKnownNonArray` (targets `Array`/`ReadonlyArray`, non-targets = keyed collections + typed arrays), so `Int8Array#find` / `#findIndex` / `#filter` are all skipped. You are right that the projectService/project explanation in the PR description was wrong — I will correct it.

One data point while verifying: this divergence is upstream `main`, not v72. Post-v72 upstream replaced `isKnownNonArray` with `isKnownNonIndexedCollection` in all three branches and moved typed arrays out of the non-target set, and `main` now has `function f(array: Int8Array) { return array.filter(fn).length > 0; }` as an *invalid* snapshot case. The v72.0.0 tag has neither. Since this PR ports v72 I went with v72 behavior; adopting the newer upstream semantics deliberately would be a reasonable follow-up.

internal/plugins/unicorn/rules/prefer_array_some/prefer_array_some_upstream_test.go
elecmonkey: Removed in cafa38ee. The case moved to the extras file as a valid case (together with the `find` / `findIndex` typed-array forms), so the upstream file stays 1:1 with v72.

elecmonkey:
All seven points addressed, one commit each, on top of a rebase onto current main (the branch also needed the #1507 `embedfs.Root` fixtures adaptation).

- 3afd96c4 unbalanced parens in the `.length` removal
- c8aaa32e parenthesized `0`
- 63e10fac `typeof` / `void` / `delete` filter arguments
- 7847c642 class heritage no longer widens to `Array`
- a1a18009 comment scan deferred into the suggestion builder
- fde067b4 nits: `parenthesizedEndOf` alias dropped, `constInitializer` now uses `ctx.Refs.Resolve`
- cafa38ee typed arrays aligned with v72 (see the inline reply — the divergence turned out to be upstream `main` behavior, not a v72 mismatch)

internal/plugins/unicorn/rules/prefer_array_some/prefer_array_some.go
elecmonkey: Correction to my last reply — I had this backwards, and cafa38ee is dropped.

The typed-array behavior tracks upstream `9c2b0bb1` ("Harden rules", #3585, 2026-07-24), which replaced `isKnownNonArray` with `isKnownNonIndexedCollection` in all three branches and moved typed arrays out of the non-target set. Its test suite has `function f(array: Int8Array) { return array.filter(fn).length > 0; }` as invalid. That commit landed after the v72.0.0 tag, so `eslint-plugin-unicorn@72` still has the old behavior — hence the no-report you measured. The port is against that commit, not the release; I have corrected the PR description, which was the source of the confusion.

Your other point stands and is fixed: the `projectService` vs `project` attribution was wrong. The upstream baseline is now named in the PR description, in the `prefer_array_some_upstream_test.go` header, and the typed-array coverage is in `prefer_array_some.md` (ea3ba14b).

internal/plugins/unicorn/rules/prefer_array_some/prefer_array_some_upstream_test.go
elecmonkey: Kept, and the file header now names the upstream baseline commit (ea3ba14b) so the case reads as part of the migrated suite rather than an addition. See the reply on the typed-array thread.

fansenze (approved):

PR #1512, opened 2026-08-03
perf: port no-undef to ctx.Refs.Resolve, drop RequiresTypeInfo

swwind:
## Summary

- `no-undef` now resolves identifiers through `ctx.Refs.Resolve` (a binder scope walk, falling back to the file's TypeChecker only when one is available) instead of requiring a TypeChecker unconditionally. This matches ESLint's own scope-based `no-undef`, and makes the rule usable on files with no type info at all (plain JS, gap files, LSP inferred-project files).
- Recognized "free" globals are now the real ECMAScript built-ins list (`utils.IsECMAScriptGlobal`) rather than whatever happens to be visible through `lib.d.ts`; `languageOptions.globals` / `/* global */` comments take priority, so an explicit `"off"` entry still reports even for a builtin name.
- `shouldSkip` gained explicit handling for `new.target`/`import.meta`, import attributes (`with { type: "json" }`), and JSX (lowercase intrinsics vs. component references) — cases the removed TypeChecker call handled incidentally and would otherwise have silently mishandled.
- Fixes a real bug in `utils.GetReferenceSymbol` found while building this: for shorthand properties (`{x}`), when `GetShorthandAssignmentValueSymbol` returned nil (undeclared `x`), it fell through to `GetSymbolAtLocation`, which always returns the property symbol, masking the undefined reference. This function is shared by 13 other rules.
- Adds `eval` to the ECMAScript globals list (it was missing; verified against real ESLint that `eval(...)` without config is not flagged).

Drops `RequiresTypeInfo` from `no-undef`.

## Benchmark

Synthetic corpus: 2000 `.ts` files (~8 MiB), each with 20 locally-declared `var`s read 200 times inside a function body, plus a fixed handful of checker-only globals (`console`, `window`, `document`) per file — weighted toward the case this change targets: identifiers resolvable from the binder alone, which the old implementation still paid a `TypeChecker.GetSymbolAtLocation` call for on every single one.

Per-rule time from `--timing all` (summed across workers), 10 runs after a warmup:

| | mean | median |
| --- | --- | --- |
| `main` | 69.5 ms | 69.3 ms |
| this PR | 48.9 ms | 48.7 ms |

Rule time: **-29.6%** (mean), **-29.7%** (median).

End-to-end wall clock on the same corpus (hyperfine, 12 runs + 2 warmup): 369.0 ms ± 18.6 → 358.0 ms ± 17.0, **-3.0%**; smaller than the rule-time delta because this is a type-aware corpus where parse/bind/typecheck dominate total time.

Both binaries produced 0 diagnostics on this corpus (all identifiers resolve), confirming no false positives from the refactor on the exercised patterns. This PR is a behavior change as well as a perf one (drops `RequiresTypeInfo`, fixes the shorthand-property bug, changes which globals are free), so unlike a pure refactor this isn't a byte-identical-diagnostics claim across all inputs — see the Go test suite for the specific behavior changes.

## Validation

- `go test ./internal/rules/no_undef/...`
- `go test` on all 13 other callers of `utils.GetReferenceSymbol` (no regressions from the shorthand-property fix): `no_ex_assign`, `prefer_arrow_callback`, `no_param_reassign`, `no_implied_eval`, `no_import_assign`, `no_loop_func`, `react_hooks/rules/exhaustive_deps`, `react_hooks/rules/rules_of_hooks`, `unicorn/rules/new_for_builtins`, `typescript/rules/no_unused_vars`, `typescript/rules/no_unused_private_class_members`, `react/rules/no_access_state_in_setstate`, `react/rules/jsx_fragments`
- `go test ./internal/utils/...`, `go test ./internal/rule/...`
- `pnpx cspell lint` and `golangci-lint run --new-from-rev=origin/main` on changed files

## Checklist

### DIFF
--- a/internal/plugins/typescript/rules/no_redeclare/no_redeclare_test.go
+++ b/internal/plugins/typescript/rules/no_redeclare/no_redeclare_test.go
@@ -501,6 +501,13 @@ func TestNoRedeclareRule(t *testing.T) {
 					{MessageId: "redeclaredAsBuiltin", Line: 1, Column: 5},
 				},
 			},
+			// eval is in the shared ECMAScript built-in table.
+			{
+				Code: "var eval = 0;",
+				Errors: []rule_tester.InvalidTestCaseError{
+					{MessageId: "redeclaredAsBuiltin", Line: 1, Column: 5},
+				},
+			},
 			// `Promise` lives in lib.es2015.promise — covered by TS lib detection.
 			{
 				Code: "var Promise = 0;",
--- a/internal/rules/no_global_assign/no_global_assign_test.go
+++ b/internal/rules/no_global_assign/no_global_assign_test.go
@@ -260,6 +260,14 @@ func TestNoGlobalAssignRule(t *testing.T) {
 				},
 			},
 
+			// eval is a read-only ECMAScript built-in
+			{
+				Code: `eval = 1;`,
+				Errors: []rule_tester.InvalidTestCaseError{
+					{MessageId: "globalShouldNotBeModified", Line: 1, Column: 1},
+				},
+			},
+
 			// Postfix increment on builtin
 			{
 				Code: `String++;`,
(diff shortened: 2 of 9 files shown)


elecmonkey (approved):
LGTM

fansenze (approved):

PR #1531, opened 2026-08-03
feat: port rule no-unreachable-loop

swwind:
## Summary

Port the `no-unreachable-loop` rule from ESLint. It reports a loop whose body
allows only one iteration — every path through the body leaves the loop.

- Refactored the control-flow graph built for `no-useless-assignment` into a
  shared `internal/cfg`, parameterized over a consumer-supplied event type, so
  more than one rule can build on it. Consumers plug in through `Hooks[E]`.
- Added `Hooks.Statement`, which runs as each statement is reached from a
  position control can reach, and `Hooks.Loop`, which runs where control flows
  back into a loop for another iteration.
- Ported `no-unreachable-loop` on top of those two hooks.

One behavioral difference from ESLint: a loop carrying more than one label
repeats when a `continue` names any of them, so
`outer: inner: while (a) { continue outer; }` is accepted.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/no-unreachable-loop

## Checklist


internal/cfg/helpers.go
```diff
@@ -0,0 +1,95 @@
...
+// builder keeps the edge.)
+func labelsOf(node *ast.Node) []string {
+	var labels []string
+	current := node
+	for parent := current.Parent; parent != nil && parent.Kind == ast.KindLabeledStatement; parent = parent.Parent {
+		labeled := parent.AsLabeledStatement()
+		if labeled.Statement != current {
+			break
+		}
+		labels = append(labels, labeled.Label.Text())
```
elecmonkey, line 34: **[P1] Keep only ESLint's loop label when recording continue targets**

Collecting every directly wrapping label causes a `continue` targeting an outer label to mark the loop as repeatable, while ESLint's code-path analyzer only associates the innermost label with that loop.

For example:

```js
outer: inner: while (condition) {
  continue outer;
}
```

ESLint 10.8.0 reports `invalid`, but Rslint reports nothing. The same mismatch occurs for `do...while`, `for`, and `for...of`, and for either of the outer labels on a triply labeled loop. A `continue` targeting the innermost label remains valid in both implementations.

This behavior is intentionally locked in by the extras tests, but the rule port should reproduce ESLint's observable code-path behavior. Please avoid letting outer wrapping labels generate the `Loop` event used by `no-unreachable-loop`, and move these cases to the invalid set.

internal/cfg/statements.go
```diff
@@ -0,0 +1,574 @@
...
+	if label != nil {
+		name = label.Text()
+	}
+	for i := len(b.jumps) - 1; i >= 0; i-- {
+		target := b.jumps[i]
+		if target.continueTo == nil {
+			continue
+		}
+		if name == "" || slices.Contains(target.labels, name) {
+			b.loop(target.loop)
```
elecmonkey, line 522: **[P1] Do not mark finally-continue paths as repeats when ESLint does not**

Calling the loop hook for every resolved `continue` makes Rslint treat a `continue` in `finally` as a back edge even when the `try` or `catch` path reached the `finally` only through `return` or `throw`.

```js
function run() {
  for (;;) {
    try {
      return;
    } finally {
      continue;
    }
  }
}
```

ESLint 10.8.0 reports `invalid`; Rslint reports nothing. The mismatch also occurs for `throw`, conditional `continue`, `while`, `for`, `for...in`, `for...of`, nested `finally` blocks, and multi-label variants. `do...while` does not have this mismatch because its repeat event comes from the test rather than directly from the `continue`.

Although the Rslint result reflects the JavaScript runtime outcome, this rule is defined by ESLint's code-path events. Please only emit the loop hook in the same target-segment states where ESLint raises `onCodePathSegmentLoop`, and change the intentional-difference cases to expect diagnostics.

internal/cfg/helpers.go
```diff
@@ -0,0 +1,95 @@
...
+	if node == nil {
+		return false
+	}
+	switch node.Kind {
+	case ast.KindTrueKeyword, ast.KindRegularExpressionLiteral:
+		return true
+	case ast.KindNumericLiteral:
+		return utils.NormalizeNumericLiteral(node.AsNumericLiteral().Text) != "0"
+	case ast.KindBigIntLiteral:
+		text := node.AsBigIntLiteral().Text
```
elecmonkey: **[P2] Normalize BigInt literals before deciding whether a test is truthy**

Comparing the raw tsgo BigInt text only against `"0n"` and `"0"` misclassifies hexadecimal zero literals as truthy:

```js
while (0x0n) {
  visit();
}

for (;;) {
  break;
}
```

ESLint treats `0x0n` as falsy, so the second loop is reachable and reports `invalid`. Rslint treats the first loop as infinite, considers the second loop unreachable, and reports nothing.

This affects `0x0n`, `0X0n`, padded forms such as `0x00n`, numeric separators such as `0x0_0n`, and both `while` and `for` tests. Decimal, binary, and octal zero forms currently happen to align because tsgo normalizes those spellings differently.

Please use `utils.NormalizeBigIntLiteral` here and compare the normalized value against `"0"` instead of inspecting the raw text.

elecmonkey (changes requested):
I compared this port against ESLint 10.8.0 at its current `main`, ran the complete migrated upstream suite, and exercised the shared CFG with an additional 303-case differential matrix.

The existing `no-unreachable-loop` and `no-useless-assignment` tests pass. Of the generated differential cases, 269 match ESLint and 34 mismatches group into the three issues below. Two groups are intentionally locked in as differences, but this port is expected to preserve upstream rule behavior rather than substitute runtime-natural control-flow semantics.

internal/cfg/statements.go
elecmonkey: resolve as Rslint's behavior more reasonable

internal/cfg/helpers.go
swwind: Fixed in 6d4b5a8d. Agreed this is a plain bug rather than a difference: `0x0n` is falsy, so `while (0x0n)` should not mark the loop infinite.

`isAlwaysTruthyTest` now runs the literal through `utils.NormalizeBigIntLiteral` and compares the result against `"0"`. Added invalid cases for `0n`, `0x0n`, `0X0n`, `0x00n`, `0x0_0n`, `0b0n`, `0o0n` and the `for (;0x0n;)` test position, plus valid cases for `1n`, `0x1n` and `0x10000000000000000n` so the arbitrary-precision path stays covered.

internal/cfg/statements.go
swwind: Resolved per your follow-up — keeping the current behavior, since a `continue` in `finally` really does override the `return` or `throw` that left the `try` and control flows back into the loop.

The two cases stay marked as an intentional difference in `no_unreachable_loop_extras_test.go`.

internal/cfg/helpers.go
swwind: I'd like to keep this one as a difference as well, on the same grounds as the `finally` case. Both labels in `outer: inner: while (a) { continue outer; }` name the same `while` statement, so `continue outer` really does begin another iteration:

```js
let n = 0;
outer: inner: while (n < 3) { n++; continue outer; }
// n === 3
```

The loop runs to completion, so "body allows only one iteration" is a false positive upstream: its code path analysis attaches only the innermost label to the loop, and the back edge of a jump naming an outer one is lost. Reproducing it would mean reproducing the false positive, and it would leave the rule inconsistent with the `finally` case we just settled.

There is a second cost too. `labelsOf` collecting every wrapping label is what makes `break outer` resolve to the right target; only the `Loop` event is in question here, so matching upstream would mean a separate innermost-label-only lookup for that one event.

The cases stay marked as an intentional difference in `no_unreachable_loop_extras_test.go`.

elecmonkey (approved):
All three points are settled from my side.

- **BigInt normalization** — fixed in `6d4b5a8d`, and the added cases cover every spelling I raised.
- **`continue` in `finally`** — agreed earlier that Rslint's reading is the more reasonable one.
- **Multi-label `continue`** — I accept keeping this as an intentional difference. Both labels name the same loop statement, so the back edge is real and upstream's innermost-label-only attachment is the false positive; not worth a separate lookup just for the `Loop` event when `labelsOf` is also what makes `break outer` resolve correctly.

Clearing my earlier CHANGES_REQUESTED.

swwind:
closed as splitted

PR #1548, opened 2026-08-04
feat(rstest): add no-conditional-expect rule

elecmonkey:
## Summary

- add `rstest/no-conditional-expect` and enable it in the recommended preset
- share conditional-expect traversal between Jest and Rstest; this changes Jest behavior in two ways, detailed below
- support Rstest assertions from globals, `@rstest/core`, `import.meta.rstest`, test context, Browser Mode, and `@rstest/playwright`
- extend Rstest call parsing for parameterized tests, callback overloads, aliases, and official extension APIs
- add Go unit tests, JS integration tests, and rule documentation

## Jest-side impact

Extracting the traversal into `internal/utils/test_framework/rules/no_conditional_expect/` means `jest/no-conditional-expect` now runs the shared code. Two behavior changes follow, both widening what the rule reports. No case that previously reported stops reporting.

**1. `.catch` receivers that do not resolve to a name are now recognized.**

`isPromiseCatchCall` moved from `jestUtils.CalleeChainName` to `GetMemberEntries` + `JoinMemberEntries`. The two differ when the object part yields no name — `CalleeChainName` propagates `""` upward, the member-entry form drops the unresolvable prefix and keeps the `catch` entry:

```js
it('works', async () => {
  await promises[key].catch(error => expect(error).toBeInstanceOf(Error));
});
// before: not treated as a promise catch  →  after: reported
```

This aligns with upstream, whose `isCatchCall` only inspects the immediate property and never looks at the receiver. Covered by a new test.

**2. Promise-catch tracking is now a depth counter, not a bool — a deliberate divergence from `eslint-plugin-jest`.**

Upstream clears `inPromiseCatch` whenever any `.catch` call exits, which loses the enclosing state. Because chained `.catch` calls **nest in the AST** — the outermost call holds the inner ones in its callee subtree — this affects ordinary chains, not just lexically nested ones:

```js
it('works', async () => {
  await Promise.resolve()
    .catch(error => expect(error).toBeInstanceOf(Error))
    .catch(error => expect(error).toBeInstanceOf(Error))
    .catch(error => expect(error).toBeInstanceOf(Error));
});
// upstream and this rule before: 1 error
// after: 3 errors
```

Each callback is an independent assertion that only runs on its own rejection path, so three is the correct answer. `eslint-plugin-jest`'s own fixtures assert one error here, so the previous behavior was a faithful port rather than a bug — the change is intentional. Two existing test expectations were updated from 1 to 3 accordingly, and a nested-catch regression test was added.

Note that the rule already diverged from upstream on the same axis before this PR: test-case tracking has been a depth counter, where upstream uses a bool that a nested test registration ends early. Both divergences are now recorded in a package doc comment on the shared traversal so a later "restore parity" pass cannot silently revert them.

Behavior deliberately left identical to upstream: an `expect` that sits inside both a conditional and a `.catch` is still reported twice at the same range, since the two checks are independent.

## Related Links

N/A

## Checklist

### DIFF
--- a/internal/plugins/rstest/rules/no_conditional_expect/no_conditional_expect.go
+++ b/internal/plugins/rstest/rules/no_conditional_expect/no_conditional_expect.go
@@ -14,7 +14,7 @@ var NoConditionalExpectRule = shared.NewRule(shared.Config{
 		return shared.Runtime{
 			TestCallbackFunctions: callbacks.Functions,
 			ClassifyCall: func(node *ast.Node) (bool, bool) {
-				parsed := rstestUtils.ParseRstestFnCallWithOfficialExtensions(node, ctx)
+				parsed := callbacks.ParseFnCall(node)
 				return parsed != nil && parsed.Kind == rstestUtils.RstestFnTypeTest,
 					rstestUtils.IsRstestExpectCall(node, ctx, callbacks)
 			},
--- a/internal/plugins/rstest/rules/no_conditional_expect/no_conditional_expect.go
+++ b/internal/plugins/rstest/rules/no_conditional_expect/no_conditional_expect.go
@@ -1,0 +1,23 @@
+package no_conditional_expect
+
+import (
+	"github.com/microsoft/typescript-go/shim/ast"
+	rstestUtils "github.com/web-infra-dev/rslint/internal/plugins/rstest/utils"
+	"github.com/web-infra-dev/rslint/internal/rule"
+	shared "github.com/web-infra-dev/rslint/internal/utils/test_framework/rules/no_conditional_expect"
+)
+
+var NoConditionalExpectRule = shared.NewRule(shared.Config{
+	Name: "rstest/no-conditional-expect",
+	Prepare: func(ctx rule.RuleContext) shared.Runtime {
+		callbacks := rstestUtils.CollectRstestTestCallbacks(ctx)
+		return shared.Runtime{
+			TestCallbackFunctions: callbacks.Functions,
+			ClassifyCall: func(node *ast.Node) (bool, bool) {
+				parsed := rstestUtils.ParseRstestFnCallWithOfficialExtensions(node, ctx)
+				return parsed != nil && parsed.Kind == rstestUtils.RstestFnTypeTest,
+					rstestUtils.IsRstestExpectCall(node, ctx, callbacks)
+			},
+		}
+	},
+})
--- a/internal/plugins/rstest/utils/parse_rstest_expect.go
+++ b/internal/plugins/rstest/utils/parse_rstest_expect.go
@@ -31,23 +31,32 @@ func IsRstestExpectCall(
 	return isImportedOrGlobalExpectRoot(root, entries, ctx)
 }
 
+// findTopMostCallExpression walks up the call/member chain that node is the
+// head of. Ascending only continues while node stays on the callee side of its
+// parent: a call in argument or computed-key position heads its own chain.
 func findTopMostCallExpression(node *ast.Node) *ast.Node {
 	top := node
-	for parent := node.Parent; parent != nil; {
-		if parent.Kind == ast.KindParenthesizedExpression {
-			parent = parent.Parent
-			continue
-		}
-		if parent.Kind == ast.KindCallExpression {
+	current := node
+	for parent := current.Parent; parent != nil; {
+		switch parent.Kind {
+		case ast.KindParenthesizedExpression:
+		case ast.KindCallExpression:
+			if parent.AsCallExpression().Expression != current {
+				return top
+			}
 			top = parent
-			parent = parent.Parent
-			continue
+		case ast.KindPropertyAccessExpression:
+			if parent.AsPropertyAccessExpression().Expression != current {
+				return top
+			}
+		case ast.KindElementAccessExpression:
+			if parent.AsElementAccessExpression().Expression != current {
+				return top
+			}
+		default:
+			return top
 		}
-		if parent.Kind != ast.KindPropertyAccessExpression &&
-			parent.Kind != ast.KindElementAccessExpression {
-			break
-		}
-		parent = parent.Parent
+		current, parent = parent, parent.Parent
 	}
 	return top
 }
--- a/internal/plugins/rstest/utils/parse_rstest_expect.go
+++ b/internal/plugins/rstest/utils/parse_rstest_expect.go
@@ -1,0 +1,137 @@
+package utils
+
+import (
+	"github.com/microsoft/typescript-go/shim/ast"
+	"github.com/web-infra-dev/rslint/internal/rule"
+	testFramework "github.com/web-infra-dev/rslint/internal/utils/test_framework"
+)
+
+func IsRstestExpectCall(
+	node *ast.Node,
+	ctx rule.RuleContext,
+	callbacks RstestTestCallbacks,
+) bool {
+	if node == nil || node.Kind != ast.KindCallExpression || findTopMostCallExpression(node) != node {
+		return false
+	}
+
+	if isImportMetaRstestExpectCall(node) {
+		return true
+	}
+
+	entries := testFramework.GetMemberEntries(node)
+	if len(entries) == 0 || entries[0].Node == nil || entries[0].Node.Kind != ast.KindIdentifier {
+		return false
+	}
+
+	root := entries[0].Node
+	if isContextExpectRoot(root, entries, ctx, callbacks) {
+		return true
+	}
+	return isImportedOrGlobalExpectRoot(root, entries, ctx)
+}
+
+func findTopMostCallExpression(node *ast.Node) *ast.Node {
+	top := node
+	for parent := node.Parent; parent != nil; {
+		if parent.Kind == ast.KindParenthesizedExpression {
+			parent = parent.Parent
+			continue
+		}
+		if parent.Kind == ast.KindCallExpression {
+			top = parent
+			parent = parent.Parent
+			continue
+		}
+		if parent.Kind != ast.KindPropertyAccessExpression &&
+			parent.Kind != ast.KindElementAccessExpression {
+			break
+		}
+		parent = parent.Parent
+	}
+	return top
+}
+
+func isContextExpectRoot(
+	root *ast.Node,
+	entries []testFramework.MemberEntry,
+	ctx rule.RuleContext,
+	callbacks RstestTestCallbacks,
+) bool {
+	if ctx.TypeChecker == nil {
+		return false
+	}
+	symbol := ctx.TypeChecker.GetSymbolAtLocation(root)
+	if symbol == nil {
+		return false
+	}
+	if callbacks.ContextExpectNames[symbol] {
+		return true
+	}
+	return callbacks.ContextReceivers[symbol] &&
+		len(entries) > 1 &&
+		entries[1].Name == "expect"
+}
+
+func isImportedOrGlobalExpectRoot(
+	root *ast.Node,
+	entries []testFramework.MemberEntry,
+	ctx rule.RuleContext,
+) bool {
+	localName := root.AsIdentifier().Text
+	var symbol *ast.Symbol
+	if ctx.TypeChecker != nil {
+		symbol = ctx.TypeChecker.GetSymbolAtLocation(root)
+	}
+
+	if name, _, ok := resolveImportMetaRstestBinding(symbol); ok {
+		return name == "expect"
+	}
+	if isImportMetaRstestObjectAlias(symbol) {
+		return len(entries) > 1 && entries[1].Name == "expect"
+	}
+
+	for _, module := range []string{RstestImportModule, RstestPlaywrightImportModule} {
+		name, _, _ := testFramework.ResolveFunctionIdentifierReferenceFromSymbol(
+			localName,
+			root,
+			symbol,
+			ctx.SourceFile,
+			module,
+		)
+		if name == "expect" {
+			return true
+		}
+		if testFramework.IsModuleNamespaceSymbol(symbol, module) &&
+			len(entries) > 1 &&
+			entries[1].Name == "expect" {
+			return true
+		}
+	}
+	return false
+}
+
+func isImportMetaRstestExpectCall(node *ast.Node) bool {
+	call := node.AsCallExpression()
+	if call == nil {
+		return false
+	}
+	_, parts, _, ok := parseImportMetaRstestChain(call.Expression)
+	return ok && len(parts) > 0 && parts[0].name == "expect"
+}
+
+func isImportMetaRstestObjectAlias(symbol *ast.Symbol) bool {
+	if symbol == nil {
+		return false
+	}
+	for _, declaration := range symbol.Declarations {
+		if declaration == nil || declaration.Kind != ast.KindVariableDeclaration {
+			continue
+		}
+		initializer := declaration.AsVariableDeclaration().Initializer
+		if isImportMetaRstest(initializer) {
+			return true
+		}
+	}
+	return false
+}
--- a/internal/plugins/rstest/utils/parse_rstest_fn.go
+++ b/internal/plugins/rstest/utils/parse_rstest_fn.go
@@ -2,6 +2,7 @@ package utils
 
 import (
 	"github.com/microsoft/typescript-go/shim/ast"
+	"github.com/microsoft/typescript-go/shim/scanner"
 	"github.com/web-infra-dev/rslint/internal/rule"
 	internalUtils "github.com/web-infra-dev/rslint/internal/utils"
 	testFramework "github.com/web-infra-dev/rslint/internal/utils/test_framework"
@@ -38,27 +39,79 @@ type rstestResolvedAPI struct {
 	state        rstestAPIState
 	originalNode *ast.Node
 	mode         RstestImportMode
+	profile      rstestAPIProfile
 	// members collects every chain member applied to the resolved API,
 	// including members consumed while following same-file const aliases.
-	members []string
+	members           []string
+	parameterizedKind RstestParameterizedKind
 }
+
+type rstestAPIProfile uint8
+
+const (
+	rstestProfileCore rstestAPIProfile = iota
+	rstestProfilePlaywright
+)
 
 // ParseRstestFnCall parses a final Rstest test/describe registration call.
 // Factory calls such as test.each(cases), test.runIf(condition), and
 // test.extend(fixtures) are intentionally not returned.
 func ParseRstestFnCall(node *ast.Node, ctx rule.RuleContext) *ParsedRstestFnCall {
+	return parseRstestFnCall(node, ctx, false, false)
+}
+
+func ParseRstestFnCallWithOfficialExtensions(
+	node *ast.Node,
+	ctx rule.RuleContext,
+) *ParsedRstestFnCall {
+	return parseRstestFnCall(node, ctx, true, true)
+}
+
+func parseRstestFnCall(
+	node *ast.Node,
+	ctx rule.RuleContext,
+	includeImportMeta bool,
+	includePlaywright bool,
+) *ParsedRstestFnCall {
 	if node == nil || node.Kind != ast.KindCallExpression {
 		return nil
 	}
 
 	call := node.AsCallExpression()
 	root, parts, rootInvoked, ok := parseRstestChain(call.Expression)
-	if !ok || rootInvoked || root == nil {
-		return nil
+	var resolved rstestResolvedAPI
+	consumed := 0
+	if ok && !rootInvoked && root != nil {
+		resolved, consumed, ok = resolveRstestRoot(
+			root,
+			parts,
+			ctx,
+			nil,
+			0,
+			includeImportMeta,
+			includePlaywright,
+		)
+	} else if includeImportMeta {
+		var importMetaRoot *ast.Node
+		importMetaRoot, parts, rootInvoked, ok = parseImportMetaRstestChain(call.Expression)
+		if ok && !rootInvoked && importMetaRoot != nil && len(parts) > 0 {
+			state := directRstestAPIState(rstestProfileCore, parts[0].name)
+			if state != rstestAPIInvalid && parts[0].invocation == rstestNotInvoked {
+				resolved = rstestResolvedAPI{
+					name:         parts[0].name,
+					state:        state,
+					originalNode: parts[0].node,
+					mode:         RSTEST_IMPORT_MODE,
+					profile:      rstestProfileCore,
+				}
+				root = importMetaRoot
+				consumed = 1
+			} else {
+				ok = false
+			}
+		}
 	}
-
-	resolved, consumed, ok := resolveRstestRoot(root, parts, ctx, nil, 0)
-	if !ok {
+	if !ok || root == nil {
 		return nil
 	}
 
@@ -94,7 +147,10 @@ func ParseRstestFnCall(node *ast.Node, ctx rule.RuleContext) *ParsedRstestFnCall
 		})
 	}
 
-	localName := root.AsIdentifier().Text
+	localName := "import.meta.rstest"
+	if root.Kind == ast.KindIdentifier {
+		localName = root.AsIdentifier().Text
+	}
 	return &ParsedRstestFnCall{
 		ParsedCall: testFramework.ParsedCall{
 			Name:          resolved.name,
@@ -114,7 +170,8 @@ func ParseRstestFnCall(node *ast.Node, ctx rule.RuleContext) *ParsedRstestFnCall
 				},
 			},
 		},
-		Parameterized: parameterized,
+		Parameterized:     parameterized,
+		ParameterizedKind: resolved.parameterizedKind,
 	}
 }
 
@@ -186,12 +243,83 @@ func parseRstestChain(node *ast.Node) (*ast.Node, []rstestChainPart, bool, bool)
 	}
 }
 
+func parseImportMetaRstestChain(node *ast.Node) (*ast.Node, []rstestChainPart, bool, bool) {
+	node = ast.SkipParentheses(node)
+	if node == nil {
+		return nil, nil, false, false
+	}
+	if ast.IsOptionalChain(node) {
+		return nil, nil, false, false
+	}
+	if isImportMetaRstest(node) {
+		return node, nil, false, true
+	}
+
+	switch node.Kind {
+	case ast.KindPropertyAccessExpression:
+		property := node.AsPropertyAccessExpression()
+		root, parts, rootInvoked, ok := parseImportMetaRstestChain(property.Expression)
+		if !ok {
+			return nil, nil, false, false
+		}
+		nameNode := property.Name()
+		if nameNode == nil || nameNode.Kind != ast.KindIdentifier {
+			return nil, nil, false, false
+		}
+		return root, append(parts, rstestChainPart{
+			name: nameNode.AsIdentifier().Text,
+			node: nameNode,
+		}), rootInvoked, true
+	case ast.KindElementAccessExpression:
+		element := node.AsElementAccessExpression()
+		root, parts, rootInvoked, ok := parseImportMetaRstestChain(element.Expression)
+		if !ok {
+			return nil, nil, false, false
+		}
+		nameNode := ast.SkipParentheses(element.ArgumentExpression)
+		name, ok := internalUtils.GetStaticStringLiteralValue(nameNode)
+		if !ok || name == "" {
+			return nil, nil, false, false
+		}
+		return root, append(parts, rstestChainPart{
+			name: name,
+			node: nameNode,
+		}), rootInvoked, true
+	case ast.KindCallExpression:
+		call := node.AsCallExpression()
+		root, parts, rootInvoked, ok := parseImportMetaRstestChain(call.Expression)
+		if !ok {
+			return nil, nil, false, false
+		}
+		if len(parts) == 0 {
+			return nil, nil, false, false
+		}
+		if parts[len(parts)-1].invocation != rstestNotInvoked {
+			return nil, nil, false, false
+		}
+		parts[len(parts)-1].invocation = rstestCallInvoked
+		return root, parts, rootInvoked, true
+	case ast.KindTaggedTemplateExpression:
+		tagged := node.AsTaggedTemplateExpression()
+		root, parts, rootInvoked, ok := parseImportMetaRstestChain(tagged.Tag)
+		if !ok || len(parts) == 0 || parts[len(parts)-1].invocation != rstestNotInvoked {
+			return nil, nil, false, false
+		}
+		parts[len(parts)-1].invocation = rstestTagInvoked
+		return root, parts, rootInvoked, true
+	default:
+		return nil, nil, false, false
+	}
+}
+
 func resolveRstestRoot(
 	root *ast.Node,
 	parts []rstestChainPart,
 	ctx rule.RuleContext,
 	visited map[*ast.Symbol]bool,
 	depth int,
+	includeImportMeta bool,
+	includePlaywright bool,
 ) (rstestResolvedAPI, int, bool) {
 	if root == nil || root.Kind != ast.KindIdentifier || depth > 16 {
 		return rstestResolvedAPI{}, 0, false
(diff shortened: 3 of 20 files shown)


internal/plugins/rstest/utils/parse_rstest_expect.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	root := entries[0].Node
+	if isContextExpectRoot(root, entries, ctx, callbacks) {
+		return true
+	}
+	return isImportedOrGlobalExpectRoot(root, entries, ctx)
+}
+
+func findTopMostCallExpression(node *ast.Node) *ast.Node {
```
elecmonkey, line 37: **🔴 High — `expect` in argument position is never detected.**

This walks up `node.Parent` and treats **any** parent `CallExpression` as the top of the chain, without checking that the node is on the callee side. An `expect` chain passed as an argument is therefore skipped entirely:

```js
test("case", () => { if (c) foo(expect(b).toBe(1)); });
// rstest: 0 errors  |  jest: 1 error

test("case", () => { if (c) expect(a).toBe(expect.stringContaining("x")); });
// inner expect missed
```

The fix is to track the current child and only ascend while it stays in callee position:

```go
if parent.Kind == ast.KindCallExpression {
    if parent.AsCallExpression().Expression != current { return top } // argument position
    top = parent
}
```

The same guard is needed on the property/element-access branches, otherwise a computed key such as `results[expect(v).toBe(1)]` is swallowed the same way.

internal/plugins/rstest/utils/parse_rstest_fn.go
```diff
@@ -335,3 +543,65 @@ func applyRstestChainPart(state rstestAPIState, part rstestChainPart) rstestAPIS
...
 }
+
+func applyResolvedRstestChainPart(resolved *rstestResolvedAPI, part rstestChainPart) bool {
+	state := applyRstestChainPart(resolved.profile, resolved.state, part)
+	if state == rstestAPIInvalid {
+		resolved.state = rstestAPIInvalid
+		return false
+	}
+	resolved.state = state
+	resolved.members = append(resolved.members, part.name)
```
elecmonkey: **🟡 Medium — `Members` and `MemberEntries` diverge for aliased chains.**

`applyResolvedRstestChainPart` appends to `resolved.members`, and it is also called from the alias loop inside `resolveRstestRoot` — so alias-internal parts end up in `Members`. `MemberEntries` is still built only from the outer `parts[consumed:]` (line 138).

```js
const t = test.skip;
t('x', fn);
// Members = ["skip"], MemberEntries = []
```

No rstest rule reads `Members` today, so this is inert right now. But roughly a dozen Jest rules key off `Members`, and this is exactly the asymmetry the next port will trip over. Pick one: build both from the full resolved chain, or keep `Members` outer-only.

internal/utils/test_framework/rules/no_conditional_expect/no_conditional_expect.go
```diff
@@ -0,0 +1,171 @@
...
+		Id:          "conditionalExpect",
+		Description: "Avoid calling `expect` conditionally",
+	}
+}
+
+func isPromiseCatchCall(node *ast.Node) bool {
+	if node == nil || node.Kind != ast.KindCallExpression {
+		return false
+	}
+	name := calleeChainName(node.AsCallExpression().Expression)
```
elecmonkey, line 49: **🟡 Medium — untested behavior change in the shipped Jest rule.**

This swaps `jestUtils.CalleeChainName` for `GetMemberEntries` + `JoinMemberEntries`. The two diverge when the object part yields no name: `CalleeChainName` propagates `""` upward, whereas `GetMemberEntries` drops the unresolvable prefix and still returns the `catch` entry.

```js
test("case", async () => { await p[key].catch(e => expect(e).toBe(1)); });
// before: not a catch call  |  after: reported
```

This actually matches upstream more closely — `isCatchCall` only inspects the immediate property — so I read it as a fix. But it silently changes an already-released rule and no test covers it. Worth an invalid case in the Jest rule tests to pin the new behavior down.

internal/utils/test_framework/rules/no_conditional_expect/no_conditional_expect.go
```diff
@@ -0,0 +1,171 @@
...
+
+func NewRule(config Config) rule.Rule {
+	return rule.Rule{
+		Name:   config.Name,
+		Schema: rule.EmptyArraySchema,
+		Run: func(ctx rule.RuleContext, options []any) rule.RuleListeners {
+			runtime := config.Prepare(ctx)
+			testCaseDepth := 0
+			conditionalDepth := 0
+			inPromiseCatch := false
```
elecmonkey: **ℹ️ Info — parity quirk, matches upstream.**

`inPromiseCatch` is a bool, so nested catches clear it early:

```js
test("case", async () => {
  await a().catch(() => {
    b().catch(() => { expect(1).toBe(1); }); // reported
    expect(2).toBe(2);                        // missed
  });
});
```

Upstream has the same bug, so this is not a regression. But `testCaseDepth` right above is already a counter where upstream uses a bool — given that divergence is deliberate, making this a counter too would be consistent and is a one-line change.

internal/plugins/rstest/rules/no_conditional_expect/no_conditional_expect.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+var NoConditionalExpectRule = shared.NewRule(shared.Config{
+	Name: "rstest/no-conditional-expect",
+	Prepare: func(ctx rule.RuleContext) shared.Runtime {
+		callbacks := rstestUtils.CollectRstestTestCallbacks(ctx)
+		return shared.Runtime{
+			TestCallbackFunctions: callbacks.Functions,
+			ClassifyCall: func(node *ast.Node) (bool, bool) {
+				parsed := rstestUtils.ParseRstestFnCallWithOfficialExtensions(node, ctx)
```
elecmonkey: **🔵 Low — every call node is parsed twice.**

`CollectRstestTestCallbacks` (line 13) does its own full-file walk calling `ParseRstestFnCallWithOfficialExtensions` on every `CallExpression`, and then `ClassifyCall` calls it **again** on every `CallExpression` during the rule traversal.

Given the recent perf pass on `main` (#1530, #1524, #1526, ...), a `map[*ast.Node]*ParsedRstestFnCall` memo built in `Prepare` and shared by both passes is cheap and removes half the work.

elecmonkey (commented):
Reviewed at `7bf52368`. `go build`, `go vet`, the Go rule tests and the JS integration test all pass locally. Each finding below was verified with a throwaway probe test.

### Summary

| # | Severity | Finding |
|---|---|---|
| 1 | High | `expect` in argument position is never detected (rstest only) |
| 2 | Medium | `Members` and `MemberEntries` disagree for aliased chains |
| 3 | Medium | Untested behavior change in the shipped Jest rule |
| 4 | Low | Perf: every call node is parsed twice |
| 5 | Info | Two upstream-parity quirks, confirmed *not* regressions |

### Upstream-parity quirks (no action needed)

I diffed the traversal against `eslint-plugin-jest/src/rules/no-conditional-expect.ts`. Both of these match upstream exactly, so I would leave them as-is:

- An `expect` inside both a conditional **and** a `.catch` reports **twice** at the same range.
- `inPromiseCatch` is a bool rather than a stack, so in nested catches the inner exit clears the flag and a later `expect` in the outer catch is missed.

Worth noting though: upstream's `inTestCase` is also a bool while this port already promotes it to a depth counter, so the implementation has deliberately diverged toward correctness there. Making `inPromiseCatch` a counter too would be self-consistent and is a one-line change. See the inline comment.

### What works well

- The `internal/utils/test_framework/rules/` extraction is the right seam. The Jest rule collapses to a 12-line adapter with no semantic drift — the `KindBinaryExpression` pre-check it drops was already subsumed by `isConditionalNode`.
- Skipping the context binding for `.each` while wiring it at parameter index 1 for `.for` is correct: `.each` genuinely has no user-visible context parameter.
- Go test coverage is thorough — globals, named/aliased imports, require destructuring, namespace imports, all four `import.meta.rstest` shapes, context destructuring, `extend`, the options overload, hoisted function callbacks, and Playwright, plus negative cases for vitest / `@jest/globals` / `@playwright/test`.
- Registration is complete: `all.go`, the preset, the preset test, and `rstest.config.mts`. `website/rule-releases.json` is script-generated, so nothing to add there.

internal/plugins/rstest/utils/parse_rstest_expect.go
elecmonkey: Resolved in d2f1df21.

`findTopMostCallExpression` now tracks the current child node and only ascends while it stays on the callee side of its parent. A call in argument or computed-key position heads its own chain and is classified on its own.

Regression tests added for all three shapes:

```js
if (condition) log(expect(value).toBe(1));                       // argument position
if (condition) expect(value).toBe(expect.stringContaining("x")); // 2 errors, inner + outer
if (condition) results[expect(value).toBe(1)] = 1;               // computed key
```

Verified: `go vet` clean, all rstest rule tests pass, the Jest rule is unaffected, and the JS integration test still passes.

Also pushed 41f2e7d0, which converts the `part.name` if/else in `applyResolvedRstestChainPart` to a tagged switch — that was the `QF1003` staticcheck failure in CI. `golangci-lint run` over the affected packages is now clean.

internal/plugins/rstest/utils/parse_rstest_fn.go
elecmonkey: Resolved in d8e66bd4 — went with keeping `Members` outer-only.

`rstestResolvedAPI.members` and the append in `applyResolvedRstestChainPart` are gone; `Members` is now built in the same loop as `MemberEntries`, from `parts[consumed:]`. `applyResolvedRstestChainPart` is still called on the alias path, so state advancement and `parameterizedKind` are untouched — only the `Members` side effect is dropped. Net 3 lines.

Before/after, measured with a throwaway probe rule that dumps the parse result:

| input | Members (before) | Members (after) | MemberEntries | ParameterizedKind |
|---|---|---|---|---|
| `const t = test.skip; t('x', fn)` | `[skip]` | `[]` | `[]` | `""` |
| `const appTest = test.extend({}); appTest.concurrent(...)` | `[extend concurrent]` | `[concurrent]` | `[concurrent]` | `""` |
| `const forCase = test.for([...]); forCase(...)` | `[for]` | `[]` | `[]` | `"for"` |
| `test.skip.each([1])('case', cb)` | `[skip each]` | `[skip each]` | `[skip each]` | `"each"` |
| `rstest.test.skip('case', cb)` | `[skip]` | `[skip]` | `[skip]` | `""` |
| `describe.only('suite', cb)` | `[only]` | `[only]` | `[only]` | `""` |

Only the aliased rows move, and they all move into strict agreement with `MemberEntries`. Non-alias paths are byte-identical. Row 3 is the guard that matters: `ParameterizedKind` is still `"for"`, so the alias-internal `.for` is still captured — that is what places the `.for` callback context at parameter index 1, and it does not regress.

Impact: zero current consumers. `ParameterizedKind` (`test_callback.go`) and `Parameterized` (`no_identical_title.go`) are separate fields derived from `resolved.state`, and no rstest rule reads `Members` today. The Jest side is unaffected — it goes through `ParseJestFnCall`, a different producer. All rstest rule tests, `go vet` and `golangci-lint` are clean.

`Members` now means "the member chain written at the call site", with `Members[i] == MemberEntries[i].Name` holding by construction.

One thing worth recording for later ports: this makes `Members` empty for `const t = test.skip; t('x', fn)`. A future rule that decides via `slices.Contains(Members, "skip")` will therefore not follow aliases. That is the rule's call to make, and the right way to expose it is a semantic field alongside `ParameterizedKind` — not by having the rule guess whether something in `Members` came from the call site or from an alias declaration.

internal/utils/test_framework/rules/no_conditional_expect/no_conditional_expect.go
elecmonkey: Covered in cab3043c — added an invalid case next to the existing `.catch` tests in the Jest rule:

```js
it('works', async () => {
  await promises[key].catch(error => expect(error).toBeInstanceOf(Error));
});
```

With a comment noting why it holds: upstream `isCatchCall` only inspects the immediate property, so the receiver need not resolve to a name. That pins the post-refactor behavior so a future change back to name-based chain resolution cannot silently revert it.

internal/plugins/rstest/rules/no_conditional_expect/no_conditional_expect.go
elecmonkey: Addressed in 3982f466.

`RstestTestCallbacks` now carries a `*rstestFnCallCache` (`map[*ast.Node]*ParsedRstestFnCall`). The collection walk populates it as it visits each `CallExpression`, and `ClassifyCall` reads through `callbacks.ParseFnCall(node)` instead of parsing again.

The cache is miss-then-compute rather than lookup-only, so correctness does not depend on the two traversals covering exactly the same node set — a node the collection walk never reached is simply parsed on demand and memoized, instead of being misclassified as a non-test call.

Verified: `go build`, `go vet`, all rstest rule tests, `golangci-lint` (0 issues), and the JS integration test all pass.

internal/utils/test_framework/rules/no_conditional_expect/no_conditional_expect.go
elecmonkey: Fixed in 75e9f320 — went with diverging from upstream for correctness, consistent with the `testCaseDepth` precedent right above it.

`inPromiseCatch` is now `promiseCatchDepth`, incremented on entering a `.catch` call and decremented on its exit, mirroring how test-case tracking already works.

This turned out to have a wider blast radius than the nested-catch case I originally reported. Two existing Jest tests changed from 1 error to 3:

```js
it('works', async () => {
  await Promise.resolve()
    .catch(error => expect(error).toBeInstanceOf(Error))
    .catch(error => expect(error).toBeInstanceOf(Error))
    .catch(error => expect(error).toBeInstanceOf(Error));
});
```

Chained `.catch` calls **nest in the AST** — the outermost call has the inner ones in its callee subtree. Upstream's bool is therefore cleared by every inner exit, and only one of the three callbacks is visited while the flag is set. I checked `eslint-plugin-jest`'s own fixtures and it does assert exactly one error for both of these, so this port was faithfully reproducing upstream; it is not a porting mistake.

Three is the semantically correct answer: each callback is an independent assertion that only runs on its own rejection path. Both expectations updated with a comment pointing at the rationale.

Also added a package doc comment on the shared traversal recording the deliberate divergences, so a future "restore parity with upstream" pass does not silently revert either counter:

1. test-case tracking is a depth counter (upstream ends the enclosing test early on a nested registration)
2. promise-catch tracking is a depth counter (this change)

and noting what is intentionally kept identical — an `expect` inside both a conditional and a `.catch` is still reported twice at the same range, since the two checks are independent.

Regression tests added for the nested-catch shape on both the Jest and Rstest rules. Verified: `go build`, `go vet`, `golangci-lint` (0 issues), all rstest rule tests, the Jest rule tests, and both JS integration suites.

elecmonkey:
@codex review

elecmonkey:
@codex review

internal/plugins/rstest/utils/parse_rstest_fn.go
```diff
@@ -296,14 +494,18 @@ func directRstestAPIState(name string) rstestAPIState {
 	}
 }
 
-func applyRstestChainPart(state rstestAPIState, part rstestChainPart) rstestAPIState {
+func applyRstestChainPart(profile rstestAPIProfile, state rstestAPIState, part rstestChainPart) rstestAPIState {
```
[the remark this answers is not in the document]
elecmonkey: The mechanism you describe is accurate, but the premise isn't: Rstest does not support modifiers after `.each` / `.for`, so rejecting these chains is the correct behavior rather than a false negative.

Per the `@rstest/core@0.11.3` typings, modifiers live on the API object, not on what the parameterized factory returns:

- `TestAPI` (`index.d.ts:3652`) carries `each`, `for`, `only`, `skip`, `concurrent`, `sequential`, `fails`, `todo`, `runIf`, `skipIf`.
- `TestEachFn` (`:3719`) and `TestForFn` (`:3757`) both return `TestCall<...>`, a plain callable type with no members.
- `DescribeEachFn` / `DescribeForFn` (`:981`, `:990`) likewise return `DescribeCall<...>`.

So `test.for(cases).concurrent('case', fn)` is a type error. The supported spelling is `test.concurrent.for(cases)('case', fn)`, and the state machine already handles that order: `rstestAPITest` transitions on `concurrent` and then on `for` into `rstestAPIParameterizedTest`, so the callback is collected and `no-conditional-expect` reports inside it.

Two side notes: the `applyRstestChainPart` switch is pre-existing shared code that this PR does not change in this respect (the base revision has the same structure minus the `profile` parameter), and line 497 is `directRstestAPIState` rather than the function described.

Not making a change here.

swwind (commented):
Reviewed at `c45e010e`. Each finding reproduced with a throwaway test, then deleted.

Three nil-initializer panics in the new files. Same root cause: `ast.SkipParentheses` is linknamed upstream, where `IsOuterExpression` does `switch node.Kind` — nil panics instead of returning nil. A `VariableDeclaration` without initializer has `Initializer == nil`.

| Code | ESLint (jest) | This PR |
|---|---|---|
| `let logger;` + `logger();` | no error | **panic** |
| `let cb; cb = () => {}; test("case", cb);` | no error | **panic** |
| `test("case", callback);` + `let state;` | no error | **panic** |
| `test("case", ({ expect }) => { if (c) expect(v).toBe(1); }, TIMEOUT);` (`TIMEOUT` undeclared) | n/a | 0 errors, expected 1 |

Row 1 is the serious one — a bare `let x;` plus any call, no rstest import needed, and the rule ships as `'error'` in the recommended preset.

Guarding before `SkipParentheses` is the convention elsewhere: `internal/rules/one_var/one_var.go:191`, `internal/plugins/import/rules/namespace/namespace.go:165`.

Separately, I checked the `.each` context skip in `recordRstestCallback` and it is correct: `@rstest/core@0.11.3` types `TestEachFn` (`index.d.ts:3719`) as `(...args: [...T])` with no context parameter; only `TestForFn` (`:3757`) has one.

internal/plugins/rstest/utils/parse_rstest_expect.go
```diff
@@ -0,0 +1,146 @@
...
+func isImportMetaRstestObjectAlias(symbol *ast.Symbol) bool {
+	if symbol == nil {
+		return false
+	}
+	for _, declaration := range symbol.Declarations {
+		if declaration == nil || declaration.Kind != ast.KindVariableDeclaration {
+			continue
+		}
+		initializer := declaration.AsVariableDeclaration().Initializer
+		if isImportMetaRstest(initializer) {
```
swwind, line 141: **🔴 High — panics on a bare `let x;` anywhere in the file.**

```js
let logger;
export function run() { logger(); }
```

`isImportMetaRstest` calls `SkipParentheses` before its own nil check. Reached for every top-most call expression, before any module check:

```
ast.SkipParentheses (utilities.go:830)
utils.isImportMetaRstest (parse_rstest_fn.go:577)
utils.isImportMetaRstestObjectAlias (parse_rstest_expect.go:141)
```

Guard inside `isImportMetaRstest` — it has several callers.

internal/plugins/rstest/utils/test_callback.go
```diff
@@ -0,0 +1,255 @@
...
+	name := argument.AsIdentifier().Text
+	declaration := internalUtils.GetDeclaration(ctx.TypeChecker, argument)
+	if declaration == nil {
+		return rstestCallbackInfo{name: name}
+	}
+	switch declaration.Kind {
+	case ast.KindFunctionDeclaration:
+		return rstestCallbackInfo{functionNode: declaration, name: name}
+	case ast.KindVariableDeclaration:
+		initializer := ast.SkipParentheses(declaration.AsVariableDeclaration().Initializer)
```
swwind: **🟡 Medium — same nil initializer.**

```js
let cb; cb = () => {}; test("case", cb);
```

Line 116 above nil-checks `argument` only *after* `SkipParentheses`, which survives only because `arguments[1]` is never nil.

internal/plugins/rstest/utils/test_callback.go
```diff
@@ -0,0 +1,255 @@
...
+			declaration := node.AsFunctionDeclaration()
+			if declaration != nil && declaration.Name() != nil {
+				name = declaration.Name().Text()
+				function = node
+			}
+		case ast.KindVariableDeclaration:
+			declaration := node.AsVariableDeclaration()
+			if declaration != nil && declaration.Name() != nil && declaration.Name().Kind == ast.KindIdentifier {
+				name = declaration.Name().AsIdentifier().Text
+				initializer := ast.SkipParentheses(declaration.Initializer)
```
swwind: **🟡 Medium — same crash, from a full-file walk.**

Hits every `KindVariableDeclaration` with an identifier name, so any `let x;` in a file that also has an unresolvable callback name crashes:

```js
test("case", callback); // undeclared -> pending
let state;              // panic
```

internal/plugins/rstest/utils/test_callback.go
```diff
@@ -0,0 +1,255 @@
...
+	ctx rule.RuleContext,
+	parsed *ParsedRstestFnCall,
+	call *ast.CallExpression,
+) rstestCallbackInfo {
+	if call == nil || call.Arguments == nil || len(call.Arguments.Nodes) < 2 {
+		return rstestCallbackInfo{}
+	}
+
+	arguments := call.Arguments.Nodes
+	if len(arguments) >= 3 {
```
swwind, line 104: **🔵 Low — third-argument preference swallows the real callback.**

Returns the `arguments[2]` result on a function node *or* a name. An undeclared third argument yields `{name: "TIMEOUT"}`, so `arguments[1]` is never inspected:

```js
test("case", ({ expect }) => { if (condition) expect(value).toBe(1); }, TIMEOUT);
// 0 errors; expected 1
```

Only bites when `expect` comes from the test context. An imported `TIMEOUT` is fine — `GetDeclaration` returns the import specifier and it falls through. Prefer `arguments[2]` only when it resolves to a function node.

internal/plugins/rstest/utils/parse_rstest_expect.go
elecmonkey: Confirmed and fixed in 6c40393d.

Reproduced your exact snippet against the rule tester; the panic stack matches what you posted:

```
parse_rstest_expect.go:31 -> :99 -> :141 -> parse_rstest_fn.go:577
```

Root cause is as you describe: `ast.SkipParentheses` reaches `IsOuterExpression` (`typescript-go/internal/ast/utilities.go:792`), which dereferences `node.Kind` with no nil check. Your point about the module check is right too — the path is entered from `isImportedOrGlobalExpectRoot` for any identifier-rooted call, so no rstest import is needed to trigger it.

Guarded inside `isImportMetaRstest` as you suggested rather than at the call site. The other three callers were only safe by accident: `:246` passes a call expression that is never nil, and `:389` / `:453` happen to sit behind `isConstRstestVariableDeclaration`, which already checks `Initializer != nil`.

Regression test added.

internal/plugins/rstest/utils/test_callback.go
elecmonkey: Confirmed and fixed in 0d8c2404.

`let cb; cb = () => {}; test("case", cb);` panics at `test_callback.go:136`, stack `:70 -> :110 -> :136`.

Fixed both halves of what you pointed at: the nil check now precedes `SkipParentheses` at the top of `resolveRstestCallbackArgument`, and the `KindVariableDeclaration` branch checks `Initializer` before skipping parentheses. You were right that the top-of-function ordering was surviving only because `arguments[1]` is never nil — the actual crash was in the initializer below it, but both are now ordered correctly.

Regression test added.

internal/plugins/rstest/utils/test_callback.go
elecmonkey: Confirmed and fixed in adcac3ad.

`test("case", callback); let state;` panics at `test_callback.go:168`, with the stack showing the `:182`/`:183` full-file walk driving it.

The conjunction you identified is what makes it reachable: an unresolvable callback name is needed to populate `pending` and start the walk, and then any uninitialized declaration in the file crashes it. Guarded `declaration.Initializer` before `SkipParentheses`.

Regression test added using your snippet.

internal/plugins/rstest/utils/test_callback.go
elecmonkey: Confirmed and fixed in 55e2d91a.

Verified the three cases:

| input | before | after |
| --- | --- | --- |
| no third argument | 1 error | 1 error |
| `..., TIMEOUT)` with `TIMEOUT` undeclared | **0 errors** | 1 error |
| `const TIMEOUT = 5000;` then the same call | 1 error | 1 error |

Your read of the mechanism is exact: `resolveRstestCallbackArgument` returns `{name: "TIMEOUT"}` from the `declaration == nil` branch, and `|| info.name != ""` made that win over `arguments[1]`. The imported-`TIMEOUT` case behaves as you said — `GetDeclaration` yields an import specifier, neither switch case matches, and it falls through to the zero value.

Took your suggested fix (prefer `arguments[2]` only when it resolves to a function node) and kept one fallback: if `arguments[1]` resolves to nothing at all, an unresolved name in the third position is still handed to the pending walk. That preserves the `(name, options, fn)` shape, e.g. `test("case", { timeout: 1 }, handler)` with `handler` declared below. Covered by a test.

swwind (commented):
Final pass at `55e2d91a`. All four findings from the last round are fixed, each with a regression test. CI green apart from the pending macOS Go job.

### 🟡 The same nil-initializer panic still lives in the Jest twin

`internal/plugins/jest/rules/no_conditional_expect/no_conditional_expect.go:80` is the Jest copy of the pending-callback walk you fixed in adcac3ad, with the identical bug:

```js
it("case", cb);
let cb;
```

```
ast.SkipParentheses (utilities.go:830)
no_conditional_expect.resolvePendingTestCallbackNames.func1 (no_conditional_expect.go:80)
```

Reproduced against the rule tester; either declaration order triggers it, and the rule ships as `'error'` in the Jest recommended preset. Pre-existing on `main` — the diff leaves that function untouched — so your call whether to fold it in. Same guard as the rstest side: nil-check `vd.Initializer` before `SkipParentheses`.

### The chained-`.catch` divergence deserves an explicit sign-off

|---|---|---|
| `.catch(cb).catch(cb).catch(cb)` | 1 error | 3 errors |
| nested `.catch`, assertion after the inner one | 1 error | 2 errors |

I asked for the counter last round, so this is on me, but the blast radius is wider than the nested case I described. Three is the semantically correct count and it is well documented — the cost is that a released rule changes output on an ordinary chain, and the two edited expectations live in `tests/eslint-plugin-jest/`, which is ported verbatim from upstream. That suite no longer mirrors upstream there. If the project's line is "match upstream, bugs included", the counter should go back to a bool. Either answer is defensible; it should be a maintainer's call.

elecmonkey:
Folded the Jest one in — 72d7d4e4.

Confirmed first: reverted only the source change, kept the new test, and got your stack exactly.

```
utilities.go:792 -> :818 -> :830 -> no_conditional_expect.go:80 (from :88)
```

Same guard as the rstest side. Regression tests cover both declaration orders.

Worth noting the correct pattern was already present one layer down: `jestUtils.testCallbackInitializerFunction` (`parse_jest_fn.go:440-442`) nil-checks the initializer before `SkipParentheses`, which is why `ResolveTestCallbackFunction` survives the same input and the name reaches `pending` instead of crashing there. The rule-local walk was the only copy that missed it.

On the chained-`.catch` divergence: keeping the counter. Three is the semantically correct count, the package doc records the divergence and the reasoning, and the two edited expectations carry a comment pointing at it. Agreed the ported suite no longer mirrors upstream at those two cases; that is the intended trade.

elecmonkey:
Did a wider nil-deref sweep after 72d7d4e4, since the three panics you found were all the same shape. **No further reachable panics.** Details, in case it saves you a pass.

**Static scan.** Every `SkipParentheses` / `ast.Is*` / `GetMemberEntries` call site in the files this PR touches, checked for a nil-reachable argument:

| site | status |
| --- | --- |
| `test_callback.go` :130 :154 :187 | guarded (0d8c2404, adcac3ad) |
| `jest/.../no_conditional_expect.go:83` | guarded (72d7d4e4) |
| `parse_rstest_fn.go:582` (`isImportMetaRstest`) | guarded (6c40393d) |
| `parse_rstest_fn.go` :199 :271 :594 (`element.ArgumentExpression`) | not nil-reachable |
| `GetMemberEntries` x3 (`parse_rstest_expect.go:22`, `rstest.go:53`, shared rule :54) | already correct — `member_chain.go:14-20` nil-checks *before* `SkipParentheses` |
| `isConditionalNode` / `IsLogicalExpression` (shared rule :66) | only reached inside the `KindBinaryExpression` arm |

**Empirical.** 58 adversarial snippets — uninitialized declarations in every position, element access with non-literal/empty/computed keys, optional chains at each link, arity edge cases (`test()`, `test(...args)`, extra args), alias cycles (`const a = b; const b = a;`), tagged-template `.each`/`.for`, context destructuring including rest patterns, `["catch"]` and computed catch receivers. Run against `rstest/no-conditional-expect`, `jest/no-conditional-expect`, and the three sibling rules that share `parse_rstest_fn.go` (`no-identical-title`, `no-disabled-tests`, `no-commented-out-tests`, `no-mocks-import`), each subtest wrapped in `recover()`. Zero panics; the only failures were cases that legitimately report. Throwaway files, deleted.

**Two latent instances of the same pattern, currently unreachable.** `parseRstestChain` (`:171`) and `isImportMeta` (`:602`) both call `SkipParentheses` before their own nil check, exactly like `isImportMetaRstest` did. They survive only because of their call sites: `parseRstestChain` gets `call.Expression` / `property.Expression` / `element.Expression`, all non-nil, and its one initializer-fed caller (`:389`) sits behind `isConstRstestVariableDeclaration`, which already checks `Initializer != nil`; `isImportMeta` only ever gets a non-nil `.Expression`. Same accident that was keeping the other three callers of `isImportMetaRstest` alive. Not fixing them in this PR since nothing reaches them today, but happy to fold in the two-line hardening if you'd rather not leave the footgun loaded.

elecmonkey:
Folded in the hardening — 936b68ab. Footgun unloaded.

Turned out to be three helpers, not two: `parseImportMetaRstestChain` (`:241`) has the same shape as `parseRstestChain` (`:170`) and `isImportMeta` (`:604`). All three now nil-check before `SkipParentheses`.

Since nothing reaches them through the rule tester, I pinned them with white-box tests in `package utils` (`parse_rstest_fn_nil_test.go`) rather than snippet fixtures — calling each helper with `nil` directly. Confirmed they fail without the change:

```
--- FAIL: TestParseRstestChainAcceptsNil
panic: runtime error: invalid memory address or nil pointer dereference
	parse_rstest_fn.go:171
```

The file also covers `isImportMetaRstest`, which passes either way now — it doubles as a regression guard for 6c40393d, so the guard you asked for there cannot quietly come back out.

That package had no test files before, so this is a new `parse_rstest_fn_nil_test.go`.

Verified: gofmt clean, `go vet` clean, golangci-lint 0 issues, full jest + rstest + test_framework suites green.

internal/plugins/rstest/utils/parse_rstest_fn.go
```diff
@@ -186,12 +238,86 @@ func parseRstestChain(node *ast.Node) (*ast.Node, []rstestChainPart, bool, bool)
...
 
+func parseImportMetaRstestChain(node *ast.Node) (*ast.Node, []rstestChainPart, bool, bool) {
+	if node == nil {
+		return nil, nil, false, false
+	}
+	node = ast.SkipParentheses(node)
+	if node == nil {
+		return nil, nil, false, false
+	}
+	if ast.IsOptionalChain(node) {
```
swwind: 🔵 Low — one `?.` anywhere in the chain turns the rule off for that call.

`NodeFlagsOptionalChain` is set on every node of an optional chain, not only the one carrying `?.`, so this bail rejects the whole expression:

```js
if (import.meta.rstest) {
  import.meta.rstest.test("case", () => {
    if (condition) import.meta.rstest?.expect(value).toBe(1); // 0 errors
  });
}
```

`parseRstestChain` has no equivalent bail, so `test?.("case", fn)` and `const api = import.meta.rstest; api?.expect(v)` are both recognised — the two paths disagree on the same syntax. If the bail is deliberate, a one-line comment saying why would keep a later reader from removing it.

internal/plugins/rstest/utils/test_callback.go
```diff
@@ -0,0 +1,275 @@
...
+	if callbacks.fnCalls == nil {
+		return nil
+	}
+	return callbacks.fnCalls.parse(node)
+}
+
+type rstestCallbackInfo struct {
+	functionNode *ast.Node
+	name         string
+	parsed       *ParsedRstestFnCall
```
swwind: 🔵 Low — `parsed` is write-only.

It is assigned at lines 108, 118 and 122 and never read: `CollectRstestTestCallbacks` passes its own `parsed` to both `recordRstestCallback` and `pending`. Dropping the field also drops the `parsed` parameter of `resolveRstestTestCallback`, which exists only to feed it.

swwind (commented):
Final pass at `936b68ab`. All four nil-initializer panics from the last round are fixed, each with a regression test, and the Jest twin at `no_conditional_expect.go:80` is guarded too. Lint&Check, Test Go (ubuntu), Test WASM and Test Rust are green; the remaining jobs are still running.

Two low-severity items left, neither a blocker.

| Code | Expected | This PR |
|---|---|---|
| `import.meta.rstest.test(...)` with `import.meta.rstest?.expect(v).toBe(1)` | 1 error | **0 errors** |
| `import.meta.rstest?.test(...)` with `import.meta.rstest?.expect(...)` | 1 error | **0 errors** |
| `const api = import.meta.rstest; api?.test(...)` with `api?.expect(...)` | 1 error | 1 error |
| `test?.("case", ...)` with a conditional `expect` | 1 error | 1 error |

`parseImportMetaRstestChain` bails on `ast.IsOptionalChain`, and that flag is set on every node of an optional chain, so a single `?.` disables both test-call and expect-call recognition. `parseRstestChain` has no such bail, which is why the last two rows still work. Verified with a throwaway rule-tester run, then deleted.

Separately, the `Members` / `MemberEntries` thread is marked resolved by `d8e66bd4`, but that commit is not in the branch: `parse_rstest_fn.go:152` still reads `Members: resolved.members`, so alias-internal members are included while `MemberEntries` stays call-site-only. I think that is the right behaviour — `rstest/no-disabled-tests` relies on it for `const skipped = test.skip; skipped('x', fn)` — so no code change needed, the thread just no longer matches HEAD.

The chained-`.catch` divergence is now documented in the package doc comment, both test files and the PR body. I raised it as a maintainer call last round and have no further objection to shipping it as documented.

elecmonkey:
Both items addressed, plus a correction on the `Members` thread — you were right to flag it, but the conclusion goes the other way.

### Optional chains on `import.meta.rstest` — 439f8a81

Removed the `ast.IsOptionalChain` bail in `parseImportMetaRstestChain`. `parseRstestChain` never had one, which is exactly why the two paths disagreed on the same syntax; there is no reason for the `import.meta` path to be stricter.

Worth adding to your table: `import.meta.rstest` is typed **optional** —

```ts
// @rstest/core/importMeta.d.ts
interface ImportMeta { readonly rstest?: typeof import('@rstest/core'); }
```

so under `strictNullChecks` the `?.` is not a stylistic choice. I checked with the repo's tsc: in your snippet the outer `import.meta.rstest.test(...)` narrows fine, but the access **inside the callback** is `TS2532: Object is possibly 'undefined'` — narrowing does not survive the function boundary. The asymmetric spelling you wrote is the shape TypeScript forces.

This also means the suite was pinning the bug: the second row of your table existed verbatim in `no_conditional_expect_test.go` as a **valid** case. Moved to invalid along with the first row.

### `Members` / `MemberEntries` — d135faac

Not just a stale thread. `Members` being the call-site-only syntactic view is a recorded contract: same length and order as `MemberEntries`, `Members[i] == MemberEntries[i].Name`, no alias-internal members in either. Anything that has to hold across an alias is supposed to be a semantic field in the style of `ParameterizedKind` — explicitly *not* something a rule recovers by scanning `Members`, because it cannot tell a call-site member from an alias-internal one.

What happened is a rebase artifact of mine: the commit enforcing that contract was dropped when this branch was rebased, because `main` had independently resolved the same divergence the *other* way (`Members: resolved.members`). I read the two as equivalent. They are opposites, and I took `main`'s without noticing.

So `no-disabled-tests` reading `slices.Contains(parsed.Members, "skip")` is the anti-pattern, not the requirement — it only works because the contract is currently violated. My mistake for shipping it that way in the first place; that rule is mine too, from a PR that opened earlier and merged later than this one.

Restored:

- `Members` is built in the same loop as `MemberEntries`, from `parts[consumed:]`.
- Added `Skipped` / `Todo` on `ParsedRstestFnCall`, set in `applyResolvedRstestChainPart` next to `parameterizedKind`, so they accumulate on the alias chain too.
- `no-disabled-tests` now reads `parsed.Skipped` / `parsed.Todo`. Its alias cases (`const skipped = test.skip; skipped('case', fn)`, the two-level `base.skip`, the namespaced `rstest.describe.skip`) all still pass, now through the semantic channel.

Pinned with two probe rules in `internal/plugins/rstest/utils`: one reports whenever `Members` and `MemberEntries` disagree, one reports whenever a semantic flag is set. The contract cases are listed valid against the first, the alias cases invalid against the second — so a future change that feeds alias members back into `Members` fails the first, and one that drops them from the semantic fields fails the second. Verified both directions by reintroducing the old behaviour: 5 cases fail as expected, clean again after reverting.

### Write-only `parsed` field — 416b34e6

Confirmed and removed. Nothing read it: the two `.parsed` hits elsewhere in the file are `rstestFnCallCache.parsed`, an unrelated field that happens to share the name. `CollectRstestTestCallbacks` passes its own `parsed` to both `recordRstestCallback` and `pending`, so the copy on `rstestCallbackInfo` was never consulted.

Dropped the field, the three assignments, and the `parsed` parameter of `resolveRstestTestCallback` that existed only to feed it. Pure deletion, no behaviour change — one insertion, six deletions.

Verified: gofmt, `go vet`, golangci-lint 0 issues, full jest + rstest + test_framework suites, and the three JS suites (12/12).

swwind (dismissed):

swwind (approved):

PR #1551, opened 2026-08-04
feat: resolve globals to off/readonly/writable levels

swwind:
## Summary

`languageOptions.globals` and `/* global */` comments resolved to a bool — declared or not — so nothing downstream could tell a read-only global from a writable one. They now resolve to ESLint's three access levels: off, readonly and writable. Besides matching ESLint, this lets the `globals` npm package be used as-is, since it spells the two declared levels as booleans.

Details:

- Config values, inline comment settings, `ctx.Globals` and `ctx.ConfigGlobals` carry an access level instead of a bool. A name neither source mentions reads back as unset, so a rule indexing the map falls back to what it already knows about that name.
- `no-global-assign` no longer reports assignments to a global declared writable, and now reports assignments to one declared read-only even when it is not an ECMAScript built-in. Both match ESLint.
- An inline setting that spells none of the three levels is dropped instead of declaring the name, leaving the config or built-in setting in place. A name written without a setting stays read-only.
- The remaining rules reading `ctx.Globals` were updated mechanically and behave as before.
- Rule test cases author globals the way an ESLint config does, and fail on a value ESLint would reject.
- The Node plugin worker already seeded all three levels into its scope manager, so it needed no changes.
- Docs: the config reference gains guidance on declaring runtime globals with the `globals` package, `no-global-assign` documents the access levels, and the architecture and rule-porting notes describe the new context type.

## Checklist


elecmonkey:
@codex review

elecmonkey:
why conflict... we are so easy to be conflicted..

fansenze (approved):

PR #1558, opened 2026-08-04
feat(typescript-eslint): implement restrict-template-expressions

swwind:
## Summary

`restrict-template-expressions` was registered, but its listener body was a placeholder that never reported — the rule appeared enabled while passing every input. This implements it against typescript-eslint 8.65.0.

- Union constituents must all be allowed, intersection constituents need one allowed, `StringLike` always passes.
- `allowAny` / `allowArray` / `allowBoolean` / `allowNever` / `allowNullish` / `allowNumber` / `allowRegExp`, with upstream's defaults — `allowArray` and `allowNever` off, the rest on. `allowArray` recurses into the number index type of arrays and tuples.
- `allow` accepts both the string shorthand and the `TypeOrValueSpecifier` object form, and matches a type's base types as well, so subclasses of an allowed type are allowed. Defaults to `Error`, `URL` and `URLSearchParams` from `lib`.
- Tagged template literals are skipped.
- Added `restrict_template_expressions.schema.json` and declared it on the rule. Options are parsed from the plain `options []any` array via `utils.GetOptionsMap`.

The option set follows upstream's schema exactly, so `allowInline` — which upstream does not have — is not part of it; `allow` accepts bare strings and covers the same case.

The rule docs listed `${num}` as incorrect, which contradicts `allowNumber` defaulting to `true`. Rewritten to describe actual behavior and document the options.

`tests/typescript-eslint/rules/restrict-template-expressions.test.ts` is enabled in `rstest.config.mts`. Its 19 snapshots were cross-checked against the report positions and message text declared in the upstream test file.

## Checklist

### DIFF
--- a/internal/plugins/typescript/rules/restrict_template_expressions/restrict_template_expressions.go
+++ b/internal/plugins/typescript/rules/restrict_template_expressions/restrict_template_expressions.go
@@ -154,7 +154,7 @@ var RestrictTemplateExpressionsRule = rule.CreateRule(rule.Rule{
 				return true
 			}
 
-			if matchesTypeOrBaseType(ctx.TypeChecker, func(t *checker.Type) bool {
+			if len(opts.Allow) > 0 && matchesTypeOrBaseType(ctx.TypeChecker, func(t *checker.Type) bool {
 				return utils.TypeMatchesSomeSpecifier(t, opts.Allow, nil, ctx.Program)
 			}, innerType, map[*checker.Type]struct{}{}) {
 				return true
--- a/internal/plugins/typescript/rules/restrict_template_expressions/restrict_template_expressions.go
+++ b/internal/plugins/typescript/rules/restrict_template_expressions/restrict_template_expressions.go
@@ -1,117 +1,207 @@
 package restrict_template_expressions
 
 import (
+	_ "embed"
+	"encoding/json"
+	"fmt"
+
 	"github.com/microsoft/typescript-go/shim/ast"
+	"github.com/microsoft/typescript-go/shim/checker"
 	"github.com/web-infra-dev/rslint/internal/rule"
+	"github.com/web-infra-dev/rslint/internal/utils"
 )
 
+//go:embed restrict_template_expressions.schema.json
+var schemaJSON []byte
+
 type RestrictTemplateExpressionsOptions struct {
-	AllowNumber  bool     `json:"allowNumber"`
-	AllowBoolean bool     `json:"allowBoolean"`
-	AllowAny     bool     `json:"allowAny"`
-	AllowNullish bool     `json:"allowNullish"`
-	AllowRegExp  bool     `json:"allowRegExp"`
-	AllowNever   bool     `json:"allowNever"`
-	AllowArray   bool     `json:"allowArray"`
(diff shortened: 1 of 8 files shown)


internal/plugins/typescript/rules/restrict_template_expressions/restrict_template_expressions.go
```diff
@@ -1,117 +1,207 @@
...
-			if allowNumber, ok := optsMap["allowNumber"].(bool); ok {
-				opts.AllowNumber = allowNumber
+			if utils.IsTypeFlagSet(innerType, checker.TypeFlagsStringLike) {
+				return true
 			}
-			if allowBoolean, ok := optsMap["allowBoolean"].(bool); ok {
-				opts.AllowBoolean = allowBoolean
+
+			if len(opts.Allow) > 0 && matchesTypeOrBaseType(ctx.TypeChecker, func(t *checker.Type) bool {
+				return utils.TypeMatchesSomeSpecifier(t, opts.Allow, nil, ctx.Program)
```
elecmonkey, line 158: `allow` is not fully upstream-compatible. On macOS, upstream accepts `{ from: 'file', path: './LOCAL-TYPES.ts', name: 'Local' }` for a type declared in `local-types.ts`, while Rslint reports it because the shared matcher compares paths case-sensitively. Package matching also differs (`demo` matches `demo-pkg` upstream but not here). Please align these specifier semantics.

internal/plugins/typescript/rules/restrict_template_expressions/restrict_template_expressions.go
```diff
@@ -1,117 +1,207 @@
...
-				// }
-			}
-		},
-	}
-}
+				for _, span := range node.AsTemplateExpression().TemplateSpans.Nodes {
+					expression := span.AsTemplateSpan().Expression
+					expressionType := utils.GetConstrainedTypeAtLocation(ctx.TypeChecker, expression)
+					if !recursivelyCheckType(expressionType) {
+						ctx.ReportNode(expression, buildInvalidTypeMessage(ctx.TypeChecker.TypeToString(expressionType)))
```
elecmonkey, line 201: `TypeToString` does not always match upstream message text. For `input: object | undefined; const value = input ?? 'fallback'`, upstream reports `object | "fallback"`, but Rslint reports `"fallback" | object`. Please preserve upstream ordering for diagnostic parity.

elecmonkey (changes requested):
Two undocumented upstream compatibility gaps remain.

internal/plugins/typescript/rules/restrict_template_expressions/restrict_template_expressions.go
swwind: This is a difference between TypeScript and typescript-go, not something we can fix here.

elecmonkey (dismissed):

internal/plugins/typescript/rules/restrict_template_expressions/restrict_template_expressions.go
swwind: fixed

fansenze (approved):

PR #1560, opened 2026-08-04
feat: port rule unicode-bom

swwind:
## Summary

Port the `unicode-bom` rule from ESLint. `"never"` (the default) disallows a leading Unicode byte order mark, U+FEFF; `"always"` requires one. Only the first position counts — a U+FEFF anywhere else is ordinary text. Diagnostics land on the zero-width range at offset 0, which renders as line 1, column 1, where ESLint reports it.

The mark is consumed while a file's bytes are decoded, so it reaches neither `SourceFile.Text()` nor any offset measured against it. `ctx.HasBOM` answers the question instead, reading it from whichever layer removed it: a VFS that tracks the mark answers directly, and anything else falls back to the file's own bytes. UTF-16 marks count, since those files are decoded to UTF-8 and lose their mark the same way. Source supplied by a caller rather than read from disk — an editor buffer, an API request, a test fixture — is stripped to match, so both shapes agree.

Fixes follow ESLint's own: inserting a mark is an insertion at the front of the text, and removing one is a range over `[-1, 0]`, a position ahead of the text because the mark is not in it. `ApplyRuleFixes` splits a leading mark off before applying fixes and puts it back afterwards, unless a fix asked for its removal.

### Behavior changes beyond the new rule

- **A byte order mark now survives `--fix`.** Fixes were applied to the decoded text and written back, so any file rewritten by any rule silently lost its mark. The CLI and the API now re-attach it before fixing.
- **A byte order mark no longer shifts columns.** Positions were measured against text that still counted the mark, putting every line-1 column one past where ESLint reports it. Two existing tests asserted the shifted values and have been corrected; `jsx_curly_spacing` had already flagged the divergence in a comment.
- **`fix.range` from the API matches ESLint**, including the `[-1, 0]` removal range, which is no longer clamped to 0.

### Editor behavior

The language server does not run this rule. An editor's document holds decoded text — VS Code turns a leading mark into the document's encoding, shown in the status bar as "UTF-8 with BOM" — so the mark is not in the document and no text edit adds or removes it. That leaves the file on disk as the only witness, and an unsaved buffer may already disagree with it. `rslint --fix`, which rewrites the file itself, is where the rule applies. A VS Code end-to-end suite pins this, starting with the fact it rests on: the fixture's bytes begin `EF BB BF` while its document text does not.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/unicode-bom
- Source code: https://github.com/eslint/eslint/blob/main/lib/rules/unicode-bom.js

## Checklist

### DIFF
--- a/internal/lsp/service.go
+++ b/internal/lsp/service.go
@@ -41,9 +41,12 @@ type lintPassResult struct {
 }
 
 // ruleFixToTextEdit converts a rule fix into an LSP TextEdit using the
-// source file's line map for position encoding.
+// source file's line map for position encoding. A fix reaching back before the
+// text — ESLint's [-1, 0] for a byte order mark — starts at the document's
+// first position: an editor document holds decoded text, and the mark is part
+// of the file's encoding rather than of that text.
 func ruleFixToTextEdit(sourceFile ast.SourceFileLike, fix rule.RuleFix) *lsproto.TextEdit {
-	startLine, startChar := scanner.GetECMALineAndUTF16CharacterOfPosition(sourceFile, fix.Range.Pos())
+	startLine, startChar := scanner.GetECMALineAndUTF16CharacterOfPosition(sourceFile, max(0, fix.Range.Pos()))
 	endLine, endChar := scanner.GetECMALineAndUTF16CharacterOfPosition(sourceFile, fix.Range.End())
 	return &lsproto.TextEdit{
 		Range: lsproto.Range{
--- a/internal/lsp/service.go
+++ b/internal/lsp/service.go
@@ -1052,6 +1052,10 @@ func lintSingleFile(
 
 	// Create collector function
 	diagnosticCollector := func(d rule.RuleDiagnostic) {
+		// A unicode-bom fix is not expressible as an editor text edit, so skip it.
+		if d.RuleName == "unicode-bom" {
+			d.FixesPtr = nil
+		}
 		diagnosticsLock.Lock()
 		defer diagnosticsLock.Unlock()
 		diagnostics = append(diagnostics, d)
--- a/internal/lsp/service.go
+++ b/internal/lsp/service.go
@@ -41,12 +41,9 @@ type lintPassResult struct {
 }
 
 // ruleFixToTextEdit converts a rule fix into an LSP TextEdit using the
-// source file's line map for position encoding. A fix reaching back before the
-// text — ESLint's [-1, 0] for a byte order mark — starts at the document's
-// first position: an editor document holds decoded text, and the mark is part
-// of the file's encoding rather than of that text.
+// source file's line map for position encoding.
 func ruleFixToTextEdit(sourceFile ast.SourceFileLike, fix rule.RuleFix) *lsproto.TextEdit {
-	startLine, startChar := scanner.GetECMALineAndUTF16CharacterOfPosition(sourceFile, max(0, fix.Range.Pos()))
+	startLine, startChar := scanner.GetECMALineAndUTF16CharacterOfPosition(sourceFile, fix.Range.Pos())
 	endLine, endChar := scanner.GetECMALineAndUTF16CharacterOfPosition(sourceFile, fix.Range.End())
 	return &lsproto.TextEdit{
 		Range: lsproto.Range{
@@ -1017,6 +1014,38 @@ func runLintWithProgramLoader(
 	), nil
 }
 
+// rulesSkippedInEditors names the rules the language server never runs. A rule
+// belongs here when what it checks is a property of the file's bytes rather
+// than of its text: an editor's document holds text that has already been
+// decoded, so such a property is neither visible in the document nor reachable
+// by a text edit.
+var rulesSkippedInEditors = map[string]bool{
+	// unicode-bom checks for a leading byte order mark. An editor decodes the
+	// mark into the document's encoding — VS Code shows it in the status bar as
+	// "UTF-8 with BOM" — so the document text never carries it and no text edit
+	// adds or removes it. The file on disk is the only remaining witness, and an
+	// unsaved buffer may already disagree with it. `rslint --fix`, which rewrites
+	// the file itself, is where the rule applies.
+	"unicode-bom": true,
+}
+
+// rulesServedToEditors drops the rules the language server never runs. The
+// input is a cached slice shared across files, so filtering builds a new one
+// and an unaffected configuration keeps the original.
+func rulesServedToEditors(rules []linter.ConfiguredRule) []linter.ConfiguredRule {
+	skipped := func(r linter.ConfiguredRule) bool { return rulesSkippedInEditors[r.Name] }
+	if !slices.ContainsFunc(rules, skipped) {
+		return rules
+	}
+	served := make([]linter.ConfiguredRule, 0, len(rules))
+	for _, configured := range rules {
+		if !skipped(configured) {
+			served = append(served, configured)
+		}
+	}
+	return served
+}
+
 func lintSingleFile(
 	program *compiler.Program,
 	sourceFile *ast.SourceFile,
@@ -1052,10 +1081,6 @@ func lintSingleFile(
 
 	// Create collector function
 	diagnosticCollector := func(d rule.RuleDiagnostic) {
-		// A unicode-bom fix is not expressible as an editor text edit, so skip it.
-		if d.RuleName == "unicode-bom" {
-			d.FixesPtr = nil
-		}
 		diagnosticsLock.Lock()
 		defer diagnosticsLock.Unlock()
 		diagnostics = append(diagnostics, d)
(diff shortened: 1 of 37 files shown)


internal/rules/unicode_bom/unicode_bom.go
```diff
@@ -0,0 +1,84 @@
...
+
+// UnicodeBomRule disallows a Unicode BOM.
+// https://eslint.org/docs/latest/rules/unicode-bom
+var UnicodeBomRule = rule.Rule{
+	Name:   "unicode-bom",
+	Schema: rule.NewSchema(schemaJSON),
+	Run: func(ctx rule.RuleContext, options []any) rule.RuleListeners {
+		// `never` is the only option the schema admits, and it is also the
+		// default, so there is nothing to read out of options.
+		if !strings.HasPrefix(ctx.SourceFile.Text(), bom) && !startsWithBOM(ctx.SourceFile.FileName()) {
```
elecmonkey: This reads the physical file even when `SourceFile` comes from an overlay. For example, a clean unsaved LSP/API buffer is still reported if the stale disk file has a BOM. Please detect the BOM from the active source frame instead of unconditionally consulting disk.

internal/rules/unicode_bom/unicode_bom.go
```diff
@@ -0,0 +1,84 @@
...
+		// wait for.
+		start := core.NewTextRange(0, 0)
+		ctx.ReportRangeWithDeferredFixes(start, rule.RuleMessage{
+			Id:          "unexpected",
+			Description: "Unexpected Unicode BOM (Byte Order Mark).",
+		}, func() []rule.RuleFix {
+			// An empty edit. Nothing in the text needs to change — the mark is
+			// not in it — but a file with any fix at all is rewritten from its
+			// decoded text, and that text has no mark. The write is the fix.
+			return []rule.RuleFix{rule.RuleFixReplaceRange(start, "")}
```
elecmonkey: This empty edit does not remove a BOM that is still present in overlay input, so `lintText({ fix: true })` returns unchanged code. It also fails through `lintFiles()`, whose JS result conversion re-adds the original disk BOM. Please make the fix produce BOM-free output across the public API paths.

elecmonkey (changes requested):
Two undocumented API/LSP compatibility issues remain. Please address the inline comments before merging.

swwind:
Both review points are real, and they trace back to one fact about how source text reaches a rule.

**Detection cannot come from the active source frame**

tsgo's VFS decodes a file's bytes before handing text to the parser, and that decode consumes a leading BOM (`decodeBytes`, `internal/vfs/internal/internal.go`). A file read from disk therefore carries no mark in `SourceFile.Text()` and none in any offset measured against it. Reading the file's own bytes is the only witness left.

Consulting disk *only* for files genuinely read from disk requires knowing whether the source came from an overlay. Nothing on `RuleContext` carries that. The only route is `ctx.Program.Host().FS()`, and the overlay sits beneath `programMetadataFS` and `parallelProgramFS`, neither of which forwards such a query. Answering it means a new overlay-identity interface in `internal/utils` plus forwarding in every wrapper — shared surface that each future VFS layer must remember to forward, or the false positive silently returns.

**The fix cannot produce BOM-free output through both API paths**

Two different causes:

- `lintText`: the overlay keeps the mark inside the text, so the empty edit at `[0, 0]` does not remove it. This half is a one-line rule change.
- `lintFiles`: Go's `Output` is already BOM-free, but `rslint.ts` deliberately re-prepends the BOM to `result.output` for any disk file that had one, so `outputFixes` writes the mark straight back. That is the correct default for every other rule and wrong only for this one.

The only rule-agnostic way to express "this fix removes the mark" is ESLint's own convention — `fixer.removeRange([-1, 0])`, a fix range starting ahead of the text. Four shared consumers then have to learn about negative offsets: `ApplyRuleFixes`, the API wire offset conversion, the LSP `TextEdit` conversion, and the JS re-prepend. A consumer that misses the clamp indexes a string with `-1`. The alternative, special-casing the rule name in the API layer, is worse.

A prototype covering both capabilities is about 130 lines across 8 shared files, none of it about `unicode-bom`, before the rule and its tests.

**Effect on shared test infrastructure**

An empty fix makes `RunRuleTester`'s multi-pass loop run its entire pass budget, because it breaks on `fixed == false` rather than on the code coming back unchanged. That is why this PR modifies `rule_tester.go`.

**Conclusion**

`unicode-bom` asks rslint two questions it is not currently built to answer: whether a given text came from disk or from a caller, and how a fix expresses removing something that is not in the text. Landing the port without both means shipping known false positives in the editor and an autofix that silently does nothing through `lintFiles`. We would rather withdraw this port than land either half-measure.

internal/utils/source_bom.go
```diff
@@ -0,0 +1,58 @@
...
+// text uses by the time rslint sees it.
+const BOM = "\uFEFF"
+
+// BOMSource is implemented by a VFS layer that knows whether the text it hands
+// out for a path began with a byte order mark. Source text always reaches the
+// linter with the mark already removed — ts-go's file reader decodes it away,
+// and [OverlayVFS] strips it from caller-supplied content so both shapes agree
+// — which leaves the layer that removed it as the only one able to answer.
+//
+// A wrapping VFS must forward this explicitly. Wrappers embed the vfs.FS
```
elecmonkey, line 20: **[P1] Forward the active overlay’s BOM state through `lintTrackingFS`**

`BOMSource` requires every wrapping VFS to forward `SourceHasBOM`, but `lintTrackingFS` does not implement it. Persistent standalone LSP programs use this wrapper as their `CompilerHost` filesystem, so `ctx.HasBOM()` fails the interface assertion and falls back to reading the physical file. This makes an unsaved clean buffer over a BOM-prefixed disk file report `unicode-bom`, while an unsaved BOM-prefixed buffer over a clean disk file is missed. Please forward the query to the current inner overlay, for example:

```go
func (fs *lintTrackingFS) SourceHasBOM(path string) bool {
    return utils.SourceHasBOM(fs.Inner, path)
}
```

A regression test should cover both stale-disk directions through the persistent LSP program path.

internal/lsp/service.go
```diff
@@ -1049,6 +1052,10 @@ func lintSingleFile(
 
 	// Create collector function
 	diagnosticCollector := func(d rule.RuleDiagnostic) {
+		// A unicode-bom fix is not expressible as an editor text edit, so skip it.
+		if d.RuleName == "unicode-bom" {
```
elecmonkey: **[P2] Preserve the expressible `"always"` fix in LSP**

This removes every fix produced by `unicode-bom`, but only the `"never"` fix is unrepresentable in an editor because its range is `[-1, 0]`. The `"always"` mode emits a normal `[0, 0]` insertion of `U+FEFF`, which `ruleFixToTextEdit` can represent and ESLint exposes as an autofix. Clearing `FixesPtr` by rule name therefore prevents both quick fix and fix-all from satisfying `"always"`. Please filter only fixes with negative positions—without mutating a potentially shared backing slice—and retain valid edits such as the `[0, 0]` insertion. Add an LSP regression test for `["error", "always"]` while keeping the existing `"never"` behavior covered.

elecmonkey (changes requested):
Please address the two inline findings before merging.

swwind:
Both findings pointed at the same thing: the editor path was being asked to serve a rule about a file's bytes. This revision drops `unicode-bom` from the language server entirely rather than trimming its fix.

An editor's document holds decoded text. VS Code turns a leading mark into the document's *encoding* — the status bar's "UTF-8 with BOM" — so the mark is not in the document, and no text edit adds or removes it. That leaves the file on disk as the only witness, and for an unsaved buffer the buffer may already disagree with it, which is exactly the divergence P1 describes. Neither direction of the rule can be shown or acted on there, so the server no longer produces the diagnostic. `rslint --fix`, which rewrites the file itself, is where the rule applies; the CLI and the API are unchanged.

**P2** — the rule-name check is gone. `lintSingleFile` filters the rule out of the active set before linting, building a fresh slice so the cached one the resolver hands out is untouched. Go tests cover both modes, each paired with the same config run off the editor path where it does report — the `[-1, 0]` removal under `never` and the `[0, 0]` insertion under `always` — so the silence is the server's doing and not a rule with nothing to say. The `max(0, …)` clamp in `ruleFixToTextEdit` went with it: no negative position reaches the editor path anymore.

**P1** — `ctx.HasBOM()` has exactly one caller, this rule, and it no longer runs in the language server, so `lintTrackingFS` never reaches that query. Forwarding it now would add a method with no live caller. Should a second rule ever need the mark in an editor, the forwarding and its stale-disk regression test belong with it.

The VS Code suite now enables `no-var` alongside and the fixture source declares one, so every assertion of silence runs against a file the server is demonstrably linting.

elecmonkey (approved):

fansenze (approved):

PR #1561, opened 2026-08-04
feat: port rule import/no-restricted-paths

swwind:
## Summary

Port the `import/no-restricted-paths` rule from eslint-plugin-import to rslint.

The rule enforces import boundaries between directories: each zone declares a `target` set of files and a `from` set they may not import, with optional `except` and `message`. `target`/`from` accept a directory path, a glob pattern, or an array of either; relative zone paths resolve against `basePath`, which defaults to the current working directory.

Extended glob syntax (`!(a)`, `@(a|b)`, `+(a)`, `?(a)`, `*(a)`) is matched literally rather than as a pattern, so zones written with it stay inactive. This is documented under "Differences from ESLint" in the rule doc, together with the supported alternatives.

## Related Links

- ESLint rule: https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-restricted-paths.md
- Source code: https://github.com/import-js/eslint-plugin-import/blob/main/src/rules/no-restricted-paths.js

## Checklist

### DIFF
--- a/internal/plugins/import/rules/no_restricted_paths/no_restricted_paths.go
+++ b/internal/plugins/import/rules/no_restricted_paths/no_restricted_paths.go
@@ -50,11 +50,12 @@ var NoRestrictedPathsRule = rule.Rule{
 		}
 
 		basePath := resolveBasePath(ctx, opts.basePath)
+		caseSensitive := ctx.Program.Host().FS().UseCaseSensitiveFileNames()
 
 		currentFilename := import_utils.GetPhysicalFilename(ctx)
 		matchingZones := make([]zone, 0, len(opts.zones))
 		for _, z := range opts.zones {
-			if isMatchingZone(z, basePath, currentFilename) {
+			if isMatchingZone(z, basePath, currentFilename, caseSensitive) {
 				matchingZones = append(matchingZones, z)
 			}
 		}
@@ -74,7 +75,7 @@ var NoRestrictedPathsRule = rule.Rule{
 
 			for i := range matchingZones {
 				if !built[i] {
-					validators[i] = makePathValidators(matchingZones[i].fromPaths, matchingZones[i].except, basePath)
+					validators[i] = makePathValidators(matchingZones[i].fromPaths, matchingZones[i].except, basePath, caseSensitive)
 					built[i] = true
 				}
 
@@ -165,23 +166,23 @@ func toStringSlice(raw any) []string {
 	return nil
 }
 
-func isMatchingZone(z zone, basePath string, currentFilename string) bool {
+func isMatchingZone(z zone, basePath string, currentFilename string, caseSensitive bool) bool {
 	for _, target := range z.targets {
-		if isMatchingTargetPath(currentFilename, tspath.ResolvePath(basePath, target)) {
+		if isMatchingTargetPath(currentFilename, tspath.ResolvePath(basePath, target), caseSensitive) {
 			return true
 		}
 	}
 	return false
 }
 
-func isMatchingTargetPath(fileName string, targetPath string) bool {
+func isMatchingTargetPath(fileName string, targetPath string, caseSensitive bool) bool {
 	if isGlob(targetPath) {
 		return utils.MatchGlob(targetPath, fileName)
 	}
-	return containsPath(fileName, targetPath)
+	return containsPath(fileName, targetPath, caseSensitive)
 }
 
-func makePathValidators(fromPaths []string, except []string, basePath string) []pathValidator {
+func makePathValidators(fromPaths []string, except []string, basePath string, caseSensitive bool) []pathValidator {
 	anyGlob := false
 	anyNonGlob := false
 	for _, from := range fromPaths {
@@ -209,7 +210,7 @@ func makePathValidators(fromPaths []string, except []string, basePath string) []
 		if anyGlob {
 			validators = append(validators, computeGlobPatternPathValidator(absoluteFrom, except))
 		} else {
-			validators = append(validators, computeAbsolutePathValidator(absoluteFrom, except))
+			validators = append(validators, computeAbsolutePathValidator(absoluteFrom, except, caseSensitive))
 		}
 	}
 	return validators
@@ -250,10 +251,10 @@ func computeGlobPatternPathValidator(absoluteFrom string, except []string) pathV
 	return validator
 }
 
-func computeAbsolutePathValidator(absoluteFrom string, except []string) pathValidator {
+func computeAbsolutePathValidator(absoluteFrom string, except []string, caseSensitive bool) pathValidator {
 	validator := pathValidator{
 		isPathRestricted: func(absoluteImportPath string) bool {
-			return containsPath(absoluteImportPath, absoluteFrom)
+			return containsPath(absoluteImportPath, absoluteFrom, caseSensitive)
 		},
 		hasValidExceptions: true,
 		invalidException: rule.RuleMessage{
@@ -268,7 +269,7 @@ func computeAbsolutePathValidator(absoluteFrom string, except []string) pathVali
 		// Upstream classifies the `from`-relative exception path with
 		// importType and rejects it when it comes out as "parent", which is
 		// exactly the set of paths that escape `from`.
-		if !containsPath(absoluteException, absoluteFrom) {
+		if !containsPath(absoluteException, absoluteFrom, caseSensitive) {
 			validator.hasValidExceptions = false
 			return validator
 		}
@@ -277,7 +278,7 @@ func computeAbsolutePathValidator(absoluteFrom string, except []string) pathVali
 
 	validator.isPathException = func(absoluteImportPath string) bool {
 		for _, absoluteException := range absoluteExceptions {
-			if containsPath(absoluteImportPath, absoluteException) {
+			if containsPath(absoluteImportPath, absoluteException, caseSensitive) {
 				return true
 			}
 		}
@@ -288,18 +289,35 @@ func computeAbsolutePathValidator(absoluteFrom string, except []string) pathVali
 
 // containsPath reports whether filePath is target itself or one of its
 // descendants, mirroring upstream's `path.relative(target, filePath)` check.
-func containsPath(filePath string, target string) bool {
+// Components are compared the way the host file system resolves names, which
+// is what Node's `path.win32.relative` does on a case-insensitive platform.
+func containsPath(filePath string, target string, caseSensitive bool) bool {
 	fileComponents := pathComponents(filePath)
 	targetComponents := pathComponents(target)
 	if len(targetComponents) > len(fileComponents) {
 		return false
 	}
 	for i, component := range targetComponents {
-		if component != fileComponents[i] {
+		if !equalPathComponent(component, fileComponents[i], caseSensitive) {
 			return false
 		}
 	}
+	// `path.relative` joins the components left over, and upstream reads a
+	// leading `..` in that relative path as "outside target". A first leftover
+	// component such as `..secret.js` therefore keeps the file out of target
+	// even though it names a real child of it.
+	if len(fileComponents) > len(targetComponents) &&
+		strings.HasPrefix(fileComponents[len(targetComponents)], "..") {
+		return false
+	}
 	return true
+}
+
+func equalPathComponent(a string, b string, caseSensitive bool) bool {
+	if caseSensitive {
+		return a == b
+	}
+	return strings.EqualFold(a, b)
 }
 
 // pathComponents splits a path into its root plus segments, dropping the empty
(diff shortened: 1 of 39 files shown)


internal/plugins/import/rules/no_restricted_paths/no_restricted_paths.md
```diff
@@ -0,0 +1,141 @@
+# import/no-restricted-paths
```
elecmonkey: hello?

swwind: Looks like our skill needs to be updated

swwind:
`basePath` resolution, for the record.

Upstream computes `options.basePath || process.cwd()` and then `path.resolve(basePath, target)`, so the base is:

| config | base |
| --- | --- |
| no `basePath` | `process.cwd()` |
| relative `basePath` | `process.cwd()` + `basePath` |
| absolute `basePath` | `basePath` |

It is always the process working directory — not the config file's directory, and not `context.cwd`.

This port previously used `ctx.Program.Host().GetCurrentDirectory()`, which is the owning tsconfig's directory. Two consequences: in a monorepo with one tsconfig per package the same zone resolved against a different base for each file, and a relative `basePath` was never made absolute, so `containsPath` could not match an absolute file name and the zone silently never fired.

Rules now read a new `RuleContext.Cwd`, plumbed through `linter.RunLinterOptions` / `LintSingleFileOptions` and set at each entry point to its real working directory:

| entry point | `Cwd` |
| --- | --- |
| CLI (multi-config, single config, `--config`, `--fix` passes) | `os.Getwd()` |
| API / IPC | `req.WorkingDirectory`, else `os.Getwd()` |
| LSP | the server's `os.Getwd()` — not the nested JS config directory, which is only the config path space |
| rule tests | unset; falls back to the Program directory, i.e. the in-memory fixture root |

With that in place the resolution table matches upstream row for row.

elecmonkey (changes requested):
I found several undocumented differences from upstream:

- Incomplete `import()` calls panic due to indexing `Arguments[0]` without checking the length.
- `require()` imports in `.ts` files are not resolved and therefore are missed.
- Parenthesized calls such as `(require)("./x")` are reported upstream but ignored here.
- `require("./x", extra)` is ignored upstream but may be reported here.
- `containsPath` differs from Node’s `path.relative` behavior, including `..name` path segments and Windows case-insensitive paths.

Please align these behaviors or document any intentional differences.

internal/plugins/import/rules/no_restricted_paths/no_restricted_paths.go
```diff
@@ -0,0 +1,478 @@
...
+		if len(matchingZones) == 0 {
+			return rule.RuleListeners{}
+		}
+
+		validators := make([][]pathValidator, len(matchingZones))
+		built := make([]bool, len(matchingZones))
+		applicable := make([]int, 0, 4)
+
+		return import_utils.VisitModules(func(source *ast.StringLiteralLike, node *ast.Node) {
+			absoluteImportPath, ok := import_utils.Resolve(source, ctx)
```
elecmonkey, line 75: This differs from upstream for `require()` calls in TypeScript files.

Repro:

```ts
// client/a.ts
const b = require("../server/b");
```

with:

```js
{ target: "./client", from: "./server" }
```

`eslint-plugin-import` reports `Unexpected path "../server/b" imported in restricted zone.`, while Rslint reports nothing.

Rslint resolves the specifier only through TypeScript's cached module resolutions, but TypeScript does not collect ordinary `require()` calls from `.ts` files. The migrated test avoids this by changing the fixture to `.js`, which hides the behavioral difference.

internal/plugins/import/rules/no_restricted_paths/no_restricted_paths.go
```diff
@@ -0,0 +1,478 @@
...
+					}
+				}
+				for _, j := range applicable {
+					if validators[i][j].hasValidExceptions && !validators[i][j].isPathException(absoluteImportPath) {
+						ctx.ReportNode(source, unexpectedPathMessage(source.Text(), matchingZones[i].message))
+					}
+				}
+			}
+		}, import_utils.VisitModulesOptions{
+			Commonjs: true,
```
elecmonkey, line 105: The CommonJS visitor differs from upstream in two observable cases.

Parenthesized callee:

```js
(require)("../server/b.js");
```

ESLint reports this because the ESTree callee is still an `Identifier`. Rslint preserves the `ParenthesizedExpression` and ignores it. The new extras test currently records this upstream-invalid case as valid.

Extra arguments:

```js
import "../server/b.js";
require("../server/b.js", "extra");
```

Upstream requires exactly one argument and reports only the import. Rslint accepts one or more arguments and, when the same specifier has already been resolved, reports both calls.

internal/plugins/import/rules/no_restricted_paths/no_restricted_paths.go
```diff
@@ -0,0 +1,478 @@
...
+			ESModule: true,
```
elecmonkey, line 106: This can crash Rslint on incomplete source code:

```js
const p = import()
```

ESLint returns a syntax error and does not run the rule. Rslint's parser recovers with a zero-argument call, after which `VisitModules` indexes `Arguments[0]` and panics with `index out of range`.

Please check the argument count before accessing the first argument. This is especially important for LSP usage while a file is being edited.

internal/plugins/import/rules/no_restricted_paths/no_restricted_paths.go
```diff
@@ -0,0 +1,478 @@
...
+			}
+		}
+		return false
+	}
+	return validator
+}
+
+// containsPath reports whether filePath is target itself or one of its
+// descendants, mirroring upstream's `path.relative(target, filePath)` check.
+func containsPath(filePath string, target string) bool {
```
elecmonkey: This does not match the upstream `path.relative()` behavior.

For example, with `from: "./server"` and a real file named `server/..secret.js`:

```js
import secret from "../server/..secret.js";
```

upstream does not report because `path.relative(server, file)` is `"..secret.js"` and therefore starts with `..`; Rslint's component-prefix check reports it.

There is also a Windows difference: Node's `path.win32.relative()` compares paths case-insensitively, so `C:\Project\client` matches `c:\PROJECT\client\a.ts`, while the exact component comparison here does not. This can silently leave a target zone inactive on Windows.

swwind:
Checked each against `eslint-plugin-import@2.32` first.

- Incomplete `import()` panic — fixed.
- `require()` in `.ts` files — not fixed here. A specifier resolves through the module resolution TypeScript recorded for it, and `ForEachDynamicImportOrRequireCall` gates `require` collection on `isJavaScriptFile`, so in a `.ts` file nothing ever reaches the cache. The same gate is why a standalone `(require)("./x")` stays unresolved — `IsRequireCall` wants a bare identifier callee and exactly one argument. Closing this means giving the shared `import_utils.Resolve` the extension-probing fallback `ResolveModuleReferenceFromSourceFile` already has, which also changes `no-self-import`, so I'd rather do it in its own PR. Documented under "Differences from ESLint" for now, with a valid test case stating the reason instead of switching the fixture to `.js`. fix in #1577
- Parenthesized `(require)("./x")` callee — fixed.
- `require("./x", extra)` — fixed.
- `containsPath` vs `path.relative` (`..name` segments, case-insensitive hosts) — fixed.

elecmonkey (commented):
Re-reviewed at 61a63ba.

The four fixed items check out: the `import()` argument guard, the parenthesized `require` callee, the exact-one-argument requirement (upstream's `!== 1`), and the `..`-prefixed leftover in `containsPath` all match upstream now, and the new tests cover them. Deferring `require()` in `.ts` files to #1577 is fine — the root cause is in the shared resolver and touches `no-self-import`, so it does belong in its own PR, and the valid test case states the reason instead of hiding it behind a `.js` fixture.

Two new issues, both introduced by the `containsPath` fix. Details inline.

internal/plugins/import/rules/no_restricted_paths/no_restricted_paths.go
```diff
@@ -0,0 +1,496 @@
...
+	Name:   "import/no-restricted-paths",
+	Schema: rule.NewSchema(schemaJSON),
+	Run: func(ctx rule.RuleContext, options []any) rule.RuleListeners {
+		opts := parseOptions(options)
+		if len(opts.zones) == 0 || ctx.Program == nil {
+			return rule.RuleListeners{}
+		}
+
+		basePath := resolveBasePath(ctx, opts.basePath)
+		caseSensitive := ctx.Program.Host().FS().UseCaseSensitiveFileNames()
```
elecmonkey: Case-insensitivity keyed off the file system diverges from upstream on macOS.

Node dispatches on the platform, not on the file system: only `path.win32.relative` lowercases before comparing, and on darwin `path.relative` is the posix one, which is case-sensitive no matter what APFS does.

`UseCaseSensitiveFileNames()` returns `false` on a default macOS checkout — `osvfs` decides by stating its own executable under swapped case (`internal/vfs/osvfs/os.go`), and a case-insensitive volume answers yes. I confirmed the `false` on a stock machine here.

So with `target: "./Client"` and a real file `client/a.ts`, this port now matches the zone and reports, while `eslint-plugin-import` on the same machine reports nothing — a false positive, on the platform most likely to hit it.

Please gate on `runtime.GOOS == "windows"` instead, which is what `path.relative` itself keys on.

internal/plugins/import/rules/no_restricted_paths/no_restricted_paths.go
```diff
@@ -0,0 +1,496 @@
...
+		},
+	}
+
+	absoluteExceptions := make([]string, 0, len(except))
+	for _, exception := range except {
+		absoluteException := tspath.ResolvePath(absoluteFrom, exception)
+		// Upstream classifies the `from`-relative exception path with
+		// importType and rejects it when it comes out as "parent", which is
+		// exactly the set of paths that escape `from`.
+		if !containsPath(absoluteException, absoluteFrom, caseSensitive) {
```
elecmonkey: The new `..`-prefix rule leaked into exception validation, where upstream uses a different predicate.

Upstream does not reuse `containsPath` here. `isValidExceptionPath` is `importType(relativeExceptionPath, context) !== 'parent'`, and `importType`'s `isRelativeToParent` is the regex `/^\.\.$|^\.\.[\\/]/`, which matches only `..` and `../…` — not a name that merely starts with two dots:

```js
path.relative('/p/server', path.resolve('/p/server', '..secret')) // '..secret'
/^\.\.$|^\.\.[\\/]/.test('..secret')                             // false -> valid exception
```

With `{ from: "./server", except: ["..secret"] }` upstream accepts the exception, while this port now sets `hasValidExceptions = false` and reports `Restricted path exceptions must be descendants of the configured 'from' path for that zone.` on every restricted import in the zone.

The comment above was accurate for the old component-prefix check, but `containsPath` now also rejects `..name` leftovers, so it is no longer "exactly the set of paths that escape `from`". These need to be two separate predicates: keep `containsPath` for `isPathRestricted` / `isPathException`, and give exception validation the `^\.\.$|^\.\.[/\\]` test on the `from`-relative path.

internal/plugins/import/rules/no_restricted_paths/no_restricted_paths.go
swwind: Fixed in fa3b6f56. `caseSensitive` is now `runtime.GOOS != "windows"`, matching what `path.relative` itself keys on, so a case-insensitive volume on macOS no longer makes the comparison case-insensitive.

internal/plugins/import/rules/no_restricted_paths/no_restricted_paths.go
swwind: Fixed in fa3b6f56. The two predicates are now separate: `isPathWithin` is the plain component-prefix comparison and backs exception validation, matching `importType`'s `^\.\.$|^\.\.[\\/]` which only classifies a whole `..` component as "parent"; `containsPath` builds on it and adds the `..`-prefixed-leftover rejection for `isPathRestricted` / `isPathException`, matching upstream's `!relative.startsWith('..')`. So `{ from: "./server", except: ["..secret"] }` is accepted again while `import "../server/..secret.js"` stays unreported. `TestIsPathWithin` covers both shapes.

elecmonkey (dismissed):
**Approving.** Both regressions from the previous round are cleanly resolved in `fa3b6f56`:

1. **Case-sensitivity** now keys off `runtime.GOOS != "windows"` rather than the host FS, matching how Node's `path.relative` dispatches (only the win32 flavor lowercases). This removes the macOS false positive where `./Client` matched `client/a.ts`.

2. **Exception validation** is split into two predicates with distinct upstream semantics: `containsPath` (`relative === '' || !relative.startsWith('..')`) for restriction/exception matching, and `isPathWithin` (`importType !== 'parent'`, regex `^\.\.$|^\.\.[\\/]`) for exception validity. `..secret` now correctly stays a valid exception, matching ESLint.

The new `TestIsPathWithin` pins the divergence between the two predicates, and the three call sites each use the correct one. All earlier findings are addressed — the four upstream-alignment fixes plus these two, with `require()` in `.ts` files deferred to #1577 with a sound root-cause explanation. LGTM.

fansenze (approved):

Issue #223, opened 2025-08-11
[Tracking]: eslint core rules


hardfist:
This is a tracking issue for porting the existing rules from [`eslint`](https://eslint.org/docs/latest/rules/).

There are **199** rules (+ 93 deprecated).

| Category | ✅ Done | 🚫 Won't | Total |
| :--- | :---: | :---: | :---: |
| Recommended | 64 | 0 | 64 |
| Not recommended | 135 | 0 | 135 |
| Deprecated | 2 | 91 | 93 |
| **Total** | **201** | **91** | **292** |

> ✅ = Implemented &nbsp;|&nbsp; 🚫 = Not intending to implement

---

## Recommended rules

<details open>
<summary>✅ 64 &nbsp;|&nbsp; 🚫 0 &nbsp;/&nbsp; Total: 64</summary>

| Status | Name |
| :----: | :--- |
| ✅ | [constructor-super](https://eslint.org/docs/latest/rules/constructor-super) |
| ✅ | [for-direction](https://eslint.org/docs/latest/rules/for-direction) |
| ✅ | [getter-return](https://eslint.org/docs/latest/rules/getter-return) |
| ✅ | [no-async-promise-executor](https://eslint.org/docs/latest/rules/no-async-promise-executor) |
| ✅ | [no-case-declarations](https://eslint.org/docs/latest/rules/no-case-declarations) |
| ✅ | [no-class-assign](https://eslint.org/docs/latest/rules/no-class-assign) |
| ✅ | [no-compare-neg-zero](https://eslint.org/docs/latest/rules/no-compare-neg-zero) |
| ✅ | [no-cond-assign](https://eslint.org/docs/latest/rules/no-cond-assign) |
| ✅ | [no-const-assign](https://eslint.org/docs/latest/rules/no-const-assign) |
| ✅ | [no-constant-binary-expression](https://eslint.org/docs/latest/rules/no-constant-binary-expression) |
| ✅ | [no-constant-condition](https://eslint.org/docs/latest/rules/no-constant-condition) |
| ✅ | [no-control-regex](https://eslint.org/docs/latest/rules/no-control-regex) |
| ✅ | [no-debugger](https://eslint.org/docs/latest/rules/no-debugger) |
| ✅ | [no-delete-var](https://eslint.org/docs/latest/rules/no-delete-var) |
| ✅ | [no-dupe-args](https://eslint.org/docs/latest/rules/no-dupe-args) |
[...]

PR #1562, opened 2026-08-04
feat: port rule no-return-await

swwind:
## Summary

Port the `no-return-await` rule from ESLint to rslint.

The rule reports `await` on a value that is returned straight to the caller — the argument of a `return`, a concise arrow body, either branch of a conditional, the right side of `&&` / `||` / `??`, or the last expression of a comma sequence. An `await` inside a `try` block, or inside a `catch` clause followed by `finally`, is left alone because removing it would move rejection handling outside the enclosing error handler. Each report carries a suggestion that removes the keyword, omitted when the keyword and its operand sit on different lines.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/no-return-await
- Source code: https://github.com/eslint/eslint/blob/main/lib/rules/no-return-await.js

## Checklist

### DIFF
--- a/internal/rules/all.go
+++ b/internal/rules/all.go
@@ -93,6 +93,7 @@ import (
 	"github.com/web-infra-dev/rslint/internal/rules/no_restricted_imports"
 	"github.com/web-infra-dev/rslint/internal/rules/no_restricted_syntax"
 	"github.com/web-infra-dev/rslint/internal/rules/no_return_assign"
+	"github.com/web-infra-dev/rslint/internal/rules/no_return_await"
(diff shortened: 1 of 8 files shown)


elecmonkey:
[no-return-await](https://eslint.org/docs/latest/rules/no-return-await) is a deprecated rule. see https://github.com/web-infra-dev/rslint/issues/223.

swwind:
> [no-return-await](https://eslint.org/docs/latest/rules/no-return-await) is a deprecated rule. see #223.

Oops, I completely forgot about that! Thanks for the reminder.

PR #1566, opened 2026-08-04
refactor: extract shared control-flow graph into internal/utils/cfg

swwind:
## Summary

The control-flow graph built for `no-useless-assignment` was generic apart from the two calls that recorded variable reads and writes. Move it into `internal/utils/cfg`, parameterized over a consumer-supplied event type, so more than one rule can build on it. Consumers plug in through `Hooks[E]{Read, Write}`.

- `Builder.Emit` returns the block and index it wrote to, so a consumer can record where an event landed — `no-useless-assignment` needs this because the doubled `finally` layout gives one assignment two sites.
- `Block.Index` lets a consumer keep per-block analysis state in a slice rather than on the block itself.
- `no-useless-assignment` keeps only its events and the backward liveness dataflow, now in `liveness.go`.

No behavior change: the graph is byte-for-byte the code that shipped with `no-useless-assignment`, only relocated and parameterized.

## Related Links

- Split out of #1531. The `no-unreachable-loop` port follows in a separate PR stacked on this one.

## Checklist


elecmonkey (commented):
I verified the "no behavior change" claim mechanically: diffing every top-level declaration of the old `no_useless_assignment/cfg.go` + `no_useless_assignment.go` against the new package, with the renames normalized (`builder`→`Builder[E]`, `block`→`Block[E]`, `isCodePathRoot`→`IsRoot`, `rootOf`→`RootOf`), leaves 42 functions identical and 10 differing only mechanically — the `index:` field in `newBlock`, `successors`→`Successors`, `jumpTarget[E]`, the inline map lookups becoming `b.read`/`b.write`, and two comment rewordings. The liveness rewrite from per-block scratch fields to a `[]blockState` indexed by `Block.Index()` is a faithful translation, and it allocates once per root instead of resetting four fields per block per symbol. Build, vet, gofmt and the rule/utils test packages are all clean locally. Nice extraction — the package doc-comment stating the ESLint-shape invariants is exactly the context a future consumer needs, and it was previously buried in a rule.

Two non-blocking notes:

**1. The contract on `Emit` contradicts itself.** `Hooks` says "Both are called only from a reachable position, so `Builder.Emit` always lands somewhere" (cfg.go:55-56), but `Emit` documents and implements a nil-block return (cfg.go:167-177), and `liveness.go:63` branches on it. The guarantee is real today — every `read`/`write` call site is guarded by a `b.cur == nil` early return at function entry (expressions.go:8, patterns.go:12,51,80), and `firstThrowableFork` only ever assigns a non-nil block. But as written a consumer can't tell whether that nil check is required. Worth either dropping the guarantee sentence and keeping the defensive path, or keeping the guarantee and noting the nil return is defensive.

**2. The new package has no tests of its own.** `internal/utils/cfg` reports `[no test files]` while sibling files under `internal/utils` are broadly tested. Coverage is real but entirely indirect, through `no_useless_assignment`'s fixtures. That was fine when the code lived inside the rule; now it's shared infrastructure whose invariants (the doubled `finally` layout, the first-throwable fork, label resolution) can be broken by a change made for one consumer and only caught by another's tests. Not a blocker for a pure relocation, but a small direct test — build a graph over a snippet with `Hooks` that append node kinds, assert the block/successor shape — would pay for itself, as a follow-up if not here.

elecmonkey:
@codex review

elecmonkey:
make some tests for your new cfg tool? 👀 @swwind

elecmonkey (approved):

fansenze (approved):

PR #1576, opened 2026-08-04
fix(jest): correct accessor ranges in matcher chains

elecmonkey:
## Summary

`ast.Node.Pos()` returns the end of the *previous* token, so it includes a node's leading trivia. ESTree's `node.range` does not. Four code paths built fix/report ranges straight from `Pos()`, and each one misbehaves as soon as a newline or comment sits between the dot and the matcher name.

The worst case corrupts source: `expect(a)[/* c */ 'toBeCalled']()` autofixed to `expect(a)[/toHaveBeenCalled']()`, which no longer parses. The hand-written `Pos()+1 / End()-1` offsets meant to strip the quotes landed inside the comment instead.

This extracts the range handling into a framework-neutral helper (`internal/utils/test_framework/accessor.go`, the counterpart of upstream's `src/rules/utils/accessors.ts`) and routes all four call sites through it. `ctx.ReportNode*` already goes through `utils.TrimNodeTextRange` and was always correct, so only hand-built `core.NewTextRange(node.Pos(), …)` needed fixing.

### Behaviour changes

| Rule | Change |
| --- | --- |
| `jest/no-alias-methods` | autofix no longer swallows the newline, indentation or comment before the matcher; the element-access + comment case no longer emits unparseable code |
| `jest/prefer-strict-equal` | suggestion fix no longer swallows leading trivia |
| `jest/no-restricted-matchers` | reported range starts at the matcher name instead of the whitespace in front of it |
| `jest/no-restricted-jest-methods` | same |
| `jest/prefer-strict-equal` | element-access suggestion keeps the original quote characters instead of normalising them to `'` |
| `jest/no-alias-methods`, `jest/prefer-strict-equal` | an alias written as a computed key (`expect(a)[toBeCalled]()`) is still reported but no longer autofixed |

The first four are strictly corrections; downstream snapshots asserting the old column numbers will need updating.

<details>
<summary>Why the quote handling changed</summary>

This resolves an inconsistency between two rules in this repo: `no-alias-methods` already preserved the original quotes while `prefer-strict-equal` rewrote `["toEqual"]` to `['toStrictEqual']`. Normalising quotes makes an autofix introduce a fresh `quotes` violation in a double-quoted codebase, so preserving them is the behaviour both rules now share.

This does diverge from upstream's `replaceAccessorFixer`, which hardcodes `'…'`. That helper only ever aimed to produce a *valid* accessor; preserving the original delimiter was not a goal.

</details>

<details>
<summary>Why computed keys are reported but not fixed</summary>

In `const toBeCalled = 'toBeCalled'; expect(a)[toBeCalled]()` the identifier is a variable reference, not the matcher name, and the old fix rewrote it to `toHaveBeenCalled` — an identifier that does not exist, so the fixed file throws at runtime. The report is unchanged; only the fix is withheld.

Note that the chain is still *parsed* as a member named after the identifier's text. That matches `getNodeChain` in eslint-plugin-jest, which likewise does not check `MemberExpression.computed`, and upstream depends on it: their `no-restricted-matchers` suite asserts that `expect(a).not[x]()` reports `not`.

Dropping computed identifiers from the chain entirely was tried and breaks exactly that case, because an unresolvable element access discards the whole chain. Distinguishing "this segment's name is not statically known" from "abort the chain" needs a change to `MemberEntry` plus a decision about what `not.<unknown>` should match in `no-restricted-matchers` options, which is out of scope here.

</details>

### Note on the rstest side

`test_framework.MemberEntriesRange` changed signature to take the source file, so `rstest.RstestFnMemberEntriesRange` changed with it. That wrapper currently has no consumers; the rstest plugin has no behaviour change in this PR. No compatibility shim is left behind deliberately, so a stale call cannot silently keep the old behaviour.

## Checklist


packages/rslint-test-tools/tests/eslint-plugin-jest/rules/prefer-strict-equal.test.ts
```diff
@@ -51,7 +54,36 @@ ruleTester.run('prefer-strict-equal', {} as never, {
           suggestions: [
             {
               messageId: 'suggestReplaceWithStrictEqual',
-              output: "expect(something)['toStrictEqual'](somethingElse);",
+              output: 'expect(something)["toStrictEqual"](somethingElse);',
```
swwind, line 57: The `RuleTester` in `tests/eslint-plugin-jest/rule-tester.ts` only asserts diagnostic count and message text — it never reads `output`, `column` or `suggestions`. So this edited expectation, and the new `column`/`output`/`suggestions` fields added to the four `.test.ts` files in this PR, are not asserted by anything. The real coverage lives in the Go `_test.go` files, where `rule_tester` does enforce "no `Output`" ⇒ no fix and `len(Suggestions)` ⇒ exact suggestion count.

Worth flagging because these are upstream-derived parity fixtures: reading them later, the new fields look like regression protection but aren't.

internal/plugins/jest/rules/no_restricted_matchers/no_restricted_matchers.go
```diff
@@ -115,7 +115,7 @@ var NoRestrictedMatchersRule = rule.Rule{
 				}
 
 				chain := strings.Join(jestFnCall.Members, ".")
-				reportRange, ok := jestUtils.JestFnMemberEntriesRange(jestFnCall.MemberEntries)
+				reportRange, ok := jestUtils.JestFnMemberEntriesRange(ctx.SourceFile, jestFnCall.MemberEntries)
```
swwind: `JestFnMemberEntriesRange` now goes through `utils.TrimNodeTextRange` → `scanner.GetRangeOfTokenAtPosition` → `NewScanner()`, which heap-allocates a scanner; before this PR it was two struct field reads. Computing it here means every parsed `expect()` pays for it even when no restriction matches.

`no_restricted_jest_methods.go:121` already does it the other way round — the range is computed only after `restrictedMethods[method]` hits. Moving this call inside the `isChainRestricted` branch would match that.

swwind (commented):
Reviewed at `d319cf9b`. Every new column/range expectation was cross-checked against `eslint-plugin-jest@29.15.5`, including `expect(a)['not']['toBe'](b)`, `expect(a).not[x]()` and `jest['advanceTimersByTime']()` — all match. Two non-blocking notes inline.

packages/rslint-test-tools/tests/eslint-plugin-jest/rules/prefer-strict-equal.test.ts
elecmonkey: need more discussion and won't solve in this pr

internal/plugins/jest/rules/no_restricted_matchers/no_restricted_matchers.go
elecmonkey: Good catch, fixed in a183bf22. The `JestFnMemberEntriesRange` call now happens inside the loop, after `isChainRestricted` matches, so a parsed `expect()` that hits no restriction never touches the scanner. This matches the shape of `no_restricted_jest_methods.go:121`. Tests still pass.

swwind (approved):
LGTM

PR #1580, opened 2026-08-05
feat: port rule consistent-return

swwind:
## Summary

Port the `consistent-return` rule from ESLint to rslint.

The rule requires every `return` in a function to either always or never carry a value, and reports a function that returns a value on one path but runs off the end of its body on another. A class `constructor` and a function whose own name starts with an uppercase letter are exempt from the run-off-the-end half of the check. The `treatUndefinedAsUnspecified` option reads `return undefined;` and `return void 0;` as returning nothing.

Reachability comes from the shared control-flow graph in `internal/utils/cfg`, which gains `Graph.EndReachable` — whether control can run off the end of a code path root, the question ESLint answers with `isAnySegmentReachable(codePath.currentSegments)` at the root's `:exit`. It reads the reachability of the block the walk ends in, which the `no-useless-return` port this one is stacked on settles for every block.

`utils.IsES5Constructor` is extracted from the typescript `no-invalid-this` rule so both rules share one implementation. The uppercase test behind it reads the first UTF-16 code unit, the way ESLint's `s[0] !== s[0].toLocaleLowerCase()` does: a character qualifies when it has a distinct lowercase form, so `Ⅰ` counts, while a character outside the BMP is read as a lone surrogate — its own lowercase form — so `𐐀` does not.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/consistent-return
- Source code: https://github.com/eslint/eslint/blob/main/lib/rules/consistent-return.js

## Checklist

### DIFF
--- a/internal/plugins/typescript/rules/no_invalid_this/no_invalid_this.go
+++ b/internal/plugins/typescript/rules/no_invalid_this/no_invalid_this.go
@@ -4,7 +4,6 @@ import (
 	"regexp"
 	"slices"
 	"strings"
-	"unicode"
 
 	"github.com/microsoft/typescript-go/shim/ast"
 	"github.com/microsoft/typescript-go/shim/scanner"
@@ -221,7 +220,7 @@ func computeFunctionValid(node *ast.Node, sf *ast.SourceFile, capIsConstructor b
 	if hasJSDocThisTag(node, sf) {
 		return true
 	}
-	if capIsConstructor && isES5Constructor(node) {
+	if capIsConstructor && utils.IsES5Constructor(node) {
 		return true
 	}
 	return !isDefaultThisBinding(node, capIsConstructor)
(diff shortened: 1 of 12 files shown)


elecmonkey (commented):
Reviewed differentially against ESLint 10.8.0 (rule source verified byte-identical to `eslint/eslint` HEAD), stacked on top of the review of #1583. Two problems.

### 1. Private method names are quoted in the message

```
class A { #foo() { if (a) return 1; } }
rslint: Expected to return a value at the end of private method '#foo'.
eslint: Expected to return a value at the end of private method #foo.
```

Upstream's `getFunctionNameWithKind` (`lib/rules/utils/ast-utils.js:2284`) pushes `` `#${parent.key.name}` `` unquoted for a `PrivateIdentifier` key; only the `getStaticPropertyName` branch wraps the name in `'…'`. `appendCoreFunctionNameFromOwner` (`internal/utils/ts_eslint.go:597-598`) quotes it too.

All three messageIds are affected, and every private shape reproduces:

```
class A { #foo() { … } }                      →  private method '#foo'
class A { static #foo() { … } }               →  static private method '#foo'
class A { get #foo() { … } }                  →  private getter '#foo'
class A { async *#foo() { … } }               →  private async generator method '#foo'
class A { #foo = function () { … }; }         →  private method '#foo'
```

`GetFunctionNameWithKindCore` is an existing helper — `complexity`, `max-params` and `require-await` already carry this — but this PR touches that function and makes `consistent-return` its fourth consumer, so it seems like the natural place to fix it.

### 2. Two CFG issues from #1583 also surface as `consistent-return` diagnostics

Both are pre-existing in `internal/utils/cfg` and are described in detail on #1583; recording the `consistent-return` repros here so they're not lost:

```js
// `for` incrementor laid out after the body → catch/after-try wrongly unreachable
// eslint: missingReturn on `f`; rslint: nothing
function f() { try { for (var i = 0;; i++) { break; } return 1; } catch (e) {} }

// `labelsOf` keeps a back edge ESLint drops → end wrongly reachable
// rslint: missingReturn on `f`; eslint: nothing
function f() { outer: inner: do { continue outer; } while (a); if (c) return 1; }
```

fansenze (approved):

PR #1583, opened 2026-08-05
feat: port rule no-useless-return

swwind:
## Summary

Port the `no-useless-return` rule from ESLint. It reports a bare `return` that nothing runs after, and fixes it by removing the statement.

Answering that means asking what would run after a `return` if the `return` were deleted — the question ESLint's code path analysis answers by keeping the unreachable segments an abrupt exit creates. `internal/utils/cfg` dropped that code instead of laying it out, so the graph builder changes with this port.

- `Block` gains `Reachable`. `Successors` spans the whole graph, unreachable blocks included; `Reachable` says which blocks run.
- `return`, `throw`, `break`, and `continue` carry the walk on into a fresh unreachable block rather than ending it, and do nothing where control already cannot arrive.
- `firstThrowableFork` no longer consumes a `try` frame's fork from an unreachable position, which would otherwise move the point at which a reachable statement forks to the handler.
- A `try` statement's normal-completion path is always laid out, including when its `try` block ends abruptly — that chain is what carries a `return` inside the `try` block past the statement.

Hooks now run in unreachable positions too, so a consumer with no use for them checks `Builder.Current().Reachable`. An edge out of an unreachable block never marks its target reachable, so the reachable subgraph is unchanged. `no-unreachable-loop` and `no-useless-assignment` skip unreachable blocks, which keeps the behavior they had — ESLint ignores unreachable code in both.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/no-useless-return
- Source code: https://github.com/eslint/eslint/blob/main/lib/rules/no-useless-return.js

## Checklist

### DIFF
--- a/internal/rules/all.go
+++ b/internal/rules/all.go
@@ -128,6 +128,7 @@ import (
 	"github.com/web-infra-dev/rslint/internal/rules/no_useless_constructor"
 	"github.com/web-infra-dev/rslint/internal/rules/no_useless_escape"
 	"github.com/web-infra-dev/rslint/internal/rules/no_useless_rename"
+	"github.com/web-infra-dev/rslint/internal/rules/no_useless_return"
 	"github.com/web-infra-dev/rslint/internal/rules/no_var"
 	"github.com/web-infra-dev/rslint/internal/rules/no_with"
 	"github.com/web-infra-dev/rslint/internal/rules/object_shorthand"
@@ -301,6 +302,7 @@ func GetAllRules() []rule.Rule {
 		no_useless_catch.NoUselessCatchRule,
 		no_useless_escape.NoUselessEscapeRule,
 		no_useless_rename.NoUselessRenameRule,
+		no_useless_return.NoUselessReturnRule,
 		no_useless_constructor.NoUselessConstructorRule,
 		no_prototype_builtins.NoPrototypeBuiltinsRule,
 		require_await.RequireAwaitRule,
--- a/internal/rules/no_unreachable_loop/no_unreachable_loop.go
+++ b/internal/rules/no_unreachable_loop/no_unreachable_loop.go
@@ -121,8 +121,9 @@ func reportLoops(ctx *rule.RuleContext, rootOrder []*ast.Node, opts ruleOptions)
 
 // singleIterationLoops returns the loops in root whose body allows only one
 // iteration: control reaches the loop, but never flows back into it for a
-// second one. A loop the code path never reaches is left alone — whether its
-// body could repeat says nothing useful about code that does not run.
+// second one. Unreachable blocks are skipped — whether a loop that never runs
+// could repeat says nothing useful, and neither does a `continue` that never
+// runs.
 //
 // NOTE: A loop wrapped in more than one label repeats when a `continue` names
 // any of them. ESLint 10.8.0 reports the double-labelled `while` and `do` forms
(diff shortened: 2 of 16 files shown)


elecmonkey (commented):
Reviewed differentially against ESLint 10.8.0 (verified byte-identical rule source to `eslint/eslint` HEAD), comparing messageId, start/end position, message text and post-`--fix` output. Three divergences.

### 1. `clears()` is missing `TSNamespaceExportDeclaration` — we under-report

```ts
if (a) { return; }
export as namespace Lib;
```

ESLint reports `unnecessaryReturn` on the `return`; rslint does not.

`clears()` (`internal/rules/no_useless_return/no_useless_return.go:243`) enumerates the TypeScript-only declarations that fall outside upstream's listener list, and `export = x` is handled — but `export as namespace X` parses to `TSNamespaceExportDeclaration`, which is not in upstream's list either, so it should not clear. It currently falls through to `default: return true`.

Narrow to trigger (needs a top-level bare `return` plus that declaration), but it is a hole in a list that is otherwise enumerated 1:1 against upstream.

### 2. A `for` statement's incrementor is laid out after the body, which drops the throwable fork into `catch` — we under-report

```js
function f() { try { for (var i = 0;; i++) { break; } return 1; } catch (e) { return; } }
```

ESLint reports the `return` in the `catch`; rslint reports nothing.

Upstream visits `i++` in traversal order — before the body — so `makeFirstThrowablePathInTryOrCatchBlock` runs while the segment is still reachable and the edge into `catch` exists. `forStatement` (`internal/utils/cfg/statements.go:243-259`) enters the update block *after* the body; when the body always exits abruptly the update block is unreachable, `firstThrowableFork` returns early on `!b.cur.Reachable`, and the `catch` entry ends up with only an edge from an unreachable block — so the whole `catch` clause and everything after the `try` statement is treated as unreachable.

I enumerated 25 `try` bodies × 4 tails (600 cases); every mismatch came from this one family — `for (…;…;i++)`, `for (;; f())`, `for (;; i.j++)`, with a body that always exits abruptly and no throwable node in the test.

This is pre-existing: `forStatement` is not in this diff, and I rebuilt the same harness on `main` (88230a2f) to confirm `no-unreachable-loop` and `no-useless-assignment` already behave identically there. Same shape for those two rules:

```js
// no-unreachable-loop: eslint reports both loops, rslint only the first
function f() { try { for (var i = 0;; i++) { break; } return 1; } catch (e) {} for (var v of items) { return 1; } }

// no-useless-assignment: eslint reports `x = 2`, rslint reports nothing
function f() { try { for (var i = 0;; i++) { break; } return 1; } catch (e) {} let x = 1; x = 2; return x; }
```

Raising it here because this PR is the one rewriting reachability in the CFG, and because `no-useless-return` is a new rule that under-reports because of it. Laying the update block out before the body would line all four rules up at once.

### 3. `labelsOf` keeps a back edge ESLint drops — we over-report

```js
function f() { outer: inner: do { continue outer; } while (a); if (c) { return; } }
```

rslint reports `unnecessaryReturn`; ESLint does not.

The comment on `labelsOf` (`internal/utils/cfg/helpers.go:21-25`) frames this as an improvement over upstream: ESLint attaches only the innermost label and loses the back edge of a `continue` naming an outer one. The consequence is that in ESLint `continue outer` goes nowhere, so a `do…while`'s test and everything after it is unreachable and the `return` is never considered — while rslint keeps the edge, making it reachable and reportable. `while` / `for-of` don't show it because their exit doesn't depend on the back edge; `do…while` does.

Also pre-existing (`helpers.go` is untouched by this diff), and it's "we report, upstream doesn't", so it's worth either aligning `labelsOf` or restating that comment as a known deviation with the affected rules listed.

swwind:
Thanks — all three reproduce here against ESLint 10.7.

**1.** Fixed. `export as namespace X` parses to `KindNamespaceExportDeclaration`, which fell through `clears()` to `default: return true`; it now returns `false`, with a test case whose expectations are taken from ESLint's actual output.

**2.** Taking this as a separate PR against `main`, since `forStatement` is untouched here and the fix lands on `no-unreachable-loop` and `no-useless-assignment` too.

**3.** Keeping the edge. `continue outer` does target the loop, so the graph is right and ESLint's is not; reproducing the dropped edge would trade a false positive for a wrong graph. The comment on `labelsOf` no longer frames it as an improvement — it now states the deviation, the shape that surfaces it (a `do…while`, whose exit hangs off the back edge), and the three rules that read it.

fansenze (approved):

PR #1591, opened 2026-08-05
fix(prefer-for-of): count shorthand property uses of the loop index

swwind:
## Summary

`{ i }` was not counted as a use of the loop index, so a loop whose index escapes as a shorthand property was reported as convertible while the equivalent `{ i: i }` was not.

The body scan asked the TypeChecker for each identifier's symbol, and on a shorthand property name `GetSymbolAtLocation` returns the property symbol, so the reference never matched the index. It now resolves identifiers through `ctx.Refs`, which treats a shorthand property name as a real reference and still skips names that only look like the index (property keys, member names, labels, declaration names).

That leaves no TypeChecker use in the rule — the index is always resolvable by the binder — so `RequiresTypeInfo` is dropped and the rule now runs on files without type information, matching upstream.

Microbenchmark (32-file program, prefer-for-of only, median of 6): 40 loops per file 4.53 ms -> 3.88 ms; one loop in a 700-line file unchanged.

## Checklist


elecmonkey:
@codex review

fansenze (approved):

PR #1593, opened 2026-08-05
feat(jest,rstest): parse setup and teardown hooks

elecmonkey:
## Summary

- extend `ParseRstestFnCall` to classify Rstest setup and teardown hooks
- add `IsTypeOfRstestFnCall` for kind-based call checks
- share the hook-order table and the call-kind matching with the jest plugin through `test_framework`

## Details

The parser now recognizes:

- `beforeAll`
- `beforeEach`
- `afterEach`
- `afterAll`

Hooks resolve through the existing Rstest API channels, including globals, named and aliased imports, namespaces, `import.meta.rstest`, const aliases, and `@rstest/playwright`.

Execution-time APIs such as `onTestFinished` and `onTestFailed` remain excluded. Hook member chains are also rejected because Rstest hooks do not expose test-style modifiers.

Existing parser consumers continue filtering by call kind and are unaffected.

## Shared with the jest plugin

Both additions would otherwise have been byte-identical copies of code jest already has, so they live in `internal/utils/test_framework` and jest was migrated onto them in the same PR rather than left on its duplicate.

**`HooksOrder` / `IsHookName` / `HookOrderIndex`** replace `JEST_HOOKS_ORDER` + `JestHookOrderIndex` and the Rstest hook-name table. Jest and Rstest expose the same four hooks in the same order; the source comment records the Rstest audit, including why `onTestFinished` / `onTestFailed` are absent. Call sites migrated: `GetJestKind`, `jest/prefer-hooks-in-order`, `directRstestAPIState`.

**`IsCallOfKind`** replaces the shared body of `IsTypeOfJestFnCall` and `IsTypeOfRstestFnCall`. Each plugin keeps its own wrapper because the parse entry point is genuinely plugin-owned — rstest parses with official extensions enabled (`import.meta` + playwright), jest does not. What is shared is only what counts as a match, and that carries two contracts worth single-sourcing: a failed parse is `false`, and an empty `kinds` list is `false` rather than "any kind". Neither was covered by a test before; `types_test.go` now pins both.

Jest behavior is unchanged — the delegations are pure moves, and `go test ./internal/plugins/jest/...` is green.

### DIFF
--- a/internal/plugins/rstest/utils/parse_rstest_fn.go
+++ b/internal/plugins/rstest/utils/parse_rstest_fn.go
@@ -57,9 +57,10 @@ const (
 	rstestProfilePlaywright
 )
 
-// ParseRstestFnCall parses a final Rstest test/describe registration call.
-// Factory calls such as test.each(cases), test.runIf(condition), and
-// test.extend(fixtures) are intentionally not returned.
+// ParseRstestFnCall parses a final Rstest test/describe/hook registration
+// call; callers distinguish the three through Kind. Factory calls such as
+// test.each(cases), test.runIf(condition), and test.extend(fixtures) are
+// intentionally not returned.
 func ParseRstestFnCall(node *ast.Node, ctx rule.RuleContext) *ParsedRstestFnCall {
 	return parseRstestFnCall(node, ctx, false, false)
 }
--- a/internal/plugins/rstest/utils/parse_rstest_fn.go
+++ b/internal/plugins/rstest/utils/parse_rstest_fn.go
@@ -1,6 +1,8 @@
 package utils
 
 import (
+	"slices"
+
 	"github.com/microsoft/typescript-go/shim/ast"
 	"github.com/microsoft/typescript-go/shim/scanner"
 	"github.com/web-infra-dev/rslint/internal/rule"
@@ -17,6 +19,7 @@ const (
 	rstestAPIDescribe
 	rstestAPIParameterizedTest
 	rstestAPIParameterizedDescribe
+	rstestAPIHook
 )
 
 type rstestInvocation uint8
@@ -68,6 +71,19 @@ func ParseRstestFnCallWithOfficialExtensions(
 	ctx rule.RuleContext,
 ) *ParsedRstestFnCall {
 	return parseRstestFnCall(node, ctx, true, true)
+}
+
+// IsTypeOfRstestFnCall reports whether node parses as a Rstest registration
+// call of one of the given kinds. Mirrors jest's IsTypeOfJestFnCall
+// (parse_jest_fn.go:32); it parses with official extensions enabled
+// (import.meta + playwright), the same choice RstestTestCallbacks.ParseFnCall
+// makes.
+func IsTypeOfRstestFnCall(node *ast.Node, ctx rule.RuleContext, kinds ...RstestFnType) bool {
+	parsed := ParseRstestFnCallWithOfficialExtensions(node, ctx)
+	if parsed == nil || len(kinds) == 0 {
+		return false
+	}
+	return slices.Contains(kinds, parsed.Kind)
 }
 
 func parseRstestFnCall(
@@ -129,6 +145,8 @@ func parseRstestFnCall(
 		resolved.kind = RstestFnTypeTest
 	case rstestAPIDescribe, rstestAPIParameterizedDescribe:
 		resolved.kind = RstestFnTypeDescribe
+	case rstestAPIHook:
+		resolved.kind = RstestFnTypeHook
 	default:
 		return nil
 	}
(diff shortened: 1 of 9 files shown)


fansenze:
### [P2] Recognize hooks exposed on the Playwright `test` object

The new hook detection works for directly imported hooks:

```ts
import { beforeEach } from '@rstest/playwright';
beforeEach(() => {});
```

However, it misses the equivalent member API officially exposed by `PlaywrightTest`:

```ts
import { test } from '@rstest/playwright';

test.beforeAll(() => {});
test.beforeEach(() => {});
test.afterEach(() => {});
test.afterAll(() => {});

test.extend({ /* fixtures */ }).beforeEach(() => {});
```

For these calls, the root resolves to `rstestAPITestWithExtend`, but `applyRstestChainPart` treats the hook member as invalid. As a result:

- `ParseRstestFnCallWithOfficialExtensions` returns `nil` instead of a call with `Kind == RstestFnTypeHook`.
- `IsTypeOfRstestFnCall(..., RstestFnTypeHook)` returns `false`.
- Rules consuming hook calls silently miss these valid `@rstest/playwright` hooks.

This is different from an invalid chain such as `beforeEach.skip()`: in `test.beforeEach()`, `.beforeEach` selects a hook exposed by the `test` object; it is not a modifier applied to an already resolved hook.

`PlaywrightTest` explicitly declares all four hook members, and `extend()` returns another `PlaywrightTest`: [upstream type definition](https://github.com/web-infra-dev/rstest/blob/c4b67c72741ffd7e4c71f641a2e19eae50401dbf/packages/playwright/src/fixture.ts#L1470-L1481).

Please add a Playwright-only transition from the test state to `rstestAPIHook` when the member is `beforeAll`, `beforeEach`, `afterEach`, or `afterAll`. The resolved name should also become the hook member name, such as `beforeEach`, rather than remain `test`. Subsequent members on the resolved hook should continue to be rejected.

Please also add coverage for all four `test.<hook>()` forms and at least one `test.extend(...).<hook>()` case, asserting both parsing and `IsTypeOfRstestFnCall`.

internal/plugins/rstest/utils/parse_rstest_fn.go
```diff
@@ -562,6 +601,14 @@ func applyResolvedRstestChainPart(resolved *rstestResolvedAPI, part rstestChainP
 		resolved.state = rstestAPIInvalid
 		return false
 	}
+	// `test.beforeEach()` resolves to the hook the member selects, not to the
+	// test object it was read from, so the reported name and original node move
+	// to the member. Rules keyed on the hook name — prefer-hooks-in-order, for
+	// one — would otherwise see every Playwright hook as `test`.
+	if state == rstestAPIHook && resolved.state != rstestAPIHook {
```
swwind: The second clause is dead: `applyRstestChainPart` has no `rstestAPIHook` case, so any part applied while already in hook state yields `rstestAPIInvalid` and returns early above. `state == rstestAPIHook` alone is equivalent — and the extra clause suggests a hook→hook transition exists, contradicting the "hooks accept no chained members, by construction" invariant documented at `directRstestAPIState`.

swwind (commented):
No correctness issues found. I traced the new rstest state machine (including `beforeAll.skip`, `test.skip.beforeEach`, and alias chains like `const setup = test.beforeEach`) and the jest migration to `testFramework.HooksOrder` — the hook table and `IsCallOfKind` semantics are behavior-preserving, and all existing parser consumers gate on `Kind`, so hooks now parsing regresses nothing. Two small notes outside the diff, plus one inline comment:

- `internal/plugins/rstest/utils/parse_rstest_fn.go:60` — the doc comment on `ParseRstestFnCall` still says the parser returns "a final Rstest test/describe registration call", but both entry points now also return hook registrations (`Kind == RstestFnTypeHook`). Worth rewording to "test/describe/hook registration call" so future callers know to check `Kind`.

- `internal/plugins/jest/rules/no_hooks/no_hooks.go:25` (pre-existing, optional) — `allowedHooks` is now the last remaining literal copy of the hook-name set this PR centralizes as `testFramework.HooksOrder`; its only use is the membership test in `parseAllowList`, which could be `testFramework.IsHookName(s)`, letting the map be deleted.

internal/plugins/rstest/utils/parse_rstest_fn.go
elecmonkey: Confirmed, and fixed in 97b4b7aa. `applyRstestChainPart` returns `rstestAPIInvalid` for any member applied to a hook, and `applyResolvedRstestChainPart` returns early on that, so the second clause could never be false. Dropped it and replaced it with a comment recording why no guard is needed, so the invariant is stated once instead of being implied by a condition that contradicts it.

elecmonkey:
Thanks for the trace — all three taken, one commit each, pushed as 6c815a36.

- 2177922d — `ParseRstestFnCall`'s doc comment now says `test/describe/hook registration call` and points callers at `Kind`, since both entry points can return hooks from this branch on.
- 97b4b7aa — dropped the dead `resolved.state != rstestAPIHook` clause (replied inline).
- 6c815a36 — `no_hooks.allowedHooks` is gone; `parseAllowList` now calls `testFramework.IsHookName`, so the hook-name set exists in exactly one place.

The third one touches a published jest rule, so: `parseAllowList` only ever used the map as a membership test, and both sets are the same four names in the same order, so the allow-list filtering is unchanged.

swwind (approved):

PR #1597, opened 2026-08-06
feat(utils): add a minimatch port for glob matching

swwind:
## Summary

Adds `internal/utils/minimatch`, a Go port of minimatch 3.1.5 and the brace-expansion 1.1.16 it resolves to — the glob matcher an ESLint plugin compares paths with. It covers the extended glob syntax (`!(a)`, `@(a|b)`, `+(a)`, `?(a)`, `*(a)`) that `utils.MatchGlob` has no syntax for, so a rule ported from an ESLint plugin can accept the same patterns its upstream does. Rules are what this is for, which is what fixes the version at 3.x; ESLint itself moved on to minimatch 10 for the paths it matches on its own behalf.

Patterns compile to regexp2 rather than the standard library's regexp, because a negated list needs lookahead.

```go
matcher := minimatch.New(pattern, minimatch.Options{})
matcher.Match(path)

minimatch.Match(pattern, path, minimatch.Options{})
minimatch.BraceExpand("src/{a,b}/**")
```

`Options{}` is minimatch's own default behavior. A pattern that cannot be compiled yields a matcher that matches nothing rather than an error, which is what minimatch does.

Matching only: brace expansion, extended glob syntax, `**`, character classes, negated patterns, and the `Dot`, `NoCase`, `MatchBase`, `NoExt`, `NoBrace`, `NoGlobStar`, `NoNegate`, `NoComment` and `FlipNegate` options. Partial matching, `makeRe` and the filesystem traversal helpers are left out.

`utils.MatchGlob` is untouched and no existing caller changes.

The port was seeded from the Go port in koblas/swerver and rewritten against the minimatch, brace-expansion and balanced-match sources. `LICENSE` carries the ISC/MIT notices for all four.

## Conformance

Checked against minimatch 3.1.5 over 2,165,448 pattern/path/option rows — short patterns enumerated exhaustively over the metacharacter, extended-glob and character-class alphabets, plus randomized longer ones — and `BraceExpand` against brace-expansion 1.1.16 over 136,416 inputs. Brace expansion agrees everywhere. Matching differs on 2 rows, in two groups, both documented in the package comment:

- A `?` or a character class against a character outside the basic multilingual plane. Each matches one whole character here, where minimatch counts UTF-16 units and so needs `??`.
- One pattern shape where minimatch 3.1.5 splices the string `undefined` into the regexp it builds: a part that ends in a bare `[` and compiles to a regexp rather than a literal — because it carries a wildcard, or because `NoCase` is set — yields something like `(?=.)a[^/]*?\[undefined`. This port answers what the pattern means.

Those counts come from a POSIX run, which cannot observe a third divergence: minimatch 3.1.5 rewrites a `\` to a `/` when it runs on Windows, spending the escape a pattern was written with, so `src/\*.ts` names one file everywhere and picks up every `.ts` file there. This port keeps a `\` an escape on every platform — the behavior minimatch 3.1.5 gives under `allowWindowsEscape` and minimatch 10 gives by default. Paths reach a rule separated by `/` whatever the platform, so a pattern means the same thing wherever rslint runs.

A `**` walks the path by recursion, remembering the pairs of path part and pattern part it has already failed at, so a pattern carrying several of them stays cheap. minimatch instead caps how deep it recurses and answers that a path does not match once it hits the cap, which its own source calls an intentional false negative.

## Checklist


internal/utils/minimatch/minimatch.go
```diff
@@ -0,0 +1,709 @@
...
+				for _, remaining := range file[fi:] {
+					if m.isDotPart(remaining) {
+						return false
+					}
+				}
+				return true
+			}
+			for fr := fi; fr < fl; fr++ {
+				if m.matchOne(file[fr:], row[pi+1:]) {
+					return true
```
elecmonkey: [P1] Bound recursive globstar matching

When a user-supplied pattern contains several `**` components and ultimately misses, this loop explores exponentially many allocations of path segments with no recursion/depth cap. For example, matching 25 ordinary segments against twelve `**` components followed by `z` takes over five seconds here, while minimatch 3.1.5 completes immediately using its bounded globstar implementation. A lint configuration can therefore make every checked path consume excessive CPU; add equivalent bounding or memoize `(file index, pattern index)` states.

internal/utils/minimatch/minimatch.go
```diff
@@ -0,0 +1,709 @@
...
+			// translated.
+			//
+			// What decides that is whether the class compiles, so ask about the
+			// translated source rather than the pattern text it came from: a
+			// character such as `[` or `\` that a glob passes straight through
+			// carries a meaning of its own in a regexp character class.
+			if _, err := regexp2.Compile(re[reClassStart:]+"]", regexp2.None); err != nil {
+				class := pattern[classStart+1 : i]
+				source, sourceMagic, _ := m.parseSource(class, true)
+				re = re[:reClassStart] + `\[` + source + `\]`
```
elecmonkey, line 538: [P2] Preserve empty negated character classes

When the pattern is `[!]` or `[^]`, regexp2 rejects the translated `[^]`, so this fallback converts the class into the literal strings `\[!\]`/`\[\^\]`. In the claimed minimatch 3.1.5 behavior, both patterns match any single non-dot character (for example, `[!]` matches `a`), but this port returns false and instead matches the literal `[!]`. Special-case this JavaScript character-class form rather than routing it through the invalid-class fallback.

elecmonkey (changes requested):
The port passes its included corpus but has an exploitable exponential globstar path and diverges from minimatch 3.1.5 for empty negated character classes. Both affect valid user-supplied glob patterns.

internal/utils/minimatch/minimatch.go
swwind: Fixed in c95ac263. `matchOne` now memoises the `(path part, pattern part)` pairs it has already failed at: every recursion resumes at those two offsets and reads nothing else, so a pair that failed once cannot succeed later. The array is allocated only when a `**` actually has something to backtrack over.

That makes the walk polynomial without a depth cap — minimatch 3.1.5 bounds `_matchGlobStarBodySections` at `maxGlobstarRecursion`, which its own source calls an "intentional false negative, acceptable break in correctness for security". 16 `**` against a 32-part path now answers in microseconds, and agrees with minimatch on a hit, a miss on the last part and a miss in the middle; `TestMatchManyGlobstars` covers all three.

internal/utils/minimatch/minimatch.go
swwind: Fixed in c95ac263. The `]` branch now recognises a class whose translated source is exactly `[^` and emits `[\s\S]` — what JavaScript's `[^]` means — instead of falling through to the invalid-class path. `[!]` matches `a` and `]` but not `.` or `ab`, and `[!]]` matches `a]` but not `]`, which is what minimatch 3.1.5 answers: it compiles both spellings to `^(?!\.)(?=.)[^]$`.

Re-running the differential over 2,165,448 rows, this accounted for 171 of the 173 remaining divergences. The two that are left are the astral-plane `?` case and minimatch's own `undefined` splice; both stay documented in the package comment, and the PR description is updated.

elecmonkey (dismissed):

fansenze (approved):

PR #1601, opened 2026-08-06
feat(eslint-plugin-jest): add `prefer-ending-with-an-expect` rule

eryue0220:
## Summary

* Port `prefer-ending-with-an-expect` from eslint to rslint

## Related Links

eslint-plugin-jest/prefer-ending-with-an-expect [doc](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/prefer-ending-with-an-expect.md) [code](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/prefer-ending-with-an-expect.ts)

## Checklist


internal/plugins/jest/utils/jest.go
```diff
@@ -173,6 +175,90 @@ func GetJestFnMemberEntries(node *ast.Node) []ParsedJestFnMemberEntry {
...
+		parsed.AssertFunctionNames = stringList(raw)
+	}
+	if raw, ok := optsMap["additionalTestBlockFunctions"].([]interface{}); ok {
+		parsed.AdditionalTestBlockFunctions = stringList(raw)
+	}
+
+	return parsed
+}
+
+func stringList(raw []interface{}) []string {
```
swwind, line 207: `utils.ToStringSlice` ?

internal/plugins/jest/utils/jest.go
```diff
@@ -173,6 +175,90 @@ func GetJestFnMemberEntries(node *ast.Node) []ParsedJestFnMemberEntry {
...
+		if segment == "**" {
+			parts = append(parts, `[a-zA-Z0-9.]*`)
+		} else {
+			parts = append(parts, strings.ReplaceAll(segment, "*", `[a-zA-Z0-9]*`))
+		}
+	}
+
+	// Other regular-expression characters are intentionally preserved because
+	// upstream passes configured patterns directly to RegExp.
+	re, err := regexp.Compile(`(?i)^(?:` + strings.Join(parts, `\.`) + `)(?:\.|$)`)
```
swwind, line 241: Prefer to use `regexp2` (`utils.CompileRegexp2`) if we need to match js's RegExp behavior.

internal/plugins/jest/rules/prefer_ending_with_an_expect/prefer_ending_with_an_expect.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+func isAssertionCall(node *ast.Node, ctx rule.RuleContext, patterns []*regexp.Regexp) bool {
+	if node == nil {
+		return false
+	}
+	if node.Kind == ast.KindAwaitExpression {
+		node = ast.SkipParentheses(node.AsAwaitExpression().Expression)
+	}
+	if node == nil || node.Kind != ast.KindCallExpression {
```
swwind, line 54: ~~Earlier version of this comment suggested following upstream here — retracted.~~ Upstream reports optional-chained final assertions (`expect?.(1).toBe(1)`, `expect(1)?.toBe(1)`), but that turns out to be an upstream oversight, not intended behavior: its statement-level `type === CallExpression` guard misses the `ChainExpression` wrapper, while its own sibling rule expect-expect accepts all of these as assertions (verified both with the RuleTester on v29.15.5). Treating them as valid here is the right call — no change needed.

internal/plugins/jest/rules/prefer_ending_with_an_expect/prefer_ending_with_an_expect_extras_test.go
```diff
@@ -0,0 +1,76 @@
...
+			{Code: `test('works', function () { expect(1).toBe(1); });`},
+			// Locks in upstream create() branch: fewer than two arguments are ignored.
+			{Code: `test('missing callback');`},
+			// Locks in upstream create() branch: named callback references are ignored.
+			{Code: `test('named callback', callback); function callback() {}`},
+			// Locks in upstream getLastStatement() expression-body arm.
+			{Code: `test('expression body', () => verify());`, Options: []interface{}{map[string]interface{}{"assertFunctionNames": []interface{}{"verify"}}}},
+			// Locks in upstream matcher branch: a configured additional test wrapper is exact-match only.
+			{Code: `suite.case('ignored', () => {});`, Options: []interface{}{map[string]interface{}{"additionalTestBlockFunctions": []interface{}{"suite.test"}}}},
+			// ---- Real-user: #1752 awaited expect regression ----
```
swwind, line 37: `#1752` / `#1742` (line 39) are upstream eslint-plugin-jest PR numbers, but a bare `#N` in this repo reads as an rslint issue/PR reference. Spell out the plugin name next to the number, or drop the numbers.

swwind (commented):
Port of `prefer-ending-with-an-expect` from eslint-plugin-jest v29.15.5. Upstream tests are fully migrated, and the structure (shared option parsing with expect-expect, callee-chain naming, await/return/last-statement handling, report location) matches upstream; CI is green.

One nit inline on test comments.

internal/plugins/jest/utils/jest.go
swwind: updated, please use #1705's `esregexp` instead.

eryue0220:
Updated, sorry for the late.

internal/plugins/jest/utils/jest.go
```diff
@@ -165,6 +167,83 @@ func GetJestFnMemberEntries(node *ast.Node) []ParsedJestFnMemberEntry {
...
+	}
+
+	if len(options) == 0 {
+		return parsed
+	}
+	optsMap, _ := options[0].(map[string]interface{})
+	if optsMap == nil {
+		return parsed
+	}
+	if names := internalUtils.ToStringSlice(optsMap["assertFunctionNames"]); names != nil {
```
swwind: `ToStringSlice` returns `nil` both when the key is absent and when it holds an empty array, so `{ "assertFunctionNames": [] }` ends up back at the `["expect"]` default. ESLint keeps the empty list — with `expect` imported from `chai`, `test('x', () => { expect(1).to.equal(1); })` reports there and passes here. Deciding on key presence (the way `parseOptions` in `internal/utils/test_framework/rules/expect_expect` does) preserves it.

internal/plugins/jest/utils/jest.go
```diff
@@ -165,6 +167,83 @@ func GetJestFnMemberEntries(node *ast.Node) []ParsedJestFnMemberEntry {
...
 }
 
+// AssertionFunctionOptions contains the options shared by Jest rules that
+// recognize assertion calls and additional test block functions.
+type AssertionFunctionOptions struct {
+	AssertFunctionNames          []string
+	AdditionalTestBlockFunctions []string
+}
+
+// ParseAssertionFunctionOptions parses the option shape shared by
```
swwind: The comment says this is the shape shared by expect-expect and prefer-ending-with-an-expect, but `jest/expect-expect` is built on `internal/utils/test_framework/rules/expect_expect`, which carries its own `parseOptions` / `compileAssertPattern` / `matchesAssertName` — these three helpers are a second copy of them, down to the `[a-zA-Z0-9.]*` expansion and the `"iu"` flags. Export the existing ones from `test_framework` and drop this copy, so rstest's expect-expect and any future port share one implementation.

internal/plugins/jest/rules/prefer_ending_with_an_expect/prefer_ending_with_an_expect.md
```diff
@@ -0,0 +1,168 @@
+# prefer-ending-with-an-expect
+
+## Rule Details
+
+Prefer ending a test body with an assertion. A test whose last statement is not
+an `expect` (or another configured assert function) often indicates unfinished
+work that may pass silently.
+
+Skipped [`test.todo` / `it.todo`](https://jestjs.io/docs/api#testtodoname)
```
swwind: This rule has no `.todo` handling — `test.todo('x')` passes only because it has fewer than two arguments, and `test.todo('x', () => {})` reports (matching ESLint). The sentence belongs to expect-expect, where the check is real; drop it here.

internal/plugins/jest/rules/prefer_ending_with_an_expect/prefer_ending_with_an_expect_extras_test.go
```diff
@@ -0,0 +1,76 @@
...
+// upstream test suite doesn't exercise. The upstream-mirrored cases live in
+// prefer_ending_with_an_expect_upstream_test.go.
+func TestPreferEndingWithAnExpectExtras(t *testing.T) {
+	rule_tester.RunRuleTester(
+		fixtures.GetRootDir(),
+		"tsconfig.json",
+		t,
+		&prefer_ending_with_an_expect.PreferEndingWithAnExpectRule,
+		[]rule_tester.ValidTestCase{
+			// ---- Dimension 4: parenthesized callback and assertion expression ----
```
swwind: `Dimension 4` doesn't mean anything to a reader of this file — the trailing description alone (`parenthesized callback and assertion expression`) is what carries the case.

swwind (commented):
Three things left before this lands.

`assertFunctionNames: []` is not honored: `utils.ToStringSlice` returns `nil` for an empty array, so an explicitly emptied list restores the `["expect"]` default instead of clearing the patterns.

| Code (options) | ESLint | This PR |
| --- | --- | --- |
| `import { expect } from 'chai';`<br>`test('x', () => { expect(1).to.equal(1); });`<br>with `{ "assertFunctionNames": [] }` | `mustEndWithExpect` | no report |

The same option parsing, pattern compilation and matching already exist as unexported helpers in `internal/utils/test_framework/rules/expect_expect` (`parseOptions` / `compileAssertPatterns` / `matchesAssertName`), and that copy handles the empty list correctly. Lifting them into `internal/utils/test_framework` and using them from both rules removes the duplication and the divergence in one go.

The rest is one doc line and one test-comment nit, inline.

eryue0220:
> Three things left before this lands.
> 
> `assertFunctionNames: []` is not honored: `utils.ToStringSlice` returns `nil` for an empty array, so an explicitly emptied list restores the `["expect"]` default instead of clearing the patterns.
> 
> Code (options)	ESLint	This PR
> `import { expect } from 'chai';`
> `test('x', () => { expect(1).to.equal(1); });`
> with `{ "assertFunctionNames": [] }`	`mustEndWithExpect`	no report
> The same option parsing, pattern compilation and matching already exist as unexported helpers in `internal/utils/test_framework/rules/expect_expect` (`parseOptions` / `compileAssertPatterns` / `matchesAssertName`), and that copy handles the empty list correctly. Lifting them into `internal/utils/test_framework` and using them from both rules removes the duplication and the divergence in one go.
> 
> The rest is one doc line and one test-comment nit, inline.

Updated.

elecmonkey (approved):
LGTM

swwind (approved):
LGTM

PR #1602, opened 2026-08-06
feat(eslint-plugin-jest): add `prefer-importing-jest-globals` rule

eryue0220:
## Summary

* Port the `prefer-importing-jest-globals` rule from eslint-plugin-jest to rslint.

## Related Links

eslint-plugin-jest/prefer-importing-jest-globals [doc](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/prefer-importing-jest-globals.md) [code](https://github.com/jest-community/eslint-plugin-jest/blob/main/src/rules/prefer-importing-jest-globals.ts)

## Checklist

Made with [Cursor](https://cursor.com)


internal/plugins/jest/rules/prefer_importing_jest_globals/prefer_importing_jest_globals.md
```diff
@@ -0,0 +1,89 @@
...
+});
+```
+
+## Original Documentation
+
+- [jest/prefer-importing-jest-globals](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/prefer-importing-jest-globals.md)
+
+## Differences from ESLint
+
+- When `languageOptions.sourceType` is unset and the file has no `import` /
```
elecmonkey: [P3] This documented difference looks stale after `languageOptions.sourceType` support landed. `ResolveLanguageDefaults` now resolves an omitted `sourceType` for `.js` files to `module`, and I confirmed the PR-head CLI autofix inserts `import { ... } from '@jest/globals'` for a `.js` file without imports/exports, matching ESLint flat config's default behavior. Only `.cjs` defaults to CommonJS, so this note should be removed or narrowed to `.cjs`/explicit CommonJS cases.

eryue0220: Updated.

elecmonkey (approved):
LGTM

internal/plugins/jest/rules/prefer_importing_jest_globals/prefer_importing_jest_globals.go
```diff
@@ -0,0 +1,396 @@
...
+	for _, el := range pattern.Elements.Nodes {
+		if el == nil || el.Kind != ast.KindBindingElement {
+			continue
+		}
+		be := el.AsBindingElement()
+		if be.DotDotDotToken != nil {
+			continue
+		}
+
+		keyNode := be.PropertyName
```
swwind, line 229: [P1] ts-go wraps a computed binding property in `ComputedPropertyName`, but this passes that wrapper directly to `accessorValue`. As a result, fixing `const {["describe"]: context}=require("@jest/globals"); describe("s",()=>context("i",()=>test("t",()=>{})))` produces `const { describe, test } = require('@jest/globals'); ...`, deleting the still-used `context` binding. The upstream fixer sees the inner string key and preserves it as `describe: context`. Please unwrap the computed-name node before classifying the accessor, with regression coverage for computed string and template keys.

internal/plugins/jest/rules/prefer_importing_jest_globals/prefer_importing_jest_globals.md
```diff
@@ -0,0 +1,85 @@
...
+describe('suite', () => {
+  test('foo', () => {
+    expect(true).toBeDefined();
+  });
+});
+```
+
+## Original Documentation
+
+- [jest/prefer-importing-jest-globals](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/prefer-importing-jest-globals.md)
```
swwind, line 78: [P3] This points to the moving `main` documentation and does not link the corresponding source, so the exact upstream implementation underlying the copied tests and documented differences cannot be recovered later. Neighboring Jest rule docs pin an upstream release and link both documentation and source. Please do the same here using the exact eslint-plugin-jest version used for this port.

swwind: I can fix it later, nits.

eryue0220: Thank you.

PR #1605, opened 2026-08-06
feat: support `languageOptions.sourceType`

eryue0220:
## Summary

* Add ESLint-compatible `languageOptions.sourceType` parsing and validation (`module` / `script` / `commonjs`), and forward it to native rules via RuleContext.SourceType.
* Only accept the top-level field (no legacy `parserOptions.sourceType`), matching the JS language plugin; unset values stay empty so rules can fall back to structural ESM detection.
* Wire `sourceType` through config registry, linter, and rule tester so tests can simulate `languageOptions.sourceType`.

## Related Links

This is the previous PR for #1602 

## Checklist


elecmonkey:
I personally think this PR is reasonable. Could you also complete the public `defineConfig` type declaration and the related documentation?

swwind:
Thanks for the contribution! This is a valuable addition, especially since `languageOptions.sourceType` is an important part of ESLint compatibility and will allow us to address several existing rule limitations.

One thing I’d suggest is waiting for #1603 to land first. #1603 already adds support for `languageOptions.ecmaVersion`, although it currently doesn’t expose it through `ctx`. I think it would be cleaner to build `sourceType` on top of that work, and then expose the entire `languageOptions` through `ctx` rather than adding each option independently. i.e. `ctx.LanguageOptions.SourceType`

This would also allow us to address the known source-type-related issues in several rules, including `strict`, `no-invalid-this`, `no-redeclare`, `no-shadow`, `no-implied-eval`, and `prefer-object-spread`.

So perhaps we could wait for #1603 to land, and then extend that implementation with `sourceType`. I think this would make the overall implementation a bit cleaner and give us a good foundation for exposing more of `languageOptions` through `ctx` in the future.

eryue0220:
@elecmonkey @swwind updated. 

And you can review again when you're available.

internal/rule/context.go
```diff
@@ -52,6 +52,13 @@ type RuleContext struct {
 	// their effective access, and inline declaration metadata. Rules should use
 	// its accessors instead of reimplementing precedence or version selection.
 	Globals Globals
+	// SourceType is the authored `languageOptions.sourceType` for this file:
+	// "module", "script", "commonjs", or "" when unset. Rules that choose
+	// between ESM `import` and CJS `require` autofix forms should read this;
+	// when empty, fall back to structural detection such as
+	// ast.IsExternalModule. Only the top-level field is forwarded — legacy
+	// `parserOptions.sourceType` is ignored, matching the JS language plugin.
+	SourceType string
```
swwind: This is the per-option field I was hoping to avoid. `rule.LanguageOptions` already exists as the normalized per-file language-option struct, and `ExtractLanguageOptions` already populates it on `ConfiguredRule` — adding `SourceType` there and exposing the struct as `ctx.LanguageOptions` would cover this without new fields on `ConfiguredRule` and `RuleContext`, and would give the next option a place to land.

As it stands `ecmaVersion` and `sourceType` reach rules by two different routes, and only one of them is reachable from `ctx`.

Related: leaving `SourceType` empty when unset makes every consuming rule derive the default from the file extension. `ResolveLanguageDefaults` already knows the `.cjs`/`.mjs`/`.js` defaults, and ESLint resolves this at config time (`.cjs` → commonjs, `.mjs` → module), so resolving to an effective value there — an `EffectiveSourceType()` alongside `EffectiveECMAVersion()` — would keep that mapping in one place instead of in each rule that reads the field.

internal/config/config.go
```diff
@@ -327,6 +330,33 @@ func validateConfigGlobals(languageOptions *LanguageOptions) error {
 	return nil
 }
 
+// validateConfigSourceType rejects authored `languageOptions.sourceType`
+// values that are not ESLint's "module" / "script" / "commonjs". Absent is
+// fine — callers treat unset as structural ESM detection. Matches the JS
+// language plugin, which only reads the top-level field (no legacy
+// `parserOptions.sourceType`).
+func validateConfigSourceType(languageOptions *LanguageOptions) error {
```
swwind, line 338: `validateConfigSourceType` is called from `ValidateConfig` directly, while the `ecmaVersion` and `globals` checks live inside `validateLanguageOptions`. A future caller that validates a single entry through `validateLanguageOptions` would silently skip sourceType. Moving it inside keeps the language-option checks together.

internal/config/rule_registry.go
```diff
@@ -85,6 +86,7 @@ func (r *RuleRegistry) GetEnabledRulesForMergedConfig(mergedConfig *MergedConfig
 					Settings:           CloneSettings(mergedConfig.Settings),
 					LanguageOptions:    languageOptions,
 					Globals:            globals,
+					SourceType:         sourceType,
```
swwind: The config → rule path has no test covering it. The new tests exercise `ValidateConfig` and `ExtractSourceType` as pure functions, no rule reads the value off `ctx`, and the rule_tester `SourceType` field is unused by any case — deleting either this line or the assignment in linter.go:243 leaves the suite green.

A probe rule in rule_tester with one valid and one invalid case asserting the value seen by the rule matches the configured one would pin the wiring down.

internal/rule_tester/rule_tester.go
```diff
@@ -35,9 +35,13 @@ type ValidTestCase struct {
...
 	// Code are picked up automatically by the linter and merged with this — no
 	// separate field needed for those.
-	Globals  map[string]any `json:"globals"`
-	TSConfig string         `json:"tsConfig"`
-	Tsx      bool           `json:"tsx"`
+	Globals map[string]any `json:"globals"`
+	// SourceType simulates top-level `languageOptions.sourceType`
+	// ("module", "script", "commonjs"). Empty leaves RuleContext.SourceType
+	// unset so rules fall back to structural ESM detection.
+	SourceType string `json:"sourceType"`
```
swwind: `SourceType` goes straight into `ConfiguredRule` without validation, unlike `Globals` which goes through `resolveTestCaseGlobals`. On the production path `ValidateConfig` rejects anything outside `module`/`commonjs`/`script` and `ExtractSourceType` drops it, so a test writing `SourceType: "esm"` reaches the rule as `"esm"` — a value no real config can produce, which a rule could then branch on. Reusing `isValidSourceType` here and failing with `t.Fatalf` would keep tests to reachable values.

internal/rules/strict/strict.md
```diff
@@ -13,6 +13,15 @@ The rule supports four options:
 
 When the file is an ES module (detected via top-level `import` / `export`), the rule always uses module semantics: every `"use strict"` directive is reported as unnecessary and removed by autofix.
 
+## Differences from ESLint
```
swwind, line 16: "does not yet consult" reads as a roadmap note. Rule docs describe current behavior, so stating the actual basis — module state is determined by top-level import/export syntax — says the same thing without implying a pending change. Same wording in no_invalid_this.md:30.

swwind (commented):
Thanks for the update. The validation and the config-side plumbing look right — rejecting null and non-strings matches ESLint flat config, and since `languageOptions.Raw` is passed through to the Node worker, the `context.ts` comment about plugin rules reading `languageOptions.sourceType` is accurate. `defineConfig` types and docs are covered now.

The main point from my earlier comment is still open though: this exposes `sourceType` as its own `RuleContext` field rather than through `LanguageOptions`. Details inline, along with a few smaller things.

internal/config/config.go
eryue0220: Updated.

internal/rules/strict/strict.md
eryue0220: Updated.

internal/rule/language_defaults.go
```diff
@@ -67,20 +67,27 @@ var commonJSRefStoreInit = RefStoreInit{
...
+// together with effective language options. Authored options win; otherwise
+// .cjs selects commonjs while .js/.mjs select module. Other extensions retain
+// the LanguageOptions zero-value defaults.
 //
 // JavaScript module files contribute only their non-global top-level scope.
 // An exact, case-sensitive .cjs extension contributes CommonJS's four globals,
 // non-global wrapper scope, and implicit wrapper arguments binding. Other
 // extensions return zero values.
-func ResolveLanguageDefaults(fileName string) (GlobalsInit, RefStoreInit) {
+func ResolveLanguageDefaults(fileName string, languageOptions LanguageOptions) (GlobalsInit, RefStoreInit, LanguageOptions) {
```
swwind, line 99: The authored value only flows into the returned `LanguageOptions`; `GlobalsInit` / `RefStoreInit` are still chosen by extension, which produces the four divergences in the summary. It also means `sourceType` is a no-op for every shipped rule today: nothing reads `ctx.LanguageOptions`, and `strict` / `no-invalid-this` document that they ignore it.

Selecting the inits from the effective source type keeps the mapping in one place:

    if languageOptions.SourceType == "" {
        switch tspath.GetAnyExtensionFromPath(fileName, nil, false) {
        case tspath.ExtensionJs, tspath.ExtensionMjs:
            languageOptions.SourceType = "module"
        case tspath.ExtensionCjs:
            languageOptions.SourceType = "commonjs"
        }
    }
    switch languageOptions.SourceType {
    case "commonjs":
        return commonJSGlobalsInit, commonJSRefStoreInit, languageOptions
    case "module":
        return GlobalsInit{}, moduleRefStoreInit, languageOptions
    }
    return GlobalsInit{}, RefStoreInit{}, languageOptions

Worth deciding explicitly what an authored `sourceType` should do on `.ts` / `.tsx`, which get zero inits regardless of the value.

website/docs/en/config/index.md
```diff
@@ -213,6 +213,26 @@ Selects the standard ECMAScript globals exposed to native rules. Accepted number
 }
 ```
 
+#### languageOptions.sourceType
+
+- **Type:** `'module' | 'script' | 'commonjs'`
+- **Default:** unset
+
+Exposes the configured module kind to rules through their per-file language context. This lets a rule whose behavior depends on module kind distinguish ECMAScript modules, scripts, and CommonJS—for example, to choose between an `import` declaration and a `require()` call. When omitted, the context leaves the value unset so a rule can fall back to structural detection from `import` / `export` syntax.
```
swwind: "the context leaves the value unset" no longer holds for JS files: `.js` / `.mjs` resolve to `module` and `.cjs` to `commonjs` before rules see the value, and `EffectiveSourceType()` never returns empty. Only `.ts` / `.tsx` / `.jsx` keep the empty value. `validateConfigSourceType`'s comment in internal/config/config.go ("callers treat unset as structural ESM detection") is stale for the same reason.

packages/rslint/src/config/config-discovery-protocol.ts
```diff
@@ -61,7 +61,8 @@ export interface FailedConfigModuleResult {
 }
 
 export type ConfigModuleLoadResult =
-  LoadedConfigModuleResult | FailedConfigModuleResult;
+  | LoadedConfigModuleResult
```
swwind: This hunk, `define-config.ts:76` (`RuleEntry`), and `packages/vscode-extension/src/Rslint.ts:60` reformat union types to the leading-`|` form. The repo's pinned Prettier (3.9.6) wants the single-line form that is on `main`, so `pnpm format:check` fails on all three files. They are also unrelated to this change.

swwind (commented):
Thanks — the earlier points are all addressed: `ctx.LanguageOptions` replaces the per-option field, `EffectiveSourceType` / `ResolveLanguageDefaults` own the extension mapping, validation moved into `validateLanguageOptions`, the rule tester rejects unreachable values, and the rule docs read as current behavior.

One thing is still open: the authored value reaches `ctx.LanguageOptions` but never reaches the scope/globals model, so `EffectiveSourceType()` and `ctx.Globals` / `ctx.Refs` can disagree. `ResolveLanguageDefaults` now takes `languageOptions`, but still selects `GlobalsInit` / `RefStoreInit` from the file extension alone.

Checked against ESLint 10.7.0 via `Linter#verify` with flat config:

| Config | Code | ESLint | This PR |
| --- | --- | --- | --- |
| `a.js` + `sourceType: 'commonjs'` | `require('x'); global;` | clean | `no-undef`: `'require' is not defined`, `'global' is not defined` |
| `a.js` + `sourceType: 'commonjs'` | `arguments;` | clean | `no-undef`: `'arguments' is not defined` |
| `a.cjs` + `sourceType: 'module'` | `require('x'); global; arguments;` | 3 x `is not defined` | clean |
| `a.js` + `sourceType: 'script'` | `var Object = 1;` with `no-redeclare` `builtinGlobals: true` | `'Object' is already defined as a built-in global variable` | clean |

Two smaller things inline.

packages/rslint/src/config/config-discovery-protocol.ts
eryue0220: Updated.

website/docs/en/config/index.md
eryue0220: Updated.

internal/rule/language_defaults.go
eryue0220: Updated.

internal/rule/language_defaults.go
```diff
@@ -67,20 +67,32 @@ var commonJSRefStoreInit = RefStoreInit{
...
+		switch tspath.GetAnyExtensionFromPath(fileName, nil, false) {
+		case tspath.ExtensionJs, tspath.ExtensionMjs:
+			languageOptions.SourceType = "module"
+		case tspath.ExtensionCjs:
+			languageOptions.SourceType = "commonjs"
+		}
+	}
+	switch languageOptions.SourceType {
+	case "commonjs":
+		return commonJSGlobalsInit, commonJSRefStoreInit, languageOptions
```
swwind: `commonjs` returns `commonJSRefStoreInit` on every extension, which gives TypeScript files the CommonJS wrapper scope and the implicit `arguments` binding. typescript-eslint's scope manager only knows `module` and `script`, so under ESLint a `.ts` file with `sourceType: 'commonjs'` gets the four CommonJS globals but keeps a global top-level scope and no `arguments`:

- `arguments;` → `'arguments' is not defined`
- `var Object = 0;` with `no-redeclare` `builtinGlobals: true` → `'Object' is already defined as a built-in global variable`

Both are clean here. The `no_undef` valid case at no_undef_test.go:1904 (`require; arguments;` on `authored-commonjs.ts`) locks in the `arguments` half.

Selecting `commonJSGlobalsInit` with a zero `RefStoreInit` on TypeScript-flavoured extensions keeps the JS path on espree semantics and the TS path on typescript-eslint's, which is the split the rest of the resolver follows.

swwind (commented):
The scope-model gap from the last round is closed for JavaScript: `.js` + `commonjs`, `.cjs` + `module`, and `.js` + `script` all match ESLint now, and the resolver owns the mapping in one place.

One divergence left, on TypeScript extensions. Checked against ESLint 10.8.1 with `@typescript-eslint/parser` 8.67.0:

| Config | Code | ESLint | This PR |
| --- | --- | --- | --- |
| `a.ts` + `sourceType: 'commonjs'`, `no-undef` | `arguments;` | `'arguments' is not defined` | clean |
| `a.ts` + `sourceType: 'commonjs'`, `no-redeclare` `builtinGlobals: true` | `var Object = 0;` | `'Object' is already defined as a built-in global variable` | clean |

The four CommonJS globals do apply on `.ts` in ESLint, so that half matches; the wrapper half doesn't. Details inline.

Separately, `EffectiveSourceType()` returns `module` for `.ts` / `.tsx` / `.jsx`, where the resolver deliberately leaves the value empty and the scope model still derives module state from `import` / `export`. That matches typescript-eslint's default, so it is the right answer for a rule to read; it just isn't the answer `ctx.Refs` gives for a TypeScript file without module syntax. Worth a line in the `EffectiveSourceType` doc comment so the first rule that consults it doesn't assume the two agree.

internal/rule/language_defaults.go
eryue0220: Updated.

swwind (commented):
The TypeScript `commonjs` split from the last round is in, and `.ts` / `.tsx` + `commonjs` now matches typescript-eslint on both `arguments` and builtin redeclaration.

Two things the new `isJavaScriptSourceExtension` gate changed by accident, plus one value that is still inert. Checked against ESLint 10.8.1 (espree for `.jsx`, `@typescript-eslint/parser` 8.67.0 for `.ts`):

| Config | Code | ESLint | This PR |
| --- | --- | --- | --- |
| `a.jsx` + `sourceType: 'commonjs'`, `no-undef` | `arguments;` | clean | `'arguments' is not defined` |
| `a.jsx` + `sourceType: 'commonjs'`, `no-redeclare` `builtinGlobals: true` | `var Object = 0;` | clean | `'Object' is already defined as a built-in global variable` |
| `a.ts` + `sourceType: 'script'`, `no-redeclare` `builtinGlobals: true` | `export {};` / `var Object = 0;` | `'Object' is already defined as a built-in global variable` | clean |

`pnpm format:check` also still fails on the three files from last round. Details inline.

internal/rule/language_defaults.go
```diff
@@ -66,21 +66,48 @@ var commonJSRefStoreInit = RefStoreInit{
 	nonGlobalTopLevelScope:  true,
 }
 
+var javascriptSourceExtensions = []string{
+	tspath.ExtensionJs,
+	tspath.ExtensionMjs,
+	tspath.ExtensionCjs,
+}
```
swwind, line 77: `.jsx` is missing here, so it moved from the espree path to the typescript-eslint path in this commit. ESLint parses `.jsx` with espree unless the config assigns another parser, and espree's `commonjs` goal supplies the wrapper scope and the implicit `arguments` binding regardless of extension. `.cts` / `.mts` are correctly out.

```suggestion
var javascriptSourceExtensions = []string{
	tspath.ExtensionJs,
	tspath.ExtensionJsx,
	tspath.ExtensionMjs,
	tspath.ExtensionCjs,
}
```

internal/rule/language_defaults.go
```diff
@@ -66,21 +66,48 @@ var commonJSRefStoreInit = RefStoreInit{
...
+// An omitted source type is filled from the filename for JavaScript files:
+// .js/.mjs select module, .cjs selects commonjs. Other extensions, including
+// .ts/.tsx/.jsx/.cts, keep the empty value. Authored sourceType then selects
+// the inits on every extension. On JavaScript files, commonjs contributes
+// writable exports, read-only global/module/require, a non-global top-level
+// scope, and the wrapper-local arguments binding; module contributes a
+// non-global top-level scope. On TypeScript-flavoured extensions, commonjs
+// contributes only the four CommonJS globals and module contributes no
+// RefStore facts, matching typescript-eslint's scope manager. script and the
+// still-empty TypeScript/JSX value contribute no defaults.
```
swwind: "module contributes no RefStore facts" does not match `case "module"` below, which returns `moduleRefStoreInit` on every extension — language_defaults_test.go:87 asserts `nonGlobalTopLevelScope: true` for `file.ts` + `module`. The code is the right half: typescript-eslint nests a module scope whenever `sourceType === 'module'`, so `a.ts` + `module` with `var Object = 1;` and `builtinGlobals: true` is clean in ESLint too.

Folding in the `.jsx` fix above:

```suggestion
// An omitted source type is filled from the filename for JavaScript files:
// .js/.mjs select module, .cjs selects commonjs. Other extensions, including
// .jsx/.ts/.tsx/.cts, keep the empty value. Authored sourceType then selects
// the inits on every extension. module contributes a non-global top-level
// scope. commonjs contributes writable exports and read-only
// global/module/require everywhere; on espree-parsed extensions
// (.js/.jsx/.mjs/.cjs) it adds the non-global wrapper scope and the
// wrapper-local arguments binding, while TypeScript-flavoured extensions stop
// at the four globals, matching typescript-eslint's scope manager. script and
// the still-empty TypeScript/JSX value contribute no defaults.
```

internal/rule/language_defaults_test.go
```diff
@@ -64,4 +69,49 @@ func TestResolveLanguageDefaults(t *testing.T) {
...
+			nonGlobalTopLevelScope bool
+		}{
+			{fileName: "file.js", authored: "commonjs", wantSourceType: "commonjs", commonJSGlobals: true, commonJSWrapper: true, nonGlobalTopLevelScope: true},
+			{fileName: "file.js", authored: "script", wantSourceType: "script"},
+			{fileName: "file.cjs", authored: "module", wantSourceType: "module", nonGlobalTopLevelScope: true},
+			{fileName: "file.cjs", authored: "script", wantSourceType: "script"},
+			{fileName: "file.ts", authored: "commonjs", wantSourceType: "commonjs", commonJSGlobals: true},
+			{fileName: "file.ts", authored: "module", wantSourceType: "module", nonGlobalTopLevelScope: true},
+			{fileName: "file.tsx", authored: "script", wantSourceType: "script"},
+			{fileName: "file.jsx", authored: "commonjs", wantSourceType: "commonjs", commonJSGlobals: true},
```
swwind: Follows the `.jsx` fix.

```suggestion
			{fileName: "file.jsx", authored: "commonjs", wantSourceType: "commonjs", commonJSGlobals: true, commonJSWrapper: true, nonGlobalTopLevelScope: true},
```

internal/rule/language_defaults.go
```diff
@@ -66,21 +66,48 @@ var commonJSRefStoreInit = RefStoreInit{
...
+	case "commonjs":
+		if isJavaScriptSourceExtension(fileName) {
+			return commonJSGlobalsInit, commonJSRefStoreInit, languageOptions
+		}
+		return commonJSGlobalsInit, RefStoreInit{}, languageOptions
+	case "module":
+		return GlobalsInit{}, moduleRefStoreInit, languageOptions
 	}
-	return GlobalsInit{}, RefStoreInit{}
+	return GlobalsInit{}, RefStoreInit{}, languageOptions
```
swwind, line 119: `script` reaches this zero `RefStoreInit`, and `no_redeclare` computes `ast.IsExternalModule(ctx.SourceFile) || ctx.Refs.HasNonGlobalTopLevelScope()`, so module syntax always wins over the authored value. typescript-eslint keys module scope on `sourceType === 'module'` alone, so a `.ts` file with `export {}` and `sourceType: 'script'` keeps a global program scope there: `export {};` + `var Object = 0;` with `builtinGlobals: true` reports in ESLint and is clean here, same with `import x from 'y';`.

`script` is the one value whose job is to switch module semantics off, so honoring it needs a fact that forces a global top-level scope rather than an additive one. If that is out of scope here, the resolver doc comment is the place to record that `script` is inert wherever module syntax is present.

architecture.md
```diff
@@ -375,21 +381,22 @@ and the effective access after applying their precedence. Rules use
...
+type is filled from the filename for JavaScript files (`.cjs` → `commonjs`,
+`.js`/`.mjs` → `module`); `.ts`/`.tsx`/`.jsx` and other extensions keep the
+empty value. The resolver then selects inits from that effective source type:
+`commonjs` contributes writable `exports`, read-only `global`, `module`, and
+`require`, plus non-global wrapper scope and the wrapper-local `arguments`
+binding; `module` contributes a non-global top-level scope; `script` and the
+still-empty TypeScript/JSX value contribute no defaults. Authored `sourceType`
+therefore applies on every extension, including `.ts`/`.tsx`. The resolver
+does not inspect `package.json`. A rule reads
+`RuleContext.LanguageOptions` when its upstream behavior depends on them.
```
swwind, line 401: This paragraph predates the JS/TS split.

```suggestion
`commonjs` contributes writable `exports`, read-only `global`, `module`, and
`require` on every extension, plus — on espree-parsed extensions
(`.js`/`.jsx`/`.mjs`/`.cjs`) — non-global wrapper scope and the wrapper-local
`arguments` binding; `module` contributes a non-global top-level scope;
`script` and the still-empty TypeScript/JSX value contribute no defaults.
Authored `sourceType` therefore applies on every extension, including
`.ts`/`.tsx`. The resolver does not inspect `package.json`. A rule reads
`RuleContext.LanguageOptions` when its upstream behavior depends on them.
```

packages/rslint/src/config/config-discovery-protocol.ts
swwind: Still present after the last round — `pnpm format:check` fails on this file, `define-config.ts:76` (`RuleEntry`), and `packages/vscode-extension/src/Rslint.ts:60`. Prettier 3.9.6 wants the single-line union that is on `main`, and all three hunks are unrelated to `sourceType`.

internal/rule/language_defaults.go
eryue0220: Updated.

internal/rule/language_defaults.go
eryue0220: Updated.

internal/rule/language_defaults_test.go
eryue0220: Updated.

internal/rule/language_defaults.go
eryue0220: Updated.

architecture.md
eryue0220: Updated.

packages/rslint/src/config/config-discovery-protocol.ts
eryue0220: Updated.

internal/rule/language_defaults.go
```diff
@@ -61,26 +62,59 @@ var commonJSGlobalsInit = GlobalsInit{entries: languageGlobalCatalog}
...
+		case tspath.ExtensionCjs:
+			languageOptions.SourceType = "commonjs"
+		}
+	}
+	switch languageOptions.SourceType {
+	case "commonjs":
+		if isJavaScriptSourceExtension(fileName) {
+			return commonJSGlobalsInit, commonJSRefStoreInit, languageOptions
+		}
+		return commonJSGlobalsInit, RefStoreInit{}, languageOptions
```
swwind, line 113: typescript-eslint's scope manager keys module scope on `sourceType === 'module'` alone, so `commonjs` on a TypeScript extension keeps the global program scope exactly the way `script` does; `export {}` or an `import` in the file does not change that. The zero `RefStoreInit` here leaves the question to `ast.IsExternalModule`, so module syntax wins instead.

```suggestion
		return commonJSGlobalsInit, scriptRefStoreInit, languageOptions
```

With that, the `file.ts` / `file.cts` / `file.mts` + `commonjs` rows in language_defaults_test.go want `globalTopLevelScope: true`, and the doc comment's "stop at the four globals" clause wants the same "global program scope" wording used for `script`. The rest of the suite passes unchanged. If `scriptRefStoreInit` reads odd at two call sites, a name like `globalProgramRefStoreInit` covers both.

internal/rule/ref_store.go
```diff
@@ -478,6 +478,20 @@ func (s *RefStore) HasNonGlobalTopLevelScope() bool {
...
+// global program scope instead.
+func (s *RefStore) HasNonGlobalProgramScope(sourceFile *ast.SourceFile) bool {
+	if s != nil && s.init.globalTopLevelScope {
+		return false
+	}
+	if sourceFile != nil && ast.IsExternalModule(sourceFile) {
+		return true
+	}
+	return s != nil && s.init.nonGlobalTopLevelScope
+}
```
swwind, line 493: The store already holds the file it indexes, so the parameter is redundant and lets a caller pass a different file than the answer is about.

```suggestion
func (s *RefStore) HasNonGlobalProgramScope() bool {
	if s == nil || s.init.globalTopLevelScope {
		return false
	}
	return ast.IsExternalModule(s.sourceFile) || s.init.nonGlobalTopLevelScope
}
```

swwind (commented):
The `script` gap from the last round is closed, `.jsx` is back on the espree path, and `pnpm format:check` is clean. Thanks for working through all the rounds on this.

One divergence is left, the same shape as `script`: `commonjs` on a TypeScript extension still lets module syntax win. Checked against ESLint 10.8.1 with `@typescript-eslint/parser` 8.67.0, `no-redeclare` `builtinGlobals: true`:

| Config | Code | ESLint | This PR |
| --- | --- | --- | --- |
| `a.ts` + `sourceType: 'commonjs'` | `export {};` / `var Object = 0;` | `'Object' is already defined as a built-in global variable` | clean |
| `a.ts` + `sourceType: 'commonjs'` | `import x from './y';` / `var Object = 0;` | `'Object' is already defined as a built-in global variable` | clean |

`.tsx`, `.cts`, and `.mts` behave the same way. One smaller thing inline.

internal/rule/language_defaults.go
```diff
@@ -61,26 +62,59 @@ var commonJSGlobalsInit = GlobalsInit{entries: languageGlobalCatalog}
...
+// global/module/require everywhere; on espree-parsed extensions
+// (.js/.jsx/.mjs/.cjs) it adds the non-global wrapper scope and the
+// wrapper-local arguments binding, while TypeScript-flavoured extensions stop
+// at the four globals, matching typescript-eslint's scope manager. script
+// forces a global program scope even when module syntax is present; the
+// still-empty TypeScript/JSX value contributes no defaults.
+func ResolveLanguageDefaults(fileName string, languageOptions LanguageOptions) (GlobalsInit, RefStoreInit, LanguageOptions) {
+	if languageOptions.SourceType == "" {
+		switch tspath.GetAnyExtensionFromPath(fileName, nil, false) {
+		case tspath.ExtensionJs, tspath.ExtensionMjs:
```
swwind, line 102: `.jsx` sits on the espree side of the split in `javascriptSourceExtensions`, but the default fill leaves it empty, so a `.jsx` file without `import` / `export` gets a global program scope. ESLint gives an espree-parsed `.jsx` the same `module` default as `.js`, so `var Object = 0;` is clean there and reported here.

```suggestion
		case tspath.ExtensionJs, tspath.ExtensionJsx, tspath.ExtensionMjs:
```

Only the `/repo/file.jsx` row in language_defaults_test.go changes with it (`sourceType: "module"`, `nonGlobalTopLevelScope: true`); the no_redeclare, no_undef, and strict suites pass unchanged. If you would rather keep `.jsx` structural, the resolver doc comment and website/docs/en/config/language-options.md are the places to say why — both currently group it with the TypeScript extensions.

swwind (commented):
I ran the full `extension × sourceType` matrix — `.js` / `.jsx` / `.mjs` / `.cjs` / `.ts` / `.tsx` / `.cts` / `.mts` against unset / `module` / `script` / `commonjs` — on both sides, probing the three facts this resolver controls: the CommonJS globals, the wrapper-local `arguments` binding, and whether the program scope is global (`no-undef`, and `no-redeclare` with `builtinGlobals: true`, with and without module syntax). ESLint 10.8.1, espree for the JS extensions and `@typescript-eslint/parser` 8.67.0 for the TypeScript ones.

Every cell matches except the `commonjs`-on-TypeScript one from my previous comment and this one:

| Config | Code | ESLint | This PR |
| --- | --- | --- | --- |
| `a.jsx`, no `sourceType`, `no-redeclare` `builtinGlobals: true` | `var Object = 0;` | clean | `'Object' is already defined as a built-in global variable` |

Cells where espree rejects the source (`script` or `commonjs` together with `import` / `export`) are excluded, since nothing constrains them. Detail inline — that plus the earlier comment is everything I found.

eryue0220:
Updated. Also defaulted unset sourceType on .ts/.tsx/.cts/.mts to module, matching typescript-eslint and closing the last gaps vs ESLint 10.8.1.

internal/rule/ref_store.go
```diff
@@ -570,6 +570,18 @@ func (s *RefStore) HasNonGlobalTopLevelScope() bool {
 	return s != nil && s.init.nonGlobalTopLevelScope
 }
 
+// HasNonGlobalProgramScope reports whether program-level declarations should
+// be treated as living outside the global scope. Module syntax and resolved
+// language defaults can both contribute; an authored global program scope
+// (script, or commonjs on TypeScript-flavoured extensions) forces global
+// program scope instead.
+func (s *RefStore) HasNonGlobalProgramScope() bool {
```
swwind, line 571: `HasNonGlobalProgramScope` takes over the only call site `HasNonGlobalTopLevelScope` had (`no_redeclare.go:467`), so the older method is now uncalled and can go.

internal/rule/language_defaults.go
```diff
@@ -61,26 +62,59 @@ var commonJSGlobalsInit = GlobalsInit{entries: languageGlobalCatalog}
...
+// top-level scope. commonjs contributes writable exports and read-only
+// global/module/require everywhere; on espree-parsed extensions
+// (.js/.jsx/.mjs/.cjs) it adds the non-global wrapper scope and the
+// wrapper-local arguments binding, while TypeScript-flavoured extensions
+// keep a global program scope, matching typescript-eslint's scope manager.
+// script forces a global program scope even when module syntax is present.
+func ResolveLanguageDefaults(fileName string, languageOptions LanguageOptions) (GlobalsInit, RefStoreInit, LanguageOptions) {
+	if languageOptions.SourceType == "" {
+		switch tspath.GetAnyExtensionFromPath(fileName, nil, false) {
+		case tspath.ExtensionJs, tspath.ExtensionJsx, tspath.ExtensionMjs,
```
swwind: ESLint's rule is simpler than the list: every extension except `.cjs` defaults to `module` — I checked `.vue`, `.foo`, `.d.ts`, and an extension-less filename, all `module`. Filling `.cjs` → `commonjs` and everything else → `module` would match that exactly and retire the still-empty case, which no linted extension reaches today (`a.CJS` never enters a TypeScript program on a case-sensitive filesystem). Optional.

swwind (commented):
Two small code-side things inline.

eryue0220:
> Two small code-side things inline.

Updated.

internal/linter/linter.go
```diff
@@ -221,11 +221,15 @@ func runLintRulesInProgram(plan *programLintPlan, opts programRunOptions, consum
...
+		var environment rule.RuleEnvironment
+		if filePlan.environment != nil {
+			environment = *filePlan.environment
+		}
 
 		// Resolve immutable language initialization once per file. Globals and
 		// RefStore receive their own concrete data and never inspect the current
 		// selection input (the file extension) themselves.
-		globalsInit, refsInit := rule.ResolveLanguageDefaults(file.FileName())
+		globalsInit, refsInit, languageOptions := rule.ResolveLanguageDefaults(file.FileName(), environment.LanguageOptions)
```
swwind, line 232: Could we also propagate these effective language options to the ESLint-plugin execution path? That path still receives the raw config, so an omitted `sourceType` appears as `undefined` in plugin contexts. More importantly, `.cjs` files fall back to `module` in the Node runner instead of ESLint’s `commonjs` default, producing different parsing and scope behavior between native and plugin rules.

packages/rslint/src/config/define-config.ts
```diff
@@ -139,6 +139,12 @@ export type GlobalsConfig = Record<string, GlobalAccess>;
 export interface LanguageOptions {
   /** ECMAScript edition used for language globals; omitted defaults to `'latest'`. */
   ecmaVersion?: number | 'latest';
+  /**
+   * Module kind for the file. When omitted, `.js`/`.mjs` resolve to `'module'`
```
swwind: This does not match `ResolveLanguageDefaults`, which defaults every filename except lowercase `.cjs` to `module`, including unknown extensions. The website says “otherwise unset,” while the JSON Schema says rules infer the value from source syntax. Could we align all three public descriptions with the implemented behavior?

internal/linter/linter.go
eryue0220: Updated.

packages/rslint/src/config/define-config.ts
eryue0220: Updated.

swwind (approved):
LGTM. Thanks for your patience.

PR #1608, opened 2026-08-06
feat(config): ship the remaining typescript-eslint presets

swwind:
## Summary

`ts.configs` gains `recommendedTypeChecked`, `strict`, `strictTypeChecked`, `stylistic` and `stylisticTypeChecked`, alongside the existing `recommended`. Each is layered the way typescript-eslint composes its own: the base entry (plugin + typed parser options), the `eslint-recommended` override, then the preset's rules.

Rule sets and option values are taken verbatim from `typescript-eslint`. Every rule in these presets is implemented in rslint.

Also in this PR:

- The `eslint-recommended` override layer is now a shared const rather than inlined in `recommended`.
- `@typescript-eslint/no-unused-vars` keeps its `^_` ignore patterns — rslint's one deliberate divergence from upstream — across every preset that enables it, via a shared `noUnusedVars` const.
- The new presets are registered in `PLUGIN_REGISTRY`, so the docs preset table and the `/rules/?preset=` filter pick them up.
- `Rules & Presets` documents how to pick a baseline and layer a stylistic preset on top.

## Related Links

- https://typescript-eslint.io/users/configs/#recommended-configurations

## Checklist


packages/rslint/src/config/presets/typescript.ts
```diff
@@ -69,19 +180,198 @@ const recommended: RslintConfigEntry[] = [
...
+      '@typescript-eslint/prefer-literal-enum-member': 'error',
       '@typescript-eslint/prefer-namespace-keyword': 'error',
+      'prefer-promise-reject-errors': 'off',
+      '@typescript-eslint/prefer-promise-reject-errors': 'error',
+      '@typescript-eslint/prefer-reduce-type-parameter': 'error',
+      '@typescript-eslint/prefer-return-this-type': 'error',
+      '@typescript-eslint/related-getter-setter-pairs': 'error',
+      'require-await': 'off',
+      '@typescript-eslint/require-await': 'error',
+      '@typescript-eslint/restrict-plus-operands': [
```
elecmonkey, line 275: [P1] Parse the strict plus-operand options

When users select `ts.configs.strictTypeChecked`, this options object reaches Go as a `map[string]interface{}`, but `RestrictPlusOperandsRule` only type-asserts its options to `RestrictPlusOperandsOptions` and otherwise silently restores every `allow*` default to `true`. Consequently the advertised strict preset still permits cases such as `'x' + 1`, boolean/string, nullish/string, `any`, and `RegExp` operands; the native rule must decode the serialized object before this preset can rely on these overrides.

elecmonkey (commented):
The new strict type-checked preset contains an option tuple that the corresponding native rule cannot decode from normal JS/JSON configuration, so a meaningful portion of its promised strict behavior is silently disabled.

packages/rslint/src/config/presets/typescript.ts
swwind: Confirmed — on `main` the rule only type-asserts to `RestrictPlusOperandsOptions`, so a JSON-decoded `map[string]interface{}` silently falls back to all defaults and the preset's overrides are dropped.

This is being fixed in #1469, which converts the `@typescript-eslint/` rules (including `restrict-plus-operands`) to decode serialized option objects. Rather than duplicating that fix here, I'd keep this PR config-only and rebase it once #1469 lands.

fansenze (approved):

PR #1611, opened 2026-08-07
perf(import): derive the cross-file structures once per program

swwind:
## Summary

Every rule that looks past the file it is linting first has to answer the same question: which modules does a file reference, and what do they resolve to. Nothing answered it once. `import/no-cycle` ran a fresh breadth-first search over the whole reachable module graph for every file, re-collecting each visited file's references from its AST at every visit. `import/namespace` re-walked and re-merged a barrel's export map for every file that imported it.

**What changed**

- `rule.ModuleGraph`, reached as `ctx.Modules`, answers what a file references and what it resolves to, per file and per module syntax, derived once per lint run. It reports what the syntax says and nothing more — whether type-only imports count, or anything under `node_modules`, stays the asking rule's question. Module resolution moves to `utils` along the way, since it only ever needed the Program.
- What is specific to eslint-plugin-import — export maps, and the `import/` settings that filter references — stays in the plugin, on a cache keyed by a weak pointer to the Program. An entry cannot keep its Program alive, and a cleanup drops it once the Program goes.
- `import/no-cycle` builds its graph from `ctx.Modules` and runs one Tarjan pass over it. A reference is only reportable when its target reaches the importer again, which puts the two in one strongly connected group, so a file with no such reference answers from the group numbers alone and the searches that do run stay inside the group.
- Merged export maps are kept when the file's re-export closure has no cycle, so a barrel is merged once per run rather than once per importer.

**Behaviour**

Unchanged. Diagnostics are byte-identical to `main` for `import/no-cycle` on VS Code's `./src` (775) and for `import/namespace` on TypeScript's `src/compiler` (0).

Inside a re-export cycle, what a module exports genuinely depends on which module the query entered from. Those maps stay private to the query that built them and are rebuilt each time, exactly as before; `TestExportMapCyclicReexportEntryDependent` holds the three different answers the three entry points into one cycle produce today.

`import/ignore` used to drop references as they were collected and now filters them where they are consumed. Every consumer applies it, and the byte-identical runs above are what confirm the move.

`allowUnsafeDynamicCyclicDependency` withholds a file's dynamically imported targets from the search, but the groups are computed over the whole edge set rather than the smaller one the search then walks. The confinement therefore still holds — a route over the withheld set is also a route over the full one, so its every file both reaches the importer and is reached from it — and that configuration takes the same confined search as every other.

What the confinement is worth there, timed on the search alone over a directly built graph: one 200-file component whose members each import a 20-deep chain costs 63.5ms unconfined against 3.1ms confined, summed over the whole project; 200 small components of the same total size, 1.35ms against 0.18ms. Acyclic projects are unaffected either way, since no file there has a candidate reference at all.

**Numbers**

Rule time from `--timing all`, summed across workers, both binaries built from this branch's base. Synthetic projects are acyclic with fan-out 3.

| | main | this PR |
| --- | --- | --- |
| `no-cycle`, 500 files | 214.7ms | 10.8ms |
| `no-cycle`, 1000 files | 1165.2ms | 21.1ms |
| `no-cycle`, 2000 files | 6266.3ms | 37.6ms |
| `no-cycle`, 4000 files | 31023.1ms | 70.9ms |
| `no-cycle`, 1000 files + `declare namespace` | 10718.0ms | 95.7ms |
| `no-cycle`, 1000 files + dynamic `import()` | 16699.6ms | 405.2ms |
| `no-cycle`, VS Code `./src`, 8188 files | 41213.5ms | 572.8ms |
| `namespace`, 1600-module barrel × 400 importers | 716.1ms | 19.9ms |

On VS Code's `./src` that is 12.69s of wall clock against 2.54s.

The remaining time is dominated by the one-time graph build, which is attributed to whichever file asks for it first. Per-file cost after that is close to nothing.

Single-file linting from a cold Program, the path the language server takes on open, is unchanged: the graph build is within noise of constructing the Program itself.

**Known limitation, left for later.** A Program is immutable, so an edit produces a new one and the cache keyed by it starts empty. The language server therefore rebuilds the whole dependency graph on every re-lint — every file of the Program, `node_modules` `.d.ts` included — where an incremental Program build itself reuses the unchanged ASTs and costs much less. `import/no-cycle` does run in editors, so a large project pays that build per re-lint. The fix is to key the per-file edges by the source file rather than by the Program, so an unchanged file's edges survive the update; resolution answers are not obviously stable across Programs, so that needs its own change and its own benchmark. This is a separate follow-up, not a blocker here.

## Related Links

None.

## Checklist


internal/plugins/import/utils/module_index.go
```diff
@@ -0,0 +1,95 @@
...
+// answers that for every rule, not just these. This holds the part that is
+// specific to eslint-plugin-import — export maps, and the `import/` settings
+// that decide which references count.
+//
+// The index is shared by every file of the run, so it holds the Program it was
+// built for rather than taking a RuleContext per call: a context belongs to
+// one file, and nothing cached here may depend on which file happened to ask
+// first.
+type ModuleIndex struct {
+	program  *compiler.Program
```
elecmonkey: `program_cache.go:17-19` requires that nothing stored in the program cache hold
a strong reference back to its Program, so that the weak key's cleanup can drop
the entry. This field breaks that: `IndexFor` (line 49) stores the `*ModuleIndex`
through `rule.CachedByProgram`, giving
`programCaches -> *programCache -> *cachedValue -> *ModuleIndex -> *compiler.Program`.
The Program is therefore permanently reachable, the `runtime.AddCleanup` at
`program_cache.go:50` never fires, and the entry never goes away.

In the editor this leaks a Program per keystroke: `lsp/service.go:1096` lints
through `LintSingleFile`, and `lintProgramRequest.load` replaces `state.program`
with the result of `program.UpdateProgram(...)`
(`internal/lsp/lint_program_store.go:113-125`), so every buffer edit produces a
new Program. The rules that reach `IndexFor` are `import/default` (via
`HasDefaultExport`) and `import/namespace` (via `GetExportMap`); either one pins
the Program it ran under, along with that index's export maps. Consecutive
Programs share unchanged source files, so what accumulates per edit is the
Program and its caches rather than a full copy of the file set — but it
accumulates without bound.

`TestCachedByProgramReleasesCollectedPrograms` stores an `*int`, so it passes
regardless. A `weak.Pointer` here, or passing the Program per call instead of
storing it, would restore the invariant.

internal/plugins/import/rules/no_cycle/graph.go
```diff
@@ -0,0 +1,382 @@
...
+		syntax:             opts.syntax,
+		ignoreExternal:     opts.ignoreExternal,
+		allowUnsafeDynamic: opts.allowUnsafeDynamicCyclicDependency,
+	}
+	return rule.CachedByProgram(ctx.Program, key, func() *moduleGraph {
+		return buildModuleGraph(ctx, settings, opts)
+	})
+}
+
+func buildModuleGraph(ctx rule.RuleContext, settings *import_utils.ModuleSettings, opts ruleOptions) *moduleGraph {
```
elecmonkey, line 75: This builds the graph over every file of the Program — `program.SourceFiles()`,
which under the default `ignoreExternal: false` includes node_modules `.d.ts` —
and runs Tarjan over all of it. `checkSourceFile` (`no_cycle.go:124`) reaches it
unconditionally, before any check on whether the linted file has references of
its own, for every file linted.

That amortizes well over a whole-project run. It does not amortize on the
`LintSingleFile` path: a new Program per edit means the program-keyed cache
never hits across edits, so one file's lint now pays a full-program edge
collection plus an SCC pass where the old lazy search from the linted file's own
imports did almost nothing for an import-free file.

Checking `ctx.Modules.Edges(ctx.SourceFile, opts.syntax)` before building would
cover the leaf-file case; beyond that it would be good to see a single-file
measurement, since that is where the tradeoff inverts.

swwind:
A serious memory leak found in large project. Mark this as draft for now, let me digging more.

swwind:
## Memory on VS Code's `./src`, ten edits per run

| | peak RSS | live heap after GC | live source files |
| --- | --- | --- | --- |
| base `032623e9` | 11.75 GB | — | — |
| + leak fix | 20.13 GB | 10.37 GB | 43,337 |
| + prune | 20.39 GB | 4.31 GB | 17,569 |
| + `GOMEMLIMIT=8GiB` | 7.60 GB | 4.39 GB | 17,568 |

Pruning collected entries as each new Program arrives halves what is retained — 17,569 source files is about one Program for the language service and one for the lint, which is the floor — and moves the peak by nothing. A memory limit moves the peak to below base and changes retention by nothing.

The peak is therefore garbage, not retention: `buildModuleGraph` walks every file of the Program, resolving every specifier, and the editor discards the result on the next keystroke. Keying a file's edges by its source file rather than by the Program is what removes that work — an incremental Program reuses the source files it did not change. The description lists it as a follow-up; these numbers make it a blocker.

internal/plugins/import/utils/module_index.go
swwind: fixed memory leakage

internal/plugins/import/rules/no_cycle/graph.go
swwind: added the prune

swwind:
I'll leave the peak RSS issue in another PR, this only focus on the correctness of the module graph system.

elecmonkey (approved):
LGTM

fansenze (approved):

PR #1612, opened 2026-08-07
feat(typescript-eslint): port rule no-unnecessary-type-parameters

swwind:
## Summary

Port the `@typescript-eslint/no-unnecessary-type-parameters` rule from typescript-eslint to rslint.

Flags type parameters on functions, methods, and classes that aren't used more than once in the signature, since a type parameter used only once doesn't relate anything and can be replaced by its constraint (or `unknown`).

Behavior matches upstream except where upstream is limited by the TypeScript public API and rslint is not: a mapped type's `as` clause and index signatures with symbol or pattern keys count as uses here, so those false positives are not reported. The `replaceUsagesWithConstraint` suggestion also parenthesizes the constraint wherever the type grammar requires it, keeping the replacement the same type as the code it replaces. Both are described in the rule docs.

## Related Links

- ESLint rule: https://typescript-eslint.io/rules/no-unnecessary-type-parameters
- Source code: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/rules/no-unnecessary-type-parameters.ts

## Checklist


elecmonkey (approved):

elecmonkey (approved):
no issues found

PR #1614, opened 2026-08-07
feat(rstest): add expect-expect rule

elecmonkey:
## Summary

Port `eslint-plugin-jest`'s `expect-expect` to rstest: every test registration must contain at least one assertion, otherwise the test can pass without verifying anything. Added to the rstest `recommended` preset at `warn`, matching `jest/expect-expect`.

The framework-neutral ledger is extracted into a shared body under `internal/utils/test_framework/rules/expect_expect`; `jest/expect-expect` becomes a thin wrapper over it.

## Behavior change in `jest/expect-expect`

Extracting the shared body also fixes one order-dependent false positive in `jest/expect-expect`. For a named callback whose declaration is asserting, the registration is now treated as covered **regardless of source order**:

```js
function myTest() {
  expect(true).toBeDefined();
}
it('should fail', myTest);
```

Upstream `eslint-plugin-jest` reports `Test has no assertions` here because the assertion is walked before the `it(...)` registration is queued; the shared engine instead records the asserting declaration and clears the registration whenever it is seen. Since the function declaration is hoisted, this snippet is runtime-equivalent to the call-first form (`it(...); function myTest() { ... }`) that both implementations already accept, so treating the two identically is the correct behavior. The previously-invalid case is now a valid test case, and the divergence is documented in the rule doc.

## Differences from jest

rstest's assertion API is vitest/chai lineage, so the wrapper diverges from jest where the framework differs:

- **default `assertFunctionNames` is `["expect", "assert"]`** — rstest exposes both as globals; no `expectTypeOf` / `bench` handling, since rstest core has neither.
- **assertions are recognized through the shared rstest call analysis, not by callee text alone** — `context.expect(...)`, `rstest.expect(...)`, an aliased `import { expect as check }` and `import.meta.rstest.expect(...)` all count without configuration, matching what `rstest/no-conditional-expect` already recognizes.
- **the `.todo` exemption reads the semantic `ParsedRstestFnCall.Todo` field**, which survives const aliases (`const t = test.todo; t('x')`), rather than the call-site `Members`.
- **test registrations are resolved through the shared rstest call analysis (`GetRstestCallAnalysis` / `ParseFnCall`)**, covering `import.meta.rstest`, `test.extend`, `.each` / `.for`, and `@rstest/playwright`.

## Related Links

## Checklist


internal/plugins/jest/rules/expect_expect/expect_expect_test.go
```diff
@@ -22,6 +22,8 @@ func TestExpectExpectRule(t *testing.T) {
 			{Code: `test("should pass", () => expect(true).toBeDefined())`},
 			{Code: `it("should pass", () => somePromise().then(() => expect(true).toBeDefined()))`},
 			{Code: `it("should pass", myTest); function myTest() { expect(true).toBeDefined() }`},
+			{Code: `function myTest() { expect(true).toBeDefined() } it("should pass", myTest);`},
```
swwind, line 25: Behavior divergence from eslint-plugin-jest. This exact code — `function myTest() { expect(true).toBeDefined() } it("should fail", myTest);` — was previously in the invalid-case list (removed a few lines below) because upstream eslint-plugin-jest reports `Test has no assertions` for it (verified live). It's now valid because `assertedByDecl`/`assertedByName` in the shared engine are set the moment the ancestor walk from `expect(...)` crosses the function declaration, independent of whether the `it(...)` registration has been seen yet. If this order-independence is intentional (arguably a correctness improvement over upstream's order-dependent behavior), it should be called out in the rule doc and PR description rather than silently swapped from invalid to valid; if not intentional, a registration should only be skipped as "already asserted" if it was already queued when the assertion was walked.

internal/utils/test_framework/rules/expect_expect/expect_expect.go
```diff
@@ -0,0 +1,301 @@
...
+
+					if isTest || isExtraBlock {
+						if classification.IsTodo || strings.HasSuffix(calleeName, ".todo") {
+							return
+						}
+						if isTest && runtime.ResolveNamedCallback != nil {
+							named := runtime.ResolveNamedCallback(node)
+							if named.DeclarationNode != nil && assertedByDecl[named.DeclarationNode] ||
+								named.DeclarationNode == nil && named.Name != "" && assertedByName[named.Name] {
+								return
```
swwind, line 305: Root cause of the divergence noted in expect_expect_test.go. `assertedByDecl[named.DeclarationNode]` / `assertedByName[named.Name]` are checked before the call is ever added to `uncheckedByDecl`/`uncheckedByName`, so a registration is skipped whenever *any* earlier-visited assertion anywhere in the file touched that declaration/name — not just one that happened before this particular registration. That's what makes source order stop mattering.

internal/utils/test_framework/rules/expect_expect/expect_expect.go
```diff
@@ -0,0 +1,301 @@
...
+) {
+	var ancestors []*ast.Node
+	for n := assertNode.Parent; n != nil; n = n.Parent {
+		ancestors = append(ancestors, n)
+	}
+
+	for i := len(ancestors) - 1; i >= 0; i-- {
+		n := ancestors[i]
+		switch n.Kind {
+		case ast.KindFunctionDeclaration:
```
swwind, line 241: Duplicated branches. `KindFunctionDeclaration` (here) and `KindVariableDeclaration` (below) both derive a `(declNode, fnName)` pair and then run the same four-step "mark asserted, clear+delete from both unchecked maps" sequence. Worth extracting into a `markAsserted(declNode, fnName, unchecked, uncheckedByDecl, uncheckedByName, assertedByDecl, assertedByName)` helper so the two branches can't drift apart.

internal/plugins/rstest/rules/expect_expect/expect_expect.go
```diff
@@ -0,0 +1,38 @@
...
+				if parsed == nil || parsed.Kind != rstestUtils.RstestFnTypeTest {
+					return shared.TestClassification{}
+				}
+				// Todo is a semantic field that survives const aliases; the
+				// call-site Members can be empty for `const t = test.todo;
+				// t('x')`, so never derive the exemption from Members here.
+				return shared.TestClassification{IsTest: true, IsTodo: parsed.Todo}
+			},
+			ResolveNamedCallback: func(callNode *ast.Node) shared.NamedCallback {
+				function, name := analysis.TestCallback(callNode)
```
swwind, line 54: Duplicate TypeChecker resolution. `analysis.TestCallback(callNode)` reruns the full TypeChecker-based resolution that `collectRstestTestCallbacks` (shared via `analysis.Callbacks()`, used by other rstest rules like `no_conditional_expect`/`valid_title`) already performs once per test call per file. When both run in the same lint pass, this declaration gets resolved twice. Not currently exposed as a per-node cache, so likely low priority.

swwind (commented):
**Summary**

| Code | eslint-plugin-jest | This PR (`jest/expect-expect`) |
|---|---|---|
| `function myTest() { expect(true).toBeDefined() } it("should fail", myTest);` | reports `Test has no assertions` | no report |

Missed report: when an asserting named callback is declared *before* the `it(...)` call that registers it, the shared engine now marks it "asserted" permanently (`assertedByDecl`/`assertedByName`) and the later registration is skipped outright — order no longer matters, which diverges from upstream's forward-pass tracking. The PR's added test case for this exact code (now under "should pass") codifies the new behavior; the old failing case was deleted rather than kept as a documented exception. A few reuse/simplification items too. Details inline.

Also, not part of this diff so no inline anchor: `internal/plugins/jest/utils/parse_jest_fn.go`'s `ResolveNamedFunctionCallback` (line 462) has no remaining callers in this branch, and `internal/plugins/jest/rules/expect_expect/expect_expect.schema.json` is now an unreferenced duplicate of the shared package's schema — both are dead and can be deleted.

internal/plugins/jest/rules/expect_expect/expect_expect_test.go
elecmonkey: This is intentional. Since function declarations are hoisted, the declaration-first form is runtime-equivalent to the call-first form both already accept, so reporting one but not the other is an order-dependent false positive. Documented in the rule doc and PR description in `33cdb5c1`.

internal/utils/test_framework/rules/expect_expect/expect_expect.go
elecmonkey: Done in `955ad7a8` — extracted a shared `markAsserted(declNode, fnName, ...)` helper; both the FunctionDeclaration and VariableDeclaration branches now call it, so they can't drift.

internal/plugins/rstest/rules/expect_expect/expect_expect.go
elecmonkey: Agreed, low priority. `TestCallback` reuses the existing overload-aware resolver rather than adding a second collector; deduplicating against `Callbacks()` would require exposing a per-node cache on the analysis, which the shared-analysis design intentionally leaves out for now. Leaving as-is.

elecmonkey:
Removed both in `c921ae56`: the now-unreferenced `ResolveNamedFunctionCallback` in `parse_jest_fn.go` and the duplicate `jest/expect_expect/expect_expect.schema.json` (the shared package embeds the schema now).

internal/utils/test_framework/rules/expect_expect/expect_expect.go
elecmonkey: This is the mechanism behind the intentional divergence discussed above — the `assertedByDecl`/`assertedByName` memo is what makes order stop mattering, which is correct given function-declaration hoisting. See the reply on the test-file thread and the rule-doc note in `33cdb5c1`.

swwind (commented):
**Summary**

Two things, one in the rstest adapter and one in the shared engine. Details inline.

**1. `rstest/expect-expect` matches assertions by callee name only, so every non-bare `expect` is a false positive.** Same four forms, same plugin — `rstest/no-conditional-expect` recognizes all of them via `analysis.IsExpectCall`:

| `expect` form inside the test callback | `rstest/no-conditional-expect` | `rstest/expect-expect` (this PR) |
|---|---|---|
| `ctx.expect(v).toBe(1)` (test context) | recognized | `Test has no assertions` |
| `rstest.expect(v).toBe(1)` (namespace import) | recognized | `Test has no assertions` |
| `check(v).toBe(1)` (`import { expect as check }`) | recognized | `Test has no assertions` |
| `import.meta.rstest.expect(v).toBe(1)` | recognized | `Test has no assertions` |

The rule ships in the `rstest` recommended preset at `warn`, so these fire out of the box.

**2. The ledger credits a callback for assertions that never run when it is the callback.**

|---|---|---|
| `const outerCb = () => { test("inner", () => { expect(v).toBe(1) }) }; test("outer", outerCb);` | reports | no report |
| `const makeCb = () => () => { expect(v).toBe(1) }; test("x", makeCb);` | reports | no report |

The `function` equivalents of both already behave this way on `main`, so this is a pre-existing gap rather than a regression — but the new `KindVariableDeclaration` branch extends it to arrow/`const` callbacks, which is a new surface.

internal/plugins/rstest/rules/expect_expect/expect_expect.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+var ExpectExpectRule = shared.NewRule(shared.Config{
+	Name: "rstest/expect-expect",
+	// Rstest ships both `expect` and `assert` (chai) as globals
+	// (packages/core/src/utils/constants.ts globalApiList), so the default
+	// asserting functions match eslint-plugin-vitest rather than jest.
+	DefaultAssertFunctionNames: []string{"expect", "assert"},
+	Prepare: func(ctx rule.RuleContext) shared.Runtime {
+		analysis := rstestUtils.GetRstestCallAnalysis(ctx)
```
swwind, line 30: `analysis` is consulted for test *registrations* (`ClassifyTest`, `ResolveNamedCallback`) but never for *assertions* — those go through the shared engine's callee-chain match against the `assertFunctionNames` regexes, which only sees a bare `expect(...)`. That's the four false positives in the summary table.

`rstest/no-conditional-expect` already solves exactly this: it passes an `IsExpectCall` hook on its shared runtime backed by `analysis.IsExpectCall`, which resolves `ctx.expect`, namespace imports, aliased imports and `import.meta.rstest.expect`. An equivalent `IsAssertion` hook on `shared.Runtime`, tried before the name-pattern match, would close all four at once and cost nothing for the jest adapter (which just leaves it nil).

Worth noting the destructured `test("case", ({ expect }) => ...)` form is already an explicit valid test case here — `ctx.expect` is the same context object, just not destructured, so supporting one and not the other is hard to explain to a user.

internal/utils/test_framework/rules/expect_expect/expect_expect.go
```diff
@@ -0,0 +1,310 @@
...
+			decl := n.AsFunctionDeclaration()
+			if decl != nil && decl.Name() != nil {
+				markAsserted(decl.AsNode(), decl.Name().Text(),
+					unchecked, uncheckedByDecl, uncheckedByName, assertedByDecl, assertedByName)
+			}
+		case ast.KindVariableDeclaration:
```
swwind, line 247: This branch marks the declared function asserted from *any* assertion lexically nested under it, including ones that cannot run when it is used as the test callback:

```js
const outerCb = () => { test("inner", () => { expect(v).toBe(1); }); };
test("outer", outerCb);   // outer callback asserts nothing — no report

const makeCb = () => () => { expect(v).toBe(1); };
test("x", makeCb);        // callback returns a closure, never asserts — no report
```

eslint-plugin-jest reports both. To be fair this isn't introduced here — the `KindFunctionDeclaration` branch above has the same shape and the `function` versions of both already pass on `main`, since the ancestor walk marks every matching declaration it crosses regardless of what sits between the assertion and that declaration. But this branch extends the same behavior to arrow/`const` callbacks, which is new surface.

Fixing it properly means bounding the walk: stop crediting a declaration once the walk has crossed a function that is itself registered as another test's callback (case 1), and don't credit across a `return`ed inner function (case 2). Both need more than a local tweak, so this may be reasonable to leave as a known limitation — but if so it's worth a note in the doc, since the current text says assertions inside a variable function "are attributed to every test that references it" without qualifying which assertions count.

Scope-blind name reuse and overloaded function declarations were probed separately and behave correctly, so this looks like the only leak in the ledger.

internal/plugins/jest/rules/expect_expect/expect_expect.md
```diff
@@ -2,10 +2,23 @@
...
+Ensure every Jest test callback contains at least one assertion. The rule tracks test APIs such as `test`, `it`, `fit`, `xit`, and `xtest` (including chained forms like `it.each` that the Jest integration recognizes) and reports when none of the configured assertion callee patterns appear in the body. Assertions inside a named function declaration or variable function passed as the test callback are attributed to every test that references it, regardless of whether the callback is declared before or after the registration. This guards against tests that run side effects but never verify outcomes.
 
 Skipped [`test.todo` / `it.todo`](https://jestjs.io/docs/api#testtodotitle) bodies are ignored.
 
+### Divergence from `eslint-plugin-jest`
+
+For a named callback whose declaration is asserting, this rule intentionally treats the registration as covered **independent of source order**, whereas upstream `eslint-plugin-jest` only clears a registration if it was already seen when the assertion was walked. Concretely:
```
swwind: The hoisting argument is sound for the `function` form, but `assertedByDecl`/`assertedByName` are now also populated from the `KindVariableDeclaration` branch, and the section doesn't cover that case.

Upstream reports variable-declared callbacks in *both* orders (verified against eslint-plugin-jest — it only resolves `FunctionDeclaration`, so `const`/`var` callbacks are never cleared either way). This rule accepts both. So for variable callbacks the divergence isn't order-independence, it's callback resolution that upstream doesn't do at all — arguably the better behavior, but a different claim than the one written here.

The order-independent half also doesn't transfer: `it('x', myTest); const myTest = () => { expect(1).toBe(1) };` is a runtime `ReferenceError`, and `const` is not hoisted, so "runtime-equivalent to the call-first form" doesn't hold there. No user-visible harm either way — the snippet is broken code regardless — but the section currently promises more than the implementation guarantees.

internal/utils/test_framework/rules/expect_expect/expect_expect.go
elecmonkey: Agreed, and thanks for confirming both shapes predate this PR. Leaving as-is: bounding the walk risks the opposite failure mode, and a false `Test has no assertions` on valid code is worse here than missing these two shapes. Recorded as a known limitation in a doc comment on `checkCallExpressionUsed` (covers both branches) rather than the rule doc.

internal/plugins/rstest/rules/expect_expect/expect_expect.go
elecmonkey: Fixed in `c4a1a1ec`. Added an optional `IsAssertion` hook on the shared runtime; rstest backs it with `analysis.IsExpectCall`, jest leaves it nil. Two deviations: it runs *after* the pattern match, so a bare `expect()` never touches the analysis, and it is only installed when the file mentions `expect` at all, since `IsExpectCall` forces callback collection this rule otherwise never triggers. Regression cases added for all four forms, plus one asserting an unresolved `agent.expect(200)` still reports.

internal/plugins/jest/rules/expect_expect/expect_expect.md
elecmonkey: You're right on both counts. Narrowed in `c4a1a1ec`: the hoisting argument now stands on its own for function declarations, and variable callbacks are described as resolution upstream doesn't do at all rather than as order-independence.

internal/plugins/rstest/rules/expect_expect/expect_expect.md
```diff
@@ -0,0 +1,86 @@
...
+### `assertFunctionNames`
+
+Names of functions treated as assertions. Defaults to `["expect", "assert"]`,
+matching the two assertion entry points Rstest exposes as globals. Names support
+wildcards: `request.*.expect`, `request.**.expect`, `expect*`.
+
+Rstest's own `expect` counts whatever the callee looks like, so these names only
+need to cover assertion helpers the rule cannot resolve. `context.expect(...)`,
+`rstest.expect(...)`, an aliased `import { expect as check }` and
+`import.meta.rstest.expect(...)` are all recognized without configuration.
```
swwind: Because the resolving hook runs as a union with the patterns, the option became additive-only for `expect`:

| Config + code | eslint-plugin-jest | This PR |
|---|---|---|
| `assertFunctionNames: ["myAssert"]`, `test("case", () => { expect(v).toBe(1) })` | reports `Test has no assertions` | no report |

Upstream replaces the default list, so naming a custom helper there also stops crediting bare `expect`; here it cannot. The trade-off looks right — resolving `ctx.expect` is worth more than the opt-out — but this paragraph reads as if it were only about the extra forms. One sentence saying the option adds to Rstest's own `expect` rather than replacing it would save a user the surprise.

swwind (commented):
**Summary**

Final pass. Everything from the previous rounds checks out: the four unresolved `expect` forms, the two dead-code deletions, and the narrowed divergence note. Re-verified the doc claims against upstream — `eslint-plugin-jest` reports the `function`-declaration form only when the declaration comes first, and reports `const`/`var` callbacks in both orders, which is what the doc now says. Also probed the new hook for over-crediting (`ctx` forwarded into a helper, an assertion in `beforeEach`, a supertest chain under defaults) — all still report.

One leftover, non-blocking: `assertFunctionNames` can no longer turn `expect` off. Detail inline.

internal/plugins/rstest/rules/expect_expect/expect_expect.md
elecmonkey: Documented in 43564962. The `assertFunctionNames` section now says the list replaces the default rather than extending it, and that naming your own helpers cannot switch Rstest's `expect` off — with the divergence from `eslint-plugin-jest`, where dropping `expect` from the list stops it counting, called out explicitly. Thanks!

swwind (approved):

fansenze (approved):

PR #1638, opened 2026-08-09
feat: add demand-driven RefStore facts

fansenze:
## Summary

- Extend the single per-file `RefStore` with neutral reference, authored-binding, and syntax-only import facts; no parallel binding/scope store is introduced.
- Add static `RuleNeeds` ceilings plus per-file `RequestRefs` demand. A private linter collector can gather requested facts during the existing traversal, then seal them before a dedicated file-finalize listener.
- Preserve legacy lazy queries and correctness-safe early-query fallback. Partial/live collections are never published, and zero-request files keep the original traversal path.
- Document raw TypeScript symbol and ESLint scope boundaries, parameter-property dual identities, and shared-observer timing attribution.
- This PR migrates no production rules and does not touch ModuleGraph or `import/no-cycle`.

## Benchmarks

Apple M1 Max, `-cpu=1`, paired against `origin/main` (`25db8e9f`):

- Zero-request 128-small-file guard: paired directions were +0.29% and -0.75%; both sides were exactly 58,528 B/op and 794 allocs/op.
- Finalize-safe streaming: references -3.05%, binding declarations -8.92%, imports -12.07%, all facts -14.91%; every lazy/streaming pair had identical B/op and allocs/op.
- Early-query fallback overhead: +2.42% at 10%, +8.68% at 50%, +15.50% at 90%. Future migrations must veto streaming when representative configurations show meaningful fallback and must use end-to-end wall/alloc measurements rather than only a per-rule `--timing` line.

## Test Plan

- `go test ./internal/rule -run` with the RefStore, facts, lifecycle, needs, reporter-layout, and request-ceiling test set
- `go test ./internal/linter -run` with RefStore lifecycle, traversal, listener-registry, and timing tests
- `go test -p 1 ./internal/config -run ^TestRuleRegistryPropagatesRuleNeeds$ -count=1`
- `pnpm exec prettier --check architecture.md`
- Linter-level zero-request, references/fallback, and facts benchmarks with `-benchmem`

## Related Links

N/A

## Checklist

### DIFF
--- a/architecture.md
+++ b/architecture.md
@@ -252,6 +252,7 @@ Rules are defined in `internal/rule/rule.go`:
 type Rule struct {
     Name             string
     RequiresTypeInfo bool
+    Needs            RuleNeeds
     Run              func(ctx RuleContext, options []any) RuleListeners
 }
 
@@ -301,6 +302,7 @@ func (*RuleContext) ReportNodeWithDeferredFixesAndSuggestions(
 func (*RuleContext) ReportRangeWithDeferredFixesAndSuggestions(
     ..., func() []RuleFix, func() []RuleSuggestion,
 )
+func (RuleContext) RequestRefs(needs RefNeeds) bool
 ```
 
 The linter creates one short-lived `CommentStore` per file. `Comments.All()`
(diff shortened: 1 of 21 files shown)


fansenze:
Closing after re-auditing the design and validating it with real rule prototypes.

The proposed framework does not have sufficient end-to-end ROI: the existing per-file RefStore already shares one lazy reference scan across all rules, while streaming makes collection eager for every enabled file and moves observer work outside per-rule --timing. In default no-timing single-thread runs, user CPU results were -8.1%, +2.5%, and +11.0% versus main (median about +2.5% regression), so the apparent listener-time improvement is not a reliable framework gain.

The RuleNeeds/facts/file-finalize expansion is therefore not justified. No redo prototype, rule migration, or benchmark file will be pushed from this investigation. Rule-local improvements that only need the existing RefStore should be handled separately.

PR #1669, opened 2026-08-10
test: resolve rule source paths under -trimpath builds

swwind:
## Summary

`TestAllRules_NilTypeCheckerEarlyReturnImpliesRequiresTypeInfo` maps each registered rule's `Run` back to its source file via `runtime.FuncForPC(...).FileLine()`, then parses that file. Under `-trimpath` the reported name is import-path-rooted (`github.com/web-infra-dev/rslint/internal/plugins/.../foo.go`) rather than absolute, so `parser.ParseFile` cannot open it and the test aborts on the first rule.

`ruleSourceParser` now locates the enclosing `go.mod` once (module path + module root) and, for any non-absolute reported name, swaps the module path prefix for the module root. Absolute paths are used as-is, so both trimmed and untrimmed builds work. When the module root can't be found or the path lies outside the module, the test still errors out rather than silently skipping the rule.

## Checklist

### DIFF
--- a/internal/config/all_rules_requires_typeinfo_test.go
+++ b/internal/config/all_rules_requires_typeinfo_test.go
@@ -8,6 +8,9 @@ import (
 	"go/ast"
 	"go/parser"
 	"go/token"
+	"os"
+	"path"
+	"path/filepath"
 	"reflect"
 	"runtime"
 	"sort"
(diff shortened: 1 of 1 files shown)


elecmonkey (approved):
reviewed

PR #1676, opened 2026-08-10
perf(lsp): collect after a lint pass drops what it derived

swwind:
## Summary

A lint derives whole-Program structures, hands back diagnostics, and drops them. The collector paces off heap growth relative to what is live, which has no way to know that happened, so the next edit's allocation stacks on the last one's garbage and the heap climbs well past what is actually reachable.

Asking for a collection once the lint has dropped what it derived keeps that from accumulating. It runs on its own goroutine — `runtime.GC` blocks its caller and relinting happens on the main goroutine, so collecting there would stall every editor interaction for as long as marking takes. No guard is needed: `runtime.GC` waits for the in-flight mark and then starts a new cycle, and `gcStart` re-checks `gcTriggerCycle` under `work.startSema`, so concurrent callers that observed the same cycle collapse into one.

Measured with a scripted LSP client over stdio, so every variant sees a byte-identical message sequence. VS Code 1.128.0 `src` (7464 `.ts`), all core rules plus all nine plugins, one document, ten edits, `GODEBUG=gctrace=1`:

| | peak RSS | total CPU | GC CPU | forced cycles |
| --- | --- | --- | --- | --- |
| main | 7011 MB | 81.1 s | 32.8 s | 1 |
| this PR | 4182 MB | 110.7 s | 62.9 s | 12 |

Peak RSS drops 40%. Total CPU rises 36%, and the rise is entirely GC CPU: one extra full mark per lint pass, about 2.7 s at this project's live heap. Live heap is the same for both variants, ~1.97 GB in every run measured, which is what says this is garbage rather than retention.

Forced cycles come out at exactly one per lint pass plus one from upstream's idle cache clean (`typescript-go/internal/project/session.go:566`), in every run.

A soft memory limit reaches a similar peak for more CPU, and its cost varies with typing rhythm — the collector is pinned to a fixed watermark and fires whenever the heap approaches it, rather than once per lint.

## Related Links

None.

## Checklist

### DIFF
--- a/internal/lsp/service.go
+++ b/internal/lsp/service.go
@@ -8,6 +8,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"runtime"
 	"slices"
 	"strings"
 	"sync"
@@ -1687,4 +1688,7 @@ func (s *Server) pushDiagnostics(uri lsproto.DocumentUri) {
 	if !lintResult.HasSyntaxErrors {
 		s.dispatchPluginLintWithConfig(uri, generation, rslintConfig, configCwd, isJSConfig)
 	}
+
+	// The pacer cannot see that the lint just dropped what it derived.
+	go runtime.GC()
 }


elecmonkey (commented):
[P1] The collector currently drops collection requests that arrive while `runtime.GC()` is already running. That means a lint pass finishing during the active GC can allocate and then release large temporary structures, have its `collect()` call return early, and then never get an explicit follow-up collection if the user stops editing.

Go treats allocations made during an active GC cycle as live for that cycle, so this can leave the latest lint pass's garbage around until the runtime pacer triggers another GC. Please coalesce requests received while GC is running into a pending collection that runs after the active cycle completes.

swwind:
> [P1] The collector currently drops collection requests that arrive while `runtime.GC()` is already running. That means a lint pass finishing during the active GC can allocate and then release large temporary structures, have its `collect()` call return early, and then never get an explicit follow-up collection if the user stops editing.
> 
> Go treats allocations made during an active GC cycle as live for that cycle, so this can leave the latest lint pass's garbage around until the runtime pacer triggers another GC. Please coalesce requests received while GC is running into a pending collection that runs after the active cycle completes.

That sounds like a valid concern. However, I don't think adding a queue would necessarily improve either performance or correctness here.

The purpose of calling `runtime.GC()` after linting is simply to give the Go GC a hint that we have finished a heavy allocation phase and that these temporary objects can be collected. Since the Go runtime already handles concurrent `runtime.GC()` calls internally, I don't expect a measurable difference between the current `atomic.Bool` approach and a queued approach.

Moreover, I think calling `runtime.GC()` directly without either an `atomic.Bool` or a queue should theoretically behave the same as the two approaches above. The runtime already provides the necessary synchronization and coordination for GC requests, so this PR might only need a few lines of code.

I'll run some benchmarks/tests on my machine to verify this assumption and share the results.

swwind:
Measured it. VS Code 1.128.0 `src` (7464 `.ts`), one rule enabled (`import/no-cycle`), driven by a scripted LSP client over stdio so every variant sees a byte-identical message sequence. `GODEBUG=gctrace=1` counts cycles; `runtime.GC()` ones are tagged `(forced)`, so a dropped request is directly observable. 16 cores, Go 1.26.5.

Four builds, differing only in `lintCollector`: **(a)** this PR, **(b)** no guard — `go runtime.GC()` per lint, **(c)** guard plus a trailing collection, **(d)** `collect()` as a no-op, i.e. main today.

**One document, 10 edits** — 11 lint passes (2 rhythms × 2 runs, range shown):

| | peak RSS | RSS at idle+30s | total CPU | GC cycles | forced |
| --- | --- | --- | --- | --- | --- |
| (a) this PR | 4144–4162 MB | ~4138 MB | 117–121 s | 36–39 | **12** |
| (b) no guard | 4144–4160 MB | ~4137 MB | 117–120 s | 35–38 | **12** |
| (c) trailing | 4147–4161 MB | ~4139 MB | 118–123 s | 37–39 | **12** |
| (d) main today | 6847 MB | 6537 MB | 91.2 s | 27 | 1 |

**8 documents changed inside one debounce window**, so `dispatchLoop` lints them back-to-back — 24 lint passes:

| | peak RSS | RSS at idle+30s | total CPU | GC cycles | forced |
| --- | --- | --- | --- | --- | --- |
| (a) this PR | 4231 MB | 4127 MB | 247.5 s | 63 | **25** |
| (b) no guard | 4230 MB | 4125 MB | 236.3 s | 62 | **25** |
| (c) trailing | 4226 MB | 4128 MB | 239.9 s | 64 | **25** |
| (d) main today | 7150 MB | 6246 MB | 166.9 s | 37 | 1 |

Forced cycles are identical across (a), (b) and (c) — one per lint pass, plus one from upstream's idle cache clean, which already does `go func() { runtime.GC() }()` (`typescript-go/internal/project/session.go:566`). The guard never dropped a request in 35 opportunities: a forced cycle costs ~140 ms wall here, while consecutive `collect()` calls are a median 3.6 s apart, and even the tightest gap in the burst workload (0.74 s) clears it by 4.6×.

@elecmonkey the mechanism is real, but reaching it needs a lint pass to finish inside a ~140 ms window that opens once per lint. Since (b) measures identically and coalesces in the runtime rather than dropping, I'll just remove the guard — that closes this without adding state.

Separately, peak RSS reproduces (−39% / −41% vs main) but total CPU here is +31% / +42%, all of it GC CPU — unlike the flat CPU in the PR description. Probably the single-rule config exposing what the full rule set hides; I'll re-run with all core rules before touching the description.

swwind:
Re-ran with the full rule set (all core rules + all 9 plugins, 997 diagnostics on the first pass). Same corpus and harness as above; **(b)** is what's now on the branch, **(d)** is main.

**One document, 10 edits** — 11 lint passes, 2 runs:

| | peak RSS | RSS at idle+30s | total CPU | GC cycles | forced | GC CPU | live heap |
| --- | --- | --- | --- | --- | --- | --- | --- |
| (b) branch | 4177 / 4175 MB | 4162 / 4162 MB | 125.1 / 127.0 s | 37 / 38 | 12 | 70.4 / 72.2 s | 1973 MB |
| (d) main | 7260 / 7567 MB | 6683 / 4518 MB | 92.5 / 94.3 s | 26 / 25 | 1 | 39.1 / 39.3 s | 1971 MB |

**8 documents in one debounce window** — 24 lint passes:

| | peak RSS | RSS at idle+30s | total CPU | GC cycles | forced | GC CPU | live heap |
| --- | --- | --- | --- | --- | --- | --- | --- |
| (b) branch | 4248 MB | 4153 MB | 260.3 s | 64 | 26 | 154.2 s | 1970 MB |
| (d) main | 7607 MB | 6809 MB | 175.6 s | 36 | 1 | 74.2 s | 1969 MB |

Rule count is not the explanation. Across all four combinations:

| | peak RSS | total CPU | GC CPU | CPU per extra collection |
| --- | --- | --- | --- | --- |
| one document, full rules | −44% | +35% | 39.2 → 71.3 s | ~2.7 s |
| burst, full rules | −44% | +48% | 74.2 → 154.2 s | ~3.2 s |
| one document, one rule | −39% | +30% | 39.2 → 66.6 s | ~2.3 s |
| burst, one rule | −41% | +42% | 68.7 → 135.4 s | ~2.7 s |

The total-CPU delta matches the GC-CPU delta almost exactly in every run (32.7 vs 32.1 s; 84.7 vs 80.0 s), so the mechanism is just one extra full mark per lint pass — 2.3–3.2 s of CPU at ~2 GB live here, and that cost scales with the live heap rather than with the rule set. Enabling every rule barely moved total CPU on its own (45.9 s vs 47 s over three passes), because the time goes into building and checking the Program, not into running rules.

Live heap is 1969–1974 MB in every run of every variant, which does confirm the framing in the description: this is garbage, not retention.

So the peak-RSS win reproduces (−44%), but "total CPU is indistinguishable from today's" does not, under either config. CPU here is utime+stime of the server process over the whole run.

Two caveats before reading too much into that. This was benched on my personal laptop (AMD); the table in the description was benched on my work laptop (Intel), and I haven't ruled out a machine difference. And the description's numbers came from real interaction through the VS Code extension, whereas this harness drives the raw LSP API directly. So the key reason for the divergence isn't established yet — I'll test more tomorrow.

swwind:
Ran the same harness on the Intel work laptop. Same corpus and scripts as above. Intel Core Ultra 7 155H (16 cores / 22 threads), 30 GB, Go 1.26.5. `(b)` is what's on the branch, `(d)` is its merge-base with the diff reverted. 12 runs, sequential, variants alternated.

**One document, 10 edits** — 11 lint passes, 2 runs, full rules:

| | peak RSS | RSS at idle+30s | total CPU | GC cycles | forced | GC CPU | live heap |
| --- | --- | --- | --- | --- | --- | --- | --- |
| (b) branch | 4237 / 4206 MB | 4173 / 4171 MB | 79.3 / 78.3 s | 37 / 37 | 12 | 31.3 / 30.4 s | 1973 / 1970 MB |
| (d) main | 7228 / 7300 MB | 5281 / 6377 MB | 62.5 / 65.3 s | 26 / 26 | 1 | 17.1 / 19.1 s | 1973 / 1972 MB |

**8 documents in one debounce window** — 24 lint passes, full rules:

| | peak RSS | RSS at idle+30s | total CPU | GC cycles | forced | GC CPU | live heap |
| --- | --- | --- | --- | --- | --- | --- | --- |
| (b) branch | 4374 MB | 2421 MB | 174.1 s | 64 | 25 | 72.1 s | 1970 MB |
| (d) main | 7054 MB | 4818 MB | 131.4 s | 39 | 1 | 38.0 s | 1969 MB |

Against the AMD numbers, same harness on both:

| | peak RSS (AMD → Intel) | total CPU (AMD → Intel) | CPU per extra collection |
| --- | --- | --- | --- |
| one document, full rules | −44% → −42% | +35% → +23% | 2.7 → 1.4 s |
| burst, full rules | −44% → −38% | +48% → +32% | 3.2 → 1.8 s |
| one document, one rule | −39% → −41% | +30% → +25% | 2.3 → 1.4 s |
| burst, one rule | −41% → −43% | +42% → +35% | 2.7 → 1.8 s |

A forced cycle costs 1.4–1.8 s CPU here against 2.3–3.2 s on AMD, at the same ~1.97 GB live. In wall time it's a median 267 ms one-document and 352 ms burst, against ~140 ms on AMD.

Live heap is 1957–1973 MB across all 12 runs, against 1969–1974 MB on AMD. `(d)` peak RSS is 7.0–7.6 GB here and 6.8–7.6 GB on AMD. The description reports 4.3 GB live and 11.2–11.7 GB peak.

Forced cycles equal lint passes + 1 in all 12 runs — 12 for the 11-pass workload, 25 for the 24-pass burst. The tightest gap between consecutive `collect()` calls in the burst workload is 1.4 s, against a 433 ms worst-case cycle.

fansenze (approved):

PR #1677, opened 2026-08-10
refactor: parse rule options by type assertion instead of JSON round-trips

swwind:
## Summary

Replaces the `json.Marshal` / `json.Unmarshal` round-trip used to decode rule options with direct type assertions on `options[0]`, matching the option-parsing style documented in the port-rule skill.

Rules converted: `no-base-to-string`, `no-confusing-void-expression`, `no-deprecated`, `no-duplicate-type-constituents`, `no-floating-promises`, `no-meaningless-void-operator`, `no-misused-promises`, `no-misused-spread`, `no-unnecessary-boolean-literal-compare`, `no-unnecessary-type-assertion`, `only-throw-error`, `prefer-optional-chain`, `prefer-promise-reject-errors`, `promise-function-async`, `require-array-sort-compare`, `restrict-template-expressions`, `strict-boolean-expressions`, `strict-void-return`.

Along with it:

- Option structs no longer use pointer-typed fields: defaults live in the struct initializer, so `*bool` / `*string` become plain `bool` / `string`. This also covers `restrict-plus-operands`, `unbound-method` and `prefer-string-starts-ends-with`, and removes the now-unused `utils.Ref` helper and `prefer-optional-chain`'s `derefBoolDefault`.
- `json:"..."` tags are dropped from the touched option structs, since nothing decodes them anymore.
- New `utils.ParseTypeOrValueSpecifier` / `utils.ParseTypeOrValueSpecifiers` decode type specifiers (string shorthand and object form) straight from config values; the `UnmarshalJSON` implementations on `TypeOrValueSpecifier`, `NameList` and `TypeOrValueSpecifierFrom` are removed.
- `no-misused-promises` parses `checksVoidReturn`'s `boolean | object` shape with a type switch instead of a custom `UnmarshalJSON`.
- `strict-boolean-expressions` drops its intermediate pointer-typed `Options` struct and parses straight into `resolvedOptions`.

## Related Links

- #1673 — the option-parsing guidance this PR follows
- #1667 — removed the legacy options compatibility layer this builds on

## Checklist

### DIFF
--- a/internal/plugins/import/rules/no_cycle/no_cycle.go
+++ b/internal/plugins/import/rules/no_cycle/no_cycle.go
@@ -2,6 +2,7 @@ package no_cycle
 
 import (
 	_ "embed"
+	"encoding/json"
 	"fmt"
 	"math"
 	"strings"
@@ -101,6 +102,13 @@ func parseMaxDepth(raw any) (int, bool) {
 			return 0, false
 		}
 		return normalizeDepth(int(value))
+	case json.Number:
+		if i, err := value.Int64(); err == nil {
+			return normalizeDepth(int(i))
+		}
+		if f, err := value.Float64(); err == nil {
+			return normalizeDepth(int(f))
+		}
 	case string:
 		if value == "∞" {
 			return unlimitedDepth, true
--- a/internal/plugins/import/rules/no_cycle/no_cycle.go
+++ b/internal/plugins/import/rules/no_cycle/no_cycle.go
@@ -2,7 +2,6 @@ package no_cycle
 
 import (
 	_ "embed"
-	"encoding/json"
 	"fmt"
 	"math"
 	"strings"
@@ -102,13 +101,6 @@ func parseMaxDepth(raw any) (int, bool) {
 			return 0, false
 		}
 		return normalizeDepth(int(value))
-	case json.Number:
-		if i, err := value.Int64(); err == nil {
-			return normalizeDepth(int(i))
-		}
-		if f, err := value.Float64(); err == nil {
-			return normalizeDepth(int(f))
-		}
 	case string:
 		if value == "∞" {
 			return unlimitedDepth, true
--- a/internal/utils/utils.go
+++ b/internal/utils/utils.go
@@ -214,11 +214,6 @@ func TypeRecurser(t *checker.Type, predicate func(t *checker.Type) /* should sto
 	}
 }
 
-// SUPER DIRTY HACK FOR OPTIONAL FIELDS :(
-func Ref[T any](a T) *T {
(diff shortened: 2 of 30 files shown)


internal/plugins/import/rules/no_cycle/no_cycle.go
```diff
@@ -102,13 +101,6 @@ func parseMaxDepth(raw any) (int, bool) {
 			return 0, false
 		}
 		return normalizeDepth(int(value))
-	case json.Number:
```
elecmonkey, line 105: Please retain `json.Number` handling for `maxDepth`. A Go caller can construct an `RslintConfig` with `maxDepth: json.Number("2")`; schema validation accepts this as an integer, and `ValidateRuleOptions` preserves the existing numeric type. Without this case, `parseMaxDepth` silently falls back to unlimited depth, changing traversal and reporting behavior for a schema-valid configuration.

internal/utils/utils.go
```diff
@@ -214,11 +214,6 @@ func TypeRecurser(t *checker.Type, predicate func(t *checker.Type) /* should sto
 	}
 }
 
-// SUPER DIRTY HACK FOR OPTIONAL FIELDS :(
```
elecmonkey, line 217: `utils.Ref` has been removed, but this porting guide still recommends calling it. Please update or remove this example so contributors are not directed to use an API that no longer exists.

internal/plugins/import/rules/no_cycle/no_cycle.go
swwind: Confirmed and restored in 900b60dc.

I checked the claim before reverting: passing `maxDepth: json.Number("2")` through `ValidateRuleOptions` passes schema validation and comes out the other side still typed `json.Number` — `normalizeNumbers` only runs on schema-supplied defaults, not on user-supplied values. So without the branch `parseMaxDepth` returned `(0, false)` and traversal silently became unlimited.

Added a regression test in `no_cycle_extras_test.go`: a valid case using `json.Number("2")` on a depth-three import. With the branch removed it fails with `Dependency cycle via ./depth-two:1=>./depth-one:1`; with it restored the package passes.

internal/utils/utils.go
swwind: Right — fixed in 900b60dc. Removed the `utils.Ref` example from the "Other Utilities" section of `.agents/skills/port-rule/references/UTILS_REFERENCE.md`. Checked the rest of the guide for references to the other removed APIs (`derefBoolDefault`, the `UnmarshalJSON` implementations); there are none.

fansenze (approved):

PR #1693, opened 2026-08-11
docs(typescript): record each rule's aligned upstream version

swwind:
## Summary

- Every `typescript/` rule doc's "Original Documentation" section now pins a "Source code" link to the newest typescript-eslint release the rule's Go port is verified against, alongside the existing (unpinnable) typescript-eslint.io doc link.
- All 131 rules audited: 98 bump to the latest release, v8.67.0 (56 had zero upstream commits since port; 42 more had only cosmetic drift, or behavioral drift rslint's Go port already independently implements, confirmed against existing Go tests).
- The remaining 33 rules stay pinned at the tag they were last actually verified against, because upstream has real behavioral changes rslint hasn't ported yet:
  - `array-type` (v8.39.0): bare `Array`/`ReadonlyArray` no longer autofixed to `any[]`, and shadowed `Array`/`ReadonlyArray` no longer flagged.
  - `await-thenable` (v8.28.0): new check for non-awaitable values passed to `Promise.all`/`race`/`allSettled`/`any`.
  - `ban-ts-comment` (v8.46.3): multi-line directive matching now handles `\r`-only and Unicode line terminators.
  - `class-literal-property-style` (v8.39.0): field-style suggestion now preserves type annotations and is withheld on decorated getters.
  - `consistent-generic-constructors` (v8.46.3): typed-array constructors (`Float32Array`, etc.) referencing the global builtin are now exempt.
  - `consistent-type-exports` (v8.46.3): alias-chain resolution now checks the value flag before continuing, fixing misclassified re-exports.
  - `method-signature-style` (v8.58.2): readonly function-typed properties now get a suggestion instead of an autofix that emitted invalid syntax.
  - `no-base-to-string` (v8.28.0): `checkUnknown` option, `@@toPrimitive`/`valueOf` detection, base-type traversal and generic-name matching for `ignoredTypeNames`, and a multi-declaration overload fix are all unported.
  - `no-extraneous-class` (v8.46.3): `TSIndexSignature` handling — likely already correct via the Go port's default-case fallback, but no test demonstrates it, so it isn't claimed as verified.
  - `no-loop-func` (v8.59.1): deliberately does not treat `using`/`await using` as constant bindings (rslint's own test suite locks in the old behavior).
  - `no-misused-promises` (v8.28.0): async-dispose detection for `using`/`await using` declarations is unported (other behavioral fixes in this window are already ported and tested).
  - `no-shadow` (v8.59.1): new `noEnumShadow` messageId for enum-member shadowing, and a stricter direct-initializer check, are both unported (rslint's own test suite locks in the old behavior).
  - `no-unnecessary-boolean-literal-compare` (v8.29.0): autofix still uses the pre-fix algorithm with known precedence bugs.
  - `no-unnecessary-condition` (v8.58.1): `isPossiblyNullish` still excludes `void` by explicit design comment.
  - `no-unnecessary-parameter-property-assignment` (v8.59.3): deliberately mirrors the upstream bug where `this[ident] = ident` was misclassified (explicit comment in the Go source).
  - `no-unnecessary-template-expression` (v8.29.1): newline detection still only checks `\n`/`\r\n`, not the fuller upstream line-break matcher.
  - `no-unnecessary-type-arguments` (v8.29.0): skipping `TSInstantiationExpression` parents is unported.
  - `no-unnecessary-type-assertion` (v8.28.0): still uses the old equality-based check; the assignability-based rewrite (723-line upstream change) is unported, marked by the Go source's own TODO.
  - `no-unnecessary-type-conversion` (v8.59.3): shadow detection still uses the old non-scope-chain-walking check (rslint's own test suite locks in the old behavior).
  - `no-unsafe-assignment` / `no-unsafe-call` / `no-unsafe-member-access` (all v8.28.0): upstream's error-vs-any messageId split and new `unsafeObjectPattern`/`allowOptionalChaining` cases are unported.
  - `no-unsafe-enum-comparison` (v8.28.0): union-of-literals support is unported.
  - `no-unsafe-type-assertion` (v8.29.0): missing the try/catch guard upstream added around a TS-compiler crash on recursive template literal types; rslint's rule pipeline has no panic recovery either.
  - `no-useless-default-assignment` (v8.59.4): overload check still only inspects the first signature instead of scanning all contextual overloads.
  - `prefer-promise-reject-errors` (v8.28.0): `allow` option is declared in the schema but never read.
  - `prefer-readonly` (v8.54.0): computed keys are still unconditionally ignored instead of analyzing statically-known ones.
  - `prefer-readonly-parameter-types` (v8.46.3): type resolution still uses `getTypeAtLocation` instead of the alias-preserving fix, and `allow` only supports bare strings.
  - `prefer-return-this-type` (v8.29.0): still uses strict type equality instead of checking union membership.
  - `prefer-string-starts-ends-with` (v8.55.0): end-anchor detection still a naive suffix check, not the unescaped-`$` fix.
  - `prefer-ts-expect-error` (v8.57.0): same line-terminator gap as `ban-ts-comment`/`no-unnecessary-template-expression`.
  - `promise-function-async` (v8.29.1): missing messageId for hybrid promise/non-promise union return types.
  - `unbound-method` (v8.28.0): union-typed member-expression access isn't checked, and the message text predates a wording fix.

## Checklist

- [x] Tests updated (or not required) — docs-only change.

### DIFF
--- a/internal/plugins/typescript/rules/adjacent_overload_signatures/adjacent_overload_signatures.md
+++ b/internal/plugins/typescript/rules/adjacent_overload_signatures/adjacent_overload_signatures.md
@@ -34,4 +34,5 @@ class MyClass {
 
 ## Original Documentation
 
-- [typescript-eslint adjacent-overload-signatures](https://typescript-eslint.io/rules/adjacent-overload-signatures)
+- [typescript-eslint: adjacent-overload-signatures](https://typescript-eslint.io/rules/adjacent-overload-signatures)
+- [Source code](https://github.com/typescript-eslint/typescript-eslint/blob/v8.67.0/packages/eslint-plugin/src/rules/adjacent-overload-signatures.ts)
--- a/internal/plugins/typescript/rules/array_type/array_type.md
+++ b/internal/plugins/typescript/rules/array_type/array_type.md
@@ -24,4 +24,5 @@ const c: (string | number)[] = [];
 
 ## Original Documentation
 
-- [typescript-eslint array-type](https://typescript-eslint.io/rules/array-type)
+- [typescript-eslint: array-type](https://typescript-eslint.io/rules/array-type)
+- [Source code](https://github.com/typescript-eslint/typescript-eslint/blob/v8.39.0/packages/eslint-plugin/src/rules/array-type.ts)
--- a/internal/plugins/typescript/rules/await_thenable/await_thenable.md
+++ b/internal/plugins/typescript/rules/await_thenable/await_thenable.md
@@ -42,4 +42,5 @@ async function baz(iter: AsyncIterable<number>) {
 
 ## Original Documentation
 
-- [typescript-eslint await-thenable](https://typescript-eslint.io/rules/await-thenable)
+- [typescript-eslint: await-thenable](https://typescript-eslint.io/rules/await-thenable)
+- [Source code](https://github.com/typescript-eslint/typescript-eslint/blob/v8.28.0/packages/eslint-plugin/src/rules/await-thenable.ts)
--- a/internal/plugins/typescript/rules/ban_ts_comment/ban_ts_comment.md
+++ b/internal/plugins/typescript/rules/ban_ts_comment/ban_ts_comment.md
@@ -31,4 +31,5 @@ const x: number = 'hello';
 
 ## Original Documentation
 
-- [typescript-eslint ban-ts-comment](https://typescript-eslint.io/rules/ban-ts-comment)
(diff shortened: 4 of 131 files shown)


swwind:
Note that `switch-exhaustiveness-check` is a no-op stub — `switch_exhaustiveness_check.go`'s `KindSwitchStatement` handler is entirely commented-out TODO pseudocode, with no actual exhaustiveness logic, and the test file has zero invalid cases. We will change this later.

fansenze (approved):

PR #1695, opened 2026-08-12
feat(rstest): add valid-expect rule

elecmonkey:
## Summary

Port `eslint-plugin-jest`'s `valid-expect` to rstest: validate the shape of `expect(...)` assertions — argument count, that a matcher is present and actually called, and that async assertions (`resolves` / `rejects` and configured async matchers) are awaited or returned. It consumes the shared rstest call analysis (`GetRstestCallAnalysis` / `ParseExpectCall`) and reads broken-chain reasons from the parser rather than re-deriving them; added to the rstest `recommended` preset at `error`.

## Behavior change in `jest/valid-expect`

An assertion wrapped in parentheses is no longer reported. Parentheses are nodes in the TypeScript AST but not in ESTree, so the `await` or `return` that upstream sees as the assertion's direct parent sat one wrapper away here:

```js
test('t', async () => { await (expect(Promise.resolve(2)).resolves.toBeDefined()); });
test('t', () => { return (expect(Promise.resolve(2)).resolves.toBeDefined()); });
```

Both are valid under `eslint-plugin-jest`, while this rule reported `asyncMustBeAwaited` and its fix rewrote working code. An assertion nothing awaits, `(expect(p).resolves.toBe(1));`, still reports, and the fix goes inside the parentheses.

Parenthesized conditional expressions are handled consistently as well, so `return (condition ? expect(p).resolves.toBe(1) : expect(q).toBe(1))` is accepted just like the unparenthesized form.

The machinery both rules share — Promise-chain walking, the await/return check, the `Promise.all([...])` de-duplication and the async/await fixes — now lives in `internal/utils/test_framework/rules/valid_expect` rather than being copied into each rule, which is how the same false positive came to exist twice. The rule bodies stay separate: broken-chain reasons, argument allowances and matcher shapes genuinely differ between the two frameworks.

## Differences from jest

rstest's `expect` is vitest/chai lineage, so the rule follows vitest rather than jest where the framework differs:

- **a second `expect` argument is allowed** when it is a message string or template literal (`expect(actual, "msg")`), and `expect.poll(fn, options)` / `expect.element(el, options)` accept an options object. Other second arguments still report `tooManyArgs`. jest's strict `maxArgs = 1` would false-positive here.
- **Chai property matchers are valid without a call** (`expect(x).to.be.ok`); these are not treated as `matcherNotCalled`.
- **forms without an assertion factory are not subjected to argument or await checks**. This includes static members (`expect.assertions(1)`, `expect.hasAssertions()`, asymmetric matchers) and bare modifier or static chains such as `expect.resolves.toBe(1)` and `expect.toResolve()`; the rule does not insert `await` into code whose assertion factory it cannot validate.
- expect provenance covers the full rstest surface (globals, `@rstest/core` imports/aliases, `require`, namespace, `import.meta.rstest`, TestContext `expect`, `@rstest/playwright`).

No `expectTypeOf` / type-assertion handling, since that is out of scope for this rule.

## Related Links

Related #935

## Checklist

### DIFF
--- a/internal/plugins/rstest/rules/valid_expect/valid_expect.go
+++ b/internal/plugins/rstest/rules/valid_expect/valid_expect.go
@@ -294,23 +294,25 @@ var ValidExpectRule = rule.Rule{
 
 				// Static members (expect.assertions(1), expect.any(...)) carry no
 				// assertion factory, so argument and await checks do not apply.
-				if parsed.Head != nil && parsed.Head.Kind == ast.KindCallExpression {
-					args := parsed.Head.AsCallExpression().Arguments.Nodes
-					if len(args) < opts.MinArgs {
+				if parsed.Head == nil || parsed.Head.Kind != ast.KindCallExpression {
+					return
+				}
+
+				args := parsed.Head.AsCallExpression().Arguments.Nodes
+				if len(args) < opts.MinArgs {
+					ctx.ReportRange(
+						expectFactoryOpenParenRange(ctx.SourceFile, parsed.Head),
+						buildErrorNotEnoughArgsMessage(opts.MinArgs),
+					)
+				}
+				if len(args) > opts.MaxArgs {
+					// vitest allowance: a lone message string/template, or the
+					// options object of poll/element, is not an excess argument.
+					if len(args) != opts.MaxArgs+1 || !isAllowedExtraExpectArg(args[opts.MaxArgs], parsed.Entry) {
 						ctx.ReportRange(
-							expectFactoryOpenParenRange(ctx.SourceFile, parsed.Head),
-							buildErrorNotEnoughArgsMessage(opts.MinArgs),
+							tooManyArgsRange(ctx.SourceFile, args, opts.MaxArgs),
+							buildErrorTooManyArgsMessage(opts.MaxArgs),
 						)
-					}
-					if len(args) > opts.MaxArgs {
-						// vitest allowance: a lone message string/template, or the
-						// options object of poll/element, is not an excess argument.
-						if len(args) != opts.MaxArgs+1 || !isAllowedExtraExpectArg(args[opts.MaxArgs], parsed.Entry) {
-							ctx.ReportRange(
-								tooManyArgsRange(ctx.SourceFile, args, opts.MaxArgs),
-								buildErrorTooManyArgsMessage(opts.MaxArgs),
-							)
-						}
 					}
 				}
 
--- a/internal/plugins/rstest/rules/valid_expect/valid_expect.go
+++ b/internal/plugins/rstest/rules/valid_expect/valid_expect.go
@@ -176,7 +176,7 @@ func isPromiseMethodCall(node *ast.Node) bool {
 	}
 
 	callee := ast.SkipParentheses(node.AsCallExpression().Expression)
-	if !isMemberAccessNode(callee) {
+	if !rstestUtils.IsMemberAccessNode(callee) {
 		return false
 	}
 
@@ -215,7 +215,7 @@ func getParentIfPromiseChained(node *ast.Node) *ast.Node {
 	}
 
 	grandParent := node.Parent.Parent
-	if grandParent.Kind != ast.KindCallExpression || !isMemberAccessNode(grandParent.AsCallExpression().Expression) {
+	if grandParent.Kind != ast.KindCallExpression || !rstestUtils.IsMemberAccessNode(grandParent.AsCallExpression().Expression) {
 		return node
 	}
 
@@ -298,20 +298,6 @@ func reportAsyncDescriptor(
 		return
 	}
 	ctx.ReportNode(descriptor.node, msg)
-}
-
-// --- Rstest-specific helpers ---
-
-func isMemberAccessNode(node *ast.Node) bool {
-	if node == nil {
-		return false
-	}
-	switch node.Kind {
-	case ast.KindPropertyAccessExpression, ast.KindElementAccessExpression:
-		return true
-	default:
-		return false
-	}
 }
 
 // expectFactoryOpenParenRange locates the `(` of the assertion factory call so
--- a/internal/plugins/rstest/rules/valid_expect/valid_expect.go
+++ b/internal/plugins/rstest/rules/valid_expect/valid_expect.go
@@ -10,7 +10,7 @@ import (
 	rstestUtils "github.com/web-infra-dev/rslint/internal/plugins/rstest/utils"
 	"github.com/web-infra-dev/rslint/internal/rule"
 	internalUtils "github.com/web-infra-dev/rslint/internal/utils"
-	testFramework "github.com/web-infra-dev/rslint/internal/utils/test_framework"
+	sharedValidExpect "github.com/web-infra-dev/rslint/internal/utils/test_framework/rules/valid_expect"
 )
 
 //go:embed valid_expect.schema.json
@@ -164,117 +164,9 @@ func readIntOption(options map[string]interface{}, key string, defaultValue int)
 // --- Async assertion machinery ---
 //
 // The Promise-chain walking, acceptable-return detection, array de-duplication
-// and async/await fixers below are ported verbatim from
-// internal/plugins/jest/rules/valid_expect/valid_expect.go, which is itself
-// aligned with eslint-plugin-vitest. Kept as a local copy rather than shared:
-// sharing them would require touching the jest rule and re-running its
-// regression, and there is no third consumer yet.
-
-func isPromiseMethodCall(node *ast.Node) bool {
-	if node == nil || node.Kind != ast.KindCallExpression {
-		return false
-	}
-
-	callee := ast.SkipParentheses(node.AsCallExpression().Expression)
-	if !rstestUtils.IsMemberAccessNode(callee) {
-		return false
-	}
-
-	return testFramework.CalleeChainName(internalUtils.AccessExpressionObject(callee)) == "Promise"
-}
-
-func getPromiseCallExpressionNode(node *ast.Node) *ast.Node {
-	if node == nil {
-		return nil
-	}
-
-	if node.Kind == ast.KindArrayLiteralExpression && node.Parent != nil && node.Parent.Kind == ast.KindCallExpression {
-		node = node.Parent
-	}
-
-	if isPromiseMethodCall(node) {
-		return node
-	}
-
-	return nil
-}
-
-func findPromiseCallExpressionNode(node *ast.Node) *ast.Node {
-	if node == nil || node.Parent == nil || node.Parent.Parent == nil {
-		return nil
-	}
-	if node.Parent.Kind != ast.KindCallExpression && node.Parent.Kind != ast.KindArrayLiteralExpression {
-		return nil
-	}
-	return getPromiseCallExpressionNode(node.Parent)
-}
-
-func getParentIfPromiseChained(node *ast.Node) *ast.Node {
-	if node == nil || node.Parent == nil || node.Parent.Parent == nil {
-		return node
-	}
-
-	grandParent := node.Parent.Parent
-	if grandParent.Kind != ast.KindCallExpression || !rstestUtils.IsMemberAccessNode(grandParent.AsCallExpression().Expression) {
-		return node
-	}
-
-	member := grandParent.AsCallExpression().Expression
-	entries := testFramework.GetMemberEntries(member)
-	if len(entries) == 0 {
-		return node
-	}
-
-	last := entries[len(entries)-1].Name
-	if last == "then" || last == "catch" {
-		return getParentIfPromiseChained(grandParent)
-	}
-
-	return node
-}
-
-func isAcceptableReturnNode(node *ast.Node, allowReturn bool) bool {
-	if node == nil {
-		return false
-	}
-
-	if allowReturn && node.Kind == ast.KindReturnStatement {
-		return true
-	}
-	if node.Kind == ast.KindConditionalExpression {
-		return isAcceptableReturnNode(node.Parent, allowReturn)
-	}
-
-	return node.Kind == ast.KindArrowFunction || node.Kind == ast.KindAwaitExpression
-}
-
-func promiseArrayExceptionKey(sourceFile *ast.SourceFile, node *ast.Node) string {
-	if sourceFile == nil || node == nil {
-		return ""
-	}
-	r := internalUtils.TrimNodeTextRange(sourceFile, node)
-	return fmt.Sprintf("%d:%d", r.Pos(), r.End())
-}
-
-func asyncInsertFix(sourceFile *ast.SourceFile, fn *ast.Node) rule.RuleFix {
-	switch fn.Kind {
-	case ast.KindMethodDeclaration, ast.KindConstructor, ast.KindGetAccessor, ast.KindSetAccessor:
-		head := internalUtils.GetFunctionHeadLoc(sourceFile, fn)
-		return rule.RuleFixReplaceRange(core.NewTextRange(head.Pos(), head.Pos()), "async ")
-	default:
-		return rule.RuleFixInsertBefore(sourceFile, fn, "async ")
-	}
-}
-
-func awaitFix(sourceFile *ast.SourceFile, node *ast.Node, alwaysAwait bool) rule.RuleFix {
-	if alwaysAwait && node.Parent != nil && node.Parent.Kind == ast.KindReturnStatement {
-		ret := node.Parent
-		retRange := internalUtils.TrimNodeTextRange(sourceFile, ret)
-		nodeRange := internalUtils.TrimNodeTextRange(sourceFile, node)
-		return rule.RuleFixReplaceRange(core.NewTextRange(retRange.Pos(), nodeRange.Pos()), "await ")
-	}
-	return rule.RuleFixInsertBefore(sourceFile, node, "await ")
-}
+// and async/await fixers live in the shared package: they are pure syntax, with
+// no jest or rstest semantics, and keeping two copies is what let the same
+// parenthesized-assertion false positive exist twice.
 
 func reportAsyncDescriptor(
 	ctx rule.RuleContext,
@@ -287,10 +179,10 @@ func reportAsyncDescriptor(
 	var fixes []rule.RuleFix
 	if fn := ast.GetContainingFunction(descriptor.node); fn != nil {
 		if !ast.IsAsyncFunction(fn) && !asyncInserted[fn] {
-			fixes = append(fixes, asyncInsertFix(ctx.SourceFile, fn))
+			fixes = append(fixes, sharedValidExpect.AsyncInsertFix(ctx.SourceFile, fn))
 			asyncInserted[fn] = true
 		}
-		fixes = append(fixes, awaitFix(ctx.SourceFile, descriptor.node, alwaysAwait))
+		fixes = append(fixes, sharedValidExpect.AwaitFix(ctx.SourceFile, descriptor.node, alwaysAwait))
 	}
 
 	if len(fixes) > 0 {
@@ -350,37 +242,6 @@ func isAllowedExtraExpectArg(arg *ast.Node, entry rstestUtils.RstestExpectEntry)
 
 func shouldBeAwaited(parsed *rstestUtils.ParsedRstestExpectCall, asyncMatchers []string) bool {
 	return rstestUtils.ShouldRstestExpectBeAwaited(parsed, asyncMatchers)
-}
-
-// resolveAsyncAssertionReportNode mirrors the jest rule: from the complete
-// assertion expression it walks out through chained .then/.catch, notes whether
-// the assertion sits inside a Promise.all([...]) array, and decides whether the
-// resulting node must be awaited or returned. assertionNode must be the
-// outermost expression of the chain — ParsedRstestExpectCall.Expression — so
-// that a Chai chain carrying several matchers, such as
-// expect(p).resolves.to.be.a("string").that.contains("x"), is inspected as a
-// whole rather than at its first matcher.
-func resolveAsyncAssertionReportNode(
-	assertionNode *ast.Node,
-	alwaysAwait bool,
-) (reportNode *ast.Node, promiseWrapped bool, insideAssertionArray bool, shouldReport bool) {
-	if assertionNode == nil {
-		return nil, false, false, false
-	}
-
-	promiseChainedAssertionNode := getParentIfPromiseChained(assertionNode)
-	insideAssertionArray = promiseChainedAssertionNode.Parent != nil && promiseChainedAssertionNode.Parent.Kind == ast.KindArrayLiteralExpression
-	reportNode = promiseChainedAssertionNode
-	if promiseCallNode := findPromiseCallExpressionNode(promiseChainedAssertionNode); promiseCallNode != nil {
-		reportNode = promiseCallNode
-		promiseWrapped = true
-	}
-
-	if reportNode.Parent == nil || isAcceptableReturnNode(reportNode.Parent, !alwaysAwait) {
-		return reportNode, promiseWrapped, insideAssertionArray, false
-	}
-
-	return reportNode, promiseWrapped, insideAssertionArray, true
 }
 
 // reportBrokenChain reports a chain that the parser resolved without a matcher.
@@ -457,7 +318,7 @@ var ValidExpectRule = rule.Rule{
 					return
 				}
 
-				reportNode, promiseWrapped, insideAssertionArray, shouldReport := resolveAsyncAssertionReportNode(
+				reportNode, promiseWrapped, insideAssertionArray, shouldReport := sharedValidExpect.ResolveAsyncAssertionReportNode(
 					parsed.Expression,
 					opts.AlwaysAwait,
 				)
@@ -465,7 +326,7 @@ var ValidExpectRule = rule.Rule{
 					return
 				}
 
-				reportNodeKey := promiseArrayExceptionKey(ctx.SourceFile, reportNode)
+				reportNodeKey := sharedValidExpect.PromiseArrayExceptionKey(ctx.SourceFile, reportNode)
 				if shouldReport && !arrayExceptions[reportNodeKey] {
 					descriptors = append(descriptors, asyncDescriptor{
 						node:           reportNode,
--- a/internal/plugins/rstest/rules/valid_expect/valid_expect.go
+++ b/internal/plugins/rstest/rules/valid_expect/valid_expect.go
@@ -353,30 +353,18 @@ func shouldBeAwaited(parsed *rstestUtils.ParsedRstestExpectCall, asyncMatchers [
 }
 
 // resolveAsyncAssertionReportNode mirrors the jest rule: from the complete
-// matcher expression it walks out through chained .then/.catch, notes whether
+// assertion expression it walks out through chained .then/.catch, notes whether
 // the assertion sits inside a Promise.all([...]) array, and decides whether the
-// resulting node must be awaited or returned. A call-style matcher's complete
-// expression is its parent CallExpression, while a Chai property matcher is
-// already complete at the member-access expression.
+// resulting node must be awaited or returned. assertionNode must be the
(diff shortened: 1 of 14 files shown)


internal/plugins/rstest/rules/valid_expect/valid_expect.go
```diff
@@ -0,0 +1,503 @@
...
+
+func shouldBeAwaited(parsed *rstestUtils.ParsedRstestExpectCall, asyncMatchers []string) bool {
+	return rstestUtils.ShouldRstestExpectBeAwaited(parsed, asyncMatchers)
+}
+
+// resolveAsyncAssertionReportNode mirrors the jest rule: from the matcher
+// member it walks out through chained .then/.catch, notes whether the assertion
+// sits inside a Promise.all([...]) array, and decides whether the resulting node
+// must be awaited or returned.
+func resolveAsyncAssertionReportNode(
```
swwind: For a Chai property-style matcher (no trailing call, e.g. `.resolves.to.be.true`), `matcherMemberNode` is already the whole assertion expression — there's no enclosing CallExpression. But this passes `matcherMemberNode.Parent` into `getParentIfPromiseChained`, which skips one level too far and inspects whatever contains the assertion (e.g. the enclosing `ReturnStatement`) instead of the assertion itself.

Repro:
```
test("t", () => { return expect(p).resolves.to.be.true; });
// false positive: "Async assertions must be awaited or returned"
```

This works correctly for call-style matchers (`.toBe(1)`) because `matcherMemberNode.Parent` there actually is the CallExpression — the logic was ported verbatim from jest, which never has uninvoked property matchers, so this gap didn't exist there. Since rstest's parser explicitly supports `RstestExpectMatcherProperty`, this function needs to branch on whether the matcher was called (use `matcherMemberNode` itself as the assertion node when it wasn't).

internal/plugins/rstest/rules/valid_expect/valid_expect.go
```diff
@@ -0,0 +1,503 @@
...
+	if len(fixes) > 0 {
+		ctx.ReportNodeWithFixes(descriptor.node, msg, fixes...)
+		return
+	}
+	ctx.ReportNode(descriptor.node, msg)
+}
+
+// --- Rstest-specific helpers ---
+
+func isMemberAccessNode(node *ast.Node) bool {
```
swwind: Duplicates the identical unexported `isMemberAccessNode` already in `rstestUtils` (`parse_rstest_expect.go:679`). Consider exporting that one and reusing it here, the way the jest plugin's `valid-expect` reuses `utils.IsMemberAccessNode`.

swwind (commented):
| Code | Expected | This PR |
|---|---|---|
| `test("t", () => { return expect(p).resolves.to.be.true; });` | Valid (assertion is returned) | ❌ reports `asyncMustBeAwaited` |

Found one correctness bug in the async-report-node resolution; see inline comment. Everything else (arg validation, chai property-matcher parsing, the rest of the async/await machinery) looks correct against the vitest/chai semantics this rule targets.

internal/plugins/rstest/rules/valid_expect/valid_expect.go
elecmonkey: Fixed in 751fa0a1. Property matchers now use the correct assertion node, with regression tests added. Thanks!

internal/plugins/rstest/rules/valid_expect/valid_expect.go
```diff
@@ -0,0 +1,500 @@
...
+						}
+					}
+				}
+
+				if len(parsed.Matchers) == 0 || !shouldBeAwaited(parsed, opts.AsyncMatchers) {
+					return
+				}
+
+				reportNode, promiseWrapped, insideAssertionArray, shouldReport := resolveAsyncAssertionReportNode(
+					parsed.Matchers[0],
```
swwind: Only the first matcher is used to locate the assertion node, but the parser deliberately returns every matcher in the chain — `findRstestExpectModifiersAndMatchers` collects from `matcherIndex` to the end, and its comment even cites `.a("string").that.does.not.contain("x")`. For a multi-matcher chain the first matcher's expression is not the outermost one, so the await/return check inspects the wrong parent:

```js
test("t", async () => {
  await expect(p).resolves.to.be.a("string").that.contains("x");   // false positive: asyncMustBeAwaited
  await expect(p).resolves.to.be.an("object").that.is.ok;          // false positive
});
test("t", () => {
  return expect(p).resolves.to.be.a("string").that.contains("x");  // false positive
});
```

The reported range is `expect(p).resolves.to.be.a("string")` — the first matcher's call — whose parent is the `.that` member access, so it is never an `AwaitExpression` or a `ReturnStatement`.

`parsed.Expression` is already exactly the node this function reconstructs ("the outermost assertion expression … a call for method-style chains and a member access for property-style Chai assertions"), so the matcher-kind switch can go away entirely:

```go
reportNode, promiseWrapped, insideAssertionArray, shouldReport := resolveAsyncAssertionReportNode(
	parsed.Expression,
	opts.AlwaysAwait,
)
```

with `resolveAsyncAssertionReportNode` taking `assertionNode *ast.Node` and starting straight at `getParentIfPromiseChained(assertionNode)`. I tried that locally: the three cases above become valid and the existing rule tests, including the property-matcher ones added in 751fa0a1, all still pass.

swwind (commented):
| Code | Expected | This PR |
|---|---|---|
| `await expect(p).resolves.to.be.a("string").that.contains("x");` | Valid | ❌ reports `asyncMustBeAwaited` |
| `await expect(p).resolves.to.be.an("object").that.is.ok;` | Valid | ❌ reports `asyncMustBeAwaited` |
| `return expect(p).resolves.to.be.a("string").that.contains("x");` | Valid | ❌ reports `asyncMustBeAwaited` |

The property-matcher fix from the last round checks out — the new assertion-node branch and its regression tests are correct. One remaining false positive in the same function, on multi-matcher Chai chains; see inline comment.

internal/plugins/rstest/rules/valid_expect/valid_expect.go
elecmonkey: Fixed in 2b30cf1a. `resolveAsyncAssertionReportNode` now takes the assertion node and starts at `getParentIfPromiseChained`, called with `parsed.Expression`; the matcher-kind switch is gone. All three cases are valid now, with regression tests for them plus one invalid multi-matcher chain that locks the report range and the `await` fix location. Thanks!

internal/plugins/rstest/rules/valid_expect/valid_expect.go
elecmonkey: Fixed in a795801e — `rstestUtils.IsMemberAccessNode` is exported and reused; the local copy is gone.

internal/plugins/rstest/rules/valid_expect/valid_expect.go
```diff
@@ -0,0 +1,488 @@
...
+
+	promiseChainedAssertionNode := getParentIfPromiseChained(assertionNode)
+	insideAssertionArray = promiseChainedAssertionNode.Parent != nil && promiseChainedAssertionNode.Parent.Kind == ast.KindArrayLiteralExpression
+	reportNode = promiseChainedAssertionNode
+	if promiseCallNode := findPromiseCallExpressionNode(promiseChainedAssertionNode); promiseCallNode != nil {
+		reportNode = promiseCallNode
+		promiseWrapped = true
+	}
+
+	if reportNode.Parent == nil || isAcceptableReturnNode(reportNode.Parent, !alwaysAwait) {
```
swwind: `isAcceptableReturnNode` is reached with the raw `reportNode.Parent`, which in the TypeScript AST can be a `ParenthesizedExpression`. ESTree has no such node, so upstream never sees it and an assertion in parentheses stays valid there:

```js
test("t", async () => { await (expect(p).resolves.toBe(1)); });   // asyncMustBeAwaited
test("t", () => { return (expect(p).resolves.toBe(1)); });        // asyncMustBeAwaited
```

Both are valid under eslint-plugin-jest 29 (checked against the real plugin). The autofix then rewrites working code, turning the second one into `async () => { return (await expect(...)); }`.

Walking out of the wrapper before the check fixes it:

```go
outermost := internalUtils.OutermostParenthesizedExpression(reportNode)
if outermost.Parent == nil || isAcceptableReturnNode(outermost.Parent, !alwaysAwait) {
```

I tried that locally: the two cases above become valid, `(expect(p).resolves.toBe(1));` still reports and fixes to `(await expect(p).resolves.toBe(1));`, and the existing rule tests all pass.

swwind (commented):
| Code | ESLint | This PR |
|---|---|---|
| `test("t", async () => { await (expect(p).resolves.toBe(1)); });` | Valid | ❌ reports `asyncMustBeAwaited` |
| `test("t", () => { return (expect(p).resolves.toBe(1)); });` | Valid | ❌ reports `asyncMustBeAwaited` |

Both fixes from the last round check out: `resolveAsyncAssertionReportNode` now works from `parsed.Expression`, the multi-matcher Chai chains are valid with regression tests on both sides, and the exported `IsMemberAccessNode` reuse is clean. I also re-checked the then/catch chaining, `Promise.all` de-duplication (including Chai property matchers inside the array), `alwaysAwait` + `return` fixes, element-access matchers and deep Chai chains — all match upstream.

One remaining divergence, inherited from the jest port rather than introduced here: parenthesized assertions. `jest/valid-expect` in this repo reports the same two cases, so it is your call whether to fix it here or leave both rules to a follow-up. Details inline.

internal/plugins/rstest/rules/valid_expect/valid_expect.go
elecmonkey: Fixed in 1474a0a1, on the jest rule too — it had the same line, and the same false positive. The await/return check now walks out with `internalUtils.OutermostParenthesizedExpression` before looking at the parent, while the report range and the fix insertion point stay on the assertion itself. Regression tests on both sides, including the unawaited `(expect(p).resolves.toBe(1));` that still reports and fixes inside the parentheses.

Since both copies had to be fixed, the shared machinery — Promise-chain walking, the await/return check, the `Promise.all([...])` de-duplication and the async/await fixes — moved to `internal/utils/test_framework/rules/valid_expect`. The rule bodies stay separate; the PR description covers the jest behavior change. Thanks!

internal/utils/test_framework/rules/valid_expect/async_assertion.go
```diff
@@ -0,0 +1,181 @@
...
+	if promiseCallNode := findPromiseCallExpressionNode(promiseChainedAssertionNode); promiseCallNode != nil {
+		reportNode = promiseCallNode
+		promiseWrapped = true
+	}
+
+	// Parentheses around the assertion are nodes in the TypeScript AST but not
+	// in ESTree, so the await or return upstream sees as the direct parent sits
+	// one or more wrappers up here. reportNode itself stays the report range and
+	// the insertion point for the fix.
+	outermost := internalUtils.OutermostParenthesizedExpression(reportNode)
```
swwind, line 168: The walk above this line still looks at raw parents, so a parenthesized assertion is invisible to it:

```js
// getParentIfPromiseChained: node.Parent is the parenthesis, so grandParent is the `.then` member access, not a call
test("t", async () => { await (expect(p).resolves.toBe(1)).then(() => {}); });                             // jest: asyncMustBeAwaited, twice
// insideAssertionArray / findPromiseCallExpressionNode never see the array or the Promise.all around it
test("t", async () => { await Promise.all([(expect(p).resolves.toBe(1)), expect(q).resolves.toBe(2)]); });  // both rules: asyncMustBeAwaited
```

Both are valid under eslint-plugin-jest 29 (checked against the real plugin). `AwaitFix` has the same gap in its `ReturnStatement` check — with `alwaysAwait`, `return (expect(p).resolves.toBe(1));` is fixed to `return (await expect(p).resolves.toBe(1));`, where upstream drops the `return` and produces `await (expect(p).resolves.toBe(1));`. The report is correct there, only the fix differs.

Walking out of the wrapper at each of those points covers all three:

```go
func findPromiseCallExpressionNode(node *ast.Node) *ast.Node {
	node = internalUtils.OutermostParenthesizedExpression(node)
	if node == nil || node.Parent == nil || node.Parent.Parent == nil {

func getParentIfPromiseChained(node *ast.Node) *ast.Node {
	outer := internalUtils.OutermostParenthesizedExpression(node)
	if outer == nil || outer.Parent == nil || outer.Parent.Parent == nil {
		return node
	}
	grandParent := outer.Parent.Parent

// ResolveAsyncAssertionReportNode
	outerChained := internalUtils.OutermostParenthesizedExpression(promiseChainedAssertionNode)
	insideAssertionArray = outerChained.Parent != nil && outerChained.Parent.Kind == ast.KindArrayLiteralExpression

// AwaitFix
	outer := internalUtils.OutermostParenthesizedExpression(node)
	if alwaysAwait && outer.Parent != nil && outer.Parent.Kind == ast.KindReturnStatement {
		ret := outer.Parent
		retRange := internalUtils.TrimNodeTextRange(sourceFile, ret)
		nodeRange := internalUtils.TrimNodeTextRange(sourceFile, outer)
```

I tried that locally: the three cases above line up with upstream, `Promise.all([(expect(...)), ...])` without an await still reports `promisesWithAsyncAssertionsMustBeAwaited` once on the whole call, and both rules' existing suites pass.

internal/utils/test_framework/rules/valid_expect/async_assertion.go
```diff
@@ -0,0 +1,181 @@
...
+
+	callee := ast.SkipParentheses(node.AsCallExpression().Expression)
+	if !isMemberAccessNode(callee) {
+		return false
+	}
+
+	return testFramework.CalleeChainName(internalUtils.AccessExpressionObject(callee)) == "Promise"
+}
+
+func isMemberAccessNode(node *ast.Node) bool {
```
swwind: `ast.IsAccessExpression` is the same predicate and is what the rest of the repo uses; the only difference is the nil guard, and both call sites here pass a non-nil node. Also, now that the helper lives in this package, `rstestUtils.IsMemberAccessNode` is used only inside its own package again, so it can go back to unexported.

internal/plugins/rstest/rules/valid_expect/valid_expect_test.go
```diff
@@ -0,0 +1,181 @@
...
+			{Code: `expect(value).not.toBe(1);`},
+			{Code: `test("t", async () => { await expect.poll(() => value).toBe(1); });`},
+			// poll/element are not treated as async by valid-expect (matching
+			// vitest: only resolves/rejects modifiers and asyncMatchers require
+			// await), so a poll chain without await is not reported here.
+			{Code: `test("t", () => { expect.poll(() => value).toBe(1); });`},
+
+			// Provenance: non-Rstest expect and local shadow are ignored.
+			{Code: `import { expect } from "vitest"; expect(1);`},
+			{Code: `const expect = () => {}; expect(1);`},
```
swwind, line 76: No test on either side passes `Options`, so `alwaysAwait`, `asyncMatchers`, `minArgs` and `maxArgs` are unexercised for this rule. I checked all four locally and they behave correctly (`alwaysAwait` rejects a returned assertion, `minArgs: 2` reports `notEnoughArgs` on `expect(value)`, `maxArgs: 2` accepts `expect(1, 2)`, a custom `asyncMatchers` entry requires `await` while the defaults keep working) — worth locking in, since `parseOptions` is the only part of the rule with no coverage at all.

swwind (commented):
| Code | ESLint | This PR |
|---|---|---|
| `test("t", async () => { await Promise.all([(expect(p).resolves.toBe(1)), expect(q).resolves.toBe(2)]); });` | Valid | ❌ `asyncMustBeAwaited` (both rules) |
| `test("t", async () => { await (expect(p).resolves.toBe(1)).then(() => {}); });` | Valid | ❌ `asyncMustBeAwaited`, reported twice (jest only) |

The extraction is clean and the parenthesized-assertion fix is right for the shape it targets — I re-checked the two fixed cases, the unawaited `(expect(...))` that still reports and fixes inside the parentheses, `Promise.all` de-duplication, then/catch chaining, and the `alwaysAwait` + `return` fixes on both rules. `OutermostParenthesizedExpression` is applied only at the final await/return check, though, so the chain walk above it and `AwaitFix` still stop at a `ParenthesizedExpression`; the two rows above are what is left, plus a fix-output divergence under `alwaysAwait`. Both were already true of `jest/valid-expect` before this PR, so this is an incomplete fix rather than a regression — your call whether to finish it here. A verified patch, a helper nit and an options-coverage gap are inline.

internal/utils/test_framework/rules/valid_expect/async_assertion.go
elecmonkey: Fixed in `07efda2f`. Parentheses are now skipped consistently during the `.then`/`.catch` walk, `Promise.all` array and wrapper detection, and the `alwaysAwait` return replacement. I added regressions on both rules for parenthesized assertions in `Promise.all`, chained `.then`, single-report Promise fixes, and the `alwaysAwait` output. The new cases failed against the previous implementation before passing with this fix.

internal/utils/test_framework/rules/valid_expect/async_assertion.go
elecmonkey: Fixed in `52903569`. The shared code now uses `ast.IsAccessExpression`, and the Rstest parser helper is unexported again.

internal/plugins/rstest/rules/valid_expect/valid_expect_test.go
elecmonkey: Covered in `52903569`. Rstest now has cases for `alwaysAwait`, `asyncMatchers`, `minArgs`, and `maxArgs`. Jest already had coverage for all four options, so I kept the new coverage scoped to Rstest.

swwind (commented):
| Code | ESLint | This PR |
|---|---|---|
| `test("t", () => { Promise.all(([expect(p).resolves.toBe(1), expect(q).resolves.toBe(2)])); });` | one `promisesWithAsyncAssertionsMustBeAwaited` on the whole call | two `asyncMustBeAwaited`, one per assertion |

All three items from the last round check out. I re-verified the two parenthesized shapes in `.then` chains and `Promise.all`, the `alwaysAwait` return output (upstream's `sourceCodeText.replace('return', 'await')` produces exactly the same text), the `ast.IsAccessExpression` swap, the parser helper going back to unexported, and the four new option cases — which also put the schema through validation. Each new regression test fails against the pre-fix `async_assertion.go`, so they lock the behavior rather than describing it. The duplicate `asyncMustBeAwaited` on an unawaited `(expect(p).resolves.toBe(1)).then(() => {})` matches upstream, which reports it twice as well, so nothing to do there.

One parenthesized shape is still uncovered; it was equally true of `jest/valid-expect` before this PR, so it is not a regression and it is your call whether to take it. Details inline.

internal/utils/test_framework/rules/valid_expect/async_assertion.go
```diff
@@ -0,0 +1,173 @@
...
+
+	return testFramework.CalleeChainName(internalUtils.AccessExpressionObject(callee)) == "Promise"
+}
+
+func getPromiseCallExpressionNode(node *ast.Node) *ast.Node {
+	if node == nil {
+		return nil
+	}
+
+	if node.Kind == ast.KindArrayLiteralExpression && node.Parent != nil && node.Parent.Kind == ast.KindCallExpression {
```
swwind: The array-to-call step still looks at the raw parent, so parentheses around the array argument hide the `Promise.all` wrapper:

```js
test("t", () => { Promise.all(([expect(p).resolves.toBe(1), expect(q).resolves.toBe(2)])); });
```

eslint-plugin-jest 29 reports this once, `promisesWithAsyncAssertionsMustBeAwaited` over the whole `Promise.all((...))` call, and fixes it to `await Promise.all((...))`. Both rules here report `asyncMustBeAwaited` twice instead, once per assertion, and the fix awaits each element rather than the call.

```go
if outer := internalUtils.OutermostParenthesizedExpression(node); node.Kind == ast.KindArrayLiteralExpression && outer.Parent != nil && outer.Parent.Kind == ast.KindCallExpression {
	node = outer.Parent
}
```

I tried that locally: the case above becomes a single `promisesWithAsyncAssertionsMustBeAwaited` with upstream's report range and fix, and both rules' existing suites pass.

elecmonkey: Fixed in `6a15afd9`. `getPromiseCallExpressionNode` now walks through parenthesized wrappers around an array before checking its enclosing call, so `Promise.all(([...]))` is recognized as one promise wrapper. Both rules now report a single `promisesWithAsyncAssertionsMustBeAwaited` over the whole call and fix it by awaiting the call rather than the individual array elements.

I added Jest and Rstest Go/JS regressions for this shape. The new Go cases failed against the previous implementation with per-element `await` output, then passed with the shared fix. The full Jest/Rstest/utils Go sweep, `go vet`, `golangci-lint`, rebuilt JS integration tests, formatting, and spelling checks all pass.

internal/plugins/rstest/rules/valid_expect/valid_expect.go
```diff
@@ -0,0 +1,349 @@
...
+				}
+
+				if parsed.Reason != rstestUtils.RstestExpectParseReasonNone {
+					reportBrokenChain(ctx, parsed)
+					return
+				}
+
+				// Static members (expect.assertions(1), expect.any(...)) carry no
+				// assertion factory, so argument and await checks do not apply.
+				if parsed.Head != nil && parsed.Head.Kind == ast.KindCallExpression {
```
swwind: The await check still runs when `parsed.Head` is nil, because only the argument check sits behind this guard. A bare modifier chain — `expect.resolves.toBe(1)`, where the author forgot `expect(x)` — reaches `shouldBeAwaited` through `Modifiers`, and a static member whose name is in `asyncMatchers` (`expect.toResolve()`, `expect.toReject()` with the defaults) reaches it through `Matcher`:

```js
expect.resolves.toBe(1);                        // asyncMustBeAwaited, fixed to `await expect.resolves.toBe(1)`
test("t", () => { expect.toResolve(); });       // asyncMustBeAwaited
```

Both are valid under eslint-plugin-jest 29 and @vitest/eslint-plugin (checked against the real plugins), and the jest rule here mirrors upstream's early return (`valid_expect.go:343`, `expectCall.Kind != CallExpression → return`), so the two ports disagree on the same input. The fix is the awkward part: it inserts `await` and marks the enclosing function `async` in code the rule cannot validate. It also makes `valid_expect.md:46-48` and the PR description ("argument and await checks are skipped for them") true only of the argument check.

Hoisting the check into an early return matches jest exactly:

```go
if parsed.Head == nil || parsed.Head.Kind != ast.KindCallExpression {
	return
}

args := parsed.Head.AsCallExpression().Arguments.Nodes
...
```

I tried that locally: both cases above become valid, `expect.hasAssertions()` / `expect.assertions(1)` / `expect.any(Number)` are unaffected, and both rule suites pass.

internal/utils/test_framework/rules/valid_expect/async_assertion.go
```diff
@@ -0,0 +1,173 @@
...
+func isAcceptableReturnNode(node *ast.Node, allowReturn bool) bool {
+	if node == nil {
+		return false
+	}
+
+	if allowReturn && node.Kind == ast.KindReturnStatement {
+		return true
+	}
+	if node.Kind == ast.KindConditionalExpression {
+		return isAcceptableReturnNode(node.Parent, allowReturn)
```
swwind: This recursion walks to the raw parent, which in the TypeScript AST can be a `ParenthesizedExpression` — the same ESTree gap the rest of this file now handles:

```js
test("t", () => { return (cond ? expect(p).resolves.toBe(1) : expect(q).toBe(1)); });
```

Valid under eslint-plugin-jest 29 and @vitest/eslint-plugin (checked against the real plugins); both rules here report `asyncMustBeAwaited`. The unparenthesized form is already covered by the jest suite's `expect.extend` conditional cases.

This is not a regression — `main` reports it too, so it is the last of the parenthesis class rather than something this PR introduced. Your call whether to finish it here:

```go
if node.Kind == ast.KindConditionalExpression {
	return isAcceptableReturnNode(internalUtils.OutermostParenthesizedExpression(node).Parent, allowReturn)
}
```

I tried that locally alongside the other patch: the case above becomes valid on both rules and both suites stay green.

swwind (commented):
| Code | eslint-plugin-jest / vitest | This PR |
|---|---|---|
| `expect.resolves.toBe(1);` | Valid | ❌ `asyncMustBeAwaited`, fixed to `await expect.resolves.toBe(1)` |
| `test("t", () => { expect.toResolve(); });` | Valid | ❌ `asyncMustBeAwaited` (default `asyncMatchers`) |
| `test("t", () => { return (cond ? expect(p).resolves.toBe(1) : expect(q).toBe(1)); });` | Valid | ❌ `asyncMustBeAwaited` (both rules) |

The extraction is behavior-preserving: `utils.GetJestFnMemberEntries` is a straight alias of `testFramework.GetMemberEntries`, and `ast.IsAccessExpression` matches the old `IsMemberAccessNode` on both call sites, which are non-nil by construction. I re-checked the `Promise.all` de-duplication including the parenthesized-array shape, then/catch chaining through nested parens, arrow-body assertions, the `alwaysAwait` return-replacement output, the vitest message/options argument allowances, Chai property and multi-matcher chains, and registration in `all.go`, the `recommended` preset, the preset test and the rstest test config. All three rows above were checked against the real `eslint-plugin-jest@29` and `@vitest/eslint-plugin`, which agree on every one.

Two items inline: one divergence this rule introduces, and one inherited from the jest port. One nit while you are here: no rstest case asserts a report range, so the `tooManyArgs` / `notEnoughArgs` ranges rely on the jest suite's column assertions.

internal/utils/test_framework/rules/valid_expect/async_assertion.go
elecmonkey: Fixed in `64fa7516`: parenthesized conditional expressions now unwrap before the return check, with Jest and Rstest regressions.

internal/plugins/rstest/rules/valid_expect/valid_expect.go
elecmonkey: Fixed in `af801972`: forms without an assertion factory now return before argument and async checks, with regressions for both reported cases.

elecmonkey:
Also addressed the range-coverage nit in `97fcebad`: the Rstest suite now pins the exact `notEnoughArgs` and `tooManyArgs` ranges.

internal/plugins/rstest/rules/valid_expect/valid_expect.md
```diff
@@ -0,0 +1,76 @@
...
+
+- **A second argument to `expect` is allowed** when it is a message string or
+  template literal (`expect(value, "msg")`), and `expect.poll(fn, options)` /
+  `expect.element(el, options)` accept an options object. These do not trigger
+  `tooManyArgs`.
+- **Chai property matchers are valid** without a call: `expect(value).to.be.ok`,
+  `expect(spy).to.have.been.called`. They are not reported as `matcherNotCalled`.
+- **Static members carry no assertion**: `expect.assertions(1)`,
+  `expect.hasAssertions()` and asymmetric matchers such as `expect.any(Number)`
+  are not subject to argument or await checks.
```
swwind: `af801972` widened this skip past static members: a bare modifier chain (`expect.resolves.toBe(1)`) and a static async matcher (`expect.toResolve()`) are skipped too, and neither is an asymmetric matcher or an `expect.assertions`-style call. Naming the actual condition — no assertion factory — covers all of them:

```suggestion
- **Forms with no assertion factory carry no assertion**: `expect.assertions(1)`,
  `expect.hasAssertions()`, asymmetric matchers such as `expect.any(Number)` and
  bare chains such as `expect.resolves.toBe(1)` or `expect.toResolve()` are not
  subject to argument or await checks.
```

The comment at `valid_expect.go:295-296` has the same narrow framing.

swwind (commented):
Final pass — no remaining divergences. One doc nit inline, nothing blocking.

swwind (approved):
LGTM

fansenze (approved):

PR #1696, opened 2026-08-12
feat(jest,rstest): add valid-expect-in-promise rule

elecmonkey:
## Summary

Port `valid-expect-in-promise` for both jest and rstest: a promise chain that contains an assertion must be returned, awaited, or consumed by a safe promise sink, otherwise the test can finish before the assertion runs or before its failure surfaces. Both plugins bind to the shared framework-neutral engine in `internal/utils/test_framework/rules/valid_expect_in_promise`; their adapters inject callback discovery, ignored callbacks, assertion provenance, and async assertion sinks. The rules are added to both `recommended` presets at `error`.

Promise sink corrections, applied to **both** jest and rstest with explicit tests:

- **`Promise.reject(promise)` is not a safe sink** — it does not adopt the argument's state, so a floating assertion chain passed to it is still unconsumed.
- **`Promise.allSettled([promise])` is not a safe sink** — a failing assertion becomes a fulfilled `{ status: 'rejected' }` result, so the test still passes.

Analysis and reporting improvements:

- unrelated assignments no longer terminate tracking of the assertion promise;
- statically resolvable named test callbacks and promise handlers are analyzed;
- floating promise chains inside nested handlers are tracked;
- consecutive handlers in the same floating chain are de-duplicated to one diagnostic;
- completion checks stop at the current test callback boundary;
- paths that fail through an escaping `throw` do not require promise consumption, while caught or `finally`-suppressed throws remain tracked.
- promise-handler identifiers declared without an initializer (`let handler;`, `var handler;`, or ambient `declare const` / `declare var`) are treated as unresolved instead of dereferencing a nil initializer and aborting the rslint process. Flow-sensitive assignments to those declarations, including assign-in-hook patterns, remain outside this rule's analysis boundary.
- a chain reaches its binding through `||`, `??`, `&&`, either branch of a conditional, and the right-hand side of a comma, because each of those evaluates the chain on exactly the paths where the binding receives it. `let pending = fallback || load().then(...)` is tracked, and so is the longhand `pending = pending || load().then(...)`, which keeps the promise the binding already holds for the same reason `||=` does;
- an array literal in a binding holds the chain and is consumed only element-wise: `await Promise.all(pending)` and `for await (const settled of pending)` consume it, while `await pending` awaits the array itself and leaves the chain floating;
- a chain stored by a logical assignment is reported at the assignment, and a chain handed to another call is reported at the statement that call sits in.

Positions and consumption above are shared by both rules. The rest of this section compares the jest rule with `eslint-plugin-jest`.

The jest rule reports the same code at the same range as `eslint-plugin-jest` for plain and logical assignments and for a chain passed to another call. It is stricter about what counts as consumption, and none of the following is consumed, although upstream reports them inconsistently or not at all:

- `Promise.race([chain])` and `Promise.any([chain])`, neither of which preserves an assertion failure the way `Promise.all` does;
- any wrapper other than `Promise.resolve` and `Promise.all`, such as `helper(chain)`;
- `void chain` and `throw chain`;
- a property of an object literal that is not destructured straight into a binding, and an array literal that is never consumed element-wise;
- a tagged template interpolation;
- a compound arithmetic assignment such as `count += chain`, which stores no promise at all.

## Shared analysis

Promise-chain collection, binding and destructuring flow, CFG dataflow, safe wrapper and aggregator handling, nested-handler propagation, de-duplication, and reporting live in one shared implementation. The shared package does not import either plugin or branch on a framework name.

The adapters provide four framework-specific inputs:

- test callback functions;
- ignored callback functions, used for Jest's `done` bailout;
- assertion identity;
- `resolves` / `rejects` assertion sinks.

This keeps Rstest's `assert` provenance and each framework's parser integration local while ensuring fixes to the common promise analysis cannot drift between copies.

## Differences between the jest and rstest rules

- **jest keeps the `done` / ambiguous-`.each` bailout**, which is a real framework behavior. rstest has no `done` callback — a test callback's first argument is `TestContext` — so the rstest rule does not inherit that bailout.
- **rstest recognizes both official `expect` and `@rstest/core`'s Chai `assert`** as assertions. Both throw on failure, so both are unsafe in a floating handler. jest has no official `assert` API; `eslint-plugin-vitest` misses `assert`, and rstest deliberately does not follow that gap.
- the rstest rule covers TestContext `expect`, `.for` / `.each`, overloads, `import.meta.rstest`, and `@rstest/playwright`.

The upstream jest fixture suite is ported as-is; fixtures that conflict with the corrections above are moved to the opposite group with tests documenting why.

## Related Links

Related #935

## Checklist


internal/plugins/jest/rules/valid_expect_in_promise/valid_expect_in_promise.go
```diff
@@ -0,0 +1,769 @@
...
+	current := root
+	for current != nil && current.Parent != nil && current.Parent != function {
+		parent := current.Parent
+		switch parent.Kind {
+		case ast.KindParenthesizedExpression:
+			current = parent
+		case ast.KindAwaitExpression:
+			return testFramework.AbruptCompletionPropagatesFailure(parent, function)
+		case ast.KindReturnStatement:
+			return testFramework.AbruptCompletionPropagatesFailure(parent, function)
```
swwind: An enclosing `catch` never intercepts the rejection of a promise that is `return`ed — that is exactly the `return` vs `return await` distinction. Routing `KindReturnStatement` through `AbruptCompletionPropagatesFailure` therefore treats this as unconsumed:

```js
test("case", () => { try { return promise.then(value => expect(value).toBe(1)); } catch (error) {} });
```

Same for the `async` variant. The `finally` half of that helper does apply to `return`; the `catch` half should be limited to the `KindAwaitExpression` branch above.

internal/plugins/jest/rules/valid_expect_in_promise/valid_expect_in_promise.go
```diff
@@ -0,0 +1,769 @@
...
+		return false
+	}
+	current := identifier
+	for current.Parent != nil {
+		parent := current.Parent
+		switch parent.Kind {
+		case ast.KindParenthesizedExpression:
+			current = parent
+		case ast.KindAwaitExpression:
+			return testFramework.AbruptCompletionPropagatesFailure(parent, nil)
```
swwind: With a `nil` boundary the walk continues past the enclosing function all the way to the SourceFile, so it picks up `try` statements that live outside the test callback:

```js
try { test("case", async () => { const pending = promise.then(v => expect(v).toBe(1)); await pending; }); } catch (error) {}
```

A module-level `try` around test registration is enough to make every awaited promise variable in the file report. `isDirectlyConsumed` passes `function`, which is why only the variable-binding path is affected; `expressionIsAwaitedOrReturned` (575/577) passes `nil` too. The boundary should be the function whose CFG is being analyzed.

internal/plugins/rstest/rules/valid_expect_in_promise/valid_expect_in_promise.go
```diff
@@ -0,0 +1,878 @@
...
+	current := root
+	for current != nil && current.Parent != nil && current.Parent != function {
+		parent := current.Parent
+		switch parent.Kind {
+		case ast.KindParenthesizedExpression:
+			current = parent
+		case ast.KindAwaitExpression:
+			return testFramework.AbruptCompletionPropagatesFailure(parent, function)
+		case ast.KindReturnStatement:
+			return testFramework.AbruptCompletionPropagatesFailure(parent, function)
```
swwind: Same as the jest copy; reproduced here as well.

internal/plugins/rstest/rules/valid_expect_in_promise/valid_expect_in_promise.go
```diff
@@ -0,0 +1,878 @@
...
+		return false
+	}
+	current := identifier
+	for current.Parent != nil {
+		parent := current.Parent
+		switch parent.Kind {
+		case ast.KindParenthesizedExpression:
+			current = parent
+		case ast.KindAwaitExpression:
+			return testFramework.AbruptCompletionPropagatesFailure(parent, nil)
```
swwind: Same as the jest copy (`expressionIsAwaitedOrReturned` at 559/561 also passes `nil`); reproduced here as well.

swwind (commented):
Solid port overall — I read through the destructuring binding map, the promise-chain recognizer, the CFG `Expression` hook, the dataflow worklist, and the `no-conditional-expect` refactor (behaviour is preserved). Two false positives, each duplicated in both plugin copies.

| Code | ESLint | This PR |
| --- | --- | --- |
| `test("case", () => { try { return promise.then(v => expect(v).toBe(1)); } catch (e) {} });` | ok | error |
| `try { test("case", async () => { const p = promise.then(v => expect(v).toBe(1)); await p; }); } catch (e) {}` | ok | error |

Divergences I did not flag because they match the PR's stated intent: `let p = chain; if (a) { p = other; } await p;` is stricter than upstream, `await (0, pending)` is looser, and the `Promise.allSettled` / `Promise.reject` differences are documented.

internal/plugins/jest/rules/valid_expect_in_promise/valid_expect_in_promise.go
elecmonkey: Fixed in 465b8580. Returned promises now ignore enclosing catches while still respecting suppressing finally blocks. Thanks!

internal/plugins/jest/rules/valid_expect_in_promise/valid_expect_in_promise.go
elecmonkey: Fixed in 465b8580. Consumption checks now stop at the current callback boundary, including Promise.resolve and Promise.all paths. Thanks!

internal/plugins/rstest/rules/valid_expect_in_promise/valid_expect_in_promise.go
elecmonkey: Fixed in 465b8580 with the same return-vs-await correction and Rstest regressions. Thanks!

internal/plugins/rstest/rules/valid_expect_in_promise/valid_expect_in_promise.go
elecmonkey: Fixed in 465b8580 with the same callback-boundary fix for Rstest. Thanks!

internal/plugins/jest/rules/valid_expect_in_promise/valid_expect_in_promise.go
```diff
@@ -0,0 +1,759 @@
...
+			if in[successor.Index()] == nil {
+				in[successor.Index()] = state{}
+			}
+			changed := merge(in[successor.Index()], current)
+			if (firstVisit || changed) && !queued[successor.Index()] {
+				queue = append(queue, successor)
+				queued[successor.Index()] = true
+			}
+		}
+		if !reachableSuccessor {
```
swwind: A block with no reachable successor is not always a normal exit. A path that ends in `throw`, and the handler path the CFG forks to at the first throwable node inside a `try`, both arrive here with the candidate still live — and on those paths the test already fails, so consumption is not required.

```js
test("case", async () => { const pending = load().then(v => expect(v).toBe(1)); if (!ready) { throw new Error("no"); } await pending; });
```

Clearing the state at a `throw` through the `Statement` hook covers the explicit case. The `try` fork reaches here through the handler entry, which is why `try { setup(); await pending; } catch (error) { throw error; }` reports while the same body without `setup();` does not.

internal/plugins/rstest/rules/valid_expect_in_promise/valid_expect_in_promise.go
```diff
@@ -0,0 +1,868 @@
...
+			if firstVisit {
+				in[successor.Index()] = state{}
+			}
+			changed := merge(in[successor.Index()], current)
+			if (firstVisit || changed) && !queued[successor.Index()] {
+				queue = append(queue, successor)
+				queued[successor.Index()] = true
+			}
+		}
+		if !reachableSuccessor {
```
swwind: Same as the jest copy; reproduced here as well.

swwind (commented):
Re-reviewed after 465b8580. Both false positives from the last round are fixed — I re-ran them along with nested variants (`try { try { test(...) } catch {} } finally {}`, `try { describe(...test(...)) } catch {}`, aggregator and wrapper sinks under an outer `try`) and they are all clean. Dropping `Promise.any` / `Promise.race` matches upstream too: ESLint reports both single-element forms.

One false positive left, in both copies — a path that ends in a `throw` is treated as a path that failed to consume the promise.

| Code | ESLint | This PR |
| --- | --- | --- |
| `test("case", async () => { const pending = load().then(v => expect(v).toBe(1)); if (!ready) { throw new Error("no"); } await pending; });` | ok | error |

The same root cause shows up inside `try`: `try { setup(); await pending; } catch (error) { throw error; }` reports, while the identical body without `setup();` is one of the rule's own valid cases. ESLint reports that one as well, so it is parity rather than a divergence, but it makes the rethrowing-`catch` valid case fragile.

Unrelated to the above: the docs do not mention that an enclosing `catch`, or a `finally` that returns / breaks / continues, makes an awaited chain count as unconsumed — `try { await load().then(v => expect(v).toBe(1)); } catch (error) {}` is valid upstream and an error here. A sentence in both `.md` files would cover it.

internal/plugins/jest/rules/valid_expect_in_promise/valid_expect_in_promise.go
elecmonkey: Fixed in `3b89dff3`. The duplicate Jest/Rstest engines are now one framework-neutral dataflow implementation, with callback discovery, ignored callbacks, assertion identity, and async assertion sinks injected by each adapter.

The shared `Statement` hook clears live candidates only for a `throw` whose failure can escape the current callback. A throw caught inside the callback, or suppressed by an abrupt `finally`, remains live and still reports. This also covers the handler path created by the first throwable node inside a `try`.

I added Jest and Rstest Go/JS regressions for the early-throw branch, the throwable-node handler fork with a rethrowing catch, a normal `finally`, a swallowed throw, and a suppressing `finally`. The three new valid cases failed before the fix. Both rule docs now describe the `catch` / `finally` boundary.

internal/plugins/rstest/rules/valid_expect_in_promise/valid_expect_in_promise.go
elecmonkey: Fixed in `3b89dff3` by the same shared engine and regression matrix as the Jest copy. The Rstest adapter now contains only its callback/assertion bindings and `assert` provenance; the throw-path dataflow fix lives in the shared implementation.

internal/utils/test_framework/rules/valid_expect_in_promise/valid_expect_in_promise.go
```diff
@@ -0,0 +1,768 @@
...
+			return rule.RuleListeners{}
+		},
+	}
+}
+
+func (a *analyzer) collectFunction(function *ast.Node, owner *chainGroup) {
+	if function == nil {
+		return
+	}
+	key := functionOwner{function: function, owner: owner}
```
swwind: `analyzing` is keyed on `{function, owner}`, but every recursion level allocates a fresh owner: `groupFor` keys groups on `{root, owner}`, so re-entering the same function under the group it just produced never repeats a key. `handler` → group for `queue.then(handler)` under it → `handler` again under that group → … until the stack is gone.

Keying the re-entrancy guard on the function alone terminates the cycle. It stays a stack guard rather than a memo, so the two sequential `promise.then(handler)` registrations in the existing `function handler(value) { expect(value).toBe(1); }` case are not nested and both are still analyzed.

internal/utils/test_framework/rules/valid_expect_in_promise/valid_expect_in_promise.go
```diff
@@ -0,0 +1,768 @@
...
+		Run: func(ctx rule.RuleContext, _ []any) rule.RuleListeners {
+			runtime := config.Prepare(ctx)
+			analyzer := &analyzer{
+				ctx:       ctx,
+				runtime:   runtime,
+				groups:    map[groupKey]*chainGroup{},
+				analyzing: map[functionOwner]bool{},
+				message:   buildFloatingPromiseMessage(config.MessageDescription),
+			}
+			for function := range runtime.TestCallbackFunctions {
```
swwind, line 106: Seeding the work list with `for function := range runtime.TestCallbackFunctions` makes diagnostic emission order depend on Go map order. The CLI sorts by position before printing, so output is unaffected, but `rule_tester` compares errors in emission order — a multi-error case that pins `Line` / `Column` would be flaky. Sorting the callbacks by position before the loop removes the hazard.

swwind (commented):
Re-reviewed after 3b89dff3. The throw-path false positive is gone — I re-ran the early-`throw` case, the `try { setup(); await pending; } catch (e) { throw e; }` handler fork, a `switch` arm that throws while the other awaits, a `for..of` with a conditional throw, and `try { throw } catch { await pending }`; all clean. `try { await pending; } catch (e) { throw e; } finally { cleanup(); }` also stays valid, where ESLint has a false positive. I diffed both pre-refactor copies against the shared engine: the extraction is faithful, the only behaviour change is the new `flowFailureExit` event. Both `.md` files now describe the `catch` / `finally` boundary.

One crash left, in both plugins — a promise handler that is reachable from itself recurses forever and overflows the stack.

| Code | ESLint | This PR |
| --- | --- | --- |
| `function handler() { queue.then(handler); }` + `test("case", () => promise.then(handler));` | ok | stack overflow |
| `function first() { queue.then(second); }` + `function second() { queue.then(first); }` + `test("case", () => promise.then(first));` | ok | stack overflow |

No assertion is needed to trigger it — the recursion is in the collection walk, so `go test` on either rule package ends in `fatal error: stack overflow`, and in the CLI it is an unrecoverable process abort rather than a diagnostic.

internal/utils/test_framework/rules/valid_expect_in_promise/valid_expect_in_promise.go
elecmonkey: Fixed in `0e376306`. The re-entrancy guard is now keyed only by the function node and remains stack-scoped through deferred deletion. This terminates self-recursive and mutually recursive promise handlers without memoizing completed analyses, so the existing case with two sequential registrations of the same handler is still analyzed twice.

I added both cycle shapes to the Jest and Rstest Go suites, plus JS integration coverage across the two plugins. The targeted tests, the full Jest/Rstest/utils Go sweep, `go vet`, `golangci-lint`, and the rebuilt Jest/Rstest JS integration tests all pass.

internal/utils/test_framework/rules/valid_expect_in_promise/valid_expect_in_promise.go
elecmonkey: Good catch. I checked both Go rule suites: they have single-error position assertions, but no multi-error case that pins `Line` / `Column`. The only multi-error cases expect two identical message IDs, so their order is currently unobservable; CLI output is sorted separately. Since this is a latent `rule_tester`-only hazard rather than a current behavior or test failure, I’m leaving callback iteration unchanged in this PR.

internal/utils/test_framework/rules/valid_expect_in_promise/valid_expect_in_promise.go
```diff
@@ -0,0 +1,762 @@
...
+			return rule.RuleListeners{}
+		},
+	}
+}
+
+func (a *analyzer) collectFunction(function *ast.Node, owner *chainGroup) {
+	if function == nil {
+		return
+	}
+	if a.analyzing[function] {
```
swwind: **Exponential re-walk.** `analyzing` is a stack set, so it breaks cycles but not sharing: a function reachable by k distinct owner paths is walked k times and allocates k copies of every `chainGroup` in its body.

```js
function f0() { q.then(f1); q.then(f1); }
function f1() { q.then(f2); q.then(f2); }
// … 20 levels
function f20(v) { expect(v).toBe(1); }
test("t", () => q.then(f0));
```

`go test` on this single case: 0.14s at 16 levels, 0.44s at 18, 1.91s at 20 — roughly 4x per two levels, and `ordered` grows at the same rate. 30 levels is ~1 KB of source and does not finish.

The walk itself does not depend on `owner` — only the `directAssertion` propagation does. Memoizing on the function rather than on the stack, and applying a cached per-function summary to each new owner, would bound both the time and the allocation.

internal/plugins/rstest/utils/rstest_ast.go
```diff
@@ -2,80 +2,12 @@ package utils
...
-// member access of then / catch / finally with a statically known name, and
-// the argument count is within the method's arity — at least 1, at most 2 for
-// then and at most 1 for catch / finally. Mirrors eslint-plugin-jest's
-// isPromiseChainCall (valid-expect-in-promise.ts:24-47); jest's Go port has no
-// counterpart yet, so the upstream JS is the reference. Parentheses are
-// skipped where ESTree would not materialize them.
+// IsPromiseChainCall remains as a compatibility wrapper for Rstest utilities.
+// The framework-neutral implementation lives in test_framework now that both
+// Jest and Rstest consume it.
 func IsPromiseChainCall(node *ast.Node) bool {
```
swwind: **Dead wrapper.** Nothing calls `rstestUtils.IsPromiseChainCall` any more — the new rule reaches `test_framework` directly. The only remaining reference is `rstest_ast_test.go`, which duplicates `internal/utils/test_framework/promise_test.go`. Both files can be deleted.

swwind (commented):
Re-reviewed after 0e376306. The recursive-handler crash is fixed — I reverted the guard locally and confirmed the two new valid cases still overflow the stack without it. I also ran a fresh round of ~45 cases against real eslint-plugin-jest and found no new false positives or negatives: every remaining divergence is either already documented, or this PR being stricter and correct (`promise?.then(…)`, `void chain` and `return { p: chain }` are all missed upstream, and upstream's `switch`-with-every-arm-awaited and `do…while` false positives are gone here).

Two things left, one of them blocking.

**The guard stops the crash, but the analysis is still exponential.** `collectFunction` skips a function only while it is on the current analysis stack, so a function reachable by k distinct owner paths is re-walked k times, and each walk allocates a fresh `chainGroup` for every chain call inside it. Fan-out 2 at depth d gives 2^d walks and 2^d groups.

| depth | `go test` time for the single case |
| --- | --- |
| 16 | 0.14s |
| 18 | 0.44s |
| 20 | 1.91s |

That is ~4x per two levels, from under 1 KB of source. At 30 levels it is half an hour and 2^30 groups, so the rule turns into a hang or an OOM rather than a diagnostic — the same failure class as the stack overflow, reached by fan-out instead of a cycle.

**`rstestUtils.IsPromiseChainCall` is now dead.** After the move into `test_framework` nothing calls it; only its own test file keeps it alive, and that test duplicates the new `promise_test.go`.

Details inline.

internal/utils/test_framework/rules/valid_expect_in_promise/valid_expect_in_promise.go
elecmonkey: Fixed in `f7fc64b0`. Collection is now memoized per function: each function body is walked once, each promise-chain AST root owns one `chainGroup`, and repeated handler references only add owner edges. After CFG consumption is resolved, a finite worklist propagates assertion reachability from function summaries to every owner group and from safe groups to their containing functions. Collection and propagation are therefore bounded by the unique functions, chain roots, and owner edges in the source graph rather than by the number of owner paths.

I ran a local, uncommitted probe using the binary fan-out shape from this comment at depths 16, 18, 20, and 30. All four generated RuleTester cases completed in about 0.03s each, including depth 30; the complete `go test` command, including compilation, took 1.678s. The probe was removed, so no benchmark or timing threshold is committed.

The committed Jest and Rstest regressions instead lock the graph semantics: propagation to multiple owners, a single report for an unsafe nested chain, and assertion-bearing recursive safe chains. The full Jest/Rstest/utils Go test sweep, `go vet`, `golangci-lint`, and rebuilt Jest/Rstest JS integration tests all pass.

internal/plugins/rstest/utils/rstest_ast.go
elecmonkey: Removed in `f7fc64b0`. `internal/plugins/rstest/utils/rstest_ast.go` and its duplicate test are gone; `IsPromiseChainCall` now has a single implementation and test under `internal/utils/test_framework`.

internal/utils/test_framework/rules/valid_expect_in_promise/valid_expect_in_promise.go
```diff
@@ -0,0 +1,805 @@
...
+	declaration := internalUtils.GetDeclaration(a.ctx.TypeChecker, node)
+	if declaration == nil {
+		return nil
+	}
+	switch declaration.Kind {
+	case ast.KindFunctionDeclaration:
+		return declaration
+	case ast.KindVariableDeclaration:
+		initializer := declaration.AsVariableDeclaration().Initializer
+		initializer = ast.SkipParentheses(initializer)
```
swwind, line 287: `declaration.AsVariableDeclaration().Initializer` is nil for `let handler;`, `var handler;`, and any ambient `declare const` / `declare var`, and `ast.SkipParentheses` panics on nil instead of returning nil. An early `if initializer == nil { return nil }` covers every form I could reach — I applied it locally and re-ran the crashing cases and both rule suites clean.

There is a second way in, on a non-module test file: `GetDeclaration` returns `symbol.Declarations[0]`, so a locally declared handler whose name collides with an ambient DOM `var` — `name`, `status`, `top`, … — resolves to the lib declaration and crashes the same way. The nil guard turns that one into a handler that is silently not analyzed; I checked that scanning the symbol's other declarations does not recover the local function, so that residue is a small precision gap rather than something to chase in this PR.

swwind (commented):
Re-reviewed after `f7fc64b0`.

**The exponential re-walk is fixed.** The binary fan-out shape from the last round is now flat in depth: depth 60 — 2^60 distinct owner paths — analyses in the same ~0.1s as depth 16, where the previous revision took 1.91s at depth 20 and did not finish at 30. Collection is one walk per function and one `chainGroup` per chain root, with owner edges carrying assertion reachability instead of duplicated groups.

I checked the memoization is semantics-preserving and not just faster. I ran ~45 cases against real `eslint-plugin-jest@29.16.1`, weighted toward the shapes the new sharing could break: a handler reached from two owners where one chain is consumed and the other floating, three-level named handler chains with a floating link in the middle, a function that is both a test callback and a promise handler, self- and mutually-recursive handlers, and floating chains nested inside a shared handler. No new false positives or negatives — multi-owner propagation lands on the right groups in every case, and each remaining divergence is either already documented or this PR being stricter and correct.

`rstestUtils.IsPromiseChainCall` and its duplicate test are gone.

One crash left, in both plugins. It is not new to `f7fc64b0` — it has been there since `dd6d9f29` — but it is reachable from ordinary test code.

**`resolveFunction` dereferences a nil initializer.** A promise handler named by an identifier whose declaration has no initializer reaches `ast.SkipParentheses(nil)` and segfaults.

| Code | ESLint | This PR |
| --- | --- | --- |
| `let handler; test("case", () => promise.then(handler));` | ok | SIGSEGV |
| `declare const handler: () => void; test("case", () => promise.then(handler));` | ok | SIGSEGV |

The shape that will actually reach it is the assign-in-a-hook pattern, on a file that is entirely valid:

```ts
let handler: (value: unknown) => void;
beforeEach(() => { handler = value => expect(value).toBe(1); });
test("case", () => promise.then(handler));
```

`.catch` and `.finally` handlers hit it identically, and no assertion is needed to trigger it — the deref is in the collection walk. Rules run inside the goroutines `core.NewWorkGroup` spawns, which carry no `recover`, so in a normal CLI run this aborts the rslint process rather than producing a diagnostic; under `go test` it takes down the test binary before the remaining cases run.

Details inline.

internal/utils/test_framework/rules/valid_expect_in_promise/valid_expect_in_promise.go
elecmonkey: Fixed in `ab485be3`: variable declarations without an initializer now resolve as unknown instead of reaching `SkipParentheses(nil)`, with Jest and Rstest regressions.

internal/utils/test_framework/rules/valid_expect_in_promise/valid_expect_in_promise.go
```diff
@@ -0,0 +1,808 @@
...
+				return nil
+			}
+			return &chainBinding{
+				target: name,
+				symbol: a.symbolOf(name),
+			}
+		}
+		if ast.IsAssignmentExpression(parent, true) {
+			binary := parent.AsBinaryExpression()
+			if binary.OperatorToken.Kind != ast.KindEqualsToken ||
```
swwind: `bindingFor` accepts only `KindEqualsToken`, so `p ||= chain` / `p ??= chain` produces no binding, and `promiseValueFlowsThroughBinary` (548) does not list the logical-assignment operators either, so `isDirectlyConsumed` does not carry the value up to the later `await p` — the chain is reported as floating.

The semantics fit the `extends` flag that is already here: a logical assignment may leave the previous value in place, so the bind should not kill earlier candidates for the same symbol.

internal/plugins/jest/rules/valid_expect_in_promise/valid_expect_in_promise.md
```diff
@@ -0,0 +1,31 @@
...
+
+The rule follows promises stored in local bindings, including statically mappable array and object destructuring, and requires every reachable path to consume them. `Promise.resolve` and `Promise.all` preserve assertion failure. `Promise.reject` does not adopt its argument, and `Promise.allSettled` converts assertion failure into a fulfilled result, so neither is a safe sink.
+
+An awaited chain is safe only when its rejection can escape the test callback. An enclosing `catch` that may swallow the rejection, or a `finally` block that may `return`, `break`, or `continue`, does not count as safe consumption. A `catch` that always rethrows preserves the failure.
+
+Jest callbacks using `done` are not analyzed because the callback can coordinate promise completion explicitly. Named callbacks and promise-handler callbacks are analyzed when their relationship to a test is statically known.
+
+This implementation intentionally fixes upstream cases where unrelated assignments stop promise tracking, named callbacks are missed, nested handlers are missed, or `Promise.reject` / `Promise.allSettled` are incorrectly accepted.
+
+This rule has no options and does not provide an autofix.
```
swwind, line 39: #1692 gave every jest rule doc an `## Original Documentation` section with version-pinned links; this is the only one without it.

    ## Original Documentation

    - [eslint-plugin-jest: valid-expect-in-promise](https://github.com/jest-community/eslint-plugin-jest/blob/v29.16.0/docs/rules/valid-expect-in-promise.md)
    - [Source code](https://github.com/jest-community/eslint-plugin-jest/blob/v29.16.0/src/rules/valid-expect-in-promise.ts)

swwind (commented):
Re-reviewed after `ab485be3`.

I ran ~47 fresh cases against real `eslint-plugin-jest`, plus a crash sweep over ~45 shapes that reach `resolveFunction` or the collection walk from an unusual angle. Everything lands on parity or on this PR being stricter and correct, with one exception.

**Logical assignment is not recognized as a binding.**

| Code | ESLint | This PR |
| --- | --- | --- |
| `test("case", async () => { let p; p \|\|= load().then(v => expect(v).toBe(1)); await p; });` | ok | error |
| `test("case", async () => { let p; p ??= load().then(v => expect(v).toBe(1)); await p; });` | ok | error |

`p ||= chain` cannot leave the chain floating: either the right side is never evaluated, or it is assigned to `p` and the later `await p` consumes it. Details inline. This is narrow enough that leaving it is a defensible call.

One doc nit inline.

internal/utils/test_framework/rules/valid_expect_in_promise/valid_expect_in_promise.go
elecmonkey: Fixed in `b0eed54e`. `bindingFor` now accepts `||=`, `&&=` and `??=`, and
`promiseValueFlowsThroughBinary` carries the value through their right-hand
side, so `await (pending ||= chain)` is consumed too.

The two directions split rather than share the `extends` flag. A logical
assignment only evaluates its right-hand side when the store happens, so the
binding itself is exact. In the other direction the value a candidate left in
the symbol is a promise: truthy and non-nullish, so `||=` and `??=` provably
never replace it and get `extends`, while `&&=` provably always does and kills
like a plain assignment. The `Write` hook skips the overwrite event on the same
two operators.

Jest and Rstest Go and JS regressions cover both directions, including
`pending &&= other` discarding the chain and `count += chain` binding nothing.
Both rule docs describe the boundary.

internal/plugins/jest/rules/valid_expect_in_promise/valid_expect_in_promise.md
elecmonkey: Added in `39d8af8b`, with the version-pinned links from your comment.

swwind (commented):
Re-reviewed after `b0eed54e` and `39d8af8b`. I ran ~50 fresh cases against real `eslint-plugin-jest`, weighted toward the paths the logical-assignment commit touched. The `||=` / `??=` / `&&=` behaviour it adds is right, and every reverse-direction divergence I hit — upstream reporting a chain this PR proves consumed — is covered by the new `.md` paragraph.

Three things left. All three report code upstream accepts and that really is consumed, and all three live in the shared engine, so both plugins are affected.

**1. A binding reached through `||`, `??`, `&&`, `?:` or a comma is lost.**

| Code | ESLint | This PR |
| --- | --- | --- |
| `let cached; test("case", async () => { cached = cached \|\| load().then(v => expect(v).toBe(1)); await cached; });` | ok | error |
| `test("case", async () => { let p = fallback \|\| load().then(v => expect(v).toBe(1)); await p; });` | ok | error |
| `test("case", async () => { let p = fallback ?? load().then(v => expect(v).toBe(1)); await p; });` | ok | error |
| `test("case", async () => { let p = ready && load().then(v => expect(v).toBe(1)); await p; });` | ok | error |
| `test("case", async () => { let p = c ? load().then(v => expect(v).toBe(1)) : other; await p; });` | ok | error |
| `test("case", async () => { let p = (setup(), load().then(v => expect(v).toBe(1))); await p; });` | ok | error |

The first row is `pending ||= chain` written longhand, which is the form the memoisation idiom usually takes.

**2. A floating chain under a compound assignment is reported at the chain instead of at the assignment.**

| Code | ESLint | This PR |
| --- | --- | --- |
| `test("case", async () => { let p; p \|\|= load().then(v => expect(v).toBe(1)); });` | 1:35-76 | 1:41-76 |
| `test("case", async () => { let p = other; p &&= load().then(v => expect(v).toBe(1)); });` | 1:43-84 | 1:49-84 |
| `test("case", async () => { obj.p \|\|= load().then(v => expect(v).toBe(1)); });` | 1:28-73 | 1:38-73 |
| `test("case", async () => { arr[0] \|\|= load().then(v => expect(v).toBe(1)); });` | 1:28-74 | 1:39-74 |
| `test("case", () => { list.push(load().then(v => expect(v).toBe(1))); });` | 1:22-69 | 1:32-67 |
| `test("case", () => { helper(load().then(v => expect(v).toBe(1))); });` | 1:22-66 | 1:29-64 |

Plain `=` matches upstream exactly; only the compound operators and the nested-in-a-call shapes drift. The new `||=` cases assert `messageId` only, which is why this did not surface.

**3. An array held in a binding, and `for await`, are not treated as consumption.**

| Code | ESLint | This PR |
| --- | --- | --- |
| `test("case", async () => { const list = [load().then(v => expect(v).toBe(1))]; await Promise.all(list); });` | ok | error |
| `test("case", async () => { const list = [load().then(v => expect(v).toBe(1))]; for await (const x of list) {} });` | ok | error |
| `test("case", async () => { for await (const x of [load().then(v => expect(v).toBe(1))]) {} });` | ok | error |

Separately, the strict-side divergences are all upstream misses and worth keeping, but several of them are not described anywhere in the docs. One doc note inline.

Details inline.

internal/utils/test_framework/rules/valid_expect_in_promise/valid_expect_in_promise.go
```diff
@@ -0,0 +1,857 @@
...
+	if target := testFramework.StaticBindingForValue(group.root); target != nil {
+		return &chainBinding{
+			target: target,
+			symbol: a.symbolOf(target),
+		}
+	}
+	current := group.root
+	for current != nil && current.Parent != nil {
+		parent := current.Parent
+		if parent.Kind == ast.KindParenthesizedExpression {
```
swwind, line 411: **The binding walk stops at the first non-paren parent.** `isDirectlyConsumed` (462) and `isSafeIdentifierUse` (513) both carry a value up through `||`, `??`, `&&`, a `?:` branch and the right side of a comma, but `bindingFor` walks through parentheses only, so `let p = fallback || load().then(...)` produces no binding at all and the later `await p` has nothing to reach.

Walking the same operators here fixes every row in the table:

```go
if parent.Kind == ast.KindConditionalExpression {
	conditional := parent.AsConditionalExpression()
	if conditional == nil || (conditional.WhenTrue != current && conditional.WhenFalse != current) {
		return nil
	}
	current = parent
	continue
}
if !ast.IsAssignmentExpression(parent, false) && promiseValueFlowsThroughBinary(parent, current) {
	current = parent
	continue
}
```

The `IsAssignmentExpression` guard keeps an assignment parent on the existing branch below, so `p ||= chain` still binds through this function's own handling instead of being walked past. I applied this locally: all six rows become valid, and both Go rule suites stay green.

internal/utils/test_framework/rules/valid_expect_in_promise/valid_expect_in_promise.go
```diff
@@ -0,0 +1,857 @@
...
+	current := root
+	for current != nil && current.Parent != nil && current.Parent != function {
+		parent := current.Parent
+		switch parent.Kind {
+		case ast.KindParenthesizedExpression:
+			current = parent
+		case ast.KindExpressionStatement, ast.KindVariableDeclaration:
+			return parent
+		default:
+			if ast.IsAssignmentExpression(parent, true) {
```
swwind: **`IsAssignmentExpression(parent, true)` excludes the operators this commit just taught the rule to bind.** The `true` restricts it to plain `=`, so a chain stored by `||=`, `??=` or `&&=` falls through to `return current` and the diagnostic starts at the chain. Matching the operator set `bindingFor` (399) accepts keeps the two in step:

```go
if ast.IsAssignmentExpression(parent, false) {
	operator := parent.AsBinaryExpression().OperatorToken.Kind
	if operator == ast.KindEqualsToken || ast.IsLogicalOrCoalescingAssignmentOperator(operator) {
		return parent
	}
}
```

The last two rows of the table are the same function stopping early for a different reason: upstream reports the enclosing `ExpressionStatement` for `list.push(chain);` and `helper(chain);`, while this returns `current`. Continuing the walk past an intervening expression parent up to the `ExpressionStatement` / `VariableDeclaration` covers those as well.

internal/utils/test_framework/rules/valid_expect_in_promise/valid_expect_in_promise.go
```diff
@@ -0,0 +1,857 @@
...
+		switch parent.Kind {
+		case ast.KindParenthesizedExpression:
+			current = parent
+		case ast.KindAwaitExpression:
+			return testFramework.AbruptCompletionPropagatesFailure(parent, function)
+		case ast.KindReturnStatement:
+			return testFramework.AbruptCompletionPropagatesFailure(parent, function)
+		case ast.KindArrowFunction:
+			return parent.AsArrowFunction().Body == current
+		case ast.KindArrayLiteralExpression:
```
swwind, line 538: **An array literal counts as a sink only while it is still the sole argument of `Promise.all`.** `safePromiseAggregatorForExpression` (622) requires `arrayNode.Parent` to be that call, so storing the array first breaks the link even though the chain is awaited:

```ts
const list = [load().then(value => expect(value).toBe(1))];
await Promise.all(list);
```

A promise in a binding is tracked, an array of promises in a binding is not, so the group has no path to consumption.

`for await` is missing from both walkers as well — neither `isDirectlyConsumed` nor `isSafeIdentifierUse` has a `KindForInOrOfStatement` case — so `for await (const x of [chain]) {}` reports although the loop awaits every element and its rejection leaves the test callback.

internal/plugins/jest/rules/valid_expect_in_promise/valid_expect_in_promise.md
```diff
@@ -0,0 +1,38 @@
...
+
+A logical assignment stores the chain like a plain assignment does, and its right-hand side is only evaluated when the store happens, so `pending ||= load().then(...)` followed by `await pending` is consumed. In the other direction a promise is truthy and non-nullish, so a later `||=` or `??=` cannot replace the promise a binding already holds, while `&&=` always does and discards it.
+
+An awaited chain is safe only when its rejection can escape the test callback. An enclosing `catch` that may swallow the rejection, or a `finally` block that may `return`, `break`, or `continue`, does not count as safe consumption. A `catch` that always rethrows preserves the failure.
+
+Jest callbacks using `done` are not analyzed because the callback can coordinate promise completion explicitly. Named callbacks and promise-handler callbacks are analyzed when their relationship to a test is statically known.
+
+This implementation intentionally fixes upstream cases where unrelated assignments stop promise tracking, named callbacks are missed, nested handlers are missed, or `Promise.reject` / `Promise.allSettled` are incorrectly accepted.
```
swwind, line 37: **The strict side of the port is only partly described.** The docs name `Promise.resolve` / `Promise.all` as safe and `Promise.reject` / `Promise.allSettled` as unsafe, but these also report here while upstream is silent, and none of them is mentioned:

- `await Promise.race([chain])` and `await Promise.any([chain])` — upstream is inconsistent about these itself, it reports the `Promise.race([pending])` variable form
- `await helper(chain)` — any wrapper other than `Promise.resolve` / `Promise.all`, same variable-form inconsistency upstream
- `void chain;` and `throw chain;`
- `const o = { p: chain };` and `const list = [chain];`
- `` tag`${chain}` ``
- `count += chain` — a compound arithmetic assignment stores no promise

One sentence — a chain is consumed only when it is awaited, returned, wrapped in `Promise.resolve` / `Promise.all`, or handed to an `expect(...).resolves` / `.rejects` sink, and every other position is floating including the ones upstream misses — covers the whole group. Both `.md` files.

internal/utils/test_framework/rules/valid_expect_in_promise/valid_expect_in_promise.go
elecmonkey: Fixed in `cd6a151b`. `bindingFor` now walks the same operators the value walkers do: `||`, `??`, `&&`, either branch of a conditional, and the right-hand side of a comma, with the `IsAssignmentExpression` guard keeping an assignment parent on the branch below. All six rows are valid now, with Jest and Rstest regressions for each.

The same hole existed in the other direction. `p = p || chain` is the longhand of `p ||= chain`, but the write still killed an earlier candidate for `p`. `writePreservesPreviousValue` covers both now: a plain assignment whose right-hand side can evaluate to the value the symbol already holds does not kill it, while `p = p && other` still does, because a promise is truthy.

internal/utils/test_framework/rules/valid_expect_in_promise/valid_expect_in_promise.go
elecmonkey: Fixed in `cd6a151b`. `promiseReportNode` accepts `=` and the logical assignment operators, and walks up through call and member parents to the enclosing `ExpressionStatement` / `VariableDeclaration`, so all six rows match the ranges you measured. The committed `||=` case now pins `Line` / `Column` / `EndColumn`, along with `obj.pending ||=`, `list.push(chain)` and `helper(chain)`.

One consequence: two chains in the same statement now share a report node, so I dropped the report-node de-duplication. They are two floating promises and upstream reports both. Nothing else in either suite changed.

internal/utils/test_framework/rules/valid_expect_in_promise/valid_expect_in_promise.go
elecmonkey: Fixed in `cd6a151b`. A binding that holds the chain inside an array literal is tracked, and marked as consumed only element-wise; the consume event carries the same flag, so `await Promise.all(list)` and `for await (const settled of list)` consume it and `await list` does not, because awaiting an array does not await its elements. That last one stays reported.

`for await` is a consumption in both walkers, including the literal form `for await (const settled of [chain])`, and `AbruptCompletionPropagatesFailure` now treats it like an await, so a `try` / `catch` around the loop still suppresses. `Promise.all(value)` is a sink only in that element-wise sense: `Promise.all(p)` on a binding that holds a single promise is not consumption and is unchanged.

internal/plugins/jest/rules/valid_expect_in_promise/valid_expect_in_promise.md
elecmonkey: Added in `f931ca11`. Both docs carry one sentence for the whole group: a chain is consumed only when it is awaited, returned, wrapped in `Promise.resolve` or `Promise.all`, consumed element-wise by `for await`, or handed to a `resolves` / `rejects` sink, and every other position leaves it floating — `Promise.race` and `Promise.any`, any other wrapper call such as `helper(chain)`, `void chain` and `throw chain`, a property of an object literal that is not destructured straight into a binding, an array literal never consumed element-wise, a tagged template interpolation, and `count += chain`.

The Jest doc says which of those the upstream rule does not report. The Rstest doc states the same boundary without the comparison, since the Rstest rule has no upstream to differ from. The PR description is split the same way.

internal/utils/test_framework/rules/valid_expect_in_promise/valid_expect_in_promise.go
```diff
@@ -0,0 +1,1010 @@
...
+		return false
+	}
+	return false
+}
+
+// assignmentPreservesSymbol reports whether value, assigned to symbol, can
+// evaluate to the value symbol already holds. `p = p || chain` keeps a promise
+// p holds, because a promise is truthy, so the write must not kill an earlier
+// candidate for p.
+func (a *analyzer) assignmentPreservesSymbol(value *ast.Node, symbol *ast.Symbol) bool {
```
swwind, line 687: **Only the left operand of `||` / `??` preserves the old value.** The binding direction and this one are not symmetric: `x || chain` binds the chain because the chain is evaluated only on the paths where it is stored, but `p = x || p` re-reads a promise that already exists, and a truthy `x` drops it. The same for `x ?? p`, for `ready && p` where a falsy `ready` becomes the value, and for a conditional where only one branch is `p`.

```go
	case ast.KindConditionalExpression:
		conditional := value.AsConditionalExpression()
		return conditional != nil &&
			a.assignmentPreservesSymbol(conditional.WhenTrue, symbol) &&
			a.assignmentPreservesSymbol(conditional.WhenFalse, symbol)
	...
		case ast.KindBarBarToken, ast.KindQuestionQuestionToken:
			return a.assignmentPreservesSymbol(binary.Left, symbol)
		case ast.KindCommaToken:
			return a.assignmentPreservesSymbol(binary.Right, symbol)
```

`&&` drops out entirely, because neither operand survives it. I applied this locally: all four rows report, `p = p || other`, `p = p ?? other`, `p = (setup(), p)`, `p = c ? p : p` and the committed `pending = pending || chain` binding stay valid, and both rule suites are green.

internal/utils/test_framework/rules/valid_expect_in_promise/valid_expect_in_promise.go
```diff
@@ -0,0 +1,1010 @@
...
+		switch parent.Kind {
+		case ast.KindParenthesizedExpression:
+			current = parent
+		case ast.KindAwaitExpression:
+			return testFramework.AbruptCompletionPropagatesFailure(parent, function), false
+		case ast.KindReturnStatement:
+			return testFramework.AbruptCompletionPropagatesFailure(parent, function), false
+		case ast.KindForOfStatement:
+			return forAwaitConsumesExpression(current, function), true
+		case ast.KindArrayLiteralExpression:
```
swwind, line 598: **A spread breaks the link to the array.** This case is reached only when the identifier is a direct element of the literal; a `SpreadElement` parent falls through to `default` and returns `false`, so `await Promise.all([...list])` reports although every element is awaited. `[...list, b()]` and `for await (const x of [...list])` are the same shape. Walking a `SpreadElement` up to its array literal and keeping `aggregate: true` from there covers all three.

`await list[0]` is the narrower mirror image: the `KindElementAccessExpression` case (616) walks up to the await and returns `aggregate: false`, so it never matches an aggregate binding even though it does reach one element.

swwind (commented):
Re-reviewed after `cd6a151b` and `f931ca11`.

All three findings from the last round are fixed. The six binding rows, the six position rows and the three array rows now match `eslint-plugin-jest@29.16.1` exactly, ranges included. Dropping the report-node de-duplication is the right call: `helper(a().then(…), b().then(…))` really is two floating promises and upstream reports both at that range. I ran ~60 fresh cases weighted toward the new binding walk, the `aggregate` flag and the `for await` sink; every remaining divergence is this PR being stricter and is covered by the new `.md` paragraph.

Two things left, both in the shared engine, so both plugins are affected.

**1. `assignmentPreservesSymbol` accepts operands that cannot preserve the old value.** A write is skipped when the right-hand side can evaluate to what the binding already holds, but four of the operand positions it accepts drop that value on one path, so a floating chain goes unreported.

| Code | ESLint | This PR |
| --- | --- | --- |
| `test("case", async () => { let p = load().then(v => expect(v).toBe(1)); p = c ? p : other; await p; });` | error | ok |
| `test("case", async () => { let p = load().then(v => expect(v).toBe(1)); p = other \|\| p; await p; });` | error | ok |
| `test("case", async () => { let p = load().then(v => expect(v).toBe(1)); p = other ?? p; await p; });` | error | ok |
| `test("case", async () => { let p = load().then(v => expect(v).toBe(1)); p = ready && p; await p; });` | error | ok |

**2. An array binding is consumed only when the identifier is named directly.** A spread or an index between the binding and the sink loses it.

| Code | ESLint | This PR |
| --- | --- | --- |
| `test("case", async () => { const list = [load().then(v => expect(v).toBe(1))]; await Promise.all([...list]); });` | ok | error |
| `test("case", async () => { const list = [load().then(v => expect(v).toBe(1))]; await Promise.all([...list, b()]); });` | ok | error |
| `test("case", async () => { const list = [load().then(v => expect(v).toBe(1))]; for await (const x of [...list]) {} });` | ok | error |
| `test("case", async () => { const list = [load().then(v => expect(v).toBe(1))]; await list[0]; });` | ok | error |

Details inline.

elecmonkey:
Both fixed in 66ae690a: value preservation now requires every path through the right-hand side to evaluate to the old promise, so the conditional and the four `||` / `??` / `&&` rows report; and an array binding is now reached through a spread into another literal and through an element access, so the four spread and index rows are consumed.

Full jest and rstest suites are green, with the eight cases added to both rule test files.

internal/utils/test_framework/rules/valid_expect_in_promise/valid_expect_in_promise.go
```diff
@@ -0,0 +1,1036 @@
...
+				})
+			}
+		},
+		Write: func(builder *cfg.Builder[flowEvent], node *ast.Node) {
+			symbol := a.symbolOf(node)
+			if symbol == nil {
+				return
+			}
+			if boundGroups := groupsByTarget[node]; len(boundGroups) > 0 {
+				for _, group := range boundGroups {
```
swwind: `const pending = [chainA, chainB]` binds both groups to the same `pending` identifier, so this loop emits two `flowBind` events for one write. `extends` is false on both, so the second event takes the kill branch in `runDataflow` and marks the first group unsafe before recording itself — only the last chain in the literal is left alive, and `await Promise.all(pending)` reports every earlier one. Only the first bind at a target replaces what the symbol held:

```go
for index, group := range boundGroups {
	builder.Emit(flowEvent{
		kind:      flowBind,
		symbol:    symbol,
		candidate: group,
		extends:   group.binding.extends || index > 0,
	})
}
```

swwind (commented):
Re-reviewed after `66ae690a`.

Both findings from the last round are fixed. The four preservation rows report and the four spread / element-access rows are consumed, ranges included. I ran ~55 fresh cases against `eslint-plugin-jest@29.16.1`, weighted toward the narrowed `assignmentPreservesSymbol` and the widened array reach: the preserving direction still holds (`pending = pending || other`, `pending = (setup(), pending)`, `pending = condition ? pending : pending`, `pending ||= other`), and every reverse-direction divergence it creates is covered by the `.md` paragraph.

Two things left, both in the shared engine, so both plugins are affected.

**1. Only the last chain in an array binding survives.** Every chain in the literal binds to the same identifier, and each bind kills the candidates already recorded for that symbol — including the siblings bound by the same write — so all but the last are reported even though the binding is consumed.

| Code | ESLint | This PR |
| --- | --- | --- |
| `test("case", async () => { const pending = [a().then(v => expect(v).toBe(1)), b().then(v => expect(v).toBe(2))]; await Promise.all(pending); });` | ok | error on the first chain |
| `test("case", async () => { const pending = [a().then(v => expect(v).toBe(1)), b().then(v => expect(v).toBe(2))]; for await (const settled of pending) {} });` | ok | error on the first chain |
| `test("case", async () => { const pending = [a().then(v => expect(v).toBe(1)), b().then(v => expect(v).toBe(2))]; await Promise.all([...pending]); });` | ok | error on the first chain |
| `test("case", async () => { const pending = [a().then(v => expect(v).toBe(1)), b().then(v => expect(v).toBe(2)), c().then(v => expect(v).toBe(3))]; await Promise.all(pending); });` | ok | error on the first two |

A list of assertion chains awaited with `Promise.all` is the shape the array-binding support was added for, and a one-element list is the only case the suite covers.

**2. `for await` over an array literal holding the binding is not consumption.**

| Code | ESLint | This PR |
| --- | --- | --- |
| `test("case", async () => { const pending = load().then(v => expect(v).toBe(1)); for await (const settled of [pending]) {} });` | ok | error |

`await Promise.all([pending])` is accepted in that same position, and `for await (const settled of [load().then(...)])` without the binding is one of the rule's own valid cases.

Both are one-line changes; with them the rows above match upstream and the jest and rstest suites stay green. Details inline.

internal/utils/test_framework/rules/valid_expect_in_promise/valid_expect_in_promise.go
```diff
@@ -0,0 +1,1036 @@
...
+		switch parent.Kind {
+		case ast.KindParenthesizedExpression:
+			current = parent
+		case ast.KindAwaitExpression:
+			return testFramework.AbruptCompletionPropagatesFailure(parent, function), viaElement
+		case ast.KindReturnStatement:
+			return testFramework.AbruptCompletionPropagatesFailure(parent, function), viaElement
+		case ast.KindForOfStatement:
+			return forAwaitConsumesExpression(current, function), true
+		case ast.KindArrayLiteralExpression:
```
swwind, line 598: `isDirectlyConsumed` and the new `KindSpreadElement` case both accept an array literal that a `for await` iterates; this one accepts only `Promise.all([...])`, so the loop stops consuming the moment the chain moves into a binding.

```go
if safePromiseAggregatorForExpression(parent, current, function) ||
	forAwaitConsumesExpression(parent, function) {
	return true, false
}
```

elecmonkey:
Both fixed in 84448c8c: a write that binds several chains now only lets the first replace what the symbol held, so the siblings in a list survive and none of the multi-element rows report; and an array literal iterated by `for await` now consumes a binding placed in it, matching `Promise.all` in that position.

Full jest and rstest suites are green, with the five cases added to both rule test files.

swwind (approved):
Nothing from here now, LGTM

PR #1698, opened 2026-08-12
feat(typescript): port rule switch-exhaustiveness-check

swwind:
## Summary

Port the `switch-exhaustiveness-check` rule from ESLint to rslint.

Requires switch statements over union types (discriminated unions, enums) to be exhaustive, reporting missing cases and optionally requiring or forbidding a `default` clause depending on configuration.

## Related Links

- ESLint rule: https://typescript-eslint.io/rules/switch-exhaustiveness-check
- Source code: https://github.com/typescript-eslint/typescript-eslint/blob/v8.67.0/packages/eslint-plugin/src/rules/switch-exhaustiveness-check.ts

## Checklist


internal/plugins/typescript/rules/switch_exhaustiveness_check/switch_exhaustiveness_check.go
```diff
@@ -26,63 +33,478 @@ var SwitchExhaustivenessCheckRule = rule.CreateRule(rule.Rule{
...
+	var missing []*checker.Type
+	for _, unionPart := range utils.UnionTypeParts(discriminantType) {
+		for _, part := range utils.IntersectionTypeParts(unionPart) {
+			if caseTypes[part] || !isTypeLiteralLikeType(part) {
+				continue
+			}
+			if hasUndefinedCase && utils.IsTypeFlagSet(part, checker.TypeFlagsUndefined) {
+				continue
+			}
+			missing = append(missing, part)
```
elecmonkey, line 200: **[P2] Preserve upstream union constituent order**

For unions whose checker order differs from TypeScript's, such as the declared `Day = 'Monday' | 'Tuesday' | ...` case, iterating `utils.UnionTypeParts` directly produces `"Friday" | ... | "Tuesday"` instead of upstream's declaration order. This changes both the diagnostic text and generated case order; the PR currently accommodates it by modifying the migrated upstream expectations, but an ESLint v8.67.0 run still emits `"Tuesday" | "Wednesday" | ...`. Normalize these constituents to the upstream TypeScript order rather than accepting ts-go's differing order.

internal/plugins/typescript/rules/switch_exhaustiveness_check/switch_exhaustiveness_check.go
```diff
@@ -26,63 +33,478 @@ var SwitchExhaustivenessCheckRule = rule.CreateRule(rule.Rule{
...
+			checker.TypeFormatFlagsUseFullyQualifiedType,
+		nil,
+	)
+}
+
+// requiresQuoting reports whether name needs bracket-quoting
+// (`Enum['a-b']`) rather than dot access (`Enum.ab`) when read back as a
+// property name.
+func requiresQuoting(name string) bool {
+	return !scanner.IsValidIdentifier(name)
```
elecmonkey: **[P2] Match upstream quoting for astral enum member names**

When an enum member starts with an astral Unicode identifier character, for example `enum Astral { '𐐀' = 1 }`, this rune-based validity check considers the name safe and suggests `case Astral.𐐀`. Upstream v8.67.0's target-aware `requiresQuoting` examines UTF-16 code units and instead suggests `case Astral['𐐀']`, so the observable suggestion is not behavior-compatible. Port the upstream target-aware identifier check, including its UTF-16 semantics.

internal/plugins/typescript/rules/switch_exhaustiveness_check/switch_exhaustiveness_check.go
elecmonkey: This difference stems from ts-go and typescript. Please ignore this review.

internal/plugins/typescript/rules/switch_exhaustiveness_check/switch_exhaustiveness_check.go
swwind: fixed

elecmonkey (approved):
Reviewed

fansenze (approved):

PR #1701, opened 2026-08-12
feat(rstest): add no-standalone-expect rule

elecmonkey:
## Summary

Port `eslint-plugin-jest`'s `no-standalone-expect` to rstest: report assertions that appear outside a test block, such as at module scope or directly inside a `describe` callback, where they run during module evaluation or test collection rather than as part of a test case. Assertions inside helper functions remain allowed.

The rule consumes the shared rstest call analysis (`GetRstestCallAnalysis` / `ParseFnCall` / `ParseExpectCall`) without changing the parser; added to the rstest `recommended` preset at `error`.

## Behavior change in `jest/no-standalone-expect`

The shared traversal also corrects existing behavior in the Jest rule. Nested callback-less registrations such as `test.todo("inner")` and `test("inner", callback)` no longer close the enclosing test scope, so assertions later in the outer test remain valid. Tagged-template scopes now cover the callback of the call returned by the tag, and arrow-function scope pushes and pops use the same predicate.

As part of the same scope-balancing correction, an assertion anywhere in a registration's own arguments is no longer reported by the Jest rule, matching `eslint-plugin-jest`, whose scope stack covers the whole call. That covers a registration with no callback, such as `test.todo(expect(1).toBe(1))`, and one that has a callback but carries the assertion in another argument, such as `test(expect(1).toBe(1), () => {})`, `test("case", { timeout: expect(1).toBe(1) }, () => {})` and `test("case", () => {}, expect(2).toBe(2))`.

## Intentional differences from `eslint-plugin-jest`

- Method, constructor, getter, and setter bodies are treated as helper-function scopes, so assertions inside them are allowed.
- A standalone assertion after a chained registration such as `test.only("case", () => {})` is still reported. The upstream rule can leave its registration scope open and miss that assertion.
- Known negated asymmetric matcher constructors such as `expect.not.stringContaining("value")` are treated as static value constructors and allowed outside test blocks.

## Differences from jest

rstest's registration and assertion APIs have a wider surface than jest, so the rule covers the framework-specific forms through the shared analysis:

- **test and suite registrations include aliases, `.each` / `.for`, `test.extend`, all supported callback overloads, `import.meta.rstest`, and `@rstest/playwright`**.
- **expect provenance covers globals, imports and aliases, `require`, namespace APIs, `import.meta.rstest`, TestContext `expect`, and `@rstest/playwright`**.
- **Chai property assertions are recognized**, including forms such as `expect(value).to.be.ok`.
- **static matcher constructors remain allowed outside test blocks**, including known negated asymmetric matchers such as `expect.not.stringContaining(...)`; assertion-count APIs such as `expect.assertions(...)` and `expect.hasAssertions()` are still reported.
- **a registration's arguments sit outside the test case it registers**. rstest evaluates a test's name, its options object and any other argument while collecting the file, so an assertion written there is reported — including on a registration with no callback at all, such as `test.todo(expect(1).toBe(1))`, which the jest rule allows to match `eslint-plugin-jest`. A tagged-template registration is the exception: `` test.each`...`(name, callback) `` counts as a test block in its entirety, so an assertion in an interpolation or in an argument of the call the tag returns is allowed.

## Related Links

Related #935

## Checklist


swwind (commented):
Behaviour matches `jest/no-standalone-expect` on the shared cases I diffed against `eslint-plugin-jest`, but the scope stack is pushed conditionally and popped unconditionally, so the enclosing `test` scope is dropped whenever a nested test call has no inline callback. Every row below was run against `eslint-plugin-jest` with the same rule options.

| Code | ESLint | This PR |
| --- | --- | --- |
| `test("outer", () => { test.todo("inner"); expect(1).toBe(1); });` | no error | `unexpectedExpect` |
| `test("outer", () => { test("inner", cb); expect(1).toBe(1); });` | no error | `unexpectedExpect` |
| `describe("d", () => { test("outer", () => { test.todo("i"); expect(1).toBe(1); }); });` | no error | `unexpectedExpect` |
| `test.each([1])("outer", value => { test.todo("i"); expect(value).toBe(1); });` | no error | `unexpectedExpect` |
| `test("outer", () => { t("inner"); expect(1).toBe(1); });` with `additionalTestBlockFunctions: ["t"]` | no error | `unexpectedExpect` |
| `test.todo(expect(1).toBe(1));` | no error | `unexpectedExpect` |
| `test("outer", () => { expect(1).toBe(1); test.todo("inner"); });` | no error | no error |
| ``myEach`a ${1}`("t", () => { expect(1).toBe(1); });`` | no error | `unexpectedExpect` |

The second-to-last row is a control: moving the `expect` before the nested call makes the report disappear, which is the push/pop asymmetry. The last row is a separate issue in the tagged-template scope range.

Two divergences that look deliberate, just noting them:

- `getBlockType` treats `MethodDeclaration`/`Constructor`/`GetAccessor`/`SetAccessor` as a `function` scope, so `class C { m() { expect(1).toBe(1); } }` is accepted here while `eslint-plugin-jest` reports it.
- Popping on exit also closes the upstream stack leak caused by its `callee.type !== MemberExpression` guard, e.g. `describe("d", () => { test.only("a", () => {}); expect(1).toBe(1); })` is missed upstream and reported here. That reads as an improvement.

internal/plugins/rstest/rules/no_standalone_expect/no_standalone_expect.go
```diff
@@ -0,0 +1,269 @@
...
+						ctx.ReportNode(node, buildUnexpectedExpectMessage())
+					}
+					return
+				}
+
+				if isTestBlockCall(node) {
+					if start, end, ok := getTestScopeRange(node); ok {
+						pushScope(blockTypeTest, start, end)
+					}
+				}
```
swwind: The exit handler pops a `test` scope for any `isTestBlockCall(node)`, but this push only happens when `getTestScopeRange` finds an argument with a function body. A nested test call without an inline callback (`test.todo("i")`, `test("i", cb)`, or a custom `t("i")`) therefore pops the *enclosing* test's scope, and every `expect` after it in the same body is reported.

Pushing unconditionally keeps both sides symmetric, falling back to the call node's own range:

```go
if isTestBlockCall(node) {
	start, end, ok := getTestScopeRange(node)
	if !ok {
		start, end = node.Pos(), node.End()
	}
	pushScope(blockTypeTest, start, end)
}
```

That also brings `test.todo(expect(1).toBe(1))` in line with upstream, which does not report it.

internal/plugins/rstest/rules/no_standalone_expect/no_standalone_expect.go
```diff
@@ -0,0 +1,269 @@
...
+}
+
+func getTemplateScopeRange(node *ast.Node) (int, int, bool) {
+	callExpr := node.AsCallExpression()
+	if callExpr == nil || callExpr.Expression == nil || callExpr.Expression.Kind != ast.KindTaggedTemplateExpression {
+		return 0, 0, false
+	}
+
+	return callExpr.Expression.Pos(), callExpr.Expression.End(), true
+}
```
swwind: The template scope only spans `callExpr.Expression`, i.e. the `` tag`...` `` part, so it never covers the callback of the call being tagged. Upstream keeps `template` on the stack for the whole call expression, which is what makes an `expect` inside a non-test tagged-template call legal:

```js
myEach`
  a
  ${1}
`("t", () => { expect(1).toBe(1); });
```

`eslint-plugin-jest` accepts this, this PR reports `unexpectedExpect`. Returning `node.Pos(), node.End()` here matches upstream. Low impact, since `additionalTestBlockFunctions` is a workaround.

internal/plugins/rstest/rules/no_standalone_expect/no_standalone_expect.md
```diff
@@ -0,0 +1,54 @@
+# no-standalone-expect
```
swwind, line 1: `scripts/gen-rule-manifest.js` resolves docs as `internal/plugins/<plugin>/rules/<rule_dir>/<rule_dir>.md`, so this needs to be `no_standalone_expect.md` — every other rstest rule doc uses the underscore name. As it stands, `node scripts/gen-rule-manifest.js` emits `"docPath": null` for `rstest:no-standalone-expect` and the docs page is not generated.

internal/plugins/rstest/rules/no_standalone_expect/no_standalone_expect.go
elecmonkey: Fixed in `43454a8f`. Test scopes now always push when the matching exit handler can pop, falling back to the call node’s range when no inline callback body is available. I also added regression coverage for nested registrations and `test.todo(expect(...))` in both the jest and rstest rules.

internal/plugins/rstest/rules/no_standalone_expect/no_standalone_expect.go
elecmonkey: Fixed in `d49c1111`. The template scope now spans the whole call expression, with regression coverage in both the jest and rstest rules.

internal/plugins/rstest/rules/no_standalone_expect/no_standalone_expect.md
elecmonkey: Fixed in `242d6e46`. The file is now named `no_standalone_expect.md`, and the generated manifest resolves its `docPath` correctly.

swwind (commented):
Both issues from the last pass are fixed and verified. `test.todo` / callback-less registrations no longer close the enclosing test scope, and the tagged-template scope now spans the whole call. The jest rule inherits both through the shared traversal, so `` myEach`…`('t', () => { expect… }) `` and the `test.only` case behave correctly there too. Go suites for all three packages pass.

Since `a9e554a9` puts jest on the shared traversal, I re-ran a differential against `eslint-plugin-jest` v29.16.0 for the jest rule. Every row below was run with the same rule options.

**None of these are introduced by this PR** — I checked each against `main` and the behaviour is identical there. But the shared traversal is now the thing that decides them for both plugins, so this seems like the moment to either fix them or write them down.

## False positives (rslint reports, upstream does not)

| Code | eslint-plugin-jest | rslint jest |
| --- | --- | --- |
| `test('n', expect(1).toBe(1), () => {});` | no error | `unexpectedExpect` |
| `test(expect(1).toBe(1), () => {});` | no error | `unexpectedExpect` |
| `test('n', { timeout: expect(1).toBe(1) }, () => {});` | no error | `unexpectedExpect` |
| `describe('d', () => { test('t', expect(1).toBe(1), () => {}); });` | no error | `unexpectedExpect` |
| `describe('d', () => { test.only('a', () => {}); expect(1).toBe(1); });` | no error | `unexpectedExpect` |

Rows 1-4 are one cause: `getTestScopeRange` narrows the `test` scope to the callback body, so an assertion in any *other* argument falls outside it. Upstream pushes `'test'` on enter and reads top-of-stack, so its scope covers the whole call subtree regardless of which argument the assertion sits in.

Row 5 is the symmetric pop closing upstream's `callee.type !== MemberExpression` stack leak, already noted in the package doc.

## False negatives (upstream reports, rslint does not)

| Code | eslint-plugin-jest | rslint jest |
| --- | --- | --- |
| `class C { m() { expect(1).toBe(1); } }` | `unexpectedExpect` | no error |
| `const o = { m() { expect(1).toBe(1); } };` | `unexpectedExpect` | no error |
| `expect.not.stringContaining('x');` | `unexpectedExpect` | no error |

Rows 1-2 are `getBlockType`'s `MethodDeclaration` / `Constructor` / `GetAccessor` / `SetAccessor` case, already noted in the package doc.

Row 3 has a different cause, outside this PR: `utils.IsStaticExpectMatcher` has no equivalent of upstream's `members.length === 1`, so any two-member static chain is treated as a value constructor. Worth a separate issue rather than something to solve here.

## Suggestions

For rows 1-4 of the first table, the fix is one line in the shared traversal — push the whole call instead of the callback body, which is the same reasoning the template branch three lines below already carries ("the scope spans the whole call, not just the tagged template"):

```go
if isTestBlockCall(node) {
    pushScope(blockTypeTest, node.Pos(), node.End())
}
```

I tried it locally: all four rows align with upstream, every other differential row is unchanged, both the jest and rstest suites stay green, and `getTestScopeRange` becomes dead code.

One caveat if you take it. That widening is right for jest but wrong for rstest: arguments are evaluated eagerly, so `expect(1).toBe(1)` in `test('n', expect(1).toBe(1), () => {})` really does run at collection time rather than as part of the test case, which is exactly what the rule exists to catch. So the range would have to become a `Runtime` field — whole node for jest, callback body for rstest — rather than a shared constant. Everything else in the traversal (push/pop symmetry, stack handling, describe/arrow/template) stays shared; only this one decision differs.

Given that, documenting is the lighter option for most of it, and this rule's README is missing the chapter the repo already uses elsewhere (`expect_expect.md` has `### Divergence from eslint-plugin-jest`, `no_unneeded_async_expect_function.md` has `## Differences from ESLint`, both placed before `## Original Documentation`). Rows 5-7 in particular are cases where rslint is arguably more correct than upstream and you'd want to keep the behaviour anyway — they're currently written only in the Go package doc of `internal/utils/test_framework/...`, which rule users never see.

## One unflagged behaviour change

`test.todo(expect(1).toBe(1))` is reported by the jest rule on `main` and allowed after this PR. That's the fallback range doing its job and it moves jest *toward* upstream, which also allows it — so no divergence to document. But it is a behaviour change to a shipped rule and the PR description doesn't mention it; worth a line there.

For rstest that same row is arguably wrong in the other direction, for the eager-evaluation reason above, and `no_standalone_expect_test.go:85-87` justifies it with "which is where upstream also allows an assertion" — which is a reason for jest but not for rstest. If the `Runtime` field above lands, that case flips to invalid for rstest and the comment can go.

elecmonkey:
Documented the intentional Jest divergences in `42354545`, and updated the PR body to describe the shipped Jest behavior changes. The framework-specific test-scope policy remains intentionally deferred.

internal/plugins/jest/rules/no_standalone_expect/no_standalone_expect.md
```diff
@@ -54,6 +54,23 @@ expect.extend({});
...
+scopes. An `expect` inside one of those bodies is therefore allowed, just like
+an assertion inside a function declaration or arrow-function helper. The
+upstream rule reports method and accessor bodies.
+
+rslint also balances the test scope opened by every recognized registration.
+As a result, a standalone assertion after a chained registration such as
+`test.only('case', () => {})` is still reported. The upstream rule can leave the
+registration scope open and miss that assertion.
+
+Known negated asymmetric matcher constructors, such as
```
swwind: The behaviour is broader than this paragraph. `utils.IsStaticExpectMatcher` returns true for any matcher other than `assertions`/`hasAssertions` as long as the head has a member-access parent, with no equivalent of upstream's `members.length === 1`, so every two-member chain is allowed — `expect.resolves.toBe(1)` and `expect.rejects.toThrow()` at module scope are accepted here and reported upstream. The rstest binding avoids this by keying off a known asymmetric-matcher list.

None of it is introduced by this PR, but the paragraph is, so either widen the sentence to say every `expect.<modifier>.<matcher>()` chain is treated as a value constructor, or tighten the binding to match upstream — the guard is one line and `IsStaticExpectMatcher` has no other caller:

```go
if len(parsed.Modifiers) == 0 && utils.IsStaticExpectMatcher(parsed.Matcher, parsed.Head.Local.Node) {
```

That aligns jest with upstream and removes the paragraph entirely, at the cost of reporting `expect.not.stringContaining('value')`, which jest genuinely supports and which is currently a valid case at `no_standalone_expect_test.go:20`. The doc fix is the safer call if you want to keep that form allowed.

internal/plugins/rstest/rules/no_standalone_expect/no_standalone_expect.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Rule Details
+
+Disallow using `expect` outside of `it` or `test` blocks. This rule reports `expect` calls that sit directly in a `describe` block, at module scope, or in other places where Rstest will not run them as part of a test case. This helps catch assertions that look meaningful but never execute as part of a test.
+
+`expect` inside a helper function is allowed, even when the helper is defined outside the `it`/`test` callback, because the assertion still runs when the helper is invoked from a test. Static `expect` APIs such as `expect.any()` and `expect.extend()` at module scope are also allowed.
+
+Only the callback body of a registration counts as the test case. A test's name, its options object and any other argument are evaluated while Rstest collects the file, so an assertion written there runs before any case starts and is reported — including on a registration that takes no callback at all, such as `test.todo(expect(1).toBe(1))`.
```
swwind: Right for the plain call forms, but not for the tagged-template one: the template scope spans the whole call, so both the interpolation and the arguments of the call it returns sit inside a test block.

```js
test.each`
  value
  ${expect(1).toBe(1)}
`('case', ({ value }) => {});

test.each`
  value
  ${1}
`(expect(1).toBe(1), ({ value }) => {});
```

Neither is reported. That matches upstream's template handling and the form is rare, so I'd narrow the sentence rather than touch the traversal.

swwind (commented):
Third pass, focused on `1b54151f` and `8c00e9e0`. All three threads from the earlier reviews are closed and verified: a nested callback-less registration keeps the enclosing test scope, the tagged-template scope spans the whole call, and the doc filename resolves in the manifest. The `ArgumentsOutsideTestScope` split reads well — the stack stays balanced on both bindings and only the fallback range differs, which is exactly where the two frameworks disagree.

I re-ran a differential against `eslint-plugin-jest` v29.16.0 and probed the new rstest policy locally.

| Code | eslint-plugin-jest | rslint jest | rslint rstest |
| --- | --- | --- | --- |
| `test.todo(expect(1).toBe(1));` | no error | no error | `unexpectedExpect` |
| `test('c', { timeout: expect(1).toBe(1) }, () => {});` | no error | no error | `unexpectedExpect` |
| `test('c', () => {}, expect(1).toBe(1));` | no error | no error | `unexpectedExpect` |
| `test('outer', () => { test.todo(expect(1).toBe(1)); });` | no error | no error | no error |
| `expect.not.stringContaining('v');` | `unexpectedExpect` | no error | no error |
| `expect.resolves.toBe(1);` | `unexpectedExpect` | no error | `unexpectedExpect` |
| ``test.each`v ${expect(1).toBe(1)}`('c', cb);`` | no error | no error | no error |

Rows 1-3 are the new rstest policy and they land where the commit message says. Row 4 is the control: an assertion in a nested registration's arguments does run as part of the enclosing case, and it stays valid.

Two doc-level notes below, both on paragraphs this PR adds. Nothing blocking.

internal/plugins/jest/rules/no_standalone_expect/no_standalone_expect.md
elecmonkey: Widened the paragraph in `f3903eb0`, and left the binding alone: the one-line guard would start reporting `expect.not.stringContaining('value')`, which jest supports and which the suite covers as valid. Happy to do that separately if you want the upstream contract exactly.

internal/plugins/rstest/rules/no_standalone_expect/no_standalone_expect.md
elecmonkey: Confirmed both forms go unreported, so I narrowed the sentence in `f3903eb0` to name the tagged-template exception and left the traversal alone. The PR description had the same overreach and is updated too.

internal/utils/test_framework/rules/no_standalone_expect/no_standalone_expect.go
```diff
@@ -0,0 +1,338 @@
...
+	// ArgumentsOutsideTestScope reports whether the arguments of a registration
+	// sit outside the test case it registers. They are evaluated while the file
+	// is collected rather than while the case runs, so an assertion among them
+	// never runs as part of a test. Upstream eslint-plugin-jest allows it — its
+	// kind stack covers the whole call subtree — and the jest binding follows
+	// that contract; the rstest binding reports it.
+	//
+	// The flag only reaches a registration with no inline callback. Whenever one
+	// is found the scope is its body either way, so an assertion in the name or
+	// the options object is already outside it.
```
swwind: The second paragraph describes the flag as fallback-only, which is what leaves jest reporting an assertion in a registration's name, options object or trailing argument.

```suggestion
	// ArgumentsOutsideTestScope reports whether the arguments of a registration
	// sit outside the test case it registers. They are evaluated while the file
	// is collected rather than while the case runs, so an assertion among them
	// never runs as part of a test. Upstream eslint-plugin-jest allows it — its
	// kind stack covers the whole call subtree — and the jest binding follows
	// that contract; the rstest binding reports it.
	//
	// It picks the scope range, not just the fallback one: keeping the arguments
	// inside the case means the scope spans the whole call, while putting them
	// outside narrows it to the callback body, or to an empty range when the
	// registration has no callback at all.
```

swwind (commented):
Final pass on `f3903eb0`. The two doc corrections check out, and the preset, manifest and schema registration are all in place.

One thing left, and a correction to my own last review: the jest column of my previous table was wrong for two rows. Re-run against `eslint-plugin-jest` v29.15.5:

| Code | eslint-plugin-jest | rslint jest |
| --- | --- | --- |
| `test('c', expect(1).toBe(1), () => {});` | no error | `unexpectedExpect` |
| `test('c', { timeout: expect(1).toBe(1) }, () => {});` | no error | `unexpectedExpect` |
| `test('c', () => {}, expect(1).toBe(1));` | no error | `unexpectedExpect` |
| `describe('d', () => { test('c', expect(1).toBe(1), () => {}); });` | no error | `unexpectedExpect` |

So the false positives from my second review are still open on the jest side. `ArgumentsOutsideTestScope` only decides the *fallback* range, so only a registration with no inline callback reaches it — which is why `test.todo(expect(1).toBe(1))` now matches upstream while the four rows above do not. Pre-existing on `main`, not introduced here, but it makes the flag's own doc comment inaccurate and it is missing from the jest rule's `Differences from ESLint` section.

Letting the flag decide the whole scope range fixes both at once — suggestions inline. With them applied the four rows align with upstream, `test.todo(expect(1).toBe(1))` stays valid, `describe('d', () => { test.only('a', () => {}); expect(1).toBe(1); })` still reports, and both rule suites pass. `getTestScopeRange` is then reached only by the rstest binding, which is the framework that wants it. Nothing else blocking.

internal/utils/test_framework/rules/no_standalone_expect/no_standalone_expect.go
```diff
@@ -0,0 +1,338 @@
...
+							// call takes its arguments in, an empty range leaves an
+							// assertion among them with no test block around it.
+							if runtime.ArgumentsOutsideTestScope {
+								start, end = node.Pos(), node.Pos()
+							} else {
+								start, end = node.Pos(), node.End()
+							}
+						}
+						pushScope(blockTypeTest, start, end)
+					}
```
swwind, line 294:
```suggestion
					if isTestBlockCall(node) {
						// The scope spans the whole call unless the framework
						// evaluates a registration's arguments outside the case it
						// registers: narrowing it to the callback body leaves the
						// name and the options object out of the test block. A
						// registration with no inline callback still has to push
						// either way, or the exit handler pops somebody else's
						// scope; an empty range balances the stack while containing
						// no assertion.
						start, end := node.Pos(), node.End()
						if runtime.ArgumentsOutsideTestScope {
							if bodyStart, bodyEnd, ok := getTestScopeRange(node); ok {
								start, end = bodyStart, bodyEnd
							} else {
								end = node.Pos()
							}
						}
						pushScope(blockTypeTest, start, end)
					}
```

swwind (commented):
Fourth pass, on `a058331f`. Two things left, neither blocking.

The commit adds no test. Every row it fixes — an assertion in a registration's name, options object or trailing argument — is decided by the new `ArgumentsOutsideTestScope: false` branch, and neither suite exercises any of them, so the branch that changed is the one nothing pins down. Suggestion inline.

The PR description's "Behavior change in `jest/no-standalone-expect`" section still lists only `test.todo(expect(1).toBe(1))`. Those rows are now allowed by the jest rule as well, which is the same kind of change to a shipped rule — worth a line there.

internal/plugins/jest/rules/no_standalone_expect/no_standalone_expect_test.go
```diff
@@ -67,6 +67,36 @@ func TestNoStandaloneExpectRule(t *testing.T) {
...
+			{Code: `test.each([1])('outer', value => { test.todo('inner'); expect(value).toBe(1); });`},
+			{
+				Code: `test('outer', () => { t('inner'); expect(1).toBe(1); });`,
+				Options: []interface{}{
+					map[string]interface{}{"additionalTestBlockFunctions": []interface{}{"t"}},
+				},
+			},
+			// The fallback range covers the arguments of a registration that has
+			// no callback, which is where upstream also allows an assertion.
+			{Code: `test.todo(expect(1).toBe(1));`},
```
swwind: After `a058331f` the jest binding has no fallback range: with `ArgumentsOutsideTestScope: false` the scope is `node.Pos(), node.End()` for every registration, so this row takes the same path as any other one and the comment describes the previous model. Rewording it and adding the rows the commit actually fixes covers both at once:

```suggestion
			// A registration's arguments belong to the test block it opens, which
			// is where upstream also allows an assertion.
			{Code: `test.todo(expect(1).toBe(1));`},
			{Code: `test('case', expect(1).toBe(1), () => {});`},
			{Code: `test('case', { timeout: expect(1).toBe(1) }, () => {});`},
			{Code: `test('case', () => {}, expect(1).toBe(1));`},
			{Code: `describe('d', () => { test('case', expect(1).toBe(1), () => {}); });`},
```

elecmonkey: no problem. pr body edited. suggestions appied.

internal/plugins/rstest/rules/no_standalone_expect/no_standalone_expect_test.go
```diff
@@ -0,0 +1,235 @@
...
+			// so an assertion among them never runs as part of the case, whether
+			// or not the registration also takes an inline callback.
+			{
+				Code:   `test.todo(expect(1).toBe(1));`,
+				Errors: []rule_tester.InvalidTestCaseError{{MessageId: "unexpectedExpect"}},
+			},
+			{
+				Code:   `test("case", { timeout: expect(1).toBe(1) }, () => {});`,
+				Errors: []rule_tester.InvalidTestCaseError{{MessageId: "unexpectedExpect"}},
+			},
```
swwind, line 223: `test("case", () => {}, expect(1).toBe(1))` is the third shape the doc's "any other argument" sentence covers, and it takes a different path from the two rows above: the callback body is found, and the assertion sits after it rather than before. Nothing currently pins it.

```suggestion
			},
			{
				Code:   `test("case", () => {}, expect(1).toBe(1));`,
				Errors: []rule_tester.InvalidTestCaseError{{MessageId: "unexpectedExpect"}},
			},
```

swwind (commented):
One inline suggestion below, on test coverage. Nothing blocking from my side.

swwind (approved):

fansenze (approved):

PR #1705, opened 2026-08-12
refactor(utils): name each JavaScript port for what it ports

swwind:
## Summary

Splits the JavaScript-specific readings a ported rule depends on into packages named for what they reproduce, and routes every rule through them.

`internal/utils/ecmascript` holds the language's own semantics: trimming, blankness, whitespace and line terminators, the trivia scanners, `/i` case canonicalization, and Number-to-String. A helper standing in for something JavaScript exposes carries that name (`StringTrim`, `NumberToString`); everything else is named the way ECMAScript names it (`IsWhiteSpace`, `Canonicalize`, `IsBlank`). `utils.IsStrWhiteSpace` and `utils.IsECMABlankLine` are gone.

`internal/utils/ecmascript/regexp` compiles a JavaScript RegExp over regexp2, and every user-written pattern now goes through it. 26 files came off regexp2 directly; the rest were on RE2, which has no lookahead, lookbehind or backreferences and whose compile error each caller swallowed, silently dropping whatever option the pattern came from. Only two files still import regexp2: this package and the minimatch port.

regexp2's ECMAScript mode differs from JavaScript in six ways, all rewritten before compiling: a bare `.` matched U+2028 and U+2029; the `m`-flag anchors broke on a line feed alone rather than on all four line terminators; ignoring case folded by Unicode's rules rather than by `Canonicalize`; `\b` was drawn over .NET's Unicode word set where ECMAScript keeps it to ASCII; `\A`, `\a` and `\e` kept their .NET meanings where JavaScript reads the letter, and an escape .NET does not know at all refused a pattern JavaScript accepts; and `y` says where a match may start, which regexp2 has no flag for. Every compiled pattern also carries a finite match timeout.

The rewrite reads the syntax only JavaScript spells, too: `[]` and `[^]`, a `(?i-m:…)` modifier group — which it applies itself, group by group, rather than leaving to .NET's own reading of an inline flag — and a `\p{…}` property escape, whose name belongs to the escape rather than being text to widen. A `(?…` opener JavaScript has no syntax for is refused, so `(?i)`, `(?ii:…)` and `(?-u:…)` never quietly compile into something their author did not write. Ignoring case is widened into the pattern rather than handed to regexp2, except where a pattern cannot spell what it needs: a backreference, or a property escape, falls back to regexp2's own case-insensitivity.

`internal/utils/minimatch3` and `internal/utils/isglob` port the npm packages a plugin reads globs with, minimatch 3.1.5 and is-glob 4.0.3. The major version is in the first name because a single ESLint run uses two of them: plugins pin `minimatch@^3.1.2`, while ESLint matches its own flat-config `files` and `ignores` with minimatch 10, which is not ported. `no-restricted-imports` and config path matching keep doublestar.

`Canonicalize` has to answer the way Node does, so the eight uppercase mappings Unicode 16 and 17 added — which Go 1.26's Unicode 15 tables lack — sit in `unicode17.go`, with a test that fails once the toolchain reaches Unicode 17 and makes the file dead weight.

depguard was listed in both the enable and the disable block, so it never ran. It is enabled now, denying regexp2 and doublestar under `internal/rules/**` and `internal/plugins/**`. Its deny on `encoding/json` is dropped: it pointed at typescript-go's `internal/json`, which is in another module and cannot be imported from here at all, while 33 files use `encoding/json` directly.

## Behaviour changes

Three rules silently did nothing when an option held a pattern Go's RE2 refused to compile — the compile error was swallowed, so the rule reported nothing while the config looked like it worked.

| Rule | Option | Before | After |
| --- | --- | --- | --- |
| `react/boolean-prop-naming` | `rule: "^(?=is)"` | RE2 rejects the lookahead, so the rule reports nothing | the pattern compiles; `something` does not start with `is`, so it reports |
| `react/jsx-handler-names` | `eventHandlerPrefix: "[[:foo:]]"` | RE2 reads `[[:foo:]]` as a POSIX class and rejects it, so the rule reports nothing | JavaScript reads it as an ordinary character class, so the rule runs and reports |

Extended glob syntax in a path option is matched rather than taken literally.

| Rule | Option | Before | After |
| --- | --- | --- | --- |
| `import/no-restricted-paths` | `target: "./restricted-paths/!(server)/**/*"` | doublestar matches `!(server)` as a literal directory name, so no file falls in the zone and the import is allowed | minimatch reads it as every directory except `server`, so a `client` file is in the zone and the import is reported |

Ignoring case compares the way JavaScript compares, which is two different readings rather than one.

| Pattern | Before | After |
| --- | --- | --- |
| `i` without `u` | regexp2 folds by Unicode's rules, so `/k/i` matches U+212A KELVIN SIGN and `/s/i` matches U+017F LONG S | a character outside ASCII never folds into ASCII, so neither matches |
| `i` with `u` or `v` | — | simple case folding, so both do match — the rule ECMAScript switches to under those flags |

A pattern is shown in a message the way its author wrote it, not the way the engine rewrote it.

| Rule | Before | After |
| --- | --- | --- |
| `jest/valid-title`, `rstest/valid-title` | ``Invalid regular expression in `disallowedWords` option: `(?i)\b(ok|(unterminated)\b` `` | ``Invalid regular expression in `disallowedWords` option: `\b(ok|(unterminated)\b` `` |

## Agent docs

`AGENTS.md` and the `port-rule` skill now map an upstream dependency to the package that reads it the same way, say where the stdlib `regexp` may still be used — a pattern written in this repository that RE2 and JavaScript read alike and that no user input reaches — and require reporting to the user, rather than substituting or porting, when a rule reads globs with anything but minimatch 3 or is-glob.

## Checklist


internal/utils/ecmascript/regexp/rewrite.go
```diff
@@ -0,0 +1,410 @@
...
+	case 'd', 'D', 'w', 'W', 's', 'S', 'b', 'B', 'k', 'c':
+		// Names a set, a boundary or a group rather than a character.
+		return utf8.RuneError, size, true
+	case 'p', 'P':
+		// A property class, but only under `u`. Without it JavaScript reads
+		// `\p` as the letter, where .NET reads a property either way.
+		if unicode {
+			return utf8.RuneError, size, true
+		}
+		return r, size, true
```
elecmonkey: When a pattern combines `i` with `u`, `decodeEscape` leaves `\p{...}` and `\P{...}` unchanged and the rewrite stays marked exact, so `Compile` omits regexp2's `IgnoreCase`. As a result `/\p{Ll}/iu` rejects `A`, whereas JavaScript accepts it under simple case folding. The same applies to property escapes inside character classes, and it affects every ported rule that accepts such a user pattern. These escapes need case closure, or they should trigger a fallback rather than compiling as an exact, case-sensitive match.

internal/utils/ecmascript/regexp/rewrite.go
```diff
@@ -0,0 +1,410 @@
...
+	rest := source[len("(?"):]
+	for _, valid := range []string{":", "=", "!", "<="} {
+		if strings.HasPrefix(rest, valid) {
+			return nil
+		}
+	}
+	if strings.HasPrefix(rest, "<!") {
+		return nil
+	}
+	return fmt.Errorf("%w: (?%.1s", ErrUnsupportedSyntax, rest)
```
elecmonkey, line 476: On the repository's Node 24 baseline, inline modifier groups such as `(?i:a)`, `(?i-m:a)`, and `(?-i:a)` are valid JavaScript RegExp syntax, but this whitelist rejects all of them as unsupported. A valid option using this syntax then fails compilation and is silently discarded by the many callers routed through this layer, disabling or changing the affected rule instead of matching the way JavaScript does. Consider accepting the modifier-group form (`(?<flags>-<flags>:...)`) here.

internal/utils/ecmascript/regexp/rewrite.go
swwind: Fixed in 9fb620d2. It was worse than reported: outside a character class the walk read `{Ll}` as text and widened the letters under `i`, so `/\p{Ll}/iu` came out as `\p{[Ll][Ll]}` and did not compile at all. Inside one, `[\p{Ll}]`, it was the exact case-sensitive match you describe.

`decodeEscape` now takes the name along as part of the escape, so it is never read as text, and a property escape under `i` and `u` marks the rewrite inexact — which hands ignoring case to regexp2, the way a backreference already did. `/\p{Ll}/iu` and `/[\p{Ll}]/iu` both match `A` now, and `/[\p{Ll}x]/iu` still matches `X`.

`\P{…}` under `i` stays approximate, and the package doc says so. JavaScript asks whether any character comparing equal to this one falls outside the property, so `/\P{Ll}/iu.test("a")` is `true`; regexp2 asks only about the character at hand and answers `false`. Case-closing a negated property set means naming its complement, which needs a resolver for every property `\p{…}` accepts — Go's `unicode` tables cover the categories and scripts but not the binary properties, so it would be a partial answer dressed as a complete one. Happy to take that on instead of the documented gap if you would rather have it.

internal/utils/ecmascript/regexp/rewrite.go
swwind: Fixed in 9fb620d2. Confirmed against Node first: `(?i:a)`, `(?i-m:a)`, `(?-i:a)` and `(?i-:a)` all compile, while `(?ii:a)`, `(?i-i:a)`, `(?-:a)`, `(?-u:a)` and `(?i)a` do not.

The rewrite carries the flags in a stack now: `(` pushes, `)` pops, and a `(?flags-flags:` opener applies `i`, `m` and `s` over that group alone, so the literals, `.`, `^`/`$` and `\b` inside it are rewritten under the group's flags rather than the pattern's. It emits a plain `(?:` — except that a group turning `i` off inside a pattern that carries `i` emits `(?-i:`, so the case-insensitivity fallback cannot reach into it. Group numbering is untouched, since the opener is non-capturing either way.

15 cases in `regexp_test.go` check the answers against Node, nesting included: `/^(?i:a(?-i:b)c)$/` matches `AbC` and not `ABC`. `no-invalid-regexp` gained three valid cases that were false positives before — `(?i:a)`, `(?i-m:a)`, `(?-ims:a)` — and the five skipped invalid ones for `(?ii:`, `(?-ii:`, `(?i-i:`, `(?-:` and `(?-u:` are un-skipped: they passed already, and they now cover the modifier parser's own rejections.

fansenze (approved):

PR #1724, opened 2026-08-14
feat: port rule no-undefined

swwind:
## Summary

Port the `no-undefined` rule from ESLint to rslint.

Disallows using `undefined` as an identifier — flags every declaration, parameter, and reference literally named `undefined`, in both value and TypeScript-specific declaration positions (interfaces, namespaces, enums, type parameters, etc.).

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/no-undefined
- Source code: https://github.com/eslint/eslint/blob/v10.8.1/lib/rules/no-undefined.js

## Checklist


internal/rules/no_undefined/no_undefined.go
```diff
@@ -0,0 +1,125 @@
...
+	if parent == nil {
+		return true
+	}
+
+	switch parent.Kind {
+	case ast.KindPropertyAccessExpression:
+		// obj.undefined -> "undefined" only names the property, not a variable.
+		return parent.AsPropertyAccessExpression().Name() == node
+
+	case ast.KindPropertySignature, ast.KindMethodSignature, ast.KindMethodDeclaration,
```
elecmonkey, line 59: [P2] Exclude index-signature parameter names — `interface I { [undefined: string]: any }` is a valid index signature, but typescript-eslint creates neither a definition nor a reference for the parameter label, so ESLint 10.8 emits no diagnostic. Since `KindIndexSignature` is not excluded, this identifier reaches `ReportNode` and produces a false positive. Please skip parameters whose parent is `KindIndexSignature`.

internal/rules/no_undefined/no_undefined.go
```diff
@@ -0,0 +1,125 @@
...
+		return true
+	}
+	if utils.IsInJSDocSyntax(node) {
+		return true
+	}
+	if ast.IsJsxTagName(node) {
+		return true
+	}
+
+	return false
```
elecmonkey, line 123: [P2] Skip dotted namespace name segments — TypeScript-go lowers `namespace undefined.Foo {}` / `namespace Foo.undefined {}` to nested module declarations, but typescript-eslint binds neither segment as a variable, so ESLint 10.8 emits no `no-undefined` diagnostic. Both identifiers currently reach this fallback and are reported; please skip dotted module-declaration names while keeping simple `namespace undefined {}` reportable.

internal/rules/no_undefined/no_undefined.go
```diff
@@ -0,0 +1,125 @@
...
+		return isNonBindingExportSpecifierName(node, parent)
+
+	case ast.KindLabeledStatement, ast.KindBreakStatement, ast.KindContinueStatement:
+		return true
+
+	case ast.KindImportAttribute:
+		// import ... with { undefined: "json" } — an attribute key, not a variable.
+		return parent.AsImportAttribute().Name() == node
+
+	case ast.KindJsxNamespacedName:
```
elecmonkey, line 96: [P2] Report namespaced JSX tag prefixes — This blanket skip also suppresses `<undefined:foo />` and its matching closing tag. typescript-eslint exposes the tag prefix as a scope reference, so ESLint reports it; only namespaced attribute components such as `<X undefined:attr />` should be ignored. Please distinguish tag names from attribute names here.

internal/rules/no_undefined/no_undefined.go
```diff
@@ -0,0 +1,125 @@
...
+	Run: func(ctx rule.RuleContext, options []any) rule.RuleListeners {
+		return rule.RuleListeners{
+			ast.KindIdentifier: func(node *ast.Node) {
+				if node.Text() != "undefined" {
+					return
+				}
+				if isNonBindingUndefinedPosition(node) {
+					return
+				}
+				ctx.ReportNode(node, rule.RuleMessage{
```
elecmonkey: [P2] Preserve upstream ranges for typed declarations — `ReportNode(node)` narrows every diagnostic to the bare identifier. For a direct typed declaration such as `function f(undefined: number) {}`, typescript-eslint’s definition name spans `undefined: number`, so ESLint’s diagnostic extends through the annotation; the same applies to typed variables, signatures, and parameter properties. Please report the upstream-equivalent definition range rather than always the identifier range.

internal/rules/no_undefined/no_undefined.go
swwind: Confirmed and fixed in 956b19c0. Verified against ESLint 10.8 + @typescript-eslint/parser 8.65: `interface I { [undefined: string]: any }`, `type T = { [undefined: string]: number }` and `class C { [undefined: string]: any }` all produce no diagnostic.

`isNonBindingUndefinedPosition` now has a `KindParameter` arm that skips the name only when the parameter's parent is a `KindIndexSignature`; every other parameter name stays a real binding, and a parameter initializer identifier stays a reference. Locked in by two valid cases.

internal/rules/no_undefined/no_undefined.go
swwind: Confirmed and fixed in 956b19c0. Verified upstream behavior:

| Code | ESLint 10.8 |
| --- | --- |
| `namespace undefined.Foo {}` | no diagnostic |
| `namespace Foo.undefined {}` | no diagnostic |
| `namespace undefined.A.undefined {}` | no diagnostic |
| `namespace undefined {}` | reported |
| `namespace A { namespace undefined {} }` | reported |
| `namespace A.B { namespace undefined {} }` | reported |

Added a `KindModuleDeclaration` arm backed by `isDottedModuleNameSegment`: tsgo lowers `namespace A.B {}` to nested module declarations, so a dotted segment is either a declaration whose body is itself a module declaration (head position) or a declaration that *is* its parent's body (tail position). Block-nested namespaces have a `ModuleBlock` body and stay reportable. Both branches plus the block-nested counterparts are covered by new cases.

internal/rules/no_undefined/no_undefined.go
swwind: Confirmed and fixed in 956b19c0. Upstream reports *both* segments of a namespaced tag name, and neither segment of a namespaced attribute name:

| Code | ESLint 10.8 |
| --- | --- |
| `<undefined:foo />` | reported (prefix) |
| `<foo:undefined />` | reported (name) |
| `<undefined:undefined />` | reported twice |
| `<undefined:foo></undefined:foo>` | reported on both tags |
| `<X undefined:attr />` | no diagnostic |
| `<X foo:undefined />` | no diagnostic |

The arm is now `return !ast.IsJsxTagName(parent)`, so a namespaced name reports through in tag position and is skipped in attribute position. Also added the two shapes that separate this from the plain-tag path, which had no coverage before: `<undefined />` stays valid (lowercase intrinsic element, matching upstream) and `<undefined.tag />` reports its root.

internal/rules/no_undefined/no_undefined.go
swwind: Confirmed and fixed in 956b19c0. Reporting now goes through `ReportRange`, with `declarationNameEnd` extending the identifier span to the trailing marker/annotation that TSESTree folds into the declaration-name `Identifier`:

| Code | ESLint 10.8 columns | Now |
| --- | --- | --- |
| `function f(undefined: number) {}` | 12-29 | 12-29 |
| `function f(undefined: number = 1) {}` | 12-29 | 12-29 |
| `function f(undefined?) {}` | 12-22 | 12-22 |
| `let undefined!: number;` | 5-23 | 5-23 |
| `const undefined: Array<string> = [];` | 7-31 | 7-31 |
| `class C { constructor(private undefined: number) {} }` | 31-48 | 31-48 |
| `function f(...undefined: number[]) {}` | 15-24 | 15-24 |
| `const { a: undefined }: { a: number } = obj;` | 12-21 | 12-21 |

The last two are the carve-outs: a rest parameter's annotation belongs to the `RestElement` upstream and a destructured binding's annotation belongs to the pattern, so the identifier stays bare in both. Every row above is a new position-asserting test case.

elecmonkey (approved):
no more issues found

PR #1725, opened 2026-08-14
feat: port rule operator-assignment

swwind:
## Summary

Port the `operator-assignment` rule from ESLint to rslint.

Requires or disallows assignment operator shorthand where possible (e.g. `x = x + y` → `x += y`), with autofix support in both directions.

Also extends `utils.IsSameReference` with ESLint's `disableStaticComputedKey` parameter (needed by this rule) and fixes an overly-conservative case in `utils.CanTokenTextsBeAdjacent`.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/operator-assignment
- Source code: https://github.com/eslint/eslint/blob/v10.8.1/lib/rules/operator-assignment.js

## Checklist


internal/utils/ts_eslint.go
```diff
@@ -1585,61 +1585,130 @@ func GetStaticExpressionValue(node *ast.Node) (string, bool) {
...
+	// Base cases: Identifier, PrivateIdentifier, ThisKeyword, and literal keys
+	// (e.g. `x[0]` vs `x[0]`, reached recursively when disableStaticComputedKey
+	// skips the static-name fast path below).
 	if left.Kind == ast.KindIdentifier && right.Kind == ast.KindIdentifier {
 		return left.AsIdentifier().Text == right.AsIdentifier().Text
 	}
+	if left.Kind == ast.KindPrivateIdentifier && right.Kind == ast.KindPrivateIdentifier {
+		return left.AsPrivateIdentifier().Text == right.AsPrivateIdentifier().Text
+	}
 	if left.Kind == ast.KindThisKeyword && right.Kind == ast.KindThisKeyword {
```
elecmonkey: **[P2] Handle `super` in same-reference comparisons**

ESLint’s `astUtils.isSameReference` explicitly treats two `Super` nodes as the same reference, but this base-case block omits `ast.KindSuperKeyword`. As a result, `class C extends B { m() { super.x = super.x + y; } }` (and computed variants) is missed entirely. Please recognize `super` here; `canBeFixed` will still reject it, preserving ESLint’s diagnostic-without-fix behavior.

internal/rules/operator_assignment/operator_assignment.go
```diff
@@ -0,0 +1,284 @@
...
+
+		if utils.HasCommentInSpan(ctx.Comments.All(), nodeRange.Pos(), opRange.Pos()) {
+			return nil
+		}
+
+		text := sourceFile.Text()
+		leftText := text[nodeRange.Pos():opRange.Pos()]
+		plainOperatorText := scanner.TokenToString(plainOperatorKind)
+
+		rightPrecedence := ast.GetExpressionPrecedence(binExpr.Right)
```
elecmonkey: **[P2] Use ESLint precedence for TS-only RHS nodes**

This uses tsgo’s native precedence, but ESLint treats parser-specific node types as unknown precedence. In `"never"` mode, `x += y!` is therefore fixed here as `x = x + y!`, while ESLint 10.8.1 with `@typescript-eslint/parser` emits `x = x + (y!)`; unparenthesized `as` and `satisfies` RHS expressions have the same mismatch. Please use the ESLint-compatible fallback precedence rather than tsgo’s native precedence for these nodes.

elecmonkey (commented):
Found two upstream-parity issues; details inline.

internal/utils/ts_eslint.go
swwind: Fixed in 88f12b55.

`IsSameReference` now treats two `SuperKeyword` nodes as the same reference, matching upstream's `case "Super": case "ThisExpression": return true`. `canBeFixed` still rejects them, so `super.x = super.x + y` (and the computed variants) reports without a fix, exactly as ESLint does.

This also fixed a second rule: `no-self-assign` shares the same helper, so `super.x = super.x` was silently missed there too. Test cases added in both rules.

internal/rules/operator_assignment/operator_assignment.go
swwind: Fixed in 88f12b55.

The `"never"` fixer now uses `utils.EslintLikePrecedence` for the right side, plus a new `utils.EslintLikeBinaryOperatorPrecedence` for the operator it is about to write — the latter mirrors upstream's `getPrecedence({ type: "BinaryExpression", operator })`. TS-only nodes hit `getPrecedence`'s `-1` branch ("assume it has the lowest precedence"), so they get parenthesized.

Verified against ESLint 10.8.1 + `@typescript-eslint/parser`:

| Code (`"never"`) | ESLint | This PR |
| --- | --- | --- |
| `foo -= bar!` | `foo = foo - (bar!)` | same (was `foo = foo - bar!`) |
| `foo -= bar as number` | `foo = foo - (bar as number)` | same |
| `foo -= bar satisfies number` | `foo = foo - (bar satisfies number)` | same |
| `foo -= (bar!)` | `foo = foo - (bar!)` | same |
| `foo -= bar?.baz` | `foo = foo - bar?.baz` | same |

All five are locked in as test cases.

elecmonkey (dismissed):
no more issues found

internal/rules/operator_assignment/operator_assignment.go
```diff
@@ -0,0 +1,341 @@
...
+		leftText := text[nodeRange.Pos():opRange.Pos()]
+		plainOperatorText := scanner.TokenToString(plainOperatorKind)
+
+		newOperatorPrecedence := utils.EslintLikeBinaryOperatorPrecedence(plainOperatorKind)
+
+		var rightText string
+		if rightNeedsParens(binExpr.Right, newOperatorPrecedence) {
+			rightRange := utils.TrimNodeTextRange(sourceFile, binExpr.Right)
+			inner := text[rightRange.Pos():rightRange.End()]
+			if plainOperatorKind == ast.KindLessThanLessThanToken && strings.ContainsRune(inner, '<') {
```
elecmonkey: **[P2] Narrow the `<<=` parse-safety guard to generic syntax**

When `"never"` mode needs parentheses and the RHS contains any `<` character, this suppresses the fix even when the generated code is valid. For example, ESLint 10.8.1 fixes `foo <<= bar < baz` to `foo = foo << (bar < baz)`, while Rslint reports without fixing; string, regex, comment, and template contents containing `<` are also unnecessarily rejected. Detect the problematic TypeScript generic/assertion shape rather than searching the raw RHS text.

elecmonkey (commented):
The rule generally matches the upstream behavior, but its `<<=` safety guard suppresses valid autofixes for ordinary comparisons and other harmless occurrences of `<`. This is an observable autofix parity regression.

fansenze (approved):

PR #1726, opened 2026-08-14
feat: port rule init-declarations

swwind:
## Summary

Port the `init-declarations` rule from ESLint to rslint.

Requires or disallows initialization in variable declarations (`var`, `let`, `const`, `using`, `await using`), with an `ignoreForLoopInit` option.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/init-declarations
- Source code: https://github.com/eslint/eslint/blob/v10.8.1/lib/rules/init-declarations.js

## Checklist

### DIFF
--- a/internal/rules/init_declarations/init_declarations.go
+++ b/internal/rules/init_declarations/init_declarations.go
@@ -86,6 +86,23 @@ func checkVariableDeclarationList(ctx rule.RuleContext, node *ast.Node, opts ini
 	kind := utils.GetVarDeclListKind(node)
 	isConstantBinding := kind == "const" || kind == "using" || kind == "await using"
 
+	// Upstream's isInitialized() walks straight from the declarator to its
+	// enclosing VariableDeclaration to THAT node's parent, and treats a
+	// for-loop parent as init/left-slot identity — true only when the
+	// declaration literally IS the loop's init/left slot, false otherwise
+	// (never falling back to checking the initializer). An unbraced for-loop
+	// body that is itself a var declaration (`for (...) var x = 1;`) hits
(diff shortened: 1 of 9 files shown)


internal/rules/init_declarations/init_declarations.go
```diff
@@ -0,0 +1,172 @@
...
+	if declList == nil || declList.Declarations == nil {
+		return
+	}
+
+	// Upstream: `if (node.declare || insideDeclaredNamespace) return;` — an
+	// ambient declaration list (either `declare var/let/const ...` itself, or
+	// any declaration nested inside a `declare namespace`/`declare module`/
+	// ambient `.d.ts` context) is exempt in both modes.
+	if utils.IsInAmbientContext(node) {
+		return
```
elecmonkey, line 80: [P2] Match upstream instead of skipping every ambient node

When linting a `.d.ts` file, TypeScript marks every declaration list ambient, so this return suppresses diagnostics that ESLint still emits; for example, ESLint reports `var x: number;` in `"always"` mode and `var x: number = 1;` in `"never"` mode. It also diverges for nested declared namespaces: ESLint reports `y` in `declare namespace A { declare namespace B { let x: number; } let y: number; }` because its boolean resets when `B` exits, whereas this implementation skips `y`. Reproduce upstream's `node.declare` and `TSModuleDeclaration` enter/exit behavior rather than treating `NodeFlagsAmbient` as equivalent.

elecmonkey:
LGTM

elecmonkey (approved):
reviewed with no issue found

fansenze (approved):

PR #1727, opened 2026-08-14
feat: port rule max-statements

swwind:
## Summary

Port the `max-statements` rule from ESLint to rslint.

Enforces a maximum number of statements allowed in function blocks.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/max-statements
- Source code: https://github.com/eslint/eslint/blob/v10.8.1/lib/rules/max-statements.js

## Checklist


internal/rules/max_statements/max_statements.go
```diff
@@ -0,0 +1,161 @@
...
+		startFunc := func(node *ast.Node) {
+			// Overload signatures and abstract/declare members reach this
+			// listener with no body. ESLint never visits them as
+			// FunctionDeclaration/FunctionExpression (they parse as
+			// TSDeclareFunction / body-less MethodDefinition instead), so skip
+			// pushing a frame for parity — and skip popping it in endFunc.
+			if node.Body() == nil {
+				return
+			}
+			stack = append(stack, 0)
```
elecmonkey, line 67: [P2] Keep member metadata outside the method frame

`ForEachChild` visits a method's modifiers/decorators and computed name after the `KindMethodDeclaration` enter listener, so this frame makes functions in those expressions look nested. With `ignoreTopLevelFunctions: true`, `class C { [(() => {})()]() { one; two; } }` is a false negative: ESLint sees two top-level functions and reports the method, while this implementation leaves the method as the sole deferred top-level function and exempts it. Please start the function frame only when traversal reaches the member's parameters/body (or otherwise keep decorators/computed names in the enclosing frame).

internal/rules/max_statements/max_statements.go
```diff
@@ -0,0 +1,161 @@
...
+			count int
+		}
+		var topLevelFunctions []topLevelEntry
+
+		report := func(node *ast.Node, count int) {
+			if count <= opts.max {
+				return
+			}
+			name := utils.UpperCaseFirstASCII(utils.GetFunctionNameWithKindCore(node))
+			ctx.ReportRange(utils.GetFunctionHeadLoc(ctx.SourceFile, node), rule.RuleMessage{
```
elecmonkey: [P2] Use the ESTree member start for decorated methods

ESLint core's `getFunctionHeadLoc` starts `MethodDefinition` reports at `parent.loc.start`, which includes decorators. Here the shared helper intentionally calls `nodeStartSkippingDecorators`, so `class C { @dec(1) foo() { one; two; } }` starts at `foo` instead of `@dec(1)`. That makes `max-statements`' diagnostic range diverge for every decorated method/getter/setter; this rule needs a range helper/override that preserves the enclosing member's start.

internal/rules/max_statements/max_statements.go
swwind: Fixed in a96fe6d9. Confirmed against ESLint 10.7 + @typescript-eslint/parser: `class C { [(() => {})()]() { one; two; } }` with `[1, { ignoreTopLevelFunctions: true }]` reports `Method has too many statements (2)` at 1:11–1:25, and the port reported nothing.

Added `KindDecorator` / `KindComputedPropertyName` enter/exit listeners that stash the member's frame while traversing that metadata and restore it afterwards, so a function written there sees the enclosing frame — matching ESTree, where decorators and the key hang off `MethodDefinition` rather than the `FunctionExpression`. Parameter defaults deliberately stay inside the member frame (ESTree puts `params` on the function), locked in by a case where `class C { m(a = () => { x; y; }) { one; } }` still reports the arrow as nested rather than deferring it.

New tests cover the class and object-literal computed-key shapes, the decorator shape, and that the member's own count survives the stash/restore.

internal/rules/max_statements/max_statements.go
swwind: Fixed in a96fe6d9. `max-statements` is a core rule, so it follows core's `getFunctionHeadLoc` (`start = parent.loc.start`, decorators included); the shared `utils.GetFunctionHeadLoc` mirrors typescript-eslint's variant, which skips them on purpose. Added a local `functionHeadLoc` that keeps the shared end position but restores the member's own start when the member carries decorators.

Verified against ESLint: `class C { @dec(1) foo() { one; two; } }` → 1:11–1:22, `@dec(1) get foo()` → 1:11–1:26, and the multi-decorator `static async` case → 1:11–3:19. The same start applies to a decorated field's initializer (`@dec(1) foo = () => { … }` → 1:11–1:25), whose ESTree parent is the `PropertyDefinition`; all covered by new tests.

elecmonkey (dismissed):

internal/utils/ts_eslint.go
```diff
@@ -735,6 +743,53 @@ func nodeStartSkippingDecorators(sourceFile *ast.SourceFile, node *ast.Node) cor
...
+	af := node.AsArrowFunction()
+	if af.Parameters == nil || len(af.Parameters.Nodes) != 1 {
+		if parenPos := findOpenParenPos(sourceFile, node); parenPos >= 0 {
+			return parenPos
+		}
+		return af.EqualsGreaterThanToken.Pos()
+	}
+
+	paramStart := TrimNodeTextRange(sourceFile, af.Parameters.Nodes[0]).Pos()
+	if parenPos := findOpenParenPosFrom(sourceFile, node.Pos(), paramStart); parenPos >= 0 {
```
elecmonkey: [P2] Find the actual parameter-list parenthesis

For a generic class-field arrow whose constraint contains parentheses, such as `class C { f = <T extends (string | number)>(x: T) => {a;b}; }`, this scan selects the `(` before `string` rather than the parameter-list `(`. With `max-statements: [1]`, ESLint 10.8.1 reports columns 11–44, while this implementation ends at column 26; the same shared-helper regression affects other consumers such as `class-methods-use-this`. The search needs to identify the token immediately preceding the sole parameter instead of taking the first parenthesis anywhere before it.

internal/plugins/typescript/rules/class_methods_use_this/class_methods_use_this.go
```diff
@@ -274,30 +272,22 @@ func run(ctx rule.RuleContext, options []any) rule.RuleListeners {
...
-				// Direct child of PropertyDeclaration — the shared helper
-				// already handles this shape correctly.
-				loc = utils.GetFunctionHeadLoc(ctx.SourceFile, node)
-			} else {
-				// Paren-wrapped: reconstruct the upstream head loc as
-				// "<field-start>...<function's own open paren>".
-				loc = classFieldHeadLocAcrossParens(ctx.SourceFile, field, node)
-			}
 		} else {
 			name = utils.GetFunctionNameWithKind(node)
```
elecmonkey, line 288: [P1] Update the auto-accessor integration snapshots

When the full JS integration suite runs, this auto-accessor behavior change makes `class-methods-use-this > invalid 4` fail because the committed snapshot still expects `class method 'method'` at columns 3–21, while the rule now emits `class arrow function` at columns 24–26. This failure occurs on both Linux and Windows in the current head CI, so the corresponding integration snapshots must be updated to lock in the newly intended upstream-aligned output.

fansenze (approved):

PR #1731, opened 2026-08-14
feat: port rule no-warning-comments

swwind:
## Summary

Port the `no-warning-comments` rule from ESLint to rslint.

Reports comments that contain any of a configured set of warning terms (default: `todo`, `fixme`, `xxx`), matched case-insensitively either at the start of the comment or anywhere within it.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/no-warning-comments
- Source code: https://github.com/eslint/eslint/blob/v10.8.1/lib/rules/no-warning-comments.js

## Checklist


internal/rules/no_warning_comments/no_warning_comments.go
```diff
@@ -0,0 +1,306 @@
...
+			prefix = `\b`
+		}
+	}
+
+	suffix := ""
+	if r, size := utf8.DecodeLastRuneInString(term); size > 0 && isWordChar(r) {
+		suffix = `\b`
+	}
+
+	return regexp.MustCompile("(?i)" + prefix + escaped + suffix)
```
elecmonkey: [P2] Preserve Unicode-aware word boundaries

When a term is adjacent to U+212A `K` or U+017F `ſ`, RE2's ASCII-only `\b` differs from JavaScript `/iu` word boundaries. For example, with the default options rslint reports `// TODOK` while ESLint v10.8.1 does not; conversely, with `{ terms: ["kelvin"], location: "anywhere" }`, ESLint reports `// Kelvin` while rslint does not. The boundary handling needs to account for these two characters that Unicode case-fold to ASCII word characters.

swwind:
This is blocked by #1705 because of the usage of regexp. Hold short for now.

internal/rules/no_warning_comments/no_warning_comments.go
swwind: Fixed by moving the rule onto `esregexp` (#1705), which writes `\b` out as the lookaround pair ECMAScript defines it as and draws the word set the way `iu` draws it, U+212A and U+017F included. `// TODO\u212A` no longer reports, and `// \u212Aelvin` under `{ terms: ["kelvin"], location: "anywhere" }` now does — both matching ESLint v10.8.1.

internal/rules/no_warning_comments/no_warning_comments.go
```diff
@@ -0,0 +1,259 @@
...
+		opts := parseOptions(options)
+		escapedDecoration := escapeRegExp(strings.Join(opts.Decoration, ""))
+		warningRegExps := make([]*esregexp.RegExp, len(opts.Terms))
+		for i, term := range opts.Terms {
+			warningRegExps[i] = convertToRegExp(term, opts.Location, escapedDecoration)
+		}
+
+		text := ctx.SourceFile.Text()
+		for _, comment := range ctx.Comments.All() {
+			checkComment(ctx, text, comment, opts.Terms, warningRegExps)
```
elecmonkey, line 35: [P2] Include Annex B HTML comments in the scan

For JavaScript script inputs using legacy HTML comments, such as `<!-- TODO` or `--> TODO`, ESLint 10.8.1 treats the text as a line comment and reports it, but `ctx.Comments.All()` returns no entries because the TypeScript-backed comment store only collects `//` and `/* */` comments. Consequently this loop silently misses warning terms in valid Annex B comments; these comments need to be collected or handled separately for JS scripts.

swwind: Not fixable inside this rule — the gap is in the parser, not in the comment scan.

typescript-go's scanner has no Annex B HTML-like comment support at all: `<!--` and `-->` are never scanned as comment trivia, so these inputs don't merely go missing from `ctx.Comments.All()`, they fail to parse. Both shapes you named produce syntax errors:

```
<!-- TODO fix this
  file.js(1,1):  error TS1109: Expression expected.
  file.js(1,11): error TS1005: ';' expected.

--> TODO fix this
  file.js(1,3):  error TS1109: Expression expected.
  file.js(1,10): error TS1005: ';' expected.
```

rslint still lints such a file — `CreateProgramLenient` tolerates syntactic errors — but it does so over a broken AST, so every rule on that file is affected, not just this one, and `eslint-disable` directives written as HTML comments are missed too. Collecting them here would mean this rule hand-rolling a JS tokenizer over text the parser has already mis-tokenized, while every other comment consumer stayed broken.

I'd rather see this handled in the scanner / comment store than special-cased in a rule port. Happy to open a follow-up issue for Annex B comment support if you agree.

elecmonkey (dismissed):

fansenze (approved):

PR #1732, opened 2026-08-14
feat: port rule sort-keys

swwind:
## Summary

Port the `sort-keys` rule from ESLint to rslint.

Requires object keys to be sorted alphabetically (or naturally), ascending or descending, with options for case sensitivity, minimum key count, blank-line-separated groups, and computed-key handling.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/sort-keys
- Source code: https://github.com/eslint/eslint/blob/v10.8.1/lib/rules/sort-keys.js

## Checklist


internal/rules/sort_keys/sort_keys.go
```diff
@@ -0,0 +1,275 @@
...
+}
+
+// propertyStaticName mirrors upstream's getPropertyName: a statically known
+// name (literal key, or non-computed identifier key) wins; otherwise, a
+// computed key whose expression is a bare identifier falls back to that
+// identifier's own name (not its runtime value) — matching upstream's
+// `node.key.name` fallback, which reads the Identifier node's `name` field
+// regardless of whether the key is computed.
+func propertyStaticName(nameNode *ast.Node) (string, bool) {
+	if name, ok := utils.GetStaticPropertyName(nameNode); ok {
```
elecmonkey, line 68: When `caseSensitive: false` is used with characters whose JavaScript lowercase mapping expands, Go's `strings.ToLower` produces a different comparison key. For example, ESLint 10.8.1 reports `{ 'İ': 1, i: 2 }` in ascending mode because `İ`.toLowerCase() is `i\u0307`, while this implementation maps `İ` directly to `i` and accepts the pair as equal. This is separate from the documented UTF-16/code-point ordering difference, so the rule needs JavaScript-compatible case conversion.

internal/rules/sort_keys/sort_keys.go
```diff
@@ -0,0 +1,275 @@
...
+func containsBlankLine(s string) bool {
+	newlines := 0
+	for _, r := range s {
+		switch r {
+		case '\n':
+			newlines++
+			if newlines >= 2 {
+				return true
+			}
+		case '\r', ' ', '\t', '\v', '\f':
```
elecmonkey, line 139: With `allowLineSeparatedGroups: true`, blank lines made from CR-only line endings or U+2028/U+2029 are group boundaries in ESLint, but this scanner increments only for `\n` (`\r` is treated as ordinary whitespace and the Unicode separators reset the count). For example, ESLint accepts `var obj = {\r b: 1,\r\r a: 2\r};`, whereas this implementation compares `a` against `b` and reports it; the same divergence occurs with two U+2028 or U+2029 separators.

internal/rules/sort_keys/sort_keys.go
elecmonkey, line 68: For numeric keys whose JavaScript string form uses exponent notation, `GetStaticPropertyName` does not match ESLint's `String(node.value)` semantics. For example, ESLint names the key `1e-7`, so it reports `{ 1e-7: 1, '1a': 2 }` as unsorted; the shared normalization formats that key as `0.0000001`, causing this rule to accept the object. Large values such as `1e21` and negative zero have analogous formatting mismatches, so numeric property names need JavaScript-compatible number-to-string conversion here.

elecmonkey (commented):
Found a few ESLint parity issues in this port.

swwind:
Blocked by #1705 according to some unicode related issues found.

swwind:
Blocked by #1766 to provide `String.prototype.toLowerCase`.

internal/rules/sort_keys/natural_compare.go
```diff
@@ -0,0 +1,116 @@
...
+//
+// Operates on Unicode code points; upstream compares by UTF-16 code unit, so
+// non-BMP (astral) characters may order differently in rare cases.
+//
+// Returns -1 if a < b, 0 if a == b, 1 if a > b.
+func naturalCompare(a, b string) int {
+	if a == b {
+		return 0
+	}
+	ra := []rune(a)
```
elecmonkey: When `natural: true` is used with escaped lone-surrogate keys, converting the strings to `[]rune` replaces each invalid WTF-8 surrogate encoding with the same sequence of `U+FFFD` values. For example, ESLint 10.8.1 reports `{ '\uD801': 1, '\uD800': 2 }` in ascending natural order, while Rslint treats the two names as equal and reports nothing. Iterate over JavaScript UTF-16 code units, or otherwise preserve lone surrogate values, rather than decoding them as Go runes.

swwind: You are blessed.

swwind: Fixed. A lone surrogate is carried as CESU-8 in a key's name, so `[]rune` turned each of the three bytes into U+FFFD and made every lone surrogate compare equal to every other one.

Comparison now goes through `ecmascript.StringCodeUnits`, which reads that encoding back into the code unit it stands for. The rule also drops its own natural-compare copy for `utils.NaturalCompare`: #1766 already ported natural-compare-lite there on code units, so the local copy was both redundant and the weaker of the two.

The default (non-`natural`) comparison used `strings.Compare`, which is code point order — the divergence the "Differences from ESLint" note recorded. It now uses `ecmascript.CompareStrings`, so every mode compares by UTF-16 code unit and the note is gone.

elecmonkey (approved):
The current PR head matches ESLint 10.8.1 across the upstream suite and targeted differential cases covering computed keys, destructuring, grouping, Unicode, numeric keys, options, and diagnostic ranges. Relevant Go tests, dependent-rule tests, vet, lint, and required CI checks also pass.

fansenze (approved):

PR #1483, opened 2026-07-31
feat(unicorn): port rule prefer-node-protocol

elecmonkey:
## Summary

Port the `prefer-node-protocol` rule from ESLint to rslint.

Enforces the `node:` protocol when importing Node.js builtin modules, across `import`/`export` sources, dynamic `import()`, `require()`, `process.getBuiltinModule()`, and TypeScript `import(...)` type nodes. A specifier is flagged only when its bare name is a Node.js builtin with a valid `node:` counterpart, and is autofixed by inserting `node:` before the module name.

## Related Links

- ESLint rule: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-node-protocol.md
- Source code: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/rules/prefer-node-protocol.js

## Checklist

- [x] Tests updated (or not required).
- [x] Documentation updated (or not required).

PR #1478, opened 2026-07-31
feat: port rule unicorn/error-message

swwind:
## Summary

Port the `unicorn/error-message` rule from ESLint to rslint.

Enforce passing a `message` value when creating a built-in error.

## Related Links

- ESLint rule: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/error-message.md
- Source code: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/rules/error-message.js

## Checklist

- [x] Tests updated (or not required).
- [x] Documentation updated (or not required).

PR #1733, opened 2026-08-14
feat: port rule unicorn/catch-error-name

rockm3:
## Summary

Port the `unicorn/catch-error-name` rule from ESLint to rslint.

Enforce consistent error parameter names in catch clauses and Promise rejection handlers, including configurable ignore patterns and collision-safe autofixes.

## Related Links

- ESLint rule: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v73.0.0/docs/rules/catch-error-name.md
- Source code: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v73.0.0/rules/catch-error-name.js

## Checklist


internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
```diff
@@ -0,0 +1,199 @@
...
+	match, ok := unicornutil.MatchDotMethodCall(function.Parent, unicornutil.DotMethodCallOptions{
+		Method:              method,
+		ArgumentsLength:     &argumentsLength,
+		AllowOptionalMember: true,
+	})
+	return ok && match.Call.Arguments()[1] == function
+}
+
+func availableName(identifier *ast.Node, references []*ast.Node, expected string) string {
+	for candidate := expected; ; candidate += "_" {
```
swwind: Upstream's `getAvailableVariableName` runs `isValidIdentifier` (which includes the TS reserved-word list) before the loop: if the name is invalid it appends one `_`, and if it is still invalid it returns `undefined` so the rule reports without a fix. That step is missing here, so any `name` that is not a valid identifier gets written into the source verbatim:

- `{"name": "_){} evilCode; if(false"}` → `try {} catch (_){} evilCode; if(false) {}`
- `{"name": "class"}` → `try {} catch (class) {}`
- `{"name": "has_space_after "}` → the fix never converges, appending a space to the candidate on every pass

Upstream has test cases for all three.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
```diff
@@ -0,0 +1,199 @@
...
+		available := !utils.IsShadowed(identifier, candidate)
```
swwind: `utils.IsShadowed` only looks at declarations, whereas upstream's `isSafeName` is `resolveVariableName(name, scope) || isUnresolvedName(name, scope)` — a candidate that is *referenced but unresolved* in the scope (a global or an ambient declaration) is unsafe too.

`try {} catch (err) { error(err) }`: ESLint produces `error_`, this produces `error`, and the fixed code `error(error)` has captured the global `error`.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
```diff
@@ -0,0 +1,199 @@
...
+				}
+				declaration := bindingDeclaration(identifier)
+				if declaration == nil || ctx.Refs == nil || declaration.Symbol() == nil {
+					return
+				}
+				references := ctx.Refs.References(declaration.Symbol())
+				if originalName == "_" && len(references) == 0 {
+					return
+				}
+				fixedName := availableName(identifier, references, opts.name)
```
swwind: When the parameter has no references, `references` is empty and the only check left is the one on the declaration site. Walking up from a parameter skips the function body via `isDirectParameterOf`, and the body is only scanned for `var`, so a `const`/`let`/`class` in the body is invisible.

`promise.catch(bad => { const error = 1; use(error) })`: ESLint produces `error_`, this produces `error`, and the fixed `promise.catch(error => { const error = 1; use(error) })` is a SyntaxError (`Identifier 'error' has already been declared`). Catch clauses are unaffected because the catch block is its own scope.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.md
```diff
@@ -0,0 +1,20 @@
+# catch-error-name
```
swwind, line 1: The other unicorn rule docs all use `## Rule Details` + incorrect/correct code blocks + `## Original Documentation` (see `error_message.md`, `no_thenable.md`). The `## Options` / `## Upstream` structure here is inconsistent with them — worth aligning.

swwind (commented):
Overall a solid port. I exercised the `then`/`catch` matching, the `_` special case, `ignore`, shorthand expansion, and most of the shadow-avoidance scenarios (#561, enclosing-parameter shadowing, function-expression self-name shadowing) — all match ESLint. The problems are concentrated in `availableName`, which is missing the two pre-checks upstream's `getAvailableVariableName` performs.

| Code | ESLint | This PR |
| --- | --- | --- |
| `try {} catch (e) {}` + `{"name": "_){} evilCode; if(false"}` | reports, no fix | fixes to `try {} catch (_){} evilCode; if(false) {}` |
| `try {} catch (e) {}` + `{"name": "has_space_after "}` | reports, no fix | fix never converges, still appending spaces after 10 passes |
| `try {} catch (bad) {}` + `{"name": "class"}` | `catch (class_)` | `catch (class)`, a syntax error |
| `try {} catch (err) { error(err) }` | `catch (error_) { error(error_) }` | `catch (error) { error(error) }` |
| `promise.catch(bad => { const error = 1; use(error) })` | `promise.catch(error_ => …)` | `promise.catch(error => …)`, a SyntaxError |

Separately, `packages/rslint/src/config/presets/unicorn.ts` still has `// 'unicorn/catch-error-name': 'error', // not implemented`. This rule is `recommended` upstream, and the previous unicorn ports (#1478 / #1479 / #1483) all flipped that line in the same PR.

The first three rows all have dedicated upstream test cases (`// Should not run into an infinity loop`). The migrated upstream suite here is 76 lines against the original's 977, which is a large trim — worth restoring the invalid-`name` and shadow-avoidance cases.

rockm3:
Fixed the review findings in c6b5189:

- Reject invalid configured identifiers and omit autofixes when appending one underscore still cannot make them valid.
- Suffix TypeScript/JavaScript reserved words such as `class`.
- Include unresolved references and declarations inside the handler body when selecting a collision-free name.
- Enable `unicorn/catch-error-name` in the Unicorn recommended preset.
- Add focused Go and JS regressions for invalid names, reserved words, unresolved calls, and handler-local declarations.

Verified with focused Go tests, the rebuilt binary + JS integration test, full typecheck, lint, and spellcheck.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
```diff
@@ -0,0 +1,267 @@
...
+}
+
+func bodyHasNameConflict(ctx rule.RuleContext, body *ast.Node, originalSymbol *ast.Symbol, candidate string) bool {
+	conflict := false
+	var walk func(*ast.Node)
+	walk = func(node *ast.Node) {
+		if node == nil || conflict {
+			return
+		}
+		if node.Kind == ast.KindIdentifier && node.AsIdentifier().Text == candidate {
```
elecmonkey: When the handler contains a same-named declaration in a nested scope, this whole-body walk treats it as a collision even though renaming the parameter cannot capture that binding. For example, ESLint fixes `try {} catch (bad) { function f() { const error = 1 } }` to `catch (error)`, while this implementation produces `catch (error_)`; the same mismatch occurs with nested blocks, loop bindings, nested catches, and nested function parameters. Check only bindings visible from the parameter or its actual references rather than every identifier below the body.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
```diff
@@ -0,0 +1,267 @@
...
+	}
+	return result
+}
+
+func upperFirst(value string) string {
+	if value == "" {
+		return value
+	}
+	runes := []rune(value)
+	runes[0] = []rune(strings.ToUpper(string(runes[0])))[0]
```
elecmonkey: When `name` starts with an astral Unicode character that has a case mapping, Go's rune-based `upperFirst` differs from upstream's UTF-16 `charAt(0).toUpperCase() + slice(1)`. With `{name: "𐐨"}`, ESLint reports and fixes `catch (descriptive𐐀)`, but this implementation accepts it because it derives `𐐀` as the uppercase suffix. Preserve upstream's UTF-16 code-unit behavior when computing the alternate suffix.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
rockm3: Fixed in 49d8c9a. Candidate conflicts are now limited to lexical scopes containing the parameter or an actual reference, so unrelated nested declarations no longer force a suffix; nested scopes that reference the renamed binding still participate. Added regressions for both cases.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
rockm3: Fixed in 49d8c9a. upperFirst now mirrors the upstream UTF-16 charAt(0) behavior for astral initials, with a Deseret regression test.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.md
rockm3: Aligned in 49d8c9a: added Rule Details, incorrect/correct examples, and Original Documentation sections matching the other Unicorn rule docs.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
```diff
@@ -0,0 +1,320 @@
...
+func bodyHasNameConflict(ctx rule.RuleContext, body *ast.Node, scopes map[*ast.Node]bool, originalSymbol *ast.Symbol, candidate string) bool {
+	conflict := false
+	var walk func(*ast.Node)
+	walk = func(node *ast.Node) {
+		if node == nil || conflict {
+			return
+		}
+		if node.Kind == ast.KindIdentifier && node.AsIdentifier().Text == candidate {
+			if scopes[nearestLexicalScope(node)] {
+				if symbol := utils.BindingNameSymbol(node); symbol != nil {
```
swwind: `BindingNameSymbol` / `IsNonReferenceIdentifier` do not cover every identifier that is neither a value binding nor a reference, so several ordinary shapes are treated as collisions:

- `try {} catch (err) { log({error: err}) }` → `error_` (object-literal property key; nested keys too, e.g. `log({a: {error: err}})`)
- `try {} catch (err) { use(err); class A { error = 1 } }` → `error_` (class field name; a method name `error() {}` is handled correctly, only `PropertyDeclaration` is not)
- `try {} catch (err) { use(err); interface error {} }` and `type error = string` → `error_` (type space cannot collide with a value binding)

ESLint returns `error` for all of these. The first one is the one that will be hit in practice — `log({error: err})` is a very common catch-block shape.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
```diff
@@ -0,0 +1,320 @@
...
+	default:
+		return false
+	}
+}
+
+func nearestLexicalScope(node *ast.Node) *ast.Node {
+	return ast.FindAncestor(node.Parent, isLexicalScope)
+}
+
+func relevantScopes(identifier *ast.Node, references []*ast.Node, body *ast.Node) map[*ast.Node]bool {
```
swwind: Upstream's unsafe check is `resolveVariableName(name, scope) || isUnresolvedName(name, scope)` over `[variable.scope, ...references.map(r => r.from)]`, and `isUnresolvedName` walks `getScopes(scope)` — the scope **and all of its descendants**. Restricting the walk to scopes on a reference's ancestor chain drops the descendants that hold no reference:

- `try {} catch (bad) { use(bad); function f() { return error } }` → ESLint `error_`, this PR `error`, so the fixed code has `f` returning the catch parameter instead of the global.
- Same for `(() => error)()`, `(() => typeof error)()`, and the Promise form `promise.catch(bad => { use(bad); function f() { return error } })`.

A resolved declaration in a descendant scope must still be ignored (that was elecmonkey's case), so the two halves need to be separated: declarations only upwards from each relevant scope, unresolved references downwards through every descendant.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
```diff
@@ -0,0 +1,320 @@
...
+	"github.com/microsoft/typescript-go/shim/ast"
+	"github.com/microsoft/typescript-go/shim/scanner"
+	"github.com/web-infra-dev/rslint/internal/plugins/unicorn/unicornutil"
+	"github.com/web-infra-dev/rslint/internal/rule"
+	"github.com/web-infra-dev/rslint/internal/utils"
+)
+
+const messageID = "catch-error-name"
+
+var reservedWords = map[string]bool{
```
swwind, line 22: The list matches upstream's `typescriptReservedWords` exactly, but upstream also runs `isIdentifier(name)` with `checkReserved`, which rejects a few names this list does not carry — `await` is the notable one, and `scanner.IsValidIdentifier("await")` accepts it:

- `{"name": "await"}` on `try {} catch (bad) { use(bad) }` → ESLint `await_`, this PR `catch (await)`, a syntax error in module code.

Related, from the scope side rather than the word list: `{"name": "undefined"}`, `"NaN"`, `"eval"` and `"arguments"` all get a `_` from ESLint because they resolve in the global (or function) scope, while this PR writes them verbatim — `promise.catch(function (arguments) { … })` is likewise a syntax error in strict code.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
```diff
@@ -0,0 +1,320 @@
...
+			}
+			if re, err := utils.CompileRegexp2(pattern, utils.JSUnicodeRegexOptions); err == nil {
+				result.ignore = append(result.ignore, re)
+			}
+		}
+	}
+	return result
+}
+
+func upperFirst(value string) string {
```
swwind, line 111: `strings.ToUpper` applies rune-wise simple case mapping, while JS `toUpperCase` applies special casing that can grow the string (`ß` → `SS`, `ﬁ` → `FI`). With `{"name": "ßad"}`, `try {} catch (xSSad) {}` is accepted by ESLint (`upperFirst` gives `SSad`) but reported here. Minor, and the same family as the astral case already handled just above.

swwind (commented):
Re-reviewed after c6b5189 and 49d8c9a. The invalid-`name` pre-check, the reserved-word suffix, the preset flip and the doc restructure all look right, and elecmonkey's nested-declaration and astral-`upperFirst` findings are resolved. What remains is concentrated in the new `relevantScopes` / `bodyHasNameConflict` pair, which diverges from upstream's `isSafeName` in both directions: it counts identifiers that are not bindings at all, and it stops scanning at scopes that happen to contain a reference.

| Code | ESLint | This PR |
| --- | --- | --- |
| `try {} catch (err) { log({error: err}) }` | `error` | `error_` |
| `try {} catch (err) { use(err); class A { error = 1 } }` | `error` | `error_` |
| `try {} catch (err) { use(err); type error = string }` | value/type spaces do not collide | `error_` |
| `try {} catch (bad) { use(bad); function f() { return error } }` | `error_` | `error`, captures the global in `f` |
| `promise.catch(bad => { use(bad); function f() { return error } })` | `error_` | `error` |
| `try {} catch (bad) { use(bad) }` + `{"name": "await"}` | `await_` | `await`, a syntax error in a module |

One divergence goes the other way and is worth keeping: `try {} catch (bad) { const error = 1 }` (parameter unused) gets `error` from ESLint, whose own fix is an early SyntaxError, while this PR gets `error_`. Please lock that in as a test case so it does not read as an accident.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
rockm3: Fixed in 8559247. Collision scanning now ignores non-reference identifiers and same-file value declarations; regressions cover object keys, class fields, and type-only declarations.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
rockm3: Fixed in 8559247. External/unresolved references are scanned throughout all handler descendants, while local descendant declarations remain ignored. Added catch and Promise nested-function regressions.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
rockm3: Fixed in 8559247. Added await to the rejected identifier set, account for configured/language globals through ctx.Globals, and handle implicit arguments in function scopes. Added regressions for await, undefined, and arguments.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
rockm3: Fixed in 8559247 using x/text full Unicode uppercasing for BMP initials while preserving the UTF-16 astral behavior. Added the ßad → SSad acceptance regression.

rockm3:
Addressed the latest re-review in 8559247. Also documented and locked in the intentional safety divergence for an unused parameter colliding with a direct handler-body declaration. Focused Go tests, rebuilt-binary JS tests, and full typecheck pass.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name_extras_test.go
```diff
@@ -0,0 +1,72 @@
...
+			// Nested scopes without references to the renamed parameter do not collide.
+			invalid("try {} catch (bad) { function f() { const error = 1 } }", "try {} catch (error) { function f() { const error = 1 } }", "bad", "error"),
+			invalid("try {} catch (bad) { { const error = 1 } }", "try {} catch (error) { { const error = 1 } }", "bad", "error"),
+			// A nested scope that references the renamed parameter still participates.
+			invalid("try {} catch (bad) { function f() { error(bad) } }", "try {} catch (error_) { function f() { error(error_) } }", "bad", "error_"),
+			// Property keys, class fields, and type-only declarations are not value bindings.
+			invalid("try {} catch (err) { log({error: err}) }", "try {} catch (error) { log({error: error}) }", "err", "error"),
+			invalid("try {} catch (err) { use(err); class A { error = 1 } }", "try {} catch (error) { use(error); class A { error = 1 } }", "err", "error"),
+			invalid("try {} catch (err) { use(err); type error = string }", "try {} catch (error) { use(error); type error = string }", "err", "error"),
```
swwind: This expectation is wrong, and it is mine to correct: I sourced it from a run with the default parser, which cannot parse `type error = string`. Under `@typescript-eslint/parser` ESLint gives `error_` here, and likewise for `interface error {}` and for either declaration at module level.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
```diff
@@ -0,0 +1,284 @@
...
+	if !isValidVariableName(candidate) {
+		candidate += "_"
+		if !isValidVariableName(candidate) {
+			return ""
+		}
+	}
+	originalSymbol := bindingDeclaration(identifier).Symbol()
+	body := handlerBody(identifier)
+	for {
+		available := !ctx.Globals.Access(candidate).IsDeclared() && !utils.IsShadowed(identifier, candidate)
```
swwind: `utils.IsShadowed` models the value space, but typescript-eslint's scope manager registers type aliases and interfaces as ordinary variables in `scope.set`, so `resolveVariableName` resolves them and the candidate is unsafe:

- `try {} catch (err) { use(err); type error = string }` → ESLint `error_`, this PR `error`
- `try {} catch (err) { use(err); interface error {} }` → same
- `type error = string; try {} catch (bad) { use(bad) }` → same
- `interface error {} ; try {} catch (bad) { use(bad) }` → same

`enum`, `namespace`, a type-only `import {error}` and type *references* all resolve correctly already, so this is only the `type`/`interface` declaration forms. Since widening `HasShadowingDeclaration` / `HasLocalDeclarationInStatements` would change every rule that uses them, a rule-local check for the two declaration kinds is probably the right scope here.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
```diff
@@ -0,0 +1,284 @@
...
+		if available && candidate == "arguments" && ast.FindAncestor(identifier.Parent, ast.IsFunctionLikeDeclaration) != nil {
```
swwind: The `arguments` guard only fires inside a function, but upstream has no such condition — `resolveVariableName('arguments', scope)` succeeds from any scope, so `getAvailableVariableName` always suffixes it. With `{"name": "arguments"}`:

- `try {} catch (bad) { use(bad) }` → ESLint `arguments_`, this PR `catch (arguments)`, a SyntaxError in module (or any other strict) code.

ESLint gives `arguments_` under both `sourceType: "module"` and `"script"`, so dropping the `ast.FindAncestor` condition and treating `arguments` as always unavailable matches. The existing test covers the inside-a-function half only.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
```diff
@@ -0,0 +1,284 @@
...
+					break
+				}
+			}
+		}
+		if available && bodyHasExternalReference(ctx, body, originalSymbol, candidate) {
+			available = false
+		}
+		// NOTE: Unlike ESLint, avoid a known unsafe fix when an unused handler
+		// parameter would be renamed to a direct lexical declaration in its body.
+		if available && len(references) == 0 && body != nil && body.Kind == ast.KindBlock &&
```
swwind: `HasShadowingDeclaration` only looks at the block's direct statements, whereas the ESLint check this stands in for resolves the whole function scope, hoisted `var`s included:

- `promise.catch(bad => { if (x) { var error = 1 } })` → ESLint `error_`, this PR `error`
- `promise.catch(function (bad) { if (x) { var error = 1 } })` → same

`promise.catch(bad => { var error = 1 })` already matches because the declaration is a direct statement, and the catch-clause form is fine because the `var` hoists to the source file where `IsShadowed` finds it. So only a Promise handler with an unused parameter and a nested `var` is affected, and the output is still valid JS — cosmetic. A `HasHoistedVarDeclaration(body, candidate)` next to the existing call would close it.

swwind (commented):
Third pass, after 8559247. The object-key / class-field / nested-descendant / `await` / `undefined` / `arguments`-in-a-function / `ßad` cases from the previous round are all resolved. I re-ran the full test suite of this PR plus ~40 new shapes against ESLint with `@typescript-eslint/parser`, and three things are left.

| Code | ESLint | This PR |
| --- | --- | --- |
| `try {} catch (err) { use(err); type error = string }` | `error_` | `error` |
| `try {} catch (err) { use(err); interface error {} }` | `error_` | `error` |
| `type error = string; try {} catch (bad) { use(bad) }` | `error_` | `error` |
| `interface error {} ; try {} catch (bad) { use(bad) }` | `error_` | `error` |
| `try {} catch (bad) { use(bad) }` + `{"name": "arguments"}` | `arguments_` | `arguments`, a syntax error in strict code |
| `promise.catch(bad => { if (x) { var error = 1 } })` | `error_` | `error` |

The first four are my fault: the `error` I gave you for the type-space row last round came from a run with the default parser, which cannot parse those cases at all. With `@typescript-eslint/parser`, `type` and `interface` declarations get a variable in the same `scope.set` as values, so `resolveVariableName` finds them. Note the rule is already inconsistent with itself here — `enum error {}`, `namespace error {}` and `import type {error} from './x'` all correctly produce `error_`; only `type` and `interface` slip through.

The intentional divergence on `try {} catch (bad) { const error = 1 }` still holds under this parser — ESLint picks `error` and its own fix is a SyntaxError — so that one is good as it stands.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name_extras_test.go
rockm3: Corrected in 473e301. The regression now expects error_, with additional interface and module-scope type/interface cases.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
rockm3: Fixed in 473e301 with a rule-local type/interface scope check, leaving the shared value-space shadowing helpers unchanged. Added all four suggested regressions.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
rockm3: Fixed in 473e301. arguments is now always treated as unavailable, with a top-level catch-clause regression.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
rockm3: Fixed in 473e301 by checking HasHoistedVarDeclaration for unused Promise rejection parameters. Added both arrow-function and function-expression regressions.

rockm3:
Addressed the third review pass in 473e301: type/interface declarations now participate in candidate collision checks, arguments is always suffixed, and nested var declarations in unused Promise handlers are detected through hoisting. Added focused regressions for every reported shape. Verified focused Go tests, rebuilt-binary JS integration tests, and full typecheck.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
```diff
@@ -0,0 +1,317 @@
...
+			return true
+		}
+	}
+	return false
+}
+
+func hasTypeDeclarationInScope(node *ast.Node, name string) bool {
+	for current := node; current != nil; current = current.Parent {
+		switch current.Kind {
+		case ast.KindBlock:
```
swwind, line 256: The scope walk recognizes only `Block` and `SourceFile`, but typescript-eslint also opens a scope for a module/namespace body and for a switch statement, and registers type parameters as variables. So a `type`/`interface` declared in a `namespace` body, or as a bare statement in a `case` clause, is invisible here, and a type parameter never is:

- `namespace N { type error = string; try {} catch (bad) { use(bad) } }` → ESLint `error_`, this PR `error` (same for `interface`, and for an unused parameter)
- `switch (x) { case 1: type error = string; try {} catch (bad) { use(bad) } }` → ESLint `error_`, this PR `error`
- `function f<error>() { try {} catch (bad) { use(bad) } }` → ESLint `error_`, this PR `error`

Adding `ast.KindModuleBlock` and `ast.KindCaseClause`/`ast.KindDefaultClause` to the switch covers the first two; type parameters would need a separate check on the enclosing function/class.

swwind (commented):
Fourth pass, after 473e301. The `type`/`interface` scope check, the unconditional `arguments` suffix, and the hoisted-`var` check all land correctly — I re-verified every case from the previous three rounds plus the option/matching arms (`ignore`, `then`/`catch` arity, optional chaining, computed `catch`, `_`, trailing-underscore stripping) and they all match ESLint now.

One family is left, and it is the same one 473e301 addressed, just not applied everywhere: `hasTypeDeclarationInScope` treats only `Block` and `SourceFile` as scopes.

| Code | ESLint | This PR |
| --- | --- | --- |
| `namespace N { type error = string; try {} catch (bad) { use(bad) } }` | `error_` | `error` |
| `namespace N { interface error {}; try {} catch (bad) { use(bad) } }` | `error_` | `error` |
| `namespace N { type error = string; try {} catch (bad) {} }` | `error_` | `error` |
| `switch (x) { case 1: type error = string; try {} catch (bad) { use(bad) } }` | `error_` | `error` |
| `function f<error>() { try {} catch (bad) { use(bad) } }` | `error_` | `error` |

The braced form `switch (x) { case 1: { type error = string; … } }` already matches, as do function bodies, method bodies, `export type`, and the descendant-scope cases — so this is only `ModuleBlock`, bare case-clause statement lists, and type parameters.

Worth noting these are cosmetic: a type-space name can never collide with a catch parameter or a rejection-handler parameter, since both always bind in an inner scope. Every row above produces valid TypeScript under either name. So this is a parity gap rather than a broken fix, and I have no objection to it landing as is if you would rather close it in a follow-up.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
rockm3: Fixed in c0638a9. The type-scope walk now covers ModuleBlock, bare case/default clause statement lists, and enclosing function/class type parameters. Added regressions for all five reported shapes; focused Go tests, rebuilt-binary JS integration tests, and full typecheck pass.

rockm3:
Addressed the fourth review pass in c0638a9 by covering namespace bodies, bare switch clauses, and enclosing function/class type parameters in the rule-local type declaration scope walk. Added all five requested regressions and verified focused Go tests, rebuilt-binary JS integration tests, and full typecheck.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
```diff
@@ -0,0 +1,336 @@
...
+			if block != nil && block.Statements != nil && statementsHaveTypeDeclaration(block.Statements.Nodes, name) {
+				return true
+			}
+		case ast.KindModuleBlock:
+			moduleBlock := current.AsModuleBlock()
+			if moduleBlock != nil && moduleBlock.Statements != nil &&
+				statementsHaveTypeDeclaration(moduleBlock.Statements.Nodes, name) {
+				return true
+			}
+		case ast.KindCaseClause, ast.KindDefaultClause:
```
swwind, line 269: All clauses of a switch share one lexical scope, so a `type`/`interface` in a sibling clause is in scope here too — `switch (x) { case 1: type error = string; default: try {} catch (bad) { use(bad) } }` gets `error_` from ESLint and `error` here. Handling `ast.KindCaseBlock` and scanning every clause, the way `utils.HasShadowingDeclarationInCaseBlock` does for value declarations, covers both this and the containing clause.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
```diff
@@ -0,0 +1,336 @@
...
+		if available && bodyHasExternalReference(ctx, body, originalSymbol, candidate) {
+			available = false
+		}
+		// NOTE: Unlike ESLint, avoid a known unsafe fix when an unused handler
+		// parameter would collide with a declaration in its body.
+		if available && len(references) == 0 && body != nil && body.Kind == ast.KindBlock {
+			block := body.AsBlock()
+			available = !utils.HasShadowingDeclaration(body, candidate) &&
+				!utils.HasHoistedVarDeclaration(body, candidate) &&
+				(block == nil || block.Statements == nil || !statementsHaveTypeDeclaration(block.Statements.Nodes, candidate))
```
swwind: `statementsHaveTypeDeclaration` is wider than the unsafe fix this guard exists for: `try {} catch (error) { type error = string }` and `interface error {}` typecheck clean, so ESLint's `error` is right and this yields `error_`. `HasShadowingDeclaration` brings in `enum` for the same reason — `enum error {}` emits `var error`, which is allowed next to a simple catch parameter. Only `const`/`let`/`class`/`function` make the fix a SyntaxError. Worth keeping for rejection handlers though: `promise.catch(bad => { type error = string })` is `error_` in ESLint, because a function scope does contain its body's declarations.

swwind (commented):
Fifth pass, after c0638a9. The `ModuleBlock` / bare-case-clause / type-parameter coverage all land, and I re-ran every case from the previous four rounds plus ~55 new shapes — nested and duplicate catches, static blocks, getters, object methods, references in default parameters, `import` / `import type` / `import =`, `enum`, `namespace`, `declare global`, `declare const`, labels, typed parameters, `is-identifier` edge names (ZWNJ, CJK, `$`, astral, empty `name`), and the parameter shapes that must not match (defaults, rest) — all match ESLint. Two things are left, both in the TypeScript type space.

| Code | ESLint | This PR |
| --- | --- | --- |
| `switch (x) { case 1: type error = string; default: try {} catch (bad) { use(bad) } }` | `error_` | `error` |
| `switch (x) { case 1: interface error {}; default: try {} catch (bad) { use(bad) } }` | `error_` | `error` |
| `switch (x) { case 1: try {} catch (bad) { use(bad) }; case 2: type error = string }` | `error_` | `error` |
| `try {} catch (bad) { type error = string }` | `error` | `error_` |
| `try {} catch (bad) { interface error {} }` | `error` | `error_` |
| `try {} catch (bad) { enum error {} }` | `error` | `error_` |

The first three are the same shape as the last round, one level out: all clauses of a switch share a single lexical scope, so a sibling clause's `type`/`interface` is in scope at the catch too.

The last three are the unused-parameter guard reaching past the unsafe fix it exists for. `try {} catch (error) { type error = string }`, `interface error {}` and `enum error {}` all typecheck clean — type and value spaces do not collide, and the enum emit (`var error`) is permitted next to a simple catch parameter — so ESLint's `error` is correct in all three. Only `const`/`let`/`class`/`function` turn the fix into a SyntaxError, which is the divergence the doc describes.

Both are cosmetic: every row above is valid under either name. No objection to landing as is and closing them in a follow-up.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
```diff
@@ -0,0 +1,336 @@
...
+		candidate += "_"
+		if !isValidVariableName(candidate) {
+			return ""
+		}
+	}
+	originalSymbol := bindingDeclaration(identifier).Symbol()
+	body := handlerBody(identifier)
+	for {
+		available := !ctx.Globals.Access(candidate).IsDeclared() &&
+			!utils.IsShadowed(identifier, candidate) &&
```
elecmonkey, line 332: Namespace bodies still miss value-space bindings when picking the autofix name.

Repro: `namespace N { const error = 1; try {} catch (bad) { use(bad) } }`
Upstream (`eslint-plugin-unicorn@73.0.0` with `@typescript-eslint/parser`) fixes to `catch (error_)`; this PR fixes to `catch (error)`, which shadows the namespace-level `const error`.

The current walk checks `ModuleBlock` only through `hasTypeDeclarationInScope`, while `utils.IsShadowed` does not treat a `ModuleBlock` as a scope boundary. The same gap applies to namespace-level functions, classes, enums, nested namespaces, and `import =` declarations. Please use the same complete binding lookup for module-block statement lists that top-level scopes already get, rather than only checking type/interface declarations.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
```diff
@@ -0,0 +1,336 @@
...
+
+func bodyHasExternalReference(ctx rule.RuleContext, body *ast.Node, originalSymbol *ast.Symbol, candidate string) bool {
+	conflict := false
+	var walk func(*ast.Node)
+	walk = func(node *ast.Node) {
+		if node == nil || conflict {
+			return
+		}
+		if node.Kind == ast.KindIdentifier && node.AsIdentifier().Text == candidate &&
+			!utils.IsNonReferenceIdentifier(node) {
```
elecmonkey, line 217: Intrinsic JSX tag names should not count as external references here.

Repro in TSX: `try {} catch (bad) { use(bad); const x = <error /> }`
Upstream fixes the parameter to `error`, but this PR fixes it to `error_`. The lowercase `<error />` tag is an intrinsic JSX element name, not a variable reference, so it should not make `error` unavailable.

`ctx.Refs.Resolve` already excludes intrinsic JSX tags, but this code still reaches the conflict branch because `utils.IsNonReferenceIdentifier` does not have the JSX tag-name exclusion that `RefStore` has. Skipping intrinsic JSX tag identifiers before checking unresolved external references should align this with upstream.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
rockm3: Fixed in e18ed19. The CaseBlock scope now scans type/interface declarations across every sibling clause, with regressions for both declaration kinds and both source orders.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
rockm3: Fixed in e18ed19. The unused catch-parameter safety guard is now limited to direct let/const/using, class, and function declarations. Type aliases, interfaces, and enums keep the upstream name; Promise handlers retain the wider function-scope collision check.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
rockm3: Fixed in e18ed19. ModuleBlock statement lists now use the complete local value-binding lookup in addition to the rule-local type/interface check. Added namespace-level const and function regressions.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
rockm3: Fixed in e18ed19. Intrinsic JSX tag names are skipped before unresolved external-reference checks, matching RefStore behavior. Added the reported TSX regression.

rockm3:
Addressed the fifth review pass and the two additional findings in e18ed19: shared switch-clause type scopes, precise unused catch-parameter safety checks, complete namespace value bindings, and intrinsic JSX tag handling. Added focused regressions for every reported shape. Go tests, rebuilt-binary JS integration tests, and full typecheck pass.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
```diff
@@ -0,0 +1,379 @@
...
+			if block != nil && block.Statements != nil && statementsHaveTypeDeclaration(block.Statements.Nodes, name) {
+				return true
+			}
+		case ast.KindModuleBlock:
+			moduleBlock := current.AsModuleBlock()
+			if moduleBlock != nil && moduleBlock.Statements != nil &&
+				(utils.HasLocalDeclarationInStatements(moduleBlock.Statements.Nodes, name) ||
+					statementsHaveTypeDeclaration(moduleBlock.Statements.Nodes, name)) {
+				return true
+			}
```
swwind, line 268: `HasLocalDeclarationInStatements` only sees the namespace body's direct statements, but `var` hoists to the namespace scope out of any nested block, loop or `try`. `utils.HasHoistedVarDeclaration(current, name)` here covers it — it already stops at nested `ModuleDeclaration` and function boundaries, so it is safe to call on the `ModuleBlock` itself.

With value declarations in it, the function name no longer matches what it does; `hasDeclarationInEnclosingScopes` would read better.

swwind (commented):
Sixth pass, after e18ed19. Shared switch-clause type scopes, the precise unused-catch-parameter check, complete namespace value bindings and intrinsic JSX tag names all land — I re-ran every case from the previous five rounds plus ~30 new shapes (namespace `var` / `class` / `enum` / `namespace` / `import =` / `export`, switch clauses with braced bodies, arrow and class type parameters, `type` / `interface` / `enum` in a rejection handler body, JSX attribute names, member tags and closing tags, `using` declarations, labelled function declarations) and one family is left.

`hasTypeDeclarationInScope`'s `ModuleBlock` branch looks at direct statements only, so a `var` hoisted out of a nested block inside a namespace body is invisible.

| Code | ESLint | This PR |
| --- | --- | --- |
| `namespace N { { var error = 1 }; try {} catch (bad) { use(bad) } }` | `error_` | `error` |
| `namespace N { for (var error of xs) {}; try {} catch (bad) { use(bad) } }` | `error_` | `error` |
| `namespace N { if (x) { var error = 1 }; function f() { try {} catch (bad) { use(bad) } } }` | `error_` | `error` |
| `namespace N { try {} catch (e2) { var error = 1 }; try {} catch (bad) { use(bad) } }` | `error_` | `error` |
| `namespace N { if (x) { var error = 1 }; try {} catch (bad) { use(bad); log(error) } }` | `error_` | `error` |

The last row is not cosmetic: after the fix `log(error)` binds to the catch parameter instead of the namespace's `var`. The same shape at file level is already correct, because `IsShadowed`'s `SourceFile` branch runs `HasHoistedVarDeclaration` — the namespace body is missing that step.

Separately, `try {} catch (bad) { using error = f() }` produces `error_` here and `error` in ESLint. This PR is right — `error` is a TS2492 redeclaration in a catch clause — so it is worth a test case next to the other documented divergence.

elecmonkey:
I found three compatibility/correctness issues in `unicorn/catch-error-name`:

- [P2] Reject `eval` as an invalid candidate name

  When `name` is configured as `eval` and the built-in global is explicitly disabled with `languageOptions.globals: { eval: 'off' }`, the current check accepts the bare name and the fixer emits `catch (eval)`. Upstream's identifier validation rejects `eval` and falls back to `eval_`; the current Rslint output is invalid in modules and other strict-mode code.

- [P2] Include namespaced JSX identifiers in parameter references

  In TSX, references inside `JsxNamespacedName` are currently omitted. For example, `try {} catch (_) { const x = <_:foo /> }` is treated as having no references and is incorrectly exempted, while upstream reports it and fixes both occurrences to `error`. For a non-underscore parameter with another use, the rule reports but leaves the namespaced JSX occurrence unrenamed, producing a partial rename.

- [P2] Preserve RegExp-valued ignore patterns

  Upstream supports JS config such as `ignore: [/^bad$/i]`. After config serialization, that RegExp becomes an object, but the current implementation only accepts string entries and silently discards it. As a result, Rslint reports `catch (Bad)` even though upstream accepts it. Since the schema still permits object entries, the RegExp source and flags need to survive the config boundary instead of being ignored.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
rockm3: Fixed in a4b8676. ModuleBlock scope checks now include HasHoistedVarDeclaration, with nested-block and for-of regressions. I kept the broader helper name for this patch because it now intentionally covers both type and value declarations.

rockm3:
Addressed the sixth review and the three additional compatibility findings in a4b8676: namespace-hoisted var bindings, strict-mode eval validation, namespaced JSX parameter references, and RegExp-valued ignore options. RegExp source/flags are now preserved at both config-module and direct IPC serialization boundaries. Added focused Go/TSX regressions and a real /^bad$/i RuleTester case. Go tests, rebuilt Go and core JS artifacts, JS integration tests, and full typecheck pass.

packages/rslint/src/internal/node.ts
```diff
@@ -170,7 +171,10 @@ export class NodeRslintService implements RslintServiceInterface {
 
   private writeMessage(message: IpcMessage): void {
     if (this.dead) return;
-    const jsonBuffer = Buffer.from(JSON.stringify(message), 'utf8');
+    const jsonBuffer = Buffer.from(
+      JSON.stringify(message, configJSONReplacer),
```
swwind, line 175: This applies to every outgoing IPC message, not just config — a rule port should not be changing the wire format that all rules' options travel over. Please revert this together with `configJSONReplacer` and its use in `config-module-host.ts`.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
```diff
@@ -0,0 +1,424 @@
...
+	if name, ok := values["name"].(string); ok {
+		result.name = name
+	}
+	if patterns, ok := values["ignore"].([]any); ok {
+		for _, value := range patterns {
+			pattern, regexOptions, hasPattern := "", utils.JSUnicodeRegexOptions, false
+			switch value := value.(type) {
+			case string:
+				pattern, hasPattern = value, true
+			case map[string]any:
```
swwind, line 62: Revert to the string-only form along with the JS-side replacer. As a side note, if this branch does come back in a separate PR, the flags-to-`regexp2.RegexOptions` mapping is byte-identical to `validatePattern` in `internal/rules/no_invalid_regexp/no_invalid_regexp.go` and belongs next to `JSRegexOptions` in `internal/utils/regexp2.go`.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.schema.json
```diff
@@ -0,0 +1,26 @@
...
+      "additionalProperties": false,
+      "properties": {
+        "name": {
+          "type": "string",
+          "description": "The expected name for the error variable."
+        },
+        "ignore": {
+          "type": "array",
+          "items": {
+            "type": ["string", "object"],
```
swwind, line 15: Back to `{"type": "string"}` once the object form is gone — as written it also accepts `{"source": …, "flags": …}` spelled out by hand in `rslint.json`, which is not a config shape upstream has.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
```diff
@@ -0,0 +1,424 @@
...
+}
+
+func jsxNamespaceReferences(body *ast.Node, name string) []*ast.Node {
+	var references []*ast.Node
+	var walk func(*ast.Node)
+	walk = func(node *ast.Node) {
+		if node == nil {
+			return
+		}
+		if node.Kind == ast.KindJsxNamespacedName {
```
swwind, line 96: This matches every `JsxNamespacedName` in the handler body, but only a tag-position namespace is a variable reference — an attribute name like `<div bad:attr="1" />` is not one, and renaming it changes the attribute. It also does not check that the namespace binds to this parameter, so a shadowed `<bad:foo />` inside a nested function gets renamed too, and an unused `_` stops being exempt because of an occurrence it does not own. Restricting to `ast.IsJsxTagName(node.Parent)` and skipping names for which `utils.IsShadowed(namespace, name)` holds covers all four rows.

swwind (commented):
Seventh pass, after a4b8676. The namespace-hoisted `var` fix and the `eval` rejection land, and I re-ran every case from the previous six rounds against them. Two things.

First, the RegExp `ignore` support has to come out. `configJSONReplacer` is wired into `NodeRslintService.writeMessage`, so it rewrites every value in every IPC message for every rule, and into `entriesJSON` in the config module host — that changes what all rules receive for a RegExp-valued option, which is a protocol change that should not ride along in a rule port. The `map[string]any` branch in `parseOptions` and the `["string", "object"]` widening in the schema go with it, and the widened schema additionally accepts a hand-written `{"source": …, "flags": …}` in `rslint.json`, which has no upstream meaning. Please revert the whole chain and keep `ignore` string-only; if RegExp-valued options are worth supporting, that is its own PR that settles the boundary for every rule at once.

Second, the new `jsxNamespaceReferences`.

| Code (TSX) | ESLint | This PR |
| --- | --- | --- |
| `try {} catch (_) { const x = <div _:attr="1" /> }` | not reported | reported |
| `try {} catch (bad) { use(bad); const x = <div bad:attr="1" /> }` | `<div bad:attr="1" />` | `<div error:attr="1" />` |
| `try {} catch (bad) { use(bad); const x = <div error:attr="1" /> }` | `error` | `error_` |
| `try {} catch (_) { function g(_) { return <_:foo /> } }` | not reported | reported |
| `try {} catch (bad) { use(bad); [1].forEach(bad => { const x = <bad:foo /> }) }` | inner `<bad:foo />` untouched | renamed to `<error:foo />` |

A JSX namespace is a variable reference only in tag position — `<div xlink:href="…" />` is an attribute name — and the walk also does not check that the namespace it found actually binds to the handler parameter. Everything in tag position is right: `<bad:foo />`, `<bad:foo></bad:foo>`, `promise.catch(_ => <_:foo />)`, `<bad.foo />` and the `<error:foo />` collision all match ESLint.

Minor: `try {} catch (bad) { using error = f() }` still yields `error_` against ESLint's `error`, which is the second intentional divergence from the last round and still has no test.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
```diff
@@ -0,0 +1,424 @@
...
+
+func bodyHasExternalReference(ctx rule.RuleContext, body *ast.Node, originalSymbol *ast.Symbol, candidate string) bool {
+	conflict := false
+	var walk func(*ast.Node)
+	walk = func(node *ast.Node) {
+		if node == nil || conflict {
+			return
+		}
+		if node.Kind == ast.KindIdentifier && node.AsIdentifier().Text == candidate &&
+			!(ast.IsJsxTagName(node) && scanner.IsIntrinsicJsxName(node.Text())) &&
```
swwind, line 216: Same distinction on the collision side: the namespace of an attribute name reaches here as a plain identifier, so `try {} catch (bad) { use(bad); const x = <div error:attr="1" /> }` counts as an external `error` and the fix becomes `error_`, where ESLint picks `error`. The tag-position case is already correct.

packages/rslint/src/internal/node.ts
rockm3: Reverted in 013b5eb. NodeRslintService is back to plain JSON.stringify with no cross-rule wire-format change.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
rockm3: Reverted in 013b5eb. parseOptions is string-only again; the RegExp object and flags handling has been removed.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.schema.json
rockm3: Reverted in 013b5eb. The ignore schema now accepts strings only.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
rockm3: Fixed in 013b5eb. Namespace references are now collected only in JSX tag position and only when no declaration shadows the handler parameter. Added attribute, nested-function, and nested-arrow regressions.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
rockm3: Fixed in 013b5eb. Namespaced JSX attributes are skipped by the external-reference collision walk while tag-position namespaces still participate.

rockm3:
Addressed the seventh review in 013b5eb. Fully reverted the RegExp option protocol/schema changes, restricted namespaced JSX handling to unshadowed tag-position references, excluded namespaced attributes from collision checks, and added the documented using-declaration safety divergence. Focused Go tests, rebuilt-binary JS integration tests, and full typecheck pass.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
```diff
@@ -0,0 +1,413 @@
...
+	"package": true, "private": true, "protected": true, "public": true, "static": true,
+	"yield": true, "any": true, "boolean": true, "constructor": true, "declare": true,
+	"get": true, "module": true, "require": true, "number": true, "set": true,
+	"string": true, "symbol": true, "type": true, "from": true, "of": true, "eval": true,
+}
+
+//go:embed catch_error_name.schema.json
+var schemaJSON []byte
+
+var upperCaser = cases.Upper(language.Und)
```
swwind: Please use `ecmascript.StringToUpperCase()`

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
```diff
@@ -0,0 +1,413 @@
...
+	if name, ok := values["name"].(string); ok {
+		result.name = name
+	}
+	if patterns, ok := values["ignore"].([]any); ok {
+		for _, value := range patterns {
+			pattern, ok := value.(string)
+			if !ok {
+				continue
+			}
+			if re, err := utils.CompileRegexp2(pattern, utils.JSUnicodeRegexOptions); err == nil {
```
swwind: Please use `esregexp.Compile(pattern, "u")`

swwind (commented):
Please also merge origin/main and resolve conflicts (if any)

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
rockm3: Updated in 1c2edc3: `upperFirst` now uses `ecmascript.StringToUpperCase()` while retaining the UTF-16 astral-code-unit behavior required by the upstream rule.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
rockm3: Updated in 1c2edc3: ignore patterns now use `esregexp.Compile(pattern, "u")` and `TestOrTimeout()`.

internal/plugins/unicorn/rules/catch_error_name/catch_error_name.go
```diff
@@ -0,0 +1,410 @@
...
+		return false
+	}
+	parameters := function.Parameters()
+	if len(parameters) == 0 || unicornutil.PlainParameterIdentifier(parameters[0]) != identifier || function.Parent == nil {
+		return false
+	}
+
+	argumentsLength := 1
+	method := "catch"
+	if match, ok := unicornutil.MatchDotMethodCall(function.Parent, unicornutil.DotMethodCallOptions{
```
swwind: Parenthesized rejection handlers are transparent in upstream’s ESTree. For `promise.catch((bad => use(bad)))`, Unicorn v73 reports `bad` and fixes it to `promise.catch((error => use(error)))`, while rslint emits no diagnostic. Here `function.Parent` is a `ParenthesizedExpression`, so both `MatchDotMethodCall` calls fail. Please walk through the callback’s parenthesized wrappers before matching the call and compare against the wrapped argument, then add this reproducer as a regression test.

rockm3: Fixed in 7568808. The callback now walks through its outer parenthesized-expression wrappers before matching `.catch`/`.then`, and the matched argument is compared with that outer wrapper. Added the reported `promise.catch((bad => use(bad)))` regression. Focused Go tests, golangci-lint v2.12.2, rebuilt-binary JS tests, and typecheck all pass.

swwind (approved):
LGTM, thanks for your contribution.

PR #1734, opened 2026-08-14
feat: port rule no-lonely-if

swwind:
## Summary

Port the `no-lonely-if` rule from ESLint to rslint.

Disallows `if` statements as the only statement in `else` blocks, suggesting the `else if` form instead. Includes autofix.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/no-lonely-if
- Source code: https://github.com/eslint/eslint/blob/v10.8.1/lib/rules/no-lonely-if.js

## Checklist

### DIFF
--- a/internal/rules/no_lonely_if/no_lonely_if.go
+++ b/internal/rules/no_lonely_if/no_lonely_if.go
@@ -68,6 +68,13 @@ func buildLonelyIfFix(sf *ast.SourceFile, elseBlock, ifNode *ast.Node) (rule.Rul
 	innerRange := utils.BracedNodeInnerRange(sf, elseBlock)
 	nodeRange := utils.TrimNodeTextRange(sf, ifNode)
 
+	// An unterminated `else {` recovers as a Block without a closing brace, so
+	// the inner if reaches past the brace pair's interior and there is no pair
+	// to unwrap.
+	if nodeRange.Pos() < innerRange.Pos() || nodeRange.End() > innerRange.End() {
+		return rule.RuleFix{}, false
+	}
+
 	// Don't fix if there are any non-whitespace characters interfering (e.g. comments).
 	if !ecmascript.IsBlank(text[innerRange.Pos():nodeRange.Pos()]) ||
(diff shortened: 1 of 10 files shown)


internal/rules/no_lonely_if/no_lonely_if.go
```diff
@@ -0,0 +1,105 @@
...
+func buildLonelyIfFix(sf *ast.SourceFile, elseBlock, ifNode *ast.Node) (rule.RuleFix, bool) {
+	text := sf.Text()
+
+	blockRange := utils.TrimNodeTextRange(sf, elseBlock)
+	innerRange := utils.BracedNodeInnerRange(sf, elseBlock)
+	nodeRange := utils.TrimNodeTextRange(sf, ifNode)
+
+	// Don't fix if there are any non-whitespace characters interfering (e.g. comments).
+	if strings.TrimSpace(text[innerRange.Pos():nodeRange.Pos()]) != "" ||
+		strings.TrimSpace(text[nodeRange.End():innerRange.End()]) != "" {
```
elecmonkey, line 72: When an otherwise fixable block contains a BOM whitespace character, such as `if(a){}else{\uFEFFif(b)c();\uFEFF}`, ESLint 10.8.1 removes the braces because JavaScript `trim()` treats U+FEFF as whitespace, but Go's `strings.TrimSpace` does not. Rslint therefore reports the diagnostic without ESLint's autofix; use the existing ECMAScript whitespace predicate rather than `strings.TrimSpace` for these slices.

swwind:
Blocked by #1766 because of usage of ECMAScript `String.trim()`

elecmonkey (approved):
The implementation matches ESLint 10.8.1 across the reviewed structural, diagnostic, and autofix paths, including comments, ECMAScript whitespace, ASI hazards, dangling-else cases, and TypeScript-specific syntax. The shared brace-helper extraction preserves the existing curly rule behavior, and both focused and full core-rule Go test suites passed.

PR #1736, opened 2026-08-14
feat: port rule no-magic-numbers

swwind:
## Summary

Port the `no-magic-numbers` rule from ESLint to rslint.

Disallows magic numbers — numeric literals used without being assigned to a named constant — with options to ignore array indexes, default values, class field initial values, `enforceConst`, `detectObjects`, and (for TypeScript syntax) enum members, numeric literal types, readonly class properties, and type indexes.

`@typescript-eslint/no-magic-numbers` is refactored into a thin wrapper around this implementation, since its schema and behavior are now fully subsumed by the core rule.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/no-magic-numbers
- Source code: https://github.com/eslint/eslint/blob/v10.8.1/lib/rules/no-magic-numbers.js

## Checklist


internal/rules/no_magic_numbers/no_magic_numbers.go
```diff
@@ -0,0 +1,566 @@
...
+		// computed once, up front, and reused by every check below.
+		fullNumberNode := node
+		fullNumberRange := nodeRange
+		isNegative := false
+		if unary, op := findUnaryParent(node); unary != nil {
+			fullNumberNode = unary
+			fullNumberRange = utils.TrimNodeTextRange(ctx.SourceFile, unary)
+			isNegative = op == ast.KindMinusToken
+		}
+		fullRaw := ctx.SourceFile.Text()[fullNumberRange.Pos():fullNumberRange.End()]
```
elecmonkey: When parentheses occur inside a unary numeric expression (for example, `f(-(1))`), ESLint 10.8.1 reports `No magic number: -1.` because its `raw` is built from the operator plus the literal's raw text. Here `fullRaw` is sliced from the promoted tsgo range, so rslint instead reports `No magic number: -(1).` (and `-((1))` for nested parentheses). This is an observable diagnostic mismatch with the stated upstream baseline; please construct the message raw value from the sign and `ownRaw` while retaining the promoted range for location reporting.

swwind: Fixed in db9232fd.

Confirmed against ESLint 10.7: `raw` is built as `` `${parent.operator}${node.raw}` ``, so `f(-(1))`, `f(-((1)))` and `f(- 1)` all report `No magic number: -1.`. `fullRaw` is now the operator concatenated with the literal's own raw text, while `fullNumberRange` is unchanged, so the reported range still spans the whole unary expression — `-(1)` covers columns 3–7 and `-((1))` covers 3–9.

Added extras cases for `-(1)`, `-((1))`, `+(1)`, `- 1`, `-(0x1F)` and `-(1_000)` asserting both the message and the end column, and corrected the `-(7)` expectation that had locked in the old text, along with its mirror in the typescript plugin suite.

internal/rules/no_magic_numbers/no_magic_numbers.go
```diff
@@ -0,0 +1,575 @@
...
+				if bigintValue == nil {
+					bigintValue = new(big.Int)
+				}
+				if isNegative {
+					bigintValue.Neg(bigintValue)
+				}
+			} else {
+				numericValue, _ = parseRawNumericValue(ownRaw)
+				if isNegative {
+					numericValue = -numericValue
```
elecmonkey, line 250: [P2] Normalize negative zero to the ignored zero value

When `ignore` contains `0`, an occurrence such as `f(-0)` is still reported because negating the parsed value produces IEEE negative zero and `FormatFloat` creates the key `"-0"`, while the option was stored as `"0"`. ESLint uses a JavaScript `Set`, whose SameValueZero comparison treats `-0` and `0` as equal, so the same configuration emits no diagnostic; canonicalize signed zero before constructing the lookup key.

internal/rules/no_magic_numbers/no_magic_numbers.go
```diff
@@ -0,0 +1,575 @@
...
+// isOkParent checks if the parent node type is one that allows numbers without reporting.
+func isOkParent(parent *ast.Node, detectObjects bool) bool {
+	if detectObjects {
+		return false
+	}
+	switch parent.Kind {
+	case ast.KindObjectLiteralExpression:
+		return true
+	case ast.KindPropertyAssignment:
+		return true
```
elecmonkey, line 364: [P2] Allow numeric object-method keys when detection is off

With the default `detectObjects: false`, object methods such as `({ 42() {} })` are reported even though ESLint emits no diagnostic. ESTree represents this numeric key under a `Property`, which is included in upstream's allowed parent types, whereas ts-go gives it a `MethodDeclaration` parent that falls through this switch; object-literal methods and accessors need the same allowance while class methods must remain reportable.

elecmonkey (commented):
The port has two reproducible behavior mismatches with ESLint 10.8.x: ignored zero does not cover negative zero, and numeric object-method keys are incorrectly reported under the default options. Existing targeted Go tests pass, but neither edge case is covered.

internal/rules/no_magic_numbers/no_magic_numbers.go
swwind: Fixed in 0361fdf3.

Confirmed against ESLint 10.7: `f(-0)` with `ignore: [0]`, `f(0)` with `ignore: [-0]` and `f(-0)` with `ignore: [-0]` all emit no diagnostic. `normalizeIgnoreFloat` is now `normalizeFloatKey` and returns `"0"` for any zero; it is used both when parsing the option values and when building the literal's lookup key, so signed zero collapses on both sides. The bigint path already canonicalized through `big.Int`, so `-0n` against `ignore: ["0n"]` was correct.

Added valid cases for all four shapes.

internal/rules/no_magic_numbers/no_magic_numbers.go
swwind: Fixed in 0361fdf3.

Confirmed against ESLint 10.7: `({ 42() {} })`, `({ [42]() {} })`, `({ get 42() {} })`, `({ set 42(v) {} })`, `({ async 42() {} })` and `({ *42() {} })` emit no diagnostic, while `class A { 42() {} }`, `class A { [42]() {} }` and `class A { get 42() {} }` are reported, as is `({ 42() {} })` under `detectObjects: true`.

`isOkParent` now allows `MethodDeclaration`, `GetAccessor` and `SetAccessor` whose parent is an `ObjectLiteralExpression` — reached directly or through `ComputedPropertyName` — so class members remain reportable. Added valid and invalid cases for all of the above.

internal/plugins/typescript/rules/no_magic_numbers/no_magic_numbers.go
```diff
@@ -2,620 +2,22 @@ package no_magic_numbers
...
-		return rule.RuleListeners{
-			ast.KindNumericLiteral: func(node *ast.Node) {
-				handleNumericNode(node, false)
-			},
-			ast.KindBigIntLiteral: func(node *ast.Node) {
-				handleNumericNode(node, true)
-			},
-		}
-	},
+	Run:    core.RunTSESLint,
```
elecmonkey, line 23: When @typescript-eslint/no-magic-numbers uses { ignoreClassFieldInitialValues: true, ignoreReadonlyClassProperties: false }, class C { readonly x = 1 } must still be reported: typescript-eslint@8.67.0 makes the TypeScript-specific readonly decision (false) before delegating to the core rule. Delegating directly to core.RunTSESLint instead lets the class-field ignore return first, so Rslint emits no diagnostic and regresses the existing plugin rule.

elecmonkey, line 23: For @typescript-eslint/no-magic-numbers, enum E { A = +1 } with ignoreEnums: false (and the analogous readonly-property case) is reported by typescript-eslint@8.67.0 as No magic number: 1. over the literal only, because its TypeScript-specific false path promotes unary minus but not unary plus. Direct core delegation instead reports +1 and includes the plus sign in the range, producing an observable message and location mismatch.

internal/rules/no_magic_numbers/no_magic_numbers.go
```diff
@@ -0,0 +1,590 @@
...
+			parent = parent.Parent
+			continue
+		case ast.KindBinaryExpression:
+			binExpr := parent.AsBinaryExpression()
+			if binExpr.OperatorToken.Kind == ast.KindEqualsToken {
+				left := binExpr.Left
+				return left != nil && (left.Kind == ast.KindArrayLiteralExpression || left.Kind == ast.KindObjectLiteralExpression)
+			}
+			return false
+		case ast.KindForOfStatement, ast.KindForInStatement:
```
elecmonkey, line 552: With { ignoreDefaultValues: true }, ESLint 10.8.1 still reports the 1 in for (const x of [a = 1]) {} because the assignment is in the iterable expression, not in the loop's destructuring target. This walk returns true whenever it eventually reaches any ForOfStatement or ForInStatement, so Rslint incorrectly suppresses assignments in the right-hand expression as well.

internal/rules/no_magic_numbers/no_magic_numbers.go
```diff
@@ -0,0 +1,590 @@
...
+			return false
+		}
+	}
+	return false
+}
+
+// isClassFieldInitialValue checks if the fullNumberNode is the direct initializer
+// of a class field. parent is the already-resolved logical parent (parens skipped).
+func isClassFieldInitialValue(fullNumberNode *ast.Node, parent *ast.Node) bool {
+	if parent == nil || parent.Kind != ast.KindPropertyDeclaration {
```
elecmonkey, line 564: With { ignoreClassFieldInitialValues: true }, ESLint 10.8.1 through the TypeScript parser and typescript-eslint@8.67.0 both report class C { accessor x = 1 }, because its ESTree parent is AccessorProperty, not PropertyDefinition. Ts-go represents the auto-accessor as a PropertyDeclaration with ModifierFlagsAccessor, so this helper currently treats it as an ordinary field and Rslint emits no diagnostic.

internal/rules/no_magic_numbers/no_magic_numbers.go
```diff
@@ -0,0 +1,590 @@
...
+		for ancestor != nil && (ancestor.Kind == ast.KindUnionType || ancestor.Kind == ast.KindParenthesizedType) {
+			ancestor = ancestor.Parent
+		}
+		return ancestor != nil && ancestor.Kind == ast.KindTypeAliasDeclaration
+	}
+
+	return false
+}
+
+func isParentTSReadonlyPropertyDefinition(node *ast.Node) bool {
```
elecmonkey, line 478: With { ignoreReadonlyClassProperties: true }, both ESLint 10.8.1 and typescript-eslint@8.67.0 accept class C { readonly [1] = foo }, while Rslint reports 1. ESTree makes the numeric key a direct child of the readonly property, but ts-go inserts a ComputedPropertyName, so requiring the literal's immediate logical parent to be PropertyDeclaration misses this supported case.

internal/plugins/typescript/rules/no_magic_numbers/no_magic_numbers.go
swwind: Fixed in 0564a9c6.

Confirmed against typescript-eslint 8.67.0: the extension resolves the TypeScript-specific positions itself through a tri-state `isAllowed`, and when one matches with its option off it reports and returns, so the core checks never run. `class C { readonly x = 1 }` under `{ ignoreClassFieldInitialValues: true, ignoreReadonlyClassProperties: false }` is reported as `No magic number: 1.` at 1:24, while the same code with `ignoreReadonlyClassProperties: true` emits nothing.

`core.RunTSESLint` now selects that path instead of delegating straight to the core implementation: the enum, numeric literal type, type index and readonly property checks run first, in upstream's order, and a match with its option off reports there rather than falling through to `ignoreDefaultValues`, `ignoreClassFieldInitialValues`, `ignoreArrayIndexes` and the parent-type allowance.

Added valid cases for `readonly x = 1` with both options on and for `readonly [1] = foo`, and invalid cases for `readonly x = 1` with `ignoreReadonlyClassProperties: false`, `readonly x = -1`, `readonly x = +1`, `readonly [1] = foo`, `accessor x = 1` and `for (const x of [a = 1]) {}`.

swwind: Fixed in 0564a9c6.

Confirmed against typescript-eslint 8.67.0: the extension's report path promotes only unary minus, citing the base rule's own comment, so `enum E { A = +1 }` with `ignoreEnums: false` is reported as `No magic number: 1.` over columns 15–16 and `class C { readonly x = +1 }` as `No magic number: 1.` over 25–26. Unary minus still promotes — `enum E { A = -1 }` reports `-1` over 14–16.

The new extension path builds the reported range and text from the literal alone unless the effective parent is a unary minus, so the plus sign is excluded from both.

This also corrected two existing expectations in the plugin suite that had locked in the old text: `POS = +1` in the `ignoreEnums: false` enum case is now `No magic number: 1.` at 6:10, and `readonly F = +6` in the `ignoreReadonlyClassProperties: false` class case is now `No magic number: 6.` at 8:17.

internal/rules/no_magic_numbers/no_magic_numbers.go
swwind: Fixed in 0564a9c6.

Confirmed against ESLint 10.8.1: `for (const x of [a = 1]) {}` and `for (const x in [a = 1]) {}` with `ignoreDefaultValues: true` both report `No magic number: 1.` at 1:22, while `for ([a = 1] of foo) {}` and `for ([a = 1] in foo) {}` emit nothing. Upstream distinguishes the two purely by node type — `AssignmentPattern` versus `AssignmentExpression` — which the parser decides.

`isInsideDestructuringAssignment` now tracks the child it came from and returns `Initializer == child` at a for-of or for-in, so only the destructuring target counts. The same tracking fixes the assignment case: at a `=` binary expression it now requires the array or object literal to be the left operand, so `[b] = [a = 1]` reports the `1` (confirmed at 1:12) while `[a = 1] = arr` and `({ p: a = 1 } = o)` stay silent.

Added invalid cases for the two for-of/for-in shapes, `[b] = [a = 1]` and `x = [a = 1]`, and valid cases for the four destructuring-target shapes.

internal/rules/no_magic_numbers/no_magic_numbers.go
swwind: Fixed in 0564a9c6.

Confirmed against ESLint 10.8.1 through @typescript-eslint/parser and against typescript-eslint 8.67.0: `class C { accessor x = 1 }` and `class C { static accessor x = 1 }` with `ignoreClassFieldInitialValues: true` are reported by both, at 1:24 and 1:31.

`isClassFieldInitialValue` now returns false when the property carries `ModifierFlagsAccessor`, so an auto-accessor is no longer treated as an ordinary class field. Added invalid cases for both shapes.

internal/rules/no_magic_numbers/no_magic_numbers.go
swwind: Fixed in 0564a9c6.

Confirmed against ESLint 10.8.1 and typescript-eslint 8.67.0: `class C { readonly [1] = foo }` and `class C { readonly [-1] = foo }` with `ignoreReadonlyClassProperties: true` emit nothing in both, while `class C { [1] = foo }` under the same option is reported at 1:12 and `class C { readonly [1] = foo }` with the option off is reported at 1:21.

`isParentTSReadonlyPropertyDefinition` now steps over a `ComputedPropertyName` before requiring a `PropertyDeclaration`, matching the ESTree shape where the computed key is a direct child of the readonly property. Added valid cases for the two ignored shapes and invalid cases for the two that must still report.

elecmonkey (dismissed):
no more issues found

fansenze (approved):

PR #1737, opened 2026-08-14
feat: port rule no-invalid-this

swwind:
## Summary

Port the `no-invalid-this` rule from ESLint core to rslint. Also refactor `@typescript-eslint/no-invalid-this` to reuse the new rule's shared walker instead of duplicating it.

Flags `this` keywords used in contexts where the value of `this` is `undefined` under strict mode.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/no-invalid-this
- Source code: https://github.com/eslint/eslint/blob/v10.8.1/lib/rules/no-invalid-this.js

## Checklist


elecmonkey:
LGTM

internal/rules/no_invalid_this/no_invalid_this.go
```diff
@@ -0,0 +1,874 @@
...
+	}
+
+	// enterPropertyDeclaration defers the push past a computed key, exactly
+	// like `enterMethodLike`. Core ESLint gives a class field's *value*
+	// position its own implicit-function code path (`PropertyDefinition#value`
+	// — see `isDefaultThisBinding`'s doc comment) but does NOT extend that
+	// treatment to the key: a computed key (`[this.foo]`) is evaluated at
+	// class-definition time in the *enclosing* scope, not the field's own
+	// context, so it must see whatever frame was already on the stack.
+	enterPropertyDeclaration := func(node *ast.Node) {
```
elecmonkey, line 167: When a class field has a decorator in a strict context, ESLint 10.8.1 evaluates the decorator in the enclosing scope and reports `this` there (for example, `export {}; function f(){ class C { @deco(this) x = 1; } }`). This listener pushes the always-valid field frame before tsgo visits the decorator, and the `ThisKeyword` workaround explicitly excludes `PropertyDeclaration`, so rslint emits no diagnostic. Please make decorators on fields use the enclosing frame just as the implementation already does for method decorators; an oracle run reports `this` at column 42 while the current rule reports zero errors.

swwind: Confirmed against ESLint 10.8.1 — thanks. Fixed in 93ac371a.

Root cause: core scopes a field's frame to the *initializer value*, not to the `PropertyDefinition` node — `onCodePathStart` fires for the class-field-initializer code path, plus the separate `AccessorProperty > *.value` listener. Decorators are visited before the value, so they see the enclosing frame. This rule pushed on `PropertyDeclaration` entry, which covered the decorators too.

The two rules genuinely diverge here, though: `@typescript-eslint/no-invalid-this` is a copy of an older core rule whose listeners are `PropertyDefinition()` / `AccessorProperty()` on entry, so the wrapper really does swallow field-decorator reports while still reporting method-decorator ones:

| Code | ESLint core | typescript-eslint |
| --- | --- | --- |
| `class A { @deco(this) foo = 1 }` | reports | no report |
| `class A { @deco(this) accessor foo = 1 }` | reports | no report |
| `class A { @deco(this) foo() {} }` | reports | reports |

So instead of changing the shared walker outright I added a third `EngineOptions` policy point, `FieldDecoratorUsesEnclosingScope` — `true` for core, `false` for the wrapper — and widened `decoratorOfMethodLikeAncestor` (now `decoratorOfClassMemberAncestor`) to accept `PropertyDeclaration` as a decorator host. Computed-key fields keep the existing deferred push, so they need no peek.

Tests: core extras gains three invalid cases (plain field, `accessor` field, computed-key field) plus valid counterparts and a field-initializer contrast case; the wrapper's existing lock-ins now name the policy point. Flipping the flag back to `false` makes all three new invalid cases fail with "reported 0 errors", so they can't pass vacuously.

internal/rules/no_invalid_this/no_invalid_this.go
```diff
@@ -0,0 +1,882 @@
...
+			// which happens before tsgo visits the modifier list, so peek
+			// one frame deeper when `this` appears inside a decorator —
+			// but ONLY if the member's frame is actually on the stack
+			// already. A computed-key member defers its push to
+			// `ComputedPropertyName:exit`, which runs AFTER the modifier
+			// list, so at decorator visit time the stack top is already
+			// the surrounding scope and no peek is needed. Field
+			// decorators peek only when `FieldDecoratorUsesEnclosingScope`
+			// says so.
+			skip := 0
```
elecmonkey: [P2] Preserve the enclosing frame inside nested decorator expressions

When a member decorator contains a nested class or object with a computed member key, the nested member boundary makes `decoratorOfClassMemberAncestor` return false, leaving the decorated outer member's always-valid frame on top. For example, `export {}; function outer(){ class O { @dec(class I { [this](){} }) m(){} } }` produces one `unexpectedThis` diagnostic at column 56 in ESLint 10.8.1 and `@typescript-eslint/no-invalid-this` 8.65.0, but rslint reports none; the computed key is evaluated in `outer`'s strict scope, not as `O.m`'s `this`.

swwind: Confirmed against ESLint 10.8.1 and `@typescript-eslint/no-invalid-this` 8.65.0 — thanks. Fixed in eb49bd18.

Root cause: the peek was a fixed one-frame lookback, and the walk up from `this` bailed out at *any* class-member ancestor. Both assumptions break once a decorator contains another member:

- A nested member reached through its own computed key has **not** pushed its frame yet (the push is deferred to `ComputedPropertyName:exit`), so the walk has to pass straight through it instead of stopping.
- Each decorator hop costs its own frame, so `@dec(class I { @dec2(this) m(){} })` needs to skip two, not one.

`decoratorOfClassMemberAncestor` is now `decoratorFrameSkip`, which counts frames instead of returning a single boolean: a decorator hop increments the count and resumes the walk outside the decorated member; arrow functions and computed-key-entered members are transparent and cost nothing; a non-arrow function-like, a static block, or a member entered through its body/initializer owns the stack top and stops the walk. Field decorators still hop only under the policy point, so the wrapper's masking is unchanged.

New lock-ins (columns verified against the oracle): nested class/object member computed keys inside a method decorator, two-level decorator nesting, and a class decorator hosting no member frame — plus the wrapper's valid counterpart where the field frame still swallows everything. Reverting the walker makes all of them fail with "reported 0 errors".

While verifying this I found a second divergence in the same area, fixed in a0ece326: the wrapper reported `class A { [this.foo] = 1 }`, but genuine `@typescript-eslint/no-invalid-this` does not — its `PropertyDefinition()` / `AccessorProperty()` push covers the computed key just as it covers the decorators. That made the existing `FieldDecoratorUsesEnclosingScope` name too narrow, so it is now `FieldFrameScopedToValue`: core scopes the field's frame to the initializer value (decorators and computed key see the enclosing scope), the wrapper scopes it to the whole member (both are masked). The two invalid cases that locked in the old behavior moved to valid, and the rule's "Differences from ESLint" bullet — which documented the divergence — was replaced with the `sourceType` one that still holds.

| Code | ESLint core | typescript-eslint | This PR (core / wrapper) |
| --- | --- | --- | --- |
| `function o(){ class C { @dec(class I { [this](){} }) m(){} } }` | reports | reports | reports / reports |
| `function o(){ class C { @dec(class I { @dec2(this) m2(){} }) m(){} } }` | reports | reports | reports / reports |
| `function o(){ class C { @dec(class I { [this](){} }) f = 1 } }` | reports | no report | reports / no report |
| `class A { [this.foo] = 1 }` | reports | no report | reports / no report |

internal/rules/no_invalid_this/no_invalid_this.go
```diff
@@ -0,0 +1,909 @@
...
+	}
+	return !isDefaultThisBinding(node, capIsConstructor)
+}
+
+// isStrictFunction reports whether `fn`'s own body scope is strict-mode —
+// mirrors ESLint's `sourceCode.getScope(node).isStrict` for a function node.
+// True when an enclosing scope is already strict (ES module, ancestor
+// "use strict", or a class body — see utils.IsInStrictMode), OR the function
+// declares its own "use strict" directive.
+func isStrictFunction(fn *ast.Node, sf *ast.SourceFile) bool {
```
elecmonkey: [P2] Respect configured ES3 when detecting strict mode

When `languageOptions.ecmaVersion` is `3`, strict-mode directives are inert, but this function always honors them. For `function foo(){ "use strict"; this; }` with `ecmaVersion: 3`, ESLint 10.8.1 emits no diagnostic while rslint reports `unexpectedThis`. Rslint already supports and propagates `ecmaVersion`, so this can no longer be treated as an unavailable framework option; use the effective configured version when deciding whether a directive enables strict mode.

internal/rules/no_invalid_this/no_invalid_this.go
```diff
@@ -0,0 +1,909 @@
...
+			args := call.Arguments.Nodes
+			callee := call.Expression
+
+			if utils.IsSpecificMemberAccess(callee, "Reflect", "apply") {
+				if len(args) != 3 || args[0] != current {
+					return true
+				}
+				return utils.IsNullOrUndefined(args[1])
+			}
+			if utils.IsSpecificMemberAccess(callee, "Array", "from") ||
```
elecmonkey: [P2] Recognize TypedArray.from callbacks with a thisArg

For calls such as `Uint8Array.from([], function () { this; }, obj)` in strict code, ESLint 10.8.1 accepts the callback because its `isArrayFromMethod` recognizes any object name ending in `Array`, whereas this branch only recognizes the literal `Array`. Consequently both the core rule and the shared TypeScript wrapper report `unexpectedThis` for TypedArray callbacks that have an explicit `thisArg`.

internal/rules/no_invalid_this/no_invalid_this.go
```diff
@@ -0,0 +1,909 @@
...
+		return node.AsFunctionDeclaration().Name() != nil
+	case ast.KindFunctionExpression:
+		return node.AsFunctionExpression().Name() != nil
+	}
+	return false
+}
+
+// thisTagPattern is ESLint's exact pattern: `^[\s*]*@this` applied to a
+// comment's *value*, i.e. with `/*` / `*/` / `//` markers stripped.
+var thisTagPattern = regexp.MustCompile(`(?m)^[\s*]*@this\b`)
```
elecmonkey: [P2] Match ESLint's exact @this tag prefix

When a leading comment contains a prefix such as `/* @thisX */`, ESLint 10.8.1 treats it as an `@this` tag because its exact pattern is `^[\s*]*@this` with no word boundary, but this `\b` rejects continuations such as letters, digits, and `_`. Thus `export {}; function f(){ /* @thisX */ function g(){ this; } }` is accepted upstream but reports `unexpectedThis` here (and in the TypeScript wrapper that shares this helper).

internal/rules/no_invalid_this/no_invalid_this.go
swwind: Fixed in e4ffb66f. The core rule now reads the effective per-file ecmaVersion and disables strict-mode detection for ES3. I also moved the previously skipped upstream ES3 case into the valid suite and added a focused regression test. The focused package tests pass.

internal/rules/no_invalid_this/no_invalid_this.go
swwind: Fixed in e4ffb66f. `isArrayFromMethod` now mirrors ESLint's `/Array$/` receiver check for `.from`, including static bracket access, while `.fromAsync` remains limited to `Array`. Added focused core and TypeScript-wrapper regressions for `Uint8Array` and `BigInt64Array`. The focused package tests pass.

internal/rules/no_invalid_this/no_invalid_this.go
swwind: Fixed in e4ffb66f. The `@this` pattern now matches ESLint's exact prefix-based expression without a word boundary. Added focused regressions for both the core rule and the shared TypeScript wrapper. The focused package tests pass.

internal/rules/no_invalid_this/no_invalid_this.go
```diff
@@ -0,0 +1,946 @@
...
+// isStrictFunction reports whether `fn`'s own body scope is strict-mode —
+// mirrors ESLint's `sourceCode.getScope(node).isStrict` for a function node.
+// True when an enclosing scope is already strict (ES module, ancestor
+// "use strict", or a class body — see utils.IsInStrictMode), OR the function
+// declares its own "use strict" directive.
+func isStrictFunction(fn *ast.Node, sf *ast.SourceFile, ecmaVersion int) bool {
+	// Strict mode was introduced in ES5; in ES3 even directive-looking string
+	// literals are semantically inert.
+	if ecmaVersion == 3 {
+		return false
```
elecmonkey, line 413: [P2] Preserve inherited strict contexts when targeting ES3

When `ecmaVersion` is 3 but ts-go accepts newer syntax, this early return disables all strictness, including strictness inherited from class bodies. For example, `class C { m(){ function foo(){ this; } } }` produces `unexpectedThis` with ESLint 10.8.1 and `@typescript-eslint/parser`, while rslint reports nothing. ES3 should suppress only directive-based strict mode; module- and class-imposed strictness must still be checked.

internal/rules/no_invalid_this/no_invalid_this.go
```diff
@@ -0,0 +1,946 @@
...
+		return node.AsFunctionDeclaration().Name() != nil
+	case ast.KindFunctionExpression:
+		return node.AsFunctionExpression().Name() != nil
+	}
+	return false
+}
+
+// thisTagPattern is ESLint's exact pattern: `^[\s*]*@this` applied to a
+// comment's *value*, i.e. with `/*` / `*/` / `//` markers stripped.
+var thisTagPattern = regexp.MustCompile(`(?m)^[\s*]*@this`)
```
elecmonkey: [P2] Match JavaScript whitespace in @this tags

When whitespace before `@this` is non-ASCII ECMAScript whitespace, this Go regexp no longer matches the JavaScript regexp it ports. For example, `export {}; function f(){ /** U+00A0@this */ function g(){ this; } }` using U+00A0 is accepted by ESLint 10.8.1, but rslint reports `unexpectedThis`; U+FEFF and other Unicode space separators behave similarly. Use ECMAScript whitespace semantics rather than Go RE2 ASCII-only `\s`.

internal/rules/no_invalid_this/no_invalid_this.go
```diff
@@ -0,0 +1,946 @@
...
+		case ast.KindBinaryExpression:
+			// Assignment / logical / conditional — defer to the enclosing
+			// statement. Continue walking.
+			current = parent
+			continue
+		case ast.KindConditionalExpression:
+			current = parent
+			continue
+		default:
+			return false
```
elecmonkey: [P2] Continue JSDoc lookup through expression containers

When a function is nested in an expression below a declaration-level JSDoc, this default return stops the search earlier than ESLint getJSDocComment. For example, ESLint 10.8.1 accepts `export {}; /** @this */ const x = [function(){ this }];`, whereas rslint reports `unexpectedThis` because `ArrayLiteralExpression` is not traversed; unary expressions have the same problem. The parent walk should continue through these expression containers until it reaches the same declaration/property/function boundaries as upstream.

elecmonkey (commented):
Found three remaining parity issues in no-invalid-this.

internal/rules/no_invalid_this/no_invalid_this.go
swwind: Fixed in 1c75a41f. `isStrictFunction` now checks syntax-imposed strict contexts (ES modules, class bodies, and TypeScript namespaces) before applying the ES3 directive gate, so ES3 suppresses only directive-based strict mode. Added focused regressions for both module- and class-inherited strictness at `ecmaVersion: 3`. The focused core package test passes.

internal/rules/no_invalid_this/no_invalid_this.go
swwind: Fixed in 1c75a41f. Replaced the Go regexp with an explicit matcher backed by ECMAScript whitespace semantics, preserving ESLint's `/^[\s*]*@this/mu` prefix behavior for non-ASCII whitespace. Added focused core and TypeScript-wrapper regressions for U+00A0 and U+FEFF. Both focused package tests pass.

internal/rules/no_invalid_this/no_invalid_this.go
swwind: Fixed in 1c75a41f. JSDoc lookup now walks through comment-free expression containers until the same function, method, or property boundaries as ESLint, while retaining the direct `CallExpression`/`NewExpression` exception. Added focused array-literal and unary-expression regressions for both the core rule and TypeScript wrapper. Both focused package tests pass.

internal/rules/no_invalid_this/no_invalid_this.go
```diff
@@ -0,0 +1,991 @@
...
+
+func run(ctx rule.RuleContext, options []any) rule.RuleListeners {
+	opts := ParseOptions(options)
+	ecmaVersion := ctx.Globals.EffectiveECMAVersion()
+	return BuildListeners(ctx, EngineOptions{
+		CapIsConstructor: opts.CapIsConstructor,
+		// Top-level `this` refers to the global object in scripts (always
+		// valid) and is `undefined` in ES modules (always invalid) —
+		// independent of strict mode.
+		TopLevelValid: !ast.IsExternalModule(ctx.SourceFile),
```
elecmonkey: [P1] Derive module strictness from configured sourceType

With `languageOptions.sourceType` now available on `RuleContext`, deriving both top-level validity and function strictness from `ast.IsExternalModule` misclassifies normal module-configured files that contain no import/export syntax. For example, with `sourceType: "module"` (also the default for `.js`/`.ts`), `this; function f(){ this; }` produces two diagnostics in ESLint 10.8.1 but none here; an explicitly script-configured file containing accepted export syntax has the reverse mismatch. Use the effective configured source type for both decisions.

internal/rules/no_invalid_this/no_invalid_this.go
```diff
@@ -0,0 +1,991 @@
...
+// "use strict", or a class body — see utils.IsInStrictMode), OR the function
+// declares its own "use strict" directive.
+func isStrictFunction(fn *ast.Node, sf *ast.SourceFile, ecmaVersion int) bool {
+	// Modules, classes, and TypeScript namespaces impose strict mode
+	// independently of directive prologues and therefore remain strict even
+	// when the configured language version is ES3.
+	if ast.IsExternalModule(sf) {
+		return true
+	}
+	for current := fn.Parent; current != nil; current = current.Parent {
```
elecmonkey, line 383: [P2] Exclude class decorators from class-body strictness

A function expression inside a class-level decorator is evaluated in the enclosing scope rather than the class body, but this unconditional ancestor check marks it strict. With `sourceType: "script"`, `@dec(function(){ this; }) class C {}` is accepted by ESLint 10.8.1 while rslint reports `unexpectedThis`; class ancestors reached through decorator slots need to be excluded from the class-body strictness check.

internal/rules/no_invalid_this/no_invalid_this.go
```diff
@@ -0,0 +1,991 @@
...
+	if parent == nil || parent.Kind == ast.KindCallExpression || parent.Kind == ast.KindNewExpression {
+		return false
+	}
+
+	// ESLint keeps climbing while the intermediate node has no leading
+	// comments and is not a function, method, or property boundary. This is
+	// intentionally broader than a hand-picked expression list: arrays,
+	// unary expressions, nested calls, and other containers are transparent.
+	for parent != nil && parent.Kind != ast.KindSourceFile {
+		hasComments, hasTag := leadingCommentSummary(parent, comments, text)
```
elecmonkey: [P2] Validate ancestor comments as actual JSDoc

When lookup climbs to a declaration or statement, this accepts any leading comment containing `@this`, regardless of comment kind or distance. Upstream `findJSDocComment` requires an adjacent `/** ... */` block, so `export {}; /* @this */ const x = [function(){ this; }];` and a `/** @this */` separated by a blank line are reported by ESLint 10.8.1 but suppressed here. The ancestor path should distinguish JSDoc blocks and enforce the one-line proximity check.

internal/rules/no_invalid_this/no_invalid_this.go
```diff
@@ -0,0 +1,991 @@
...
+			}
+			// ESLint's walker has explicit cases for `Property` /
+			// `PropertyDefinition` / `MethodDefinition` but NOT for
+			// `AccessorProperty`. tsgo collapses both ESTree kinds onto
+			// KindPropertyDeclaration; we re-introduce the distinction here
+			// by checking `ModifierFlagsAccessor`. For auto-accessors a
+			// function-expression initializer falls through to the walker's
+			// default branch (default-bound), matching how upstream's
+			// baseRule walker treats `AccessorProperty` parents.
+			if parent.ModifierFlags()&ast.ModifierFlagsAccessor != 0 {
```
elecmonkey: [P2] Preserve the field frame for auto-accessor initializers

For a function expression used directly as an auto-accessor initializer, ESLint's `AccessorProperty > *.value` listener keeps the initializer's always-valid field frame above the function code-path frame. Consequently both ESLint core and the TypeScript wrapper accept `class C { accessor x = function(){ this; } }`, but this branch marks that function default-bound and rslint reports `unexpectedThis`. The accessor initializer needs the same effective field-frame treatment as a regular field initializer.

internal/rules/no_invalid_this/no_invalid_this.go
```diff
@@ -0,0 +1,991 @@
...
+// `FieldFrameScopedToValue` says so; otherwise its frame is already on the
+// stack for both positions and the walk stops there.
+func decoratorFrameSkip(thisNode *ast.Node, fieldFrameScopedToValue bool) int {
+	skip := 0
+	prev, current := thisNode, thisNode.Parent
+	for current != nil {
+		switch current.Kind {
+		case ast.KindDecorator:
+			member := current.Parent
+			if member == nil || !isDecoratedClassMember(member) {
```
elecmonkey, line 1054: [P2] Walk through nested class decorators when selecting frames

When a member decorator contains a decorated class expression, the inner class decorator owns no validity frame, but this return stops before reaching the outer member decorator. For `export {}; function outer() { class C { @dec((@d(this) class I {})) m(){} } }`, ESLint 10.8.1 reports the `this` from `outer`'s invalid frame while rslint reports nothing because `C.m`'s valid frame remains selected. Continue walking outside class decorators without incrementing the skip count.

internal/rules/no_invalid_this/no_invalid_this.go
swwind: blocked by #1877

internal/rules/no_invalid_this/no_invalid_this.go
swwind: Fixed in fd70129e. Both the core rule and the TypeScript wrapper now derive top-level validity and function strictness from the effective `languageOptions.sourceType` instead of `ast.IsExternalModule`. Added regressions for module-configured code without import/export syntax and script-configured code containing accepted export syntax. Both affected package test suites pass.

internal/rules/no_invalid_this/no_invalid_this.go
swwind: Fixed in fd70129e. The strictness walk now skips the decorated class body when a function is reached through a class or member decorator, while continuing outward so genuinely strict enclosing scopes still apply. Added a focused sloppy-script class-decorator regression. The core package test suite passes.

internal/rules/no_invalid_this/no_invalid_this.go
swwind: Fixed in fd70129e. I rechecked the dependency and #1877 is not a blocker: that PR skips ts-go’s synthetic JSDoc AST during linter traversal, while this issue is in `no-invalid-this`’s source-comment lookup. The climbed ancestor path now mirrors `findJSDocComment`: only the closest adjacent `/** ... */` block can supply `@this`; ordinary block comments and JSDoc separated by a blank line do not. Added focused regressions in both the core rule and TypeScript wrapper. Both affected package test suites pass.

internal/rules/no_invalid_this/no_invalid_this.go
swwind: Fixed in fd70129e. Function expressions used as auto-accessor initializer values now receive the same effective always-valid field frame as regular field initializers, accounting for ts-go collapsing `AccessorProperty` and `PropertyDefinition` into `PropertyDeclaration`. Added focused core and TypeScript-wrapper regressions and removed the previous incorrect wrapper lock-in. Both affected package test suites pass.

internal/rules/no_invalid_this/no_invalid_this.go
swwind: Fixed in fd70129e. `decoratorFrameSkip` now walks through class decorators without incrementing the frame skip count, allowing an enclosing member decorator to select the correct outer frame. Added focused regressions for both the core rule and TypeScript wrapper. Both affected package test suites pass.

internal/rules/no_invalid_this/no_invalid_this.go
```diff
@@ -0,0 +1,1052 @@
...
+	}
+	skipDecoratedClassBody := false
+	for current := fn.Parent; current != nil; current = current.Parent {
+		if current.Kind == ast.KindDecorator {
+			host := current.Parent
+			if host != nil && (ast.IsClassLike(host) || isDecoratedClassMember(host)) {
+				skipDecoratedClassBody = true
+			}
+		}
+		if current.Kind == ast.KindModuleDeclaration {
```
elecmonkey: [P2] Treat TypeScript enum scopes as strict

When linting TypeScript with `sourceType: "script"`, enum scopes are strict upstream, but this ancestor walk recognizes only namespaces and classes. For `enum E { X = (function(){ this; }, 1) }`, ESLint 10.8.1 with the TypeScript parser and `@typescript-eslint/no-invalid-this` 8.67.0 report `unexpectedThis` at column 27, while both shared Rslint rules emit no diagnostic.

internal/rules/no_invalid_this/no_invalid_this.go
```diff
@@ -0,0 +1,1052 @@
...
+	// independently of directive prologues and therefore remain strict even
+	// when the configured language version is ES3.
+	if isModule {
+		return true
+	}
+	skipDecoratedClassBody := false
+	for current := fn.Parent; current != nil; current = current.Parent {
+		if current.Kind == ast.KindDecorator {
+			host := current.Parent
+			if host != nil && (ast.IsClassLike(host) || isDecoratedClassMember(host)) {
```
elecmonkey: [P2] Keep member decorators inside the class strict scope

For a function expression inside a method or field decorator in a script, this condition skips the enclosing class and incorrectly treats the function as sloppy. For example, `class C { @dec(function(){ this; }) m(){} }` produces `unexpectedThis` at column 28 in ESLint 10.8.1 with the TypeScript parser and in `@typescript-eslint/no-invalid-this` 8.67.0, but Rslint reports nothing; only class-level decorators are visited outside the class scope, whereas member decorators inherit the class scope's strictness.

internal/rules/no_invalid_this/no_invalid_this.go
```diff
@@ -0,0 +1,1052 @@
...
+
+func isJSDocLookupBoundary(node *ast.Node) bool {
+	if ast.IsFunctionLike(node) {
+		return true
+	}
+	switch node.Kind {
+	case ast.KindMethodDeclaration, ast.KindConstructor,
+		ast.KindGetAccessor, ast.KindSetAccessor,
+		ast.KindPropertyAssignment, ast.KindShorthandPropertyAssignment,
+		ast.KindPropertyDeclaration, ast.KindBindingElement:
```
elecmonkey: [P2] Let JSDoc lookup cross array binding elements

Treating every `BindingElement` as a JSDoc boundary conflates object properties with array-pattern elements. With `/** @this */ const [x = function(){ this; }] = [];` in module code, ESLint 10.8.1 climbs through the ESTree `AssignmentPattern` and `ArrayPattern` to the variable declaration and emits no diagnostic, while Rslint stops here and reports `unexpectedThis` at column 37; array binding elements need to remain transparent even though object binding properties are boundaries.

internal/rules/no_invalid_this/no_invalid_this.go
swwind: Fixed in 05b0cb76. `isStrictFunction` now treats TypeScript enum declarations as syntax-imposed strict scopes, including when the effective source type is `script`. Added focused regressions for both the core rule and the TypeScript wrapper. Both affected package test suites pass.

internal/rules/no_invalid_this/no_invalid_this.go
swwind: Fixed in 05b0cb76. The strictness walk now skips a class boundary only for a decorator directly attached to that class; member decorators retain the enclosing class's strict scope. Added focused regressions for both the core rule and the TypeScript wrapper. Both affected package test suites pass.

internal/rules/no_invalid_this/no_invalid_this.go
swwind: Fixed in 05b0cb76. JSDoc lookup now treats only object-pattern `BindingElement` nodes as property boundaries, while array-pattern binding elements remain transparent like ESTree `AssignmentPattern` nodes. Added focused regressions for both the core rule and the TypeScript wrapper. Both affected package test suites pass.

internal/rules/no_invalid_this/no_invalid_this.go
```diff
@@ -0,0 +1,1057 @@
...
+	if parent == nil || parent.Kind == ast.KindCallExpression || parent.Kind == ast.KindNewExpression {
+		return false
+	}
+
+	// ESLint keeps climbing while the intermediate node has no leading
+	// comments and is not a function, method, or property boundary. This is
+	// intentionally broader than a hand-picked expression list: arrays,
+	// unary expressions, nested calls, and other containers are transparent.
+	for parent != nil && parent.Kind != ast.KindSourceFile {
+		hasComments, hasTag := ancestorJSDocSummary(parent, comments, sf)
```
elecmonkey, line 549: [P2] Stop JSDoc lookup before enclosing functions

The ancestor comment lookup runs before checking whether the ancestor is a function boundary, so an outer function's `@this` tag incorrectly validates nested functions. With `/** @this */ function outer(){ return function(){ this; }; }` in module mode, ESLint 10.8.1 and `@typescript-eslint/no-invalid-this` 8.67.0 report the inner `this` at column 51, while both Rslint rules report nothing. Check function/method/property boundaries before inspecting their leading comments.

internal/rules/no_invalid_this/no_invalid_this.go
```diff
@@ -0,0 +1,1057 @@
...
+	}
+	if fn.Kind != ast.KindFunctionExpression {
+		return false
+	}
+
+	// ESTree discards parentheses and TypeScript expression wrappers, so find
+	// the first semantic parent before applying its direct-call exception.
+	current := fn
+	parent := current.Parent
+	for parent != nil && ast.IsOuterExpression(parent, ast.OEKParentheses|ast.OEKAssertions) {
```
elecmonkey: [P2] Preserve TypeScript wrappers during JSDoc lookup

Skipping assertion wrappers before applying the direct-call exception does not match the AST exposed by `@typescript-eslint/parser`. For `/** @this */ foo(function(){ this; } as any);` in module mode, ESLint 10.8.1 and the TypeScript wrapper 8.67.0 accept the function, but both Rslint rules report `unexpectedThis` at column 30; `satisfies` and non-null wrappers behave similarly. Let these TypeScript nodes participate in the ancestor walk instead of treating them like erased parentheses.

internal/rules/no_invalid_this/no_invalid_this.go
```diff
@@ -0,0 +1,1057 @@
...
+			body := current.Body()
+			if body != nil && body.Kind == ast.KindBlock && utils.HasUseStrictDirective(body) {
+				return true
+			}
+		}
+	}
+
+	// Strict-mode directives were introduced in ES5. In ES3, only the
+	// syntax-imposed contexts handled above can make the function strict.
+	if ecmaVersion == 3 {
```
elecmonkey: [P2] Keep TypeScript strict directives active under ES3

The unconditional ES3 return matches Espree, but not the TypeScript parser that Rslint also replaces. With `ecmaVersion: 3`, `sourceType: "script"`, and `function f(x: number){ 'use strict'; this; }`, ESLint 10.8.1 using `@typescript-eslint/parser` and `@typescript-eslint/no-invalid-this` 8.67.0 both report column 38, while both Rslint rules emit nothing. The directive gate needs to distinguish JavaScript's ES3 semantics from TypeScript-parser semantics.

internal/plugins/typescript/rules/no_invalid_this/no_invalid_this.go
```diff
@@ -37,743 +36,14 @@ var NoInvalidThisRule = rule.CreateRule(rule.Rule{
...
-	}
-	return false
+	opts := no_invalid_this_core.ParseOptions(options)
+	if ctx.LanguageOptions.SourceType == "" {
+		// The upstream TypeScript RuleTester defaults to module semantics.
+		// Production rslint contexts already carry an effective source type.
+		ctx.LanguageOptions.SourceType = "module"
+	}
+	engineOptions := no_invalid_this_core.CoreEngineOptions(ctx, opts)
+	engineOptions.FieldFrameScopedToValue = false
```
elecmonkey, line 47: [P2] Reproduce the wrapper's auto-accessor frame lifetime

Setting only the field-boundary policy does not reproduce the upstream wrapper's interaction with core's `AccessorProperty > *.value` listener. In module code `class C { accessor x = 1; } this;`, `@typescript-eslint/no-invalid-this` 8.67.0 emits no diagnostic because the initialized accessor leaves the core value frame active after the wrapper overrides the exit listener, whereas Rslint pops its sole frame and reports column 29. This also changes later computed keys and member decorators, so the wrapper requires a separate accessor-value frame policy rather than only `FieldFrameScopedToValue = false`.

internal/rules/no_invalid_this/no_invalid_this.go
swwind: Fixed in d67c363b. The JSDoc ancestor walk now checks function, method, and property boundaries before inspecting their leading comments, so an outer function's `@this` tag no longer validates a nested function. Added focused regressions for both the core rule and the TypeScript wrapper. Both affected package tests pass.

internal/rules/no_invalid_this/no_invalid_this.go
swwind: Fixed in d67c363b. JSDoc lookup now unwraps parentheses only; TypeScript `as`, `satisfies`, and non-null assertion nodes remain in the ancestor walk, matching the nodes exposed by typescript-estree and preserving the wrapped callback behavior. Added focused coverage for all three wrappers in both affected packages. Both package tests pass.

internal/rules/no_invalid_this/no_invalid_this.go
swwind: Fixed in d67c363b. ES3 directive suppression is now limited to JavaScript/JSX source files. TypeScript sources keep strict directives active under `ecmaVersion: 3`, matching typescript-eslint parser semantics. The existing Espree ES3 case is now explicitly run as `.js`, alongside the focused `.ts` regression. Both affected package tests pass.

internal/plugins/typescript/rules/no_invalid_this/no_invalid_this.go
swwind: Deliberately not reproduced. The upstream wrapper leaks the accessor-initializer validity frame after the initialized auto-accessor has finished, suppressing diagnostics in unrelated later code. Rslint keeps the balanced frame lifetime and correctly reports the later top-level `this`. Added a focused regression and documented this intentional divergence on the rule page. The TypeScript wrapper package tests pass.

swwind: Hold there, I can't reproduce. https://eslint.org/play/#eyJ0ZXh0IjoiLyplc2xpbnQgbm8taW52YWxpZC10aGlzOiBcImVycm9yXCIqL1xuXG5jbGFzcyBDIHsgYWNjZXNzb3IgeCA9IDE7IH0gdGhpczsiLCJvcHRpb25zIjp7InJ1bGVzIjp7fSwibGFuZ3VhZ2VPcHRpb25zIjp7InBhcnNlck9wdGlvbnMiOnsiZWNtYUZlYXR1cmVzIjp7fX0sImVjbWFWZXJzaW9uIjoibGF0ZXN0IiwicGFyc2VyIjoiQHR5cGVzY3JpcHQtZXNsaW50L3BhcnNlciJ9fX0=

internal/rules/no_invalid_this/no_invalid_this.go
```diff
@@ -0,0 +1,1060 @@
...
+
+	// ESLint keeps climbing while the intermediate node has no leading
+	// comments and is not a function, method, or property boundary. This is
+	// intentionally broader than a hand-picked expression list: arrays,
+	// unary expressions, nested calls, and other containers are transparent.
+	for parent != nil && parent.Kind != ast.KindSourceFile {
+		if isJSDocLookupBoundary(parent) {
+			return false
+		}
+		hasComments, hasTag := ancestorJSDocSummary(parent, comments, sf)
```
elecmonkey, line 549: [P2] Ignore JSDoc attached to erased parentheses

This still isn't affected by #1877. That PR prunes TypeScript-Go's synthetic AST nodes that come from JSDoc comments during normal rule traversal, but this path is reading ordinary source comments through `ctx.Comments.All()` inside `hasJSDocThisTag`; no synthetic JSDoc AST node is involved. For `foo(/** @this */ (function(){ this; } as any));` in module mode, ESLint 10.8.1 and `@typescript-eslint/no-invalid-this` 8.67.0 report `unexpectedThis` at column 31, while both Rslint rules emit nothing because the comment before the erased parenthesized wrapper is treated as the function's own leading `@this` tag. `satisfies` has the same result.

internal/rules/no_invalid_this/no_invalid_this.go
```diff
@@ -0,0 +1,1060 @@
...
+	if parent == nil || parent.Kind == ast.KindCallExpression || parent.Kind == ast.KindNewExpression {
+		return false
+	}
+
+	// ESLint keeps climbing while the intermediate node has no leading
+	// comments and is not a function, method, or property boundary. This is
+	// intentionally broader than a hand-picked expression list: arrays,
+	// unary expressions, nested calls, and other containers are transparent.
+	for parent != nil && parent.Kind != ast.KindSourceFile {
+		if isJSDocLookupBoundary(parent) {
```
elecmonkey, line 540: [P2] Check eligible ancestor JSDoc before stopping

This also isn't changed by #1877: the issue is the rule's own ancestor-comment lookup order, not traversal over synthetic JSDoc syntax. #1877 hides reparsed comment AST from rule listeners, but here `hasJSDocThisTag` is explicitly walking real parent nodes and deciding whether their real leading comments apply. In module mode, `const outer = /** @this */ function(){ return function(){ this; }; };` is accepted by ESLint 10.8.1 and the TypeScript wrapper, but both Rslint rules report the inner `this`; property-level cases such as `const o = { /** @this */ p: [function(){ this; }] };` fail the same way. The walk should still perform upstream-equivalent JSDoc lookup on eligible function-expression and property ancestors before stopping.

internal/rules/no_invalid_this/no_invalid_this.go
```diff
@@ -0,0 +1,1060 @@
...
+		},
+
+		// Class static block: own `this` context bound to the class — VALID.
+		ast.KindClassStaticBlockDeclaration:                      func(*ast.Node) { push(true) },
+		rule.ListenerOnExit(ast.KindClassStaticBlockDeclaration): func(*ast.Node) { pop() },
+
+		// Arrow function: lexical `this`. Intentionally NOT registered so
+		// the enclosing frame governs.
+
+		ast.KindThisKeyword: func(node *ast.Node) {
```
elecmonkey: [P2] Report this expressions inside TypeScript type queries

This is independent of #1877 as well. `type T = typeof this;` contains authored TypeScript syntax, not syntax synthesized from JSDoc, so the linter-level JSDoc pruning in #1877 does not remove or add this node. Upstream `@typescript-eslint/parser` exposes the `this` inside `typeof this` as a `ThisExpression`, so ESLint 10.8.1 and `@typescript-eslint/no-invalid-this` 8.67.0 report it in module mode at column 17; Rslint emits nothing because this listener only handles `KindThisKeyword` and misses ts-go's `KindThisType` representation inside type queries.

internal/rules/no_invalid_this/no_invalid_this.go
swwind: Fixed in 0af02dc0. JSDoc lookup now uses the first token retained by the corresponding ESTree node as its semantic start. A comment separated from the function by erased parentheses is therefore no longer treated as the function’s own leading `@this` tag. Added focused `as` and `satisfies` regressions for both the core rule and the TypeScript wrapper. Both affected package test suites pass.

internal/rules/no_invalid_this/no_invalid_this.go
swwind: Fixed in 0af02dc0. The ancestor walk now checks eligible `FunctionExpression` and object-property boundaries for their own adjacent JSDoc before stopping, while still stopping before an enclosing function declaration’s leading comment. Added focused function-expression and property regressions for both the core rule and the TypeScript wrapper. Both affected package test suites pass.

internal/rules/no_invalid_this/no_invalid_this.go
swwind: Fixed in 0af02dc0. The shared listener now handles `this` directly under a TypeScript `TypeQuery` across ts-go’s `ThisType` and identifier shapes, without treating ordinary `this` types as expressions. Added focused `type T = typeof this` regressions for both the core rule and the TypeScript wrapper. Both affected package test suites pass.

internal/rules/no_invalid_this/no_invalid_this.go
```diff
@@ -0,0 +1,1106 @@
...
+
+		// Arrow function: lexical `this`. Intentionally NOT registered so
+		// the enclosing frame governs.
+
+		ast.KindThisKeyword: reportThis,
+		ast.KindThisType: func(node *ast.Node) {
+			// typescript-estree represents the operand in `typeof this` as
+			// ThisExpression, while tsgo uses ThisType. Ordinary `this` types
+			// remain TSThisType upstream and must not be treated as expressions.
+			if node.Parent != nil && node.Parent.Kind == ast.KindTypeQuery {
```
elecmonkey: [P2] Walk through qualified `typeof this` operands

The new TypeQuery handling only reports when `this` is the direct child of `TypeQuery`, so qualified queries such as `type T = typeof this.x;` are missed. With module semantics, ESLint 10.8.1 and `@typescript-eslint/no-invalid-this` 8.67.0 both report `unexpectedThis` at column 17, while both shared Rslint rules report zero diagnostics. Please recognize `this` at the root of a qualified TypeQuery entity name as well.

internal/rules/no_invalid_this/no_invalid_this.go
```diff
@@ -0,0 +1,1106 @@
...
+	// like `enterMethodLike` — but only under `FieldFrameScopedToValue`.
+	// Core ESLint gives a class field's *value* position its own
+	// implicit-function code path (`PropertyDefinition#value` — see
+	// `isDefaultThisBinding`'s doc comment) but does NOT extend that
+	// treatment to the key: a computed key (`[this.foo]`) is evaluated at
+	// class-definition time in the *enclosing* scope, not the field's own
+	// context, so it must see whatever frame was already on the stack. The
+	// wrapper's `PropertyDefinition()` / `AccessorProperty()` listeners push
+	// on entry unconditionally, so there the key is covered too.
+	enterPropertyDeclaration := func(node *ast.Node) {
```
elecmonkey, line 196: [P2] Delay the core field frame until the initializer

For the core rule, entering every non-computed `PropertyDeclaration` pushes the always-valid initializer frame before ts-go visits the field's type annotation. Consequently `class C { x: typeof this; }` in module mode produces zero diagnostics here, while ESLint 10.8.1 reports `unexpectedThis` at column 21; `accessor x: typeof this = 1` has the same mismatch. The core field frame must cover only the initializer value, leaving annotations in the enclosing context.

internal/rules/no_invalid_this/no_invalid_this.go
```diff
@@ -0,0 +1,1106 @@
...
+	if parent == nil || parent.Kind == ast.KindCallExpression || parent.Kind == ast.KindNewExpression {
+		return false
+	}
+
+	// ESLint keeps climbing while the intermediate node has no leading
+	// comments and is not a function, method, or property boundary. This is
+	// intentionally broader than a hand-picked expression list: arrays,
+	// unary expressions, nested calls, and other containers are transparent.
+	for parent != nil && parent.Kind != ast.KindSourceFile {
+		if isJSDocLookupBoundary(parent) {
```
elecmonkey, line 540: [P2] Honor JSDoc on all eligible ancestor boundaries

The ancestor walk checks JSDoc only on `FunctionExpression` and `PropertyAssignment`, but upstream checks the reached parent unless it is a `FunctionDeclaration` or `Program`. This causes false positives in both shared rules: for example, ESLint 10.8.1 and typescript-eslint 8.67.0 accept `const outer = /** @this */ () => function(){ this; };`, `const { /** @this */ p = [function(){ this; }] } = obj;`, and `class C { /** @this */ x = [function(){ this; }]; }`, while Rslint reports the inner `this` in each case. Arrow functions, binding properties, and field-property boundaries need the same final adjacent-JSDoc lookup.

internal/rules/no_invalid_this/no_invalid_this.go
swwind: Fixed in 429a3eb15. The shared listener now recognizes `this` at the left root of a qualified TypeQuery entity name, such as `typeof this.x`, while rejecting right-side property segments. Focused regressions cover both the core rule and the TypeScript wrapper, and both affected package test suites pass.

internal/rules/no_invalid_this/no_invalid_this.go
swwind: Fixed in 429a3eb15. For the core policy, a `this` report inside a class field’s type annotation now looks past the physically active field frame and uses the enclosing frame, matching core’s initializer-value-only code path. Focused cases cover both a regular field and an initialized auto-accessor. The TypeScript wrapper retains its intentional whole-member frame behavior, with valid lock-ins for both forms. Both affected package test suites pass.

internal/rules/no_invalid_this/no_invalid_this.go
swwind: Fixed in 429a3eb15. The JSDoc ancestor walk now checks every reached eligible boundary before stopping, except `FunctionDeclaration` and Program as upstream does. It also resolves the closest adjacent comment from the boundary’s ESTree token start, including a parenthesized arrow’s semantic parameter-list `(`. Focused regressions cover arrow functions, object-binding properties, and class fields in both shared rules. Both affected package test suites pass.

internal/rules/no_invalid_this/no_invalid_this.go
```diff
@@ -0,0 +1,1134 @@
...
+// Together these match every case the upstream test suite exercises:
+//   - `/** @this */ function foo()` (own JSDoc)
+//   - `function foo() { /** @this */ return function bar() {} }` (parent ReturnStatement JSDoc)
+//   - `foo(/* @this */ function(){})` (leading comment between `(` and `function`)
+//
+// Out of scope (and correctly NOT matched): `/** @this */ foo(function(){})`
+// — the JSDoc here belongs to the enclosing CallExpression, not the function
+// argument; ESLint's `getJSDocComment` stops walking at CallExpression
+// parents, so we do too.
+func hasJSDocThisTag(fn *ast.Node, comments []*ast.CommentRange, sf *ast.SourceFile) bool {
```
elecmonkey, line 502: [P2] Preserve @this comments placed after export

When an `@this` comment appears between an export keyword and a function declaration, such as `export /* @this */ function foo(){ this; }` or the equivalent `export default` form, ESLint 10.8.1 and `@typescript-eslint/no-invalid-this` 8.67.0 report nothing, while both Rslint rules report `unexpectedThis`. The exported ts-go declaration starts at `export`, so this leading-comment scan stops before reaching the comment attached to the function token. Please also inspect comments immediately preceding the function token for exported declarations.

swwind: Fixed. `hasJSDocThisTag` now inspects contiguous comments between the export modifiers and the ESTree declaration start, covering both `export` and `export default` function declarations. Focused regressions were added for the core rule and the TypeScript wrapper, and both affected package tests pass.

internal/rules/no_invalid_this/no_invalid_this.go
```diff
@@ -0,0 +1,1172 @@
...
+	if hasThisTagInLeadingComments(fn, comments, sf) {
```
elecmonkey: [P2] Restrict comments before export to actual JSDoc

For an exported declaration such as `/* @this */ export function f(){ this; }` (or a preceding `// @this`), `hasThisTagInLeadingComments` scans from ts-go's declaration position at `export` and treats the ordinary comment as a callback-style leading comment, so both native rules emit no diagnostic. ESLint 10.8.1 and `@typescript-eslint/no-invalid-this` 8.67.0 instead report `unexpectedThis` at the inner `this`, because comments before the export wrapper count only through `findJSDocComment`, which requires an adjacent `/** ... */` block. Keep the broad leading-comment scan scoped to the ESTree function start and handle comments before the export wrapper with JSDoc-only semantics.

swwind: Fixed in c3a335bc9. Exported function comments are now split at the export modifiers: broad leading-comment matching applies only between `export`/`default` and the function token, while comments before the export wrapper use adjacent `/** ... */` JSDoc semantics. Added focused core and TypeScript-wrapper regressions, plus an extended parity matrix covering `export default`, comment shadowing, blank-line adjacency, multiple leading comments, and async modifier boundaries. Both affected package test suites pass.

internal/rules/no_invalid_this/no_invalid_this.go
```diff
@@ -0,0 +1,1182 @@
...
+		if !valid {
+			ctx.ReportNode(node, msg)
+		}
+	}
+
+	return rule.RuleListeners{
+		// Non-arrow function-like containers — push a frame whose validity
+		// depends on strict mode, parameter shape, JSDoc, name, and
+		// surrounding context.
+		ast.KindFunctionDeclaration:                      pushFunction,
```
elecmonkey, line 234: [P2] Bodyless TypeScript signatures should not push a `this` frame here. For example, `function F(x: typeof this): void;` is exposed upstream as `TSDeclareFunction`, so ESLint 10.8.1 and `@typescript-eslint/no-invalid-this` 8.67.0 report `unexpectedThis` on the `typeof this`, but Rslint reports nothing because ts-go exposes the signature as `KindFunctionDeclaration` and this listener pushes a valid frame via the capitalized name. The same false negative happens with `this` parameters, `@this` JSDoc, ambient `declare` functions, and bodyless/abstract class method signatures; declarations without bodies should leave type-query `this` in the enclosing frame.

swwind: Fixed in 3ae15a6f9. Bodyless function declarations, methods, constructors, and accessors now leave the enclosing validity frame active, including computed-name members; abstract properties likewise skip the `PropertyDefinition` frame. Focused regressions cover capitalized and ambient functions, `this` parameters, `@this` JSDoc, abstract methods, getters and properties, and computed bodyless methods in both the core and TypeScript wrapper packages. `go test ./internal/rules/no_invalid_this/` and `go test ./internal/plugins/typescript/rules/no_invalid_this/` both pass.

elecmonkey (approved):
LGTM

PR #1738, opened 2026-08-14
feat: port rule func-names

swwind:
## Summary

Port the `func-names` rule from ESLint to rslint.

Requires or disallows named `function` expressions, with `always`/`as-needed`/`never` modes and an independent `generators` sub-option.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/func-names
- Source code: https://github.com/eslint/eslint/blob/v10.8.1/lib/rules/func-names.js

## Checklist


internal/rules/func_names/func_names.go
```diff
@@ -0,0 +1,186 @@
...
+		// Covers a plain or compound assignment (`foo = function(){}`,
+		// `foo ??= function(){}`) and a destructuring default inside a
+		// non-shorthand property or array element (`{key: foo =
+		// function(){}}`, `[foo = function(){}]`, always plain `=`): tsgo
+		// parses all of these as BinaryExpression. Upstream's
+		// AssignmentExpression check doesn't gate on operator either — see
+		// the "quux ??= function() {};" as-needed example in the rule doc.
+		be := parent.AsBinaryExpression()
+		return ast.IsAssignmentOperator(be.OperatorToken.Kind) &&
+			be.Left.Kind == ast.KindIdentifier &&
```
elecmonkey: With `as-needed`, assignments whose left identifier is parenthesized are inferred-name contexts in ESLint 10.8.1: both `(foo) = function() {};` and `((foo)) = function() {};` produce no diagnostic because Espree exposes the left side as the identifier `foo`. tsgo retains those parentheses, but this branch checks `be.Left.Kind` directly, so rslint reports the function as unnamed. Please unwrap the assignment LHS before the identifier check, as is already done for the RHS.

swwind: Good catch, fixed in b8dfe8cd.

`hasInferredName` now unwraps the assignment target before the identifier check:

```go
return ast.IsAssignmentOperator(be.OperatorToken.Kind) &&
    ast.SkipParentheses(be.Left).Kind == ast.KindIdentifier &&
    ast.SkipParentheses(be.Right) == node
```

Added three `as-needed` valid cases to `func_names_extras_test.go` covering `(foo) = function() {};`, `((foo)) = function() {};` and `(foo) ??= function() {};`. All three fail without the change.

The other inferred-name branches don't need the same treatment — `VariableDeclaration`, `Parameter` and `BindingElement` names are binding positions, so the grammar doesn't allow parentheses there.

internal/rules/func_names/func_names.go
```diff
@@ -0,0 +1,186 @@
...
+	case ast.KindVariableDeclaration:
+		vd := parent.AsVariableDeclaration()
+		return parent.Name().Kind == ast.KindIdentifier &&
+			vd.Initializer != nil && ast.SkipParentheses(vd.Initializer) == node
+	case ast.KindPropertyAssignment:
+		pa := parent.AsPropertyAssignment()
+		return pa.Initializer != nil && ast.SkipParentheses(pa.Initializer) == node
+	case ast.KindPropertyDeclaration:
+		pd := parent.AsPropertyDeclaration()
+		return pd.Initializer != nil && ast.SkipParentheses(pd.Initializer) == node
```
elecmonkey: PropertyDeclaration is broader than ESLints PropertyDefinition here. In tsgo, both a normal class field and a TS auto-accessor (`accessor foo = ...`) land on PropertyDeclaration, but upstream @typescript-eslint/parser maps the auto-accessor parent to AccessorProperty, so ESLint does not treat it as an inferred-name context. As a result, `class C { accessor foo = function() {} }` should still report `Unexpected unnamed function.` in `as-needed` mode. Please gate this branch on `!ast.HasSyntacticModifier(parent, ast.ModifierFlagsAccessor)` before treating the initializer as inferred-name.

swwind: Good catch, fixed in bec175b0.

The `PropertyDeclaration` branch now skips auto-accessors:

```go
case ast.KindPropertyDeclaration:
	pd := parent.AsPropertyDeclaration()
	return !ast.HasSyntacticModifier(parent, ast.ModifierFlagsAccessor) &&
		pd.Initializer != nil && ast.SkipParentheses(pd.Initializer) == node
```

Verified against ESLint 10.8.1 with `@typescript-eslint/parser` 8.67: `class C { accessor foo = function() {}; }` under `as-needed` reports `Unexpected unnamed function.` — plain `function`, not `method 'foo'`, and the range covers only the `function` keyword, because `getFunctionNameWithKind` and `getFunctionHeadLoc` don't recognise `AccessorProperty` either. The shared `GetFunctionNameWithKindCore` / `GetFunctionHeadLoc` helpers already gate on `ModifierFlagsAccessor`, so only `hasInferredName` needed the fix.

Added two `as-needed` invalid cases to `func_names_extras_test.go` covering `accessor foo = function() {}` and `static accessor foo = function() {}`; both fail without the change.

elecmonkey (approved):
no more issues found

fansenze (approved):

PR #1739, opened 2026-08-14
fix: align block-scoped-var with ESLint

fansenze:
## Summary

- Normalize merged `var` identities and declaration occurrences so diagnostics, counts, and definition locations match ESLint 10.8.
- Model declaration, parameter, destructuring, TypeScript, and TSX reference events with exact TSESTree ranges while reusing binder, RefStore, and shared traversal helpers.
- Add adversarial regression coverage for the original review cases and the discovered JavaScript, TypeScript, and TSX boundaries.

### Go benchmark

Command: `go test ./internal/rules/block_scoped_var -run "^$" -bench "^BenchmarkBlockScopedVar$" -benchmem -benchtime=100x -count=5`

Median of five samples:

| Case | ns/op | B/op | allocs/op |
| --- | ---: | ---: | ---: |
| NoVar | 5,052 → 3,499 | 0 → 0 | 0 → 0 |
| InScopeUnique | 62,547 → 63,408 | 80,161 → 109,025 | 742 → 773 |
| OutOfScopeUnique | 75,393 → 91,005 | 97,604 → 133,561 | 999 → 1,027 |
| MergedAndDuplicate | 112,007 → 117,619 | 89,901 → 128,495 | 881 → 1,012 |
| DestructuringAndCatch | 228,487 → 158,620 | 147,190 → 118,215 | 1,577 → 851 |
| JSXAndParameterDefaults | 41,964 → 69,643 | 26,400 → 62,482 | 208 → 578 |

The slower matched cases perform reference and diagnostic work that the prior implementation omitted. The zero-allocation no-match path is 30.7% faster, and the destructuring/catch path is 30.6% faster.

## Related Links

N/A

## Checklist

### DIFF
--- a/internal/plugins/typescript/rules/fixtures/tsconfig.block-scoped-var-jsx.json
+++ b/internal/plugins/typescript/rules/fixtures/tsconfig.block-scoped-var-jsx.json
@@ -1,0 +1,11 @@
+{
+  "compilerOptions": {
+    "jsx": "preserve",
+    "jsxFactory": "preact.h",
+    "jsxFragmentFactory": "preact.Fragment",
+    "target": "esnext",
+    "module": "commonjs",
+    "strict": true,
+    "types": []
+  }
+}
--- a/internal/rules/block_scoped_var/block_scoped_var.go
+++ b/internal/rules/block_scoped_var/block_scoped_var.go
@@ -2,13 +2,18 @@ package block_scoped_var
 
 import (
 	"fmt"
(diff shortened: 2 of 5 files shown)


swwind:
I’m not convinced we need to pursue such extensive ESLint parity—the implementation growing from 177 to ~2,540 lines seems like a very high cost for behavior that provides little practical value. And even copy the bug of eslint.

PR #1740, opened 2026-08-14
feat: port rule consistent-this

swwind:
## Summary

Port the `consistent-this` rule from ESLint to rslint.

It enforces two things about variables with a designated alias name for `this`: a declared alias must be initialized or assigned `this` in the same scope as its declaration, and any variable assigned `this` must use one of the designated alias names.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/consistent-this
- Source code: https://github.com/eslint/eslint/blob/v10.8.1/lib/rules/consistent-this.js

## Checklist


internal/rules/consistent_this/consistent_this.go
```diff
@@ -0,0 +1,320 @@
...
+}
+
+// referenceScopeNode is the scope a reference belongs to, standing in for
+// upstream's `reference.from`. It is the nearest enclosing block-scope
+// container, except that a loop whose header declares no block-scoped
+// binding is walked past: tsgo counts every `for`/`for-in`/`for-of`
+// statement as a container, while eslint-scope opens a scope for one only
+// when its header declares with `let`, `const` or `using`.
+func referenceScopeNode(ref *ast.Node) *ast.Node {
+	scopeNode := ast.GetEnclosingBlockScopeContainer(ref)
```
elecmonkey, line 261: [P2] Skip object method scopes for computed keys

When an assignment appears in a computed object-literal key, such as `function f() { var self; ({ [self = this]() {} }); }`, ESLint treats the write as belonging to `f` and reports nothing, but `GetEnclosingBlockScopeContainer` returns the method declaration here, so Rslint incorrectly reports `self` as unassigned. The scope mapping needs to walk past object methods/accessors when the reference is inside their computed key.

internal/rules/consistent_this/consistent_this.go
```diff
@@ -0,0 +1,320 @@
...
+	for scopeNode != nil && isLoopWithoutOwnScope(scopeNode) {
+		scopeNode = ast.GetEnclosingBlockScopeContainer(scopeNode)
+	}
+	return scopeNode
```
elecmonkey, line 268: [P2] Account for `with` scopes when matching writes

For valid sloppy-mode code such as `function f() { var self; with (obj) self = this; }`, ESLint reports the declaration because the write's `reference.from` is the `with` scope rather than the function scope. Since ts-go's block-scope helper does not include `with` statements, this function returns the enclosing function and Rslint incorrectly accepts the declaration as assigned in the same scope.

internal/rules/consistent_this/consistent_this.go
swwind: Confirmed against ESLint — `function f() { var self; ({ [self = this]() {} }); }` is clean upstream, and the same for `get`/`set` computed keys, while Rslint reported the declaration.

Fixed by `isComputedKeyOfObjectLiteralMember`: a reference inside an object-literal member's computed key now walks past that member to the scope the object literal is written in. A class member's computed key is deliberately left alone — it belongs to the class scope, which is likewise never the enclosing function's, so `class C { [self = this]() {} }` keeps reporting, matching upstream.

internal/rules/consistent_this/consistent_this.go
swwind: Confirmed — ESLint reports the declaration for `function f() { var self; with (obj) self = this; }` and Rslint accepted it.

Fixed by `enclosingWithStatement`: a `with` body between the reference and the nearest tsgo block-scope container is now the reference's scope. The object expression is not matched, since it is evaluated in the enclosing scope — `with (self = this) {}` stays clean, as upstream.

internal/rules/consistent_this/consistent_this.go
```diff
@@ -0,0 +1,363 @@
...
+		// alias among scopeNode's own directly-declared bindings and, unless
+		// already initialized or later assigned `this` in that exact scope,
+		// reports every declaration of it.
+		checkWasAssigned := func(scopeNode *ast.Node, alias string) {
+			locals := ast.GetLocals(scopeNode)
+			if locals == nil {
+				return
+			}
+			sym := locals[alias]
+			if sym == nil || sym.Flags&aliasSymbolFlags == 0 {
```
elecmonkey: [P2] Include TypeScript declarations in alias checks

When TypeScript syntax is parsed with `@typescript-eslint/parser`, declarations such as `interface self {}`, `type self = string`, and `namespace self {}` create scope variables, so ESLint 10.8.1 reports `aliasNotAssignedToThis` when `self` is configured as an alias. The symbol-flag filter here excludes those declarations, causing Rslint to report nothing; the added `type self = string` valid test therefore locks in a mismatch rather than upstream behavior.

internal/rules/consistent_this/consistent_this.go
```diff
@@ -0,0 +1,363 @@
...
+
+// referenceScopeNode is the scope a reference belongs to, standing in for
+// upstream's `reference.from`. It is the nearest enclosing block-scope
+// container, with three corrections where tsgo's containers and
+// eslint-scope's scopes disagree: a loop whose header declares no
+// block-scoped binding is walked past, an object-literal member is walked
+// past for a reference in its computed key, and a `with` body — no container
+// of tsgo's at all — is a scope of its own.
+func referenceScopeNode(ref *ast.Node) *ast.Node {
+	scopeNode := ast.GetEnclosingBlockScopeContainer(ref)
```
elecmonkey, line 261: [P2] Treat enum initializers as a separate reference scope

For TypeScript enum initializers such as `function f() { var self; enum E { A = (self = this, 1) } }`, `@typescript-eslint/parser` places the write in a `tsEnum` scope, so ESLint 10.8.1 reports the `self` declaration as unassigned in the function scope. `GetEnclosingBlockScopeContainer` does not recognize enums and this mapping returns the function, so Rslint incorrectly accepts the declaration with no diagnostic.

internal/rules/consistent_this/consistent_this.go
swwind: Confirmed against ESLint 10.8.1 — `type self = string`, `interface self {}` and `namespace self {}` all report `aliasNotAssignedToThis`, and so does a function type parameter (`function f<self>() {}`), while Rslint reported nothing.

Fixed by dropping the symbol-flag filter: every local of a tsgo scope is a variable of that scope as typescript-eslint's scope manager sees it, type-space ones included. The `type self = string` case moved from valid to invalid, with `interface` / `namespace` / `export namespace` / type-parameter cases locked in alongside it. A class's or interface's own type parameter (`class C<self> {}`) still reports nothing, matching upstream, since a class scope is never one this rule checks.

internal/rules/consistent_this/consistent_this.go
swwind: Confirmed — ESLint reports the declaration for `function f() { var self; enum E { A = (self = this, 1) } }` and Rslint accepted it.

Fixed by widening the `with` handling into `enclosingUnmappedScope`, which now matches an enum body too: both are scopes for eslint-scope and no container at all for tsgo. An enum's own name is bound in the enclosing scope, so only its members are matched — `function f() { var self; enum E { A = 1 } self = this; }` stays valid.

fansenze (approved):

PR #1745, opened 2026-08-16
feat: port rule id-length

swwind:
## Summary

Port the `id-length` rule from ESLint to rslint.

Enforces a minimum and/or maximum identifier length convention, exempting property names via `properties`, an `exceptions` list, and `exceptionPatterns` regular expressions.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/id-length
- Source code: https://github.com/eslint/eslint/blob/v10.8.1/lib/rules/id-length.js

## Checklist


internal/rules/id_length/id_length.go
```diff
@@ -0,0 +1,575 @@
...
+		for _, e := range arr {
+			if s, ok := e.(string); ok {
+				opts.exceptions[s] = true
+			}
+		}
+	}
+	if arr, ok := m["exceptionPatterns"].([]any); ok {
+		for _, p := range arr {
+			s, ok := p.(string)
+			if !ok {
```
elecmonkey, line 239: Configurations using JavaScript-only regexp syntax are silently ignored here. For example, ESLint 10.8.1 accepts `exceptionPatterns: ["(?=x)"]` and exempts `var x = 1`, while Go's `regexp.Compile` rejects the lookahead and this implementation reports `x`. The schema permits arbitrary strings, so this is reachable in normal ESLint-compatible configs (lookbehind and backreferences have the same issue). Please compile/match these patterns with JavaScript RegExp semantics rather than dropping patterns Go cannot parse.

swwind: Fixed in 47323c01. The branch now has #1705 merged in, so `exceptionPatterns` compiles through `internal/utils/ecmascript/regexp` with the `u` flag — the same as upstream's `new RegExp(pattern, "u")` — and the schema's items carry `format: "regex"`, so a pattern that genuinely does not compile is rejected at config validation rather than silently dropped.

Checked against ESLint 10.8.1:

| Config | ESLint | Before | After |
| --- | --- | --- | --- |
| `exceptionPatterns: ["(?=x)"]`, `var x = 1` | exempt | reports `x` | exempt |
| `exceptionPatterns: ["(?<=^)x"]`, `var x = 1` | exempt | reports `x` | exempt |
| `min: 3, exceptionPatterns: ["^(.)\\1$"]`, `var xx = 1; var ab = 2;` | reports `ab` only | reports both | reports `ab` only |
| `exceptionPatterns: ["^\\p{Ll}$"]`, `var e = 1` | exempt | exempt | exempt |

All four are locked in as test cases; the last one covers the `u` flag itself, since without it `\p` is an identity escape and the pattern would only match the literal `p{Ll}`.

internal/rules/id_length/id_length.go
```diff
@@ -0,0 +1,580 @@
...
+		return isValidImportSpecifier(effectiveParent, wrapped)
+
+	case ast.KindImportClause:
+		return effectiveParent.Name() == wrapped
+
+	case ast.KindNamespaceImport:
+		return effectiveParent.Name() == wrapped
+
+	case ast.KindParameter:
+		return effectiveParent.Name() == wrapped
```
elecmonkey: Handling every tsgo KindParameter unconditionally diverges from upstream for TS parameter properties. With class Foo { constructor(private x: number) {} }, ESLint 10.8.1 plus @typescript-eslint/parser 8.65.0 emits no id-length diagnostic because the identifier parent is TSParameterProperty, which this rule does not support. Rslint reports x here, so this introduces a TypeScript-only false positive.

internal/rules/id_length/id_length.go
```diff
@@ -0,0 +1,580 @@
...
+
+	case ast.KindClassDeclaration:
+		return effectiveParent.Name() == wrapped
+
+	case ast.KindFunctionDeclaration, ast.KindFunctionExpression:
+		return effectiveParent.Name() == wrapped
+
+	case ast.KindMethodDeclaration, ast.KindGetAccessor, ast.KindSetAccessor:
+		return effectiveParent.Name() == wrapped
```
elecmonkey, line 369: This unconditional method branch is applying MethodDefinition semantics to object-literal methods too. Under { properties: never }, ESLint accepts const obj = { x() {} }, but Rslint reports x; with default options ESLint also accepts const obj = { [x]() {} }, while this implementation reports the computed key. tsgo reuses the same kinds for object and class members here, so the dispatch needs to inspect the containing node before treating this as a class MethodDefinition.

elecmonkey, line 369: KindConstructor is missing from this unconditional-method dispatch, so class constructors are never checked against max. For example, class Foo { constructor() {} } with { min: 0, max: 5 } produces one ESLint diagnostic for constructor but none here.

internal/rules/id_length/id_length.go
```diff
@@ -0,0 +1,580 @@
...
+// that upstream would see as ObjectPattern's Property branch — i.e. the
+// enclosing object literal is (transitively) a destructuring-assignment
+// target.
+func isValidPatternPropertyValue(opts idLengthOptions, keyNode, valueNode *ast.Node, node *ast.Node) bool {
+	return isValidPatternPropertyValueNodes(opts, keyNode, valueNode, node)
+}
+
+func isValidPatternPropertyValueNodes(opts idLengthOptions, keyNode, valueNode *ast.Node, node *ast.Node) bool {
+	if sameIdentifierText(keyNode, valueNode) {
+		return node == keyNode && opts.properties
```
elecmonkey, line 488: This sameIdentifierText shorthand path makes all object binding names depend on properties, but upstream treats AssignmentPattern and RestElement unconditionally. With { properties: never }, ESLint still reports x in both const { x = 1 } = obj and const { ...x } = obj, while this returns false for those cases because key and value text match. Defaulted and rest bindings need to stay independent of the properties option.

internal/rules/id_length/id_length.go
swwind: Fixed in 5d2bb505. A `KindParameter` whose declaration is a parameter property (`ast.IsParameterPropertyDeclaration`) and carries no initializer is now skipped, matching TSParameterProperty's absence from SUPPORTED_EXPRESSIONS. A default value is still checked, since typescript-eslint then nests an AssignmentPattern between the two and that entry is unconditional.

Checked against ESLint 10.8.1 with @typescript-eslint/parser 8.67.0:

| Code | ESLint | Before | After |
| --- | --- | --- | --- |
| `class Foo { constructor(private x: number) {} }` | no report | reports `x` | no report |
| `class Foo { constructor(public readonly x: number) {} }` | no report | reports `x` | no report |
| `class Foo { constructor(private x = 1) {} }` | reports `x` | reports `x` | reports `x` |
| `class Foo { constructor(x: number) {} }` | reports `x` | reports `x` | reports `x` |

All four are locked in as test cases.

internal/rules/id_length/id_length.go
swwind: Fixed in 5d2bb505. The method/accessor branch now dispatches on the containing node: a class member keeps MethodDefinition semantics, an object-literal member goes through the same `isValidPlainProperty` path as a `key: value` pair (so it honours both `properties` and `isImportAttributeKey`), and an interface or type-literal member is a TSMethodSignature and is not checked at all. `isValidComputedPropertyNameOwner` is restricted to class members for the same reason.

| Code | ESLint | Before | After |
| --- | --- | --- | --- |
| `const obj = { x() {} };` + `properties: never` | no report | reports `x` | no report |
| `const obj = { get x() { return 1; } };` + `properties: never` | no report | reports `x` | no report |
| `const obj = { [x]() {} };` | no report | reports `x` | no report |
| `const obj = { get [x]() { return 1; } };` | no report | reports `x` | no report |
| `import('foo.json', { with: { a() {} } });` | no report | reports `a` | no report |
| `class C { x() {} }` + `properties: never` | reports `C`, `x` | reports `C`, `x` | reports `C`, `x` |

The same containing-node check turned up two more TS-only false positives, fixed in the same commit: `abstract` and `accessor` class members are TSAbstractMethodDefinition / TSAbstractPropertyDefinition / TSAbstractAccessorProperty / AccessorProperty rather than MethodDefinition / PropertyDefinition, and a body-less function declaration (an overload signature or `declare function`) is a TSDeclareFunction rather than a FunctionDeclaration. Neither is a supported parent, so ESLint checks none of them.

swwind: Fixed in 5d2bb505, though not as a missing switch case: tsgo's ConstructorDeclaration has no name node at all — `constructor` is a keyword token, so no Identifier listener ever fires for it. The rule now carries a dedicated `ast.KindConstructor` listener that runs the same length and exception check and reports on the keyword's own range, located by scanning past any modifiers.

| Code (with `{ min: 0, max: 5 }`) | ESLint | Before | After |
| --- | --- | --- | --- |
| `class Foo { constructor() {} }` | reports at 1:13-24 | no report | reports at 1:13-24 |
| `class Foo { public constructor() {} }` | reports at 1:20-31 | no report | reports at 1:20-31 |
| `const Cls = class { constructor() {} };` | reports at 1:21-32 | no report | reports at 1:21-32 |
| `class C2 { constructor(); constructor(a?: number) {} }` | two reports | no report | two reports |
| plus `exceptions: ["constructor"]` | no report | no report | no report |
| `const o2 = { constructor() {} };` | reports at 1:14 | reports at 1:14 | reports at 1:14 |

internal/rules/id_length/id_length.go
swwind: Fixed in 5d2bb505. An object-binding element carrying a rest token or an initializer now returns early and checks the binding name unconditionally, before the key/value comparison is reached — upstream's pattern conversion hands those to RestElement and AssignmentPattern respectively, and neither consults `properties`. The `({ x = 1 } = obj)` spelling gets the same treatment on the ShorthandPropertyAssignment branch via its ObjectAssignmentInitializer.

| Code (all with `properties: never`) | ESLint | Before | After |
| --- | --- | --- | --- |
| `const { x = 1 } = obj;` | reports `x` | no report | reports `x` |
| `const { ...x } = obj;` | reports `x` | no report | reports `x` |
| `function ff({ x = 1 }) {}` | reports `x` | no report | reports `x` |
| `({ x = 1 } = obj);` | reports `x` | no report | reports `x` |
| `const { longName: b = 1 } = obj;` | reports `b` only | reports `b` only | reports `b` only |
| `const { x } = obj;` | no report | no report | no report |

fansenze (approved):

PR #1749, opened 2026-08-16
feat: port rule no-inline-comments

swwind:
## Summary

Port the `no-inline-comments` rule from ESLint to rslint.

Disallows comments placed on the same line as code on either side of them, with a JSX exception for a comment that is the sole content of a JSX expression container, and an `ignorePattern` option to exempt comments matching a user-supplied regex.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/no-inline-comments
- Source code: https://github.com/eslint/eslint/blob/v10.8.1/lib/rules/no-inline-comments.js

## Checklist


swwind:
Blocked by #1705

internal/rules/no_inline_comments/no_inline_comments.go
```diff
@@ -0,0 +1,120 @@
...
+
+			// Don't report ESLint directive comments.
+			if utils.IsDirectiveComment(comment.Kind, ecmascript.StringTrim(value)) {
+				return
+			}
+
+			ctx.ReportRange(comment.TextRange, unexpectedInlineCommentMessage)
+		}
+
+		for _, comment := range ctx.Comments.All() {
```
elecmonkey, line 114: When linting script source such as `var x = 1<!--foo`, ESLint 10.8.1 treats `<!--foo` as a line comment and reports it at columns 10-17. This loop only receives slash-prefixed comments from `ctx.Comments.All()`, so the PR head emits no `no-inline-comments` diagnostic for this valid Annex B form. Include HTML-style line comments in this rule's comment stream or handle them separately.

internal/rules/no_inline_comments/no_inline_comments.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+func parseOptions(raw []any) options {
+	opts := options{}
+	if len(raw) == 0 {
+		return opts
+	}
+	m, _ := raw[0].(map[string]any)
+	if pattern, ok := m["ignorePattern"].(string); ok && pattern != "" {
+		if re, err := esregexp.Compile(pattern, "u"); err == nil {
```
elecmonkey, line 34: This now uses Unicode-mode regexp compilation, but the error path is still silently ignored. For example, `ignorePattern: "\\a"` is valid without flags but invalid for `new RegExp(pattern, "u")`; ESLint 10.8.1 throws during rule initialization, while this branch leaves `ignorePattern` nil and linting continues as if no ignore pattern was configured. Please surface this compile failure through option validation / rule initialization instead of dropping it.

internal/rules/no_inline_comments/no_inline_comments.go
swwind: That's tsgo's problem, they do not parse `<!--` at all. Nothing we can do here.

internal/rules/no_inline_comments/no_inline_comments.go
swwind: That's a potential bug here, but currently all other rules do a silent ignore on regex that is valid without "u" but invalid with "u". So we just follow other rules.

elecmonkey (approved):

PR #1750, opened 2026-08-16
feat: port rule no-restricted-properties

swwind:
## Summary

Port the `no-restricted-properties` rule from ESLint to rslint.

Disallows specified properties on specified objects. Supports restricting a specific object+property pair, a whole object (every property), or a property globally (on every object) — each with an optional `allowObjects`/`allowProperties` escape hatch and a custom message. Applies to dot/bracket property access and to destructuring.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/no-restricted-properties
- Source code: https://github.com/eslint/eslint/blob/v10.8.1/lib/rules/no-restricted-properties.js

## Checklist


elecmonkey:
LGTM

internal/rules/no_restricted_properties/no_restricted_properties.go
```diff
@@ -0,0 +1,360 @@
...
+		}
+	}
+
+	if matched.found && (!matched.hasAllowList || !isAllowedName(propertyName, matched.allowList)) {
+		allowedPropertiesMessage := ""
+		if matched.hasAllowList {
+			allowedPropertiesMessage = " Only these properties are allowed: " + strings.Join(matched.allowList, ", ") + "."
+		}
+		messageSuffix := ""
+		if matched.hasMessage {
```
elecmonkey: When a restriction explicitly sets `message: ""`, ESLint 10.8.1 treats the empty string as falsy and emits the normal message with no suffix, but this `hasMessage` check appends a trailing space. For example, `{ object: "foo", property: "bar", message: "" }` on `foo.bar` yields `'foo.bar' is restricted from being used. ` here instead of ESLint's `'foo.bar' is restricted from being used.`. Please gate the suffix on a non-empty message value (and likewise in the property-only branch).

swwind: Good catch, fixed in 15591ef2.

Dropped the `hasMessage` flag from all three restriction structs and gated the suffix on `message != ""` instead, which matches upstream's falsy check in both the object-property branch and the property-only branch. Verified against ESLint 10.8.1:

| Options | ESLint | This PR |
| --- | --- | --- |
| `{object:"foo",property:"bar",message:""}` | `'foo.bar' is restricted from being used.` | same |
| `{object:"foo",message:""}` | `'foo.bar' is restricted from being used.` | same |
| `{property:"bar",message:""}` | `'bar' is restricted from being used.` | same |

Added the three cases to the extras test.

internal/rules/no_restricted_properties/no_restricted_properties.go
```diff
@@ -0,0 +1,352 @@
...
+// (`this.#foo`) is deliberately never static: it's a distinct name-space from
+// string-keyed properties, matching upstream's astUtils.getStaticPropertyName
+// (which only special-cases plain Identifier member names).
+func staticMemberPropertyName(node *ast.Node) (string, bool) {
+	switch node.Kind {
+	case ast.KindPropertyAccessExpression:
+		return utils.GetStaticPropertyName(node.AsPropertyAccessExpression().Name())
+	case ast.KindElementAccessExpression:
+		arg := ast.SkipParentheses(node.AsElementAccessExpression().ArgumentExpression)
+		return utils.GetStaticExpressionValue(arg)
```
elecmonkey, line 244: [P2] Stringify numeric keys with JavaScript semantics

When a computed numeric key crosses JavaScript's exponential-format thresholds, this produces the wrong property name. For example, `foo[1e21]` with `{ property: "1000000000000000000000" }` is reported here because `GetStaticExpressionValue` formats the number in fixed notation, whereas ESLint 10.8.1 stringifies it as `1e+21` and reports nothing; conversely, `{ property: "1e+21" }` is missed. The same mismatch occurs below `1e-6` and for computed destructuring keys, so numeric literals need JavaScript Number-to-string semantics.

swwind: Good catch, fixed in 46e976fb.

The root cause was in the shared helper `utils.NormalizeNumericLiteral`, which formatted every literal with `strconv.FormatFloat(f, 'f', -1, 64)` — always fixed notation. It now goes through `ecmascript.NumberToString`, the existing Number::toString implementation, which switches to exponential notation outside `[1e-6, 1e21)`. That path backs both element access and computed destructuring keys, so both spellings line up now. Verified against ESLint 10.8.1:

| Code | Options | ESLint | This PR |
| --- | --- | --- | --- |
| `foo[1e21]` | `{property:"1e+21"}` | reported | same |
| `foo[1e21]` | `{property:"1000000000000000000000"}` | not reported | same |
| `foo[1e-7]` | `{property:"1e-7"}` | reported | same |
| `foo[1e-7]` | `{property:"0.0000001"}` | not reported | same |
| `let {[1e21]: x} = foo;` | `{property:"1e+21"}` | reported | same |
| `let {[1e21]: x} = foo;` | `{property:"1000000000000000000000"}` | not reported | same |
| `foo[1e21]` | `{object:"foo",property:"1e+21"}` | reported | same |

All seven are in the extras test.

internal/rules/no_restricted_properties/no_restricted_properties.go
```diff
@@ -0,0 +1,352 @@
...
+				objectExpr = node.AsElementAccessExpression().Expression
+			}
+			objectName, hasObjectName := staticObjectName(objectExpr)
+			propertyName, hasPropertyName := staticMemberPropertyName(node)
+			checkPropertyAccess(ctx, r, node, objectName, hasObjectName, propertyName, hasPropertyName)
+		}
+
+		return rule.RuleListeners{
+			ast.KindPropertyAccessExpression: checkMemberAccess,
+			ast.KindElementAccessExpression:  checkMemberAccess,
```
elecmonkey, line 405: When linting TSX such as `<foo.bar />` with `{ object: "foo", property: "bar" }`, ts-go represents the tag name as a `PropertyAccessExpression`, so this listener reports it; ESLint 10.8.1 represents it as a `JSXMemberExpression` and reports nothing because the upstream rule only visits `MemberExpression`. A paired tag such as `<foo.bar></foo.bar>` is consequently reported twice. Please exclude every property-access link that belongs to a JSX tag-name chain before calling `checkMemberAccess`.

swwind: Good catch, fixed in 7e6f05d3c.

Added `utils.IsInJsxTagName` to exclude every property-access link in dotted JSX tag-name chains, and reused it in the existing rules with the same ESTree/ts-go distinction.

Added focused regression coverage for:

- self-closing `<foo.bar />`
- paired `<foo.bar></foo.bar>`
- chained `<foo.bar.baz />`
- ordinary member access inside JSX attributes and children, which must still report

Verified the matrix against ESLint 10.8.1 and ran the affected Go package tests successfully.

elecmonkey (commented):
Found one remaining parity issue.

internal/rules/no_restricted_properties/no_restricted_properties.go
```diff
@@ -0,0 +1,356 @@
...
+// (`this.#foo`) is deliberately never static: it's a distinct name-space from
+// string-keyed properties, matching upstream's astUtils.getStaticPropertyName
+// (which only special-cases plain Identifier member names).
+func staticMemberPropertyName(node *ast.Node) (string, bool) {
+	switch node.Kind {
+	case ast.KindPropertyAccessExpression:
+		return utils.GetStaticPropertyName(node.AsPropertyAccessExpression().Name())
+	case ast.KindElementAccessExpression:
+		arg := ast.SkipParentheses(node.AsElementAccessExpression().ArgumentExpression)
+		return utils.GetStaticExpressionValue(arg)
```
elecmonkey, line 244: [P2] Normalize large radix element keys from the raw token

When bracket access uses a binary, octal, or hexadecimal literal above 2^53, this path re-parses `NumericLiteral.Text` through `GetStaticExpressionValue`, which can round differently from Espree's numeric value. For `foo[0x1000000000000281]` with `{ property: "1152921504606847500" }`, ESLint 10.8.1 reports, while this branch derives `1152921504606847700` and misses it (and incorrectly reports for that latter option). Computed destructuring already avoids this through the source-aware `GetStaticPropertyName` path; element access needs equivalent normalization.

internal/rules/no_restricted_properties/no_restricted_properties.go
```diff
@@ -0,0 +1,356 @@
...
+			ast.KindPropertyAccessExpression: checkMemberAccess,
+			ast.KindElementAccessExpression:  checkMemberAccess,
+			ast.KindObjectBindingPattern: func(node *ast.Node) {
+				checkPattern(ctx, r, node)
+			},
+			ast.KindObjectLiteralExpression: func(node *ast.Node) {
+				if !utils.IsInDestructuringAssignment(node) {
+					return
+				}
+				checkPattern(ctx, r, node)
```
elecmonkey, line 413: [P2] Limit assignment handling to actual object patterns

When an object literal occurs inside a member-expression assignment target, `IsInDestructuringAssignment` considers it part of the surrounding pattern even though ESTree keeps it as an `ObjectExpression`. For valid `let x; ({ a: ({bad}).x } = foo);` with `{ property: "bad" }`, ESLint 10.8.1 reports nothing, but this listener visits `{bad}` and emits `restrictedProperty`. The guard should admit only object literals that correspond to actual `ObjectPattern` nodes.

elecmonkey (commented):
Found one remaining parity issue.

internal/rules/no_restricted_properties/no_restricted_properties.go
swwind: Good catch, fixed in cffce481.

I moved source-aware radix normalization into `utils.GetStaticExpressionValue`, so element-access callers consistently recover the raw token before applying JavaScript Number rounding above 2^53, rather than fixing only this rule.

Added regression coverage for hexadecimal, binary, octal, uppercase/separated, parenthesized, optional-chain, and object-specific forms. The tests also verify that the incorrectly rounded value does not match and that radix BigInt keys retain exact integer semantics.

`go test ./internal/rules/no_restricted_properties` and `go test ./internal/utils` pass.

internal/rules/no_restricted_properties/no_restricted_properties.go
swwind: Good catch, fixed in 0c913e43.

The `ObjectLiteralExpression` listener now uses ts-go's `IsArrayLiteralOrObjectLiteralDestructuringPattern` check, so only literals that correspond to actual ESTree `ObjectPattern` nodes are visited. Ordinary object expressions nested inside assignment targets are skipped.

Added the reported `let x; ({ a: ({bad}).x } = foo);` case as a focused regression test. `go test ./internal/rules/no_restricted_properties` passes.

internal/rules/no_restricted_properties/no_restricted_properties.go
```diff
@@ -0,0 +1,356 @@
...
+		}
+
+		return rule.RuleListeners{
+			ast.KindPropertyAccessExpression: checkMemberAccess,
+			ast.KindElementAccessExpression:  checkMemberAccess,
+			ast.KindObjectBindingPattern: func(node *ast.Node) {
+				checkPattern(ctx, r, node)
+			},
+			ast.KindObjectLiteralExpression: func(node *ast.Node) {
+				if !ast.IsArrayLiteralOrObjectLiteralDestructuringPattern(node) {
```
elecmonkey: [P2] Include for-in assignment patterns

When an assignment-form object pattern is used as a `for-in` initializer, such as `for ({bad} in xs) {}` with `{ property: "bad" }`, ESLint 10.8.1 visits the `ObjectPattern` and reports `restrictedProperty`, but this guard rejects it because `IsArrayLiteralOrObjectLiteralDestructuringPattern` recognizes direct `for-of` initializers but not `for-in`. Consequently, property-only restrictions are silently bypassed for this valid syntax; include direct `ForInStatement` initializer objects in the pattern check.

swwind: Good catch, fixed in 65449d06.

The `ObjectLiteralExpression` listener now recognizes direct `ForInStatement` initializers as destructuring patterns while retaining structural traversal for nested object and array patterns. This covers both `for ({bad} in xs) {}` and nested `for ({outer: {bad}} in xs) {}` without re-admitting ordinary object expressions inside assignment targets.

Added both as focused regression cases. `go test ./internal/rules/no_restricted_properties` passes.

internal/utils/ts_eslint.go
```diff
@@ -1857,7 +1860,7 @@ func GetStaticExpressionValue(node *ast.Node) (string, bool) {
 	case ast.KindStringLiteral:
 		return node.AsStringLiteral().Text, true
 	case ast.KindNumericLiteral:
-		return NormalizeNumericLiteral(node.AsNumericLiteral().Text), true
+		return numericLiteralPropertyName(node), true
```
elecmonkey, line 1866: **Parse radix literals without incremental float rounding.** When this path handles a sufficiently large binary, octal, or hexadecimal key, `radixLiteralValue` accumulates each digit directly into a `float64` (`value = value*radix + digit`), which can round differently from JavaScript's correctly-rounded literal conversion.

For example, with `{ property: "23363847825694777000" }`, ESLint 10.8.1 reports both `foo[0b10100010000111101000011111100111101111100110100011110110000010100]` and the equivalent computed destructuring key, but Rslint reports them only for the incorrect key `23363847825694780000`.

Parse the complete integer exactly (e.g. via `big.Int`) before converting to `float64` so the single rounding step matches ECMAScript.

internal/rules/no_restricted_properties/no_restricted_properties.go
```diff
@@ -0,0 +1,386 @@
...
+			return statement != nil && statement.Initializer == current
+		case ast.KindArrayLiteralExpression, ast.KindObjectLiteralExpression:
+			current = parent
+		case ast.KindPropertyAssignment:
+			property := parent.AsPropertyAssignment()
+			if property == nil || property.Initializer != current || parent.Parent == nil {
+				return false
+			}
+			current = parent.Parent
+		default:
```
elecmonkey, line 369: **Traverse object patterns nested under array rest elements.** When a property-only restriction targets an object pattern used as an array rest target — e.g. `([...{bad}] = xs);` with `{ property: "bad" }` — ESLint 10.8.1 visits the nested `ObjectPattern` and reports `restrictedProperty`. Here the ancestry walk reaches ts-go's `SpreadElement`, hits the `default:` branch, and returns `false`, so Rslint emits nothing.

The same omission occurs for `for ([...{bad}] of xs) {}` and `for ([...{bad}] in xs) {}`. Treat a spread element that belongs to a destructuring array as a transparent pattern ancestor and continue the walk to its parent.

elecmonkey (commented):
Two behavioral mismatches versus ESLint 10.8.1 on the new `no-restricted-properties` rule.

internal/utils/ts_eslint.go
swwind: Fixed in fb66eef0.

I added regression coverage for both the reported element-access case and the equivalent computed destructuring key.

Inspecting the pinned Espree/Acorn implementation showed that its radix reader rounds the multiplication and addition as separate JavaScript Number operations. Go could contract the original single expression into a fused multiply-add, producing a different result. Splitting those operations yields `23363847825694777000` for the reported case while preserving the existing large-radix parity cases.

`go test ./internal/rules/no_restricted_properties` and `go test ./internal/utils` pass.

internal/rules/no_restricted_properties/no_restricted_properties.go
swwind: Fixed in fb66eef0.

A `SpreadElement` is now treated as a transparent pattern ancestor only when it directly contains the current target and belongs to an array literal. The ancestry walk also recognizes the final `=` left-hand side and both `for-of` and `for-in` initializers.

Added focused regression cases for `([...{bad}] = xs)`, `for ([...{bad}] of xs)`, and `for ([...{bad}] in xs)`.

`go test ./internal/rules/no_restricted_properties` passes.

internal/rules/no_restricted_properties/no_restricted_properties.go
```diff
@@ -0,0 +1,395 @@
...
+	case ast.KindObjectLiteralExpression:
+		props := pattern.AsObjectLiteralExpression().Properties
+		if props != nil {
+			elements = props.Nodes
+		}
+	}
+
+	for _, el := range elements {
+		propertyName, hasPropertyName := staticPatternPropertyName(el)
+		checkPropertyAccess(ctx, r, pattern, objectName, hasObjectName, propertyName, hasPropertyName)
```
elecmonkey, line 333: [P2] Include TS annotations in object-pattern diagnostic ranges

For typed destructuring such as `const {bar}: T = foo` with `{ property: "bar" }`, ESLint 10.8.1 with `@typescript-eslint/parser` reports the `ObjectPattern` at columns 7–15, including `: T`, while Rslint ends the range at column 12 after `}` because this passes the raw ts-go binding pattern to `ReportNode`. The same mismatch affects typed parameters and longer or multiline annotations, so extend the reported pattern range through its owning variable/parameter type annotation.

swwind: Fixed in c5f96871.

Object binding pattern diagnostics now extend through an authored type annotation when the pattern directly names a variable declaration or parameter. The source-position guard keeps JSDoc-hosted type nodes from extending the ESTree-compatible range.

Added focused coverage for a typed variable destructuring pattern, a parameter with a multiline type annotation, and a JavaScript JSDoc boundary case.

`go test ./internal/rules/no_restricted_properties` passes.

elecmonkey (approved):
good job but why Cloudflare build longer than 5 hrs

PR #1751, opened 2026-08-16
feat: port rule no-object-constructor

swwind:
## Summary

Port the `no-object-constructor` rule from ESLint to rslint.

Disallows calling the `Object` constructor without an argument (`Object()` / `new Object()`), suggesting the object literal `{}` instead.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/no-object-constructor
- Source code: https://github.com/eslint/eslint/blob/v10.8.1/lib/rules/no-object-constructor.js

## Checklist


elecmonkey (dismissed):

internal/rules/no_object_constructor/no_object_constructor.go
```diff
@@ -0,0 +1,137 @@
...
+
+// needsWrappingParens mirrors upstream's needsParentheses: a bare `{}` would
+// be misparsed (as a block, or as the arrow function's own block body) in
+// either of two positions — the start of an ExpressionStatement, or the
+// concise (unparenthesized) body of an arrow function.
+func needsWrappingParens(sourceFile *ast.SourceFile, node *ast.Node) bool {
+	if utils.IsStartOfExpressionStatement(sourceFile, node) {
+		return true
+	}
+	parent := node.Parent
```
elecmonkey: **Parenthesize replacements that start compound arrow bodies**

When a zero-argument `Object()` begins—but is not itself—the concise arrow body, e.g. `const f = () => Object().x`, `node.Parent` is the property access. This returns false and suggests `const f = () => {}.x`, which is invalid syntax. ESLint 10.8.1 inspects the preceding token and suggests `const f = () => ({}).x`; for the explicitly parenthesized `const f = () => (Object().x)`, it still suggests the bare `{}`. Please recognize the call through intervening expression ancestors while stopping at explicit-parentheses boundaries.

internal/rules/no_object_constructor/no_object_constructor.go
```diff
@@ -0,0 +1,139 @@
...
+			}
+			if callee.AsIdentifier().Text != "Object" {
+				return
+			}
+			if args != nil && len(args.Nodes) > 0 {
+				return
+			}
+
+			// A local declaration shadows the global constructor.
+			if utils.IsShadowed(callee, "Object") {
```
elecmonkey, line 118: Respect type-only Object bindings in TypeScript scopes

When TypeScript declares `Object` in a type scope, such as `type Object = {}; Object()` or `function f<Object>() { Object() }`, `utils.IsShadowed` ignores the type alias, interface, or type parameter, so this falls through and reports the call. Upstream's scope variable has an identifier in these cases and ESLint 10.8.1 stays silent, making these false positives; resolve the actual scope variable or include TypeScript type bindings in this check.

elecmonkey, line 118: Check declarations inside namespace module blocks

When the call occurs in a TypeScript namespace, for example `namespace N { const Object = f; Object() }`, `utils.IsShadowed` never examines the enclosing `ModuleBlock`, so the local declaration is missed and the rule suggests replacing a call to the user's function. Upstream resolves the namespace-local variable and reports nothing; module-block declarations, including local import-equals declarations, need to participate in shadow detection.

swwind: Fixed in 4b811da5.

Confirmed against ESLint 10.8.1 + @typescript-eslint/parser: `type Object = {}; Object()`, `interface Object {} Object()`, `function f<Object>() { Object() }`, and `class C<Object> { m() { Object() } }` all stay silent upstream, while rslint reported all four.

`utils.IsShadowed` only models value-space declarations, so the rule now layers two more checks after it, mirroring how `no-label-var` already resolves the same scope-manager/TypeScript divergence:

- `ctx.Refs.IsNameDefinedInFileWithMeaning(callee, "Object", Value|Type|Namespace|Alias)` — resolves the name in every declaration space at the call site, covering type aliases, interfaces, and type parameters.
- `utils.HasEnclosingTypeParameter` — scope-manager keeps class type parameters in the lexical scope chain of static members, where TypeScript's resolver deliberately hides them, so `class C<Object> { static m() { Object() } }` needs this on top of the resolver lookup. The helper was local to `no-label-var`; it is now shared in `internal/utils`.

Added as valid cases in the extras test, plus `declare function f<Object>(): void; Object()` as an invalid case so the type-parameter check can't over-reach past its own declaration.

swwind: Fixed in 4b811da5, by the same `ctx.Refs.IsNameDefinedInFileWithMeaning(callee, "Object", Value|Type|Namespace|Alias)` lookup — the binder's scope walk enters the `ModuleBlock` that `utils.IsShadowed` never examined.

Verified against ESLint 10.8.1: `namespace N { const Object = f; Object() }`, `namespace N { function Object() {} Object() }`, `namespace N { namespace Object {} Object() }`, `namespace N { import Object = require("x"); Object() }`, and `namespace N.M { const Object = f; Object() }` are all silent upstream and are now valid cases in the extras test.

Two invalid cases lock the boundary in the other direction: `namespace N { Object(); }` still reports (with the `({})` suggestion, since the call starts an ExpressionStatement), and `namespace N { const Object = f; } Object()` reports because the namespace-local declaration doesn't reach the outer scope.

internal/rules/no_object_constructor/no_object_constructor.go
```diff
@@ -0,0 +1,156 @@
...
```
elecmonkey, line 118: [P2] Honor body bindings when checking parameter defaults

When the call is in a parameter initializer and the function body declares `Object`, as in `function f(a = Object()) { var Object; }`, ESLint 10.8.1 stays silent because its scope variable has a declaration identifier. Both `IsShadowed` and `IsNameDefinedInFileWithMeaning` deliberately step past body declarations from parameter initializers, so rslint instead reports the call and suggests `{}`; the same mismatch occurs with `let`, `const`, function/class, and type-space body declarations.

internal/rules/no_object_constructor/no_object_constructor.go
```diff
@@ -0,0 +1,156 @@
...
+			// members, where TypeScript's resolver hides them.
+			if utils.HasEnclosingTypeParameter(callee, "Object") {
+				return
+			}
+			// scope-manager also creates an `Object` variable for TypeScript
+			// type-space declarations — type aliases, interfaces, type
+			// parameters — and for declarations inside a namespace body, none
+			// of which utils.IsShadowed models. Resolving the name in every
+			// declaration space at the call site covers them.
+			if ctx.Refs != nil && ctx.Refs.IsNameDefinedInFileWithMeaning(
```
elecmonkey, line 135: [P2] Exclude namespace re-exports from shadow detection

For a call after a namespace re-export, such as `namespace N { export { Object } from 'x'; Object() }`, ESLint 10.8.1 reports the global call because the re-export creates no local scope variable. The all-space binder lookup resolves the exported alias and returns early, so rslint misses the diagnostic; `export type { Object }` and `export * as Object` inside the namespace are affected similarly.

internal/rules/no_object_constructor/no_object_constructor.go
swwind: Fixed in f30738ea and 980f0bc4.

Two separate gaps. The first is that this rule never called `utils.IsShadowedFromParameterInitializer` — the helper landed in `internal/utils` on #1753 but was only wired into `no-array-constructor`, so every parameter-default form fell through here. It is now called alongside `HasEnclosingTypeParameter`.

The second is the one your "and type-space body declarations" note points at: the helper scanned the body with `HasShadowingDeclaration`, which only recognizes `var`/`let`/`const`, function, class, enum, and namespace declarations. It now uses `hasFunctionScopeDeclaration`, which covers everything scope-manager puts in the function scope — `HasLocalDeclarationInStatements` for the value space, plus type aliases and interfaces.

Verified against ESLint 10.8.1 and locked in as valid cases: `var`, `let`, `const`, `function`, `class`, `enum`, `namespace`, `type Object = {}`, `interface Object {}`, and `import Object = require("x")` in the body of `function f(a = new Object()) { … }`. The nested-block boundary still reports, since a `let` in a nested block stays in that block's scope.

internal/rules/no_object_constructor/no_object_constructor.go
swwind: Fixed in 980f0bc4, in `RefStore` rather than in the rule — the binder scope walk is what returns the alias, so the gap belongs there.

What the walk actually resolves, for the four forms:

| code | declarations of the resolved symbol |
| --- | --- |
| `namespace N { export { Object } from "x"; Object() }` | `ExportSpecifier` |
| `namespace N { export type { Object } from "x"; Object() }` | `ExportSpecifier` |
| `namespace N { export * as Object from "x"; Object() }` | `NamespaceExport` |
| `namespace N { export { Object }; Object() }` | `ExportSpecifier` |
| `namespace N { export const Object = f; Object() }` | `VariableDeclaration` |

A top-level `export { Object } from "x"; Object()` already resolves to nothing, so only the namespace forms were affected.

`resolveName` now skips a symbol whose declarations are all `ExportSpecifier` or `NamespaceExport` and resumes from `outerExportScopeLocation`, mirroring the existing `jsdocOnlyScope` and body-only-parameter-barrier skips: a namespace re-export names what the namespace exports without declaring it inside, so scope-manager creates no variable for it.

Boundary cases in the other direction stay silent, because the symbol still carries a real declaration: `namespace N { const Object = f; export { Object }; Object(); }` and `namespace N { import Object = require("x"); export { Object }; Object(); }`.

Since `resolveName` is a core path, I swept the RefStore-heavy suites for regressions — `internal/rule`, `internal/utils`, `no-undef`, `no-use-before-define`, `no-label-var`, `no-redeclare`, `no-restricted-globals`, `prefer-object-spread`, `no-obj-calls`, `@typescript-eslint/no-shadow`, and the whole `import` plugin — all green.

internal/utils/shadowing.go
```diff
@@ -184,15 +184,134 @@ func isDirectParameterOf(fn *ast.Node, child *ast.Node) bool {
...
+}
+
+// escapesThroughParameterDecorator reports whether a reference that reached fn
+// through prevChild is out of fn's own scope because it sits in a function or
+// class nested inside one of fn's parameter decorators. scope-manager attaches
+// such a scope to the enclosing class rather than to the decorated function, so
+// none of fn's bindings — parameters, type parameters, body declarations —
+// reach the reference. A reference sitting directly in the decorator, with no
+// scope in between, still acquires fn's scope.
+func escapesThroughParameterDecorator(fn *ast.Node, prevChild *ast.Node, inParameterDecorator bool, crossedScope bool) bool {
```
swwind: Fixed in f30738ea.

Confirmed against ESLint 10.8.1 + @typescript-eslint/parser, and the scope dump explains the split. `getScope()` is AST-based, so a reference sitting *directly* in a parameter decorator does acquire the decorated function's scope — `class C { m(@dec(Array()) x: number) { var Array; } }` is silent upstream, which is what the existing `isDirectParameterOf` guard was getting right. But scope-manager parents a scope *created inside* a parameter decorator to the enclosing class, not to the decorated function, so once an arrow, function expression, or class body intervenes the whole function drops out of the chain — its body bindings, its parameters, and its type parameters alike.

New `escapesThroughParameterDecorator` helper, applied at all three sites that walk into a function through its parameter list: `IsShadowedFromParameterInitializer` (body declarations), `IsShadowed` (parameters and function name), and `HasEnclosingTypeParameter` (method type parameters).

Locked in as invalid cases, all silent before the fix: `@dec(() => Array())`, `@dec(function () { … })`, and `@dec(class { p = Array(); })` with `var Array` in the body; `m(@dec(() => Array()) x, Array: any)`; and `m<Array>(@dec(() => Array()) x)`. Valid cases hold the boundary in both directions: the direct-decorator forms stay silent, and the enclosing class is still in the chain, so `class Array { … }` and `class C<Array> { … }` shadow.

internal/rules/no_object_constructor/no_object_constructor.go
```diff
@@ -0,0 +1,160 @@
...
+				"Object",
+				ast.SymbolFlagsValue|ast.SymbolFlagsType|ast.SymbolFlagsNamespace|ast.SymbolFlagsAlias,
+			) {
+				return
+			}
+			// A config `/* global Object: off */` / `languageOptions.globals`
+			// entry un-declares the builtin, so `Object` no longer resolves to
+			// a known global — ESLint's `getVariableByName` would return
+			// undefined and the rule stays silent.
+			if !ctx.Globals.Access("Object").IsDeclared() {
```
elecmonkey, line 146: [P2] Keep inline `global ...:off` from suppressing reports

When the source contains `/* global Object:off */ Object()`, this guard returns because `Globals.Access` applies the inline override. ESLint 10.8.1 still reports `preferLiteral`: an inline `off` directive does not remove the pre-existing ECMAScript `Object` binding, unlike `languageOptions.globals: { Object: "off" }`. Consequently, these global constructor calls are missed; this check needs to distinguish the config override from the inline override.

swwind: No change here — I can't reproduce the config-vs-inline split. Both forms behave identically in ESLint 10.8.1; what changes the answer is the parser.

With espree, `addDeclaredGlobals` computes `commentValue || configValue` and skips the name entirely when that is `"off"`, so either form removes the variable, `getVariableByName` returns undefined, and the rule stays silent.

With @typescript-eslint/parser, `Object` is a lib global that scope-manager creates itself, and neither form can remove it, so both still report.

| setup | espree | @typescript-eslint/parser |
| --- | --- | --- |
| `Object()` | reports | reports |
| `globals: { Object: "off" }` | silent (`Object` is not defined) | reports |
| `/* global Object:off */` | silent (`Object` is not defined) | reports |
| `parserOptions: { lib: [] }` + `globals: { Object: "off" }` | n/a | silent |

So there is no inline-only behavior to model. `ctx.Globals.Access("Object").IsDeclared()` matches the espree column exactly, and it is the same guard the other ported global-constructor rules use — `no-new-wrappers`, `no-new-func`, `no-implied-eval`, `symbol-description`, `no-obj-calls` and about twenty more. Splitting `ConfiguredAccess` from `Access` here would match neither parser and would leave this rule alone in disagreeing with the rest of the port.

internal/utils/shadowing.go
```diff
@@ -184,15 +184,259 @@ func isDirectParameterOf(fn *ast.Node, child *ast.Node) bool {
...
+	prevChild := node
+	inParameterDecorator := false
+	crossedScope := false
+	for current := node.Parent; current != nil; current = current.Parent {
+		if current.Kind == ast.KindParameter {
+			inParameterDecorator = prevChild.Kind == ast.KindDecorator
+		}
+		isFunctionLike := ast.IsFunctionLikeDeclaration(current)
+		if (isFunctionLike || ast.IsClassLike(current)) &&
+			!escapesFunctionScope(current, prevChild, inParameterDecorator, crossedScope) {
```
elecmonkey: [P2] Stop type parameters at nested class-decorator scopes

When a call is inside a scope created within a class decorator, such as `@dec(() => Object()) class C<Object> {}`, `@typescript-eslint/scope-manager` places the arrow scope outside the class scope, so ESLint reports the global `Object()` call. `HasEnclosingTypeParameter` instead walks through the arrow and finds the class type parameter, causing Rslint to emit no diagnostic; nested class-decorator scopes need to escape the decorated class similarly to nested parameter-decorator scopes.

internal/rules/no_object_constructor/no_object_constructor.go
```diff
@@ -0,0 +1,160 @@
...
+				return
+			}
+			if args != nil && len(args.Nodes) > 0 {
+				return
+			}
+
+			// A local declaration shadows the global constructor.
+			if utils.IsShadowed(callee, "Object") {
+				return
+			}
```
elecmonkey, line 120: [P2] Stop class-expression names at nested decorator scopes

For a named class expression whose decorator creates an inner scope, such as `const C = @dec(() => Object()) class Object {}`, scope-manager places the arrow outside the class-expression scope and ESLint reports the global `Object()` call. `IsShadowed` still sees the enclosing class name, so this guard suppresses the diagnostic; class-expression names must likewise stop applying after crossing a nested class-decorator scope.

internal/utils/shadowing.go
swwind: Fixed in a66e63f3.

Confirmed against ESLint 10.8.1 + @typescript-eslint/parser, and the scope dump shows the same split the parameter-decorator case has. `getScope()` is AST-based, so a reference sitting *directly* in a class decorator still acquires the class scope — `@dec(new Object()) class C<Object> { }` is silent upstream. But scope-manager parents a scope created *inside* a class decorator to the scope holding the class, so the arrow's `upper` is the module scope and the class drops out of the chain along with its type parameters.

New `escapesClassScope` helper in `internal/utils`, applied where `HasEnclosingTypeParameter` walks into a class through its own decorators.

| code | upstream |
| --- | --- |
| `@dec(() => new Object()) class C<Object> { }` | reports |
| `@dec(function () { return new Object(); }) class C<Object> { }` | reports |
| `@dec(class { p = new Object(); }) class C<Object> { }` | reports |
| `@dec(new Object()) class C<Object> { }` | silent |
| `class C<Object> { @dec(() => new Object()) m() { } }` | silent |

All five are test cases now; the first three were silent before the fix. The last two hold the boundary in the other direction: only the class's own decorators escape the class scope, and a member's decorator is still evaluated inside it.

internal/rules/no_object_constructor/no_object_constructor.go
swwind: Fixed in a66e63f3.

Two layers, since a class expression's own name leaks through both.

`utils.IsShadowed` now stops at the same `escapesClassScope` boundary as the type parameters, but only for a class *expression*: a class declaration's name is declared in the scope holding the class, so `@dec(() => new Object()) class Object { }` stays silent upstream and that name keeps shadowing wherever the decorator resolves.

The second is `RefStore.resolveName`. TypeScript's resolver keeps a class expression's self-reference visible throughout the class node, decorators included, while a class's decorators are evaluated in the scope holding the class and cannot see the class binding at all — so the walk now steps over that binding and resumes outside the class, mirroring the existing namespace re-export skip.

| code | upstream |
| --- | --- |
| `const C = @dec(() => new Object()) class Object { };` | reports |
| `const C = @dec(class { p = new Object(); }) class Object { };` | reports |
| `const C = @dec(new Object()) class Object { };` | silent |
| `@dec(() => new Object()) class Object { }` | silent |

Since `resolveName` is a core path, I swept the RefStore-heavy suites again — `internal/rule`, `internal/utils`, `no-undef`, `no-use-before-define`, `no-label-var`, `no-redeclare`, `no-restricted-globals`, `prefer-object-spread`, `no-obj-calls`, `@typescript-eslint/no-shadow`, and the whole `import` plugin — all green.

fansenze (approved):

PR #1752, opened 2026-08-16
feat: port rule no-div-regex

swwind:
## Summary

Port the `no-div-regex` rule from ESLint to rslint.

Disallows equal signs explicitly at the beginning of regular expression literals, since `/=` at the start of a regex can be visually confused with a division-assignment operator. The rule is autofixable, wrapping the leading `=` in a character class (e.g. `/=foo/` → `/[=]foo/`).

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/no-div-regex
- Source code: https://github.com/eslint/eslint/blob/v10.8.1/lib/rules/no-div-regex.js

## Checklist


elecmonkey:
LGTM

elecmonkey (approved):

fansenze (approved):

PR #1753, opened 2026-08-16
feat: port rule no-array-constructor

swwind:
## Summary

Port the `no-array-constructor` rule from ESLint core to rslint.

Disallows using `Array()` / `new Array()` in favor of array literal notation, except when a single non-spread argument is given (the sparse-array-length case, which can't be distinguished from a regular value without type information). Supports TypeScript type-argument and optional-chain forms.

Stacked on #1751 (`no-object-constructor`), which this rule reuses `utils.NeedsPrecedingSemicolon`'s full ASI-safety logic from — this PR's diff is just the `no-array-constructor` rule itself.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/no-array-constructor
- Source code: https://github.com/eslint/eslint/blob/v10.8.1/lib/rules/no-array-constructor.js

## Checklist

### DIFF
--- a/internal/rules/no_array_constructor/no_array_constructor.go
+++ b/internal/rules/no_array_constructor/no_array_constructor.go
@@ -82,6 +82,7 @@ var NoArrayConstructorRule = rule.Rule{
 	Name:   "no-array-constructor",
 	Schema: rule.EmptyArraySchema,
 	Run: func(ctx rule.RuleContext, options []any) rule.RuleListeners {
+		var shadowCache utils.ScopeManagerShadowCache
 		check := func(node *ast.Node) {
 			callee, args, typeArgs := calleeArgsAndTypeArgs(node)
 			if callee == nil {
@@ -133,16 +134,19 @@ var NoArrayConstructorRule = rule.Rule{
 			// keeps a function's parameter initializers in the same scope as its
 			// body declarations, where both TypeScript's resolver and IsShadowed
 			// follow runtime lexical semantics instead.
-			if utils.HasEnclosingParameter(callee, "Array") ||
-				utils.HasEnclosingTypeParameter(callee, "Array") ||
+			if shadowCache.HasEnclosingParameter(callee, "Array") ||
+				shadowCache.HasEnclosingTypeParameter(callee, "Array") ||
 				utils.HasEnclosingClassExpressionName(callee, "Array") ||
-				utils.IsShadowedFromParameterInitializer(callee, "Array") {
+				shadowCache.IsShadowedFromParameterInitializer(callee, "Array") {
 				return
 			}
 			// A config `/* global Array: off */` / `languageOptions.globals`
 			// entry un-declares Espree's builtin. typescript-eslint/parser also
 			// installs its library `Array` variable, which remains visible after
-			// the override, so TypeScript-flavoured files still report.
+			// the override, so TypeScript-flavoured files still report. rslint does
+			// not currently project parserOptions.lib into native rule contexts;
+			// the documented divergence for an explicitly empty lib follows from
+			// preserving that existing API boundary.
 			if ast.IsInJSFile(callee) && !ctx.Globals.Access("Array").IsDeclared() {
 				return
 			}
--- a/internal/rules/no_array_constructor/no_array_constructor.go
+++ b/internal/rules/no_array_constructor/no_array_constructor.go
@@ -101,7 +101,7 @@ var NoArrayConstructorRule = rule.Rule{
 			}
 			// `Array<Foo>()` — explicit type arguments mean this isn't the
 			// bare-constructor call the rule targets.
-			if typeArgs != nil && len(typeArgs.Nodes) > 0 {
+			if typeArgs != nil {
 				return
 			}
 			// A single non-spread argument (`Array(9)`, `Array(x)`) may be
@@ -112,8 +112,19 @@ var NoArrayConstructorRule = rule.Rule{
 				return
 			}
 
-			// A local declaration shadows the global constructor.
-			if utils.IsShadowed(callee, "Array") {
+			// A local declaration shadows the global constructor. The binder
+			// lookup resolves the declaration once without rescanning every
+			// enclosing body for each matching call. Keep the AST fallback for
+			// manually assembled rule contexts that do not provide a RefStore.
+			if ctx.Refs == nil {
+				if utils.IsShadowed(callee, "Array") {
+					return
+				}
+			} else if ctx.Refs.IsNameDefinedInFileWithMeaning(
+				callee,
+				"Array",
+				ast.SymbolFlagsValue|ast.SymbolFlagsType|ast.SymbolFlagsNamespace|ast.SymbolFlagsAlias,
+			) {
 				return
 			}
 			// scope-manager keeps class type parameters visible inside static
@@ -122,26 +133,15 @@ var NoArrayConstructorRule = rule.Rule{
 			// declarations, where both TypeScript's resolver and IsShadowed
 			// follow runtime lexical semantics instead.
 			if utils.HasEnclosingTypeParameter(callee, "Array") ||
+				utils.HasEnclosingClassExpressionName(callee, "Array") ||
 				utils.IsShadowedFromParameterInitializer(callee, "Array") {
 				return
 			}
-			// scope-manager also creates an `Array` variable for TypeScript
-			// type-space declarations — type aliases, interfaces, type
-			// parameters, type-only imports — which utils.IsShadowed doesn't
-			// model. Resolving the name in every declaration space at the call
-			// site covers them.
-			if ctx.Refs != nil && ctx.Refs.IsNameDefinedInFileWithMeaning(
-				callee,
-				"Array",
-				ast.SymbolFlagsValue|ast.SymbolFlagsType|ast.SymbolFlagsNamespace|ast.SymbolFlagsAlias,
-			) {
-				return
-			}
 			// A config `/* global Array: off */` / `languageOptions.globals`
-			// entry un-declares the builtin, so `Array` no longer resolves to
-			// a known global — ESLint's `getVariableByName` would return
-			// undefined and the rule stays silent.
-			if !ctx.Globals.Access("Array").IsDeclared() {
+			// entry un-declares Espree's builtin. typescript-eslint/parser also
+			// installs its library `Array` variable, which remains visible after
+			// the override, so TypeScript-flavoured files still report.
+			if ast.IsInJSFile(callee) && !ctx.Globals.Access("Array").IsDeclared() {
 				return
 			}
 
--- a/internal/rules/no_array_constructor/no_array_constructor.go
+++ b/internal/rules/no_array_constructor/no_array_constructor.go
@@ -1,0 +1,216 @@
+package no_array_constructor
+
+import (
+	"github.com/microsoft/typescript-go/shim/ast"
+	"github.com/microsoft/typescript-go/shim/scanner"
+	"github.com/web-infra-dev/rslint/internal/rule"
+	"github.com/web-infra-dev/rslint/internal/utils"
+)
+
+func preferLiteralMessage() rule.RuleMessage {
+	return rule.RuleMessage{
+		Id:          "preferLiteral",
+		Description: "The array literal notation [] is preferable.",
+	}
+}
+
+func useLiteralMessage() rule.RuleMessage {
+	return rule.RuleMessage{
+		Id:          "useLiteral",
+		Description: "Replace with an array literal.",
+	}
+}
+
+func useLiteralAfterSemicolonMessage() rule.RuleMessage {
+	return rule.RuleMessage{
+		Id:          "useLiteralAfterSemicolon",
+		Description: "Replace with an array literal, add preceding semicolon.",
+	}
+}
+
+// calleeArgsAndTypeArgs extracts the callee, argument list, and type-argument
+// list shared by CallExpression and NewExpression, the two forms this rule
+// inspects.
+func calleeArgsAndTypeArgs(node *ast.Node) (callee *ast.Node, args *ast.NodeList, typeArgs *ast.NodeList) {
+	switch node.Kind {
+	case ast.KindCallExpression:
+		callExpr := node.AsCallExpression()
+		return callExpr.Expression, callExpr.Arguments, callExpr.TypeArguments
+	case ast.KindNewExpression:
+		newExpr := node.AsNewExpression()
+		return newExpr.Expression, newExpr.Arguments, newExpr.TypeArguments
+	}
+	return nil, nil, nil
+}
+
+// findOpenParen scans forward from calleeEnd (the position right after the
+// possibly-parenthesized callee) for the call's own opening paren, stopping
+// at nodeEnd. Only `?.` punctuation and trivia can appear before it, since
+// type arguments were already excluded by the caller. found is false when the
+// node has no parentheses at all, e.g. a bare `new Array`.
+func findOpenParen(sourceFile *ast.SourceFile, calleeEnd int, nodeEnd int) (pos int, found bool) {
+	text := sourceFile.Text()
+	p := calleeEnd
+	for p < nodeEnd {
+		p = scanner.SkipTrivia(text, p)
+		if p >= nodeEnd {
+			break
+		}
+		if text[p] == '(' {
+			return p, true
+		}
+		p++
+	}
+	return 0, false
+}
+
+func countNonSpread(args *ast.NodeList) int {
+	if args == nil {
+		return 0
+	}
+	count := 0
+	for _, arg := range args.Nodes {
+		if arg.Kind != ast.KindSpreadElement {
+			count++
+		}
+	}
+	return count
+}
+
+// https://eslint.org/docs/latest/rules/no-array-constructor
+var NoArrayConstructorRule = rule.Rule{
+	Name:   "no-array-constructor",
+	Schema: rule.EmptyArraySchema,
+	Run: func(ctx rule.RuleContext, options []any) rule.RuleListeners {
+		check := func(node *ast.Node) {
+			callee, args, typeArgs := calleeArgsAndTypeArgs(node)
+			if callee == nil {
+				return
+			}
+			// ESTree flattens grouping parentheses around the callee, so
+			// `(Array)()` still has an Identifier callee; TS-only wrappers
+			// (non-null assertion, `as`/`satisfies`) are deliberately left
+			// alone since ESTree never collapses those back to a bare
+			// Identifier either.
+			callee = ast.SkipParentheses(callee)
+			if callee == nil || callee.Kind != ast.KindIdentifier {
+				return
+			}
+			if callee.AsIdentifier().Text != "Array" {
+				return
+			}
+			// `Array<Foo>()` — explicit type arguments mean this isn't the
+			// bare-constructor call the rule targets.
+			if typeArgs != nil && len(typeArgs.Nodes) > 0 {
+				return
+			}
+			// A single non-spread argument (`Array(9)`, `Array(x)`) may be
+			// intentionally creating a sparse array of that length — leave
+			// it alone, since the argument's runtime type can't be known
+			// statically.
+			if args != nil && len(args.Nodes) == 1 && args.Nodes[0].Kind != ast.KindSpreadElement {
+				return
+			}
+
+			// A local declaration shadows the global constructor.
+			if utils.IsShadowed(callee, "Array") {
+				return
+			}
+			// scope-manager keeps class type parameters visible inside static
+			// members, where TypeScript's resolver hides them, and keeps a
+			// function's parameter initializers in the same scope as its body
+			// declarations, where both TypeScript's resolver and IsShadowed
+			// follow runtime lexical semantics instead.
+			if utils.HasEnclosingTypeParameter(callee, "Array") ||
+				utils.IsShadowedFromParameterInitializer(callee, "Array") {
+				return
+			}
+			// scope-manager also creates an `Array` variable for TypeScript
+			// type-space declarations — type aliases, interfaces, type
+			// parameters, type-only imports — which utils.IsShadowed doesn't
+			// model. Resolving the name in every declaration space at the call
+			// site covers them.
+			if ctx.Refs != nil && ctx.Refs.IsNameDefinedInFileWithMeaning(
+				callee,
+				"Array",
+				ast.SymbolFlagsValue|ast.SymbolFlagsType|ast.SymbolFlagsNamespace|ast.SymbolFlagsAlias,
+			) {
+				return
+			}
+			// A config `/* global Array: off */` / `languageOptions.globals`
+			// entry un-declares the builtin, so `Array` no longer resolves to
+			// a known global — ESLint's `getVariableByName` would return
+			// undefined and the rule stays silent.
+			if !ctx.Globals.Access("Array").IsDeclared() {
+				return
+			}
+
+			nodeRange := utils.TrimNodeTextRange(ctx.SourceFile, node)
+			openParen, hasParens := findOpenParen(ctx.SourceFile, callee.End(), nodeRange.End())
+			argsText := ""
+			commentBound := nodeRange.End()
+			if hasParens {
+				argsText = ctx.SourceFile.Text()[openParen+1 : nodeRange.End()-1]
+				commentBound = openParen
+			}
+			fixText := "[" + argsText + "]"
+
+			// NOTE: Unlike ESLint, utils.NeedsPrecedingSemicolon doesn't model
+			// TypeScript-only node kinds (type positions, ambient/overload
+			// function declarations, import-equals declarations), so it falls
+			// back to the conservative "needs a semicolon" answer there —
+			// e.g. after `type T = Foo` or `import Foo = Bar`. The extra `;`
+			// never changes behavior, only adds a redundant character; see
+			// the rule doc's "Differences from ESLint" section.
+			addSemicolon := utils.IsStartOfExpressionStatement(ctx.SourceFile, node) &&
+				utils.NeedsPrecedingSemicolon(ctx.SourceFile, node)
+			if addSemicolon {
+				fixText = ";" + fixText
+			}
+
+			// Replacing the call with an array literal directly (rather than
+			// as an opt-in suggestion) is only safe when doing so can't
+			// change runtime behavior or silently drop source text:
+			//   - `Array?.(...)` short-circuits to `undefined` when `Array`
+			//     is nullish; `[...]` always evaluates.
+			//   - A single spread plus at most one more argument
+			//     (`Array(5, ...args)`) is equivalent to the sparse-array
+			//     single-argument form when the spread happens to be empty
+			//     at runtime — an outcome an autofix must not risk.
+			//   - A comment between the node's start and the call's own
+			//     opening paren (`Array/*a*/()`) would be silently discarded
+			//     by the fix. Comments inside the argument list itself don't
+			//     count: the fix copies that text through verbatim.
+			shouldSuggest := ast.IsOptionalChainRoot(node) ||
+				(args != nil && len(args.Nodes) > 0 && countNonSpread(args) < 2) ||
+				utils.HasCommentInSpan(ctx.Comments.All(), nodeRange.Pos(), commentBound)
+
+			buildFixes := func() []rule.RuleFix {
+				if shouldSuggest {
+					return nil
+				}
+				return []rule.RuleFix{rule.RuleFixReplaceRange(nodeRange, fixText)}
+			}
+			buildSuggestions := func() []rule.RuleSuggestion {
+				if !shouldSuggest {
+					return nil
+				}
+				msg := useLiteralMessage()
+				if addSemicolon {
+					msg = useLiteralAfterSemicolonMessage()
+				}
+				return []rule.RuleSuggestion{{
+					Message:  msg,
+					FixesArr: []rule.RuleFix{rule.RuleFixReplaceRange(nodeRange, fixText)},
+				}}
+			}
+
+			ctx.ReportNodeWithDeferredFixesAndSuggestions(node, preferLiteralMessage(), buildFixes, buildSuggestions)
+		}
+
+		return rule.RuleListeners{
+			ast.KindCallExpression: check,
+			ast.KindNewExpression:  check,
+		}
+	},
+}
--- a/internal/rules/no_array_constructor/no_array_constructor.go
+++ b/internal/rules/no_array_constructor/no_array_constructor.go
@@ -127,12 +127,14 @@ var NoArrayConstructorRule = rule.Rule{
 			) {
 				return
 			}
-			// scope-manager keeps class type parameters visible inside static
-			// members, where TypeScript's resolver hides them, and keeps a
-			// function's parameter initializers in the same scope as its body
-			// declarations, where both TypeScript's resolver and IsShadowed
+			// scope-manager keeps function parameters visible directly inside
+			// parameter decorators and class type parameters visible inside
+			// static members, where TypeScript's resolver hides them. It also
+			// keeps a function's parameter initializers in the same scope as its
+			// body declarations, where both TypeScript's resolver and IsShadowed
 			// follow runtime lexical semantics instead.
-			if utils.HasEnclosingTypeParameter(callee, "Array") ||
+			if utils.HasEnclosingParameter(callee, "Array") ||
+				utils.HasEnclosingTypeParameter(callee, "Array") ||
 				utils.HasEnclosingClassExpressionName(callee, "Array") ||
 				utils.IsShadowedFromParameterInitializer(callee, "Array") {
 				return
--- a/internal/utils/shadowing.go
+++ b/internal/utils/shadowing.go
@@ -219,6 +219,27 @@ func HasEnclosingTypeParameter(node *ast.Node, name string) bool {
 	return false
 }
 
+// HasEnclosingClassExpressionName reports whether scope-manager exposes an
+// enclosing class expression's own name at node. A reference sitting directly
+// in the class's decorator acquires the class scope, while a scope created
+// inside that decorator is parented outside the class and cannot see the name.
+func HasEnclosingClassExpressionName(node *ast.Node, name string) bool {
+	prevChild := node
+	crossedScope := false
+	for current := node.Parent; current != nil; current = current.Parent {
+		if current.Kind == ast.KindClassExpression && !escapesClassScope(prevChild, crossedScope) {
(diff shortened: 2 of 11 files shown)


internal/rules/no_array_constructor/no_array_constructor.go
```diff
@@ -0,0 +1,211 @@
...
+			// A single non-spread argument (`Array(9)`, `Array(x)`) may be
+			// intentionally creating a sparse array of that length — leave
+			// it alone, since the argument's runtime type can't be known
+			// statically.
+			if args != nil && len(args.Nodes) == 1 && args.Nodes[0].Kind != ast.KindSpreadElement {
+				return
+			}
+
+			// A local declaration shadows the global constructor.
+			if utils.IsShadowed(callee, "Array") {
```
elecmonkey, line 116: **[P1] Handle TypeScript namespace-local `Array` bindings**

When a call is inside a TypeScript namespace/module block that declares its own `Array`, such as `namespace N { const Array = () => 123; Array(); }`, `utils.IsShadowed` never examines `KindModuleBlock`, so this branch treats the reference as the global constructor. ESLint 10.8.1 reports nothing, while rslint reports it and `--fix` rewrites the local function call to `[]`, changing runtime behavior; resolve the actual binding or include module-block scopes before reporting.

internal/rules/no_array_constructor/no_array_constructor.go
```diff
@@ -0,0 +1,211 @@
...
+}
+
+// hasLeadingComment reports whether a comment appears in [from, to) — the
+// span between the node's start and the call's own opening paren (or the
+// node's end, when it has no parentheses at all). Comments inside the
+// argument list itself don't count: they survive an autofix intact, but a
+// comment in this span would be silently dropped, so its presence forces a
+// suggestion instead.
+func hasLeadingComment(ctx rule.RuleContext, from int, to int) bool {
+	for _, c := range ctx.Comments.All() {
```
elecmonkey: **[P2] Use binary search for per-report comment checks**

In files containing many `Array` constructor calls and comments, every report starts scanning the comment list from index zero until reaching the current call, making interleaved calls/comments take O(N^2) time. The existing `utils.HasCommentInSpan(ctx.Comments.All(), from, to)` performs this lookup with binary search over the same source-ordered list and avoids this scaling problem.

internal/rules/no_array_constructor/no_array_constructor.go
```diff
@@ -0,0 +1,195 @@
...
+			// A single non-spread argument (`Array(9)`, `Array(x)`) may be
+			// intentionally creating a sparse array of that length — leave
+			// it alone, since the argument's runtime type can't be known
+			// statically.
+			if args != nil && len(args.Nodes) == 1 && args.Nodes[0].Kind != ast.KindSpreadElement {
+				return
+			}
+
+			// A local declaration shadows the global constructor.
+			if utils.IsShadowed(callee, "Array") {
```
elecmonkey, line 116: [P2] Respect TypeScript-only `Array` bindings

In TypeScript files where `Array` is declared only as a type, such as `interface Array {}; Array()` (also type aliases, type parameters, and type-only imports), ESLint 10.8.1 with `@typescript-eslint/parser` emits no diagnostic because that scope variable has a declaration. `utils.IsShadowed` ignores these bindings, so rslint reports and autofixes the call, creating an undocumented parity mismatch; resolve the same scope binding before treating `Array` as the predefined global.

elecmonkey, line 116: [P2] Match upstream shadowing for parameter initializers

When the call occurs in a default-parameter initializer and the function body declares `Array`, for example `function f(x = Array()) { var Array; }`, ESLint 10.8.1 emits no diagnostic, while `IsShadowed` deliberately excludes body declarations from parameter initializers and rslint reports and autofixes the call. Although that helper models runtime lexical semantics, this port documents only an ASI difference, so this rule needs an upstream-compatible scope lookup for this case.

elecmonkey (commented):
The implementation diverges from ESLint 10.8.1 for two reproducible shadowing classes and emits extra diagnostics and autofixes. These differences are not included in the PR's documented intentional divergence.

internal/rules/no_array_constructor/no_array_constructor.go
swwind: Fixed in c73271ac.

Confirmed against ESLint 10.8.1 + @typescript-eslint/parser: `interface Array {} Array()`, `type Array = {}; Array()`, `import type { Array } from "x"; Array()`, `function f<Array>() { Array() }`, and `class C<Array> { m() { Array() } }` are all silent upstream, while rslint reported and autofixed every one.

`utils.IsShadowed` only models value-space declarations, so the rule now layers two more checks after it:

- `ctx.Refs.IsNameDefinedInFileWithMeaning(callee, "Array", Value|Type|Namespace|Alias)` — resolves the name in every declaration space at the call site, covering type aliases, interfaces, type parameters, and type-only imports.
- `utils.HasEnclosingTypeParameter` — scope-manager keeps class type parameters in the lexical scope chain of static members, where TypeScript's resolver deliberately hides them, so `class C<Array> { static m() { Array() } }` needs this on top of the resolver lookup.

All added as valid cases in the extras test, plus `declare function f<Array>(): void; Array()` as an invalid case so the type-parameter check can't over-reach past its own declaration.

swwind: Fixed in c73271ac.

The resolver can't answer this one: TypeScript scopes a parameter initializer the same way `IsShadowed` does, with the parameter environment as a parent of the body's variable environment. scope-manager instead puts both in one function scope, so `getVariableByName` finds the body's binding from a default value. So this is a second, separate helper rather than a change to `IsShadowed` — that helper's `isDirectParameterOf` guard is still load-bearing, since `no-restricted-globals` resolves through `globalScope.through` and *does* report on `function f(x = foo) { var foo; }`.

`utils.IsShadowedFromParameterInitializer` walks out to any function whose parameter list lexically contains the call and checks exactly what scope-manager puts in that function scope: vars hoisted from any depth, plus lexical declarations at the top level of the body.

Verified against ESLint 10.8.1 and locked in as valid cases: `var Array`, `let Array`, `function Array() {}`, `class Array {}`, `if (a) { var Array; }`, the arrow-nested `function f(x = () => Array()) { var Array; }`, and the arrow form `const g = (x = Array()) => { var Array; }`. Two invalid cases lock the boundary: `function f(x = Array()) { { let Array; } }` still reports (a nested-block `let` stays in that block's scope), as does `function f(x = Array()) { }`.

Note this same false positive exists in `no-object-constructor` on #1751 (`function f(x = new Object()) { var Object; }`); since the helper lands here, that one is fixed by this PR rather than the base.

internal/utils/shadowing.go
```diff
@@ -204,11 +204,38 @@ func HasEnclosingTypeParameter(node *ast.Node, name string) bool {
...
+// model scope-manager variables need this on top of IsShadowed.
+//
+// Only the declarations scope-manager puts in that function scope count: vars
+// hoisted from any depth, plus lexical declarations at the top level of the
+// body. A `let` in a nested block stays in that block's scope.
+func IsShadowedFromParameterInitializer(node *ast.Node, name string) bool {
+	prevChild := node
+	for current := node.Parent; current != nil; current = current.Parent {
+		if ast.IsFunctionLikeDeclaration(current) && isDirectParameterOf(current, prevChild) {
+			if body := current.Body(); body != nil &&
```
elecmonkey, line 283: When `Array()` is inside a function created by a parameter decorator, such as `class C { m(@dec(() => Array()) x: number) { var Array; } }`, `isDirectParameterOf(current, prevChild)` treats the entire decorator subtree as a parameter initializer and incorrectly applies the method body's binding. ESLint 10.8.1 with `@typescript-eslint/parser` 8.58.2 reports `preferLiteral` for this call, while rslint emits nothing. Distinguish parameter decorators from actual initializer subtrees, including nested function expressions, before suppressing the diagnostic.

swwind: Fixed in f30738ea. That commit is on #1751 now — the shared `internal/utils/shadowing.go` changes moved there and this PR is just the rule port, so this thread will show as outdated here.

Confirmed against ESLint 10.8.1 + @typescript-eslint/parser, and the scope dump explains the split. `getScope()` is AST-based, so a reference sitting *directly* in a parameter decorator does acquire the decorated function's scope — `class C { m(@dec(Array()) x: number) { var Array; } }` is silent upstream, which is what the existing `isDirectParameterOf` guard was getting right. But scope-manager parents a scope *created inside* a parameter decorator to the enclosing class, not to the decorated function, so once an arrow, function expression, or class body intervenes the whole function drops out of the chain — its body bindings, its parameters, and its type parameters alike.

New `escapesThroughParameterDecorator` helper, applied at all three sites that walk into a function through its parameter list: `IsShadowedFromParameterInitializer` (body declarations), `IsShadowed` (parameters and function name), and `HasEnclosingTypeParameter` (method type parameters).

Locked in as invalid cases, all silent before the fix: `@dec(() => Array())`, `@dec(function () { … })`, and `@dec(class { p = Array(); })` with `var Array` in the body; `m(@dec(() => Array()) x, Array: any)`; and `m<Array>(@dec(() => Array()) x)`. Valid cases hold the boundary in both directions: the direct-decorator forms stay silent, and the enclosing class is still in the chain, so `class Array { … }` and `class C<Array> { … }` shadow.

internal/rules/no_array_constructor/no_array_constructor.go
```diff
@@ -0,0 +1,216 @@
...
+				return
+			}
+			// scope-manager keeps class type parameters visible inside static
+			// members, where TypeScript's resolver hides them, and keeps a
+			// function's parameter initializers in the same scope as its body
+			// declarations, where both TypeScript's resolver and IsShadowed
+			// follow runtime lexical semantics instead.
+			if utils.HasEnclosingTypeParameter(callee, "Array") ||
+				utils.IsShadowedFromParameterInitializer(callee, "Array") {
```
elecmonkey, line 125: [P2] Exclude method scopes from computed names and decorators

When `Array()` is evaluated in a method decorator or computed key, these guards incorrectly apply bindings that belong only to the method's function scope. For both `class C { @dec(Array()) m<Array>() {} }` and `class C { [Array()]() { var Array; } }`, ESLint 10.8.1 with `@typescript-eslint/parser` reports and fixes the call, while rslint emits nothing; treat decorator and computed-name subtrees as outside the method's type-parameter, parameter, and body scopes before suppressing the diagnostic.

elecmonkey, line 125: [P2] Include Annex B functions in parameter-initializer scope

When a default parameter precedes an Annex B function declaration in the body, such as `function f(x = Array()) { if (x) function Array(){} }`, ESLint 10.8.1 treats the body function as a function-scope `Array` binding and stays silent. `IsShadowedFromParameterInitializer` does not discover this declaration, so rslint emits an extra diagnostic and autofix; include the body declarations that scope-manager hoists through these single-statement constructs.

elecmonkey (commented):
The rule still diverges from ESLint 10.8.1 for reproducible method-scope and parameter-initializer cases, producing both missing and extra diagnostics. The existing tests and CI pass, but they do not cover these scope-manager boundaries.

internal/rules/no_array_constructor/no_array_constructor.go
swwind: Fixed in 05e661e9. That commit is on #1751, since the change is in the shared `internal/utils/shadowing.go`; the `no-array-constructor` lock-ins are here in 7dc110b7.

The split is AST-based. ESTree keeps a member's decorators and its computed key on the class or object member, while the type parameters, the parameters, and the body live on the function expression underneath it, so `getScope()` never reaches the method for a reference in either position — it lands in the class scope. A parameter decorator is the opposite case, and the existing `isDirectParameterOf` guard already has it right: in ESTree the parameter and its decorators sit inside the function's `params`, so a reference directly there does acquire the method's scope.

`escapesThroughParameterDecorator` is now `escapesFunctionScope`, which also returns true when the reference reached the function through one of its own decorators or through its name, and it is applied in `IsShadowed` and `HasEnclosingTypeParameter`.

Verified against ESLint 10.8.1 + @typescript-eslint/parser, all reporting upstream and silent in rslint before the fix, now invalid cases with the `[]` autofix:

- `class C { @dec(Array()) m() { var Array; } }`
- `class C { @dec(Array()) m(Array: any) { } }`
- `class C { @dec(Array()) m<Array>() { } }`
- `class C { [Array()]() { var Array; } }`
- `class C { [Array()](Array: any) { } }`
- `class C { [Array()]<Array>() { } }`
- `const o = { [Array()]() { var Array; } };`

The enclosing scope is still in the chain, so these stay silent and are locked in as valid cases: `class C<Array> { @dec(Array()) m() { } }`, `class C<Array> { [Array()]() { } }`, and `class Array { [Array()]() { } }`.

swwind: Fixed in 05e661e9 (on #1751, shared helper), covered here in 7dc110b7.

It turned out not to be Annex B specific. scope-manager defines a function declaration in the innermost scope that exists at its position, and an `if`/`else` branch, a labelled statement, and an unbraced loop body create none — so the declaration lands in the enclosing function, namespace, source file, or switch case block. The new `hasHoistedFunctionDeclaration` walks exactly those scope-less statement positions; it is wired into `hasFunctionScopeDeclaration` (the parameter-initializer path this thread is about) and into `IsShadowed`'s SourceFile, Block, ModuleBlock, and CaseBlock branches, where the same declarations were being missed.

Verified silent against ESLint 10.8.1 and added as valid cases: `function f(x = Array()) { if (a) function Array() {} }`, the `lbl: function Array() {}` form, and `while (a) function Array() {}`.

Two invalid cases hold the boundary, both still reporting upstream: `function f(x = Array()) { { function Array() {} } }`, where the block keeps the declaration to itself, and `function f(x = Array()) { for (let i; ;) function Array() {} }`, where the `let` initializer gives the loop a scope of its own.

internal/rules/no_array_constructor/no_array_constructor.go
```diff
@@ -0,0 +1,216 @@
...
+				return
+			}
+			// scope-manager also creates an `Array` variable for TypeScript
+			// type-space declarations — type aliases, interfaces, type
+			// parameters, type-only imports — which utils.IsShadowed doesn't
+			// model. Resolving the name in every declaration space at the call
+			// site covers them.
+			if ctx.Refs != nil && ctx.Refs.IsNameDefinedInFileWithMeaning(
```
elecmonkey, line 133: **[P2] Do not treat qualified namespace names as local bindings**

For `namespace N.Array { Array(); }`, this all-meanings lookup resolves the `Array` segment of the qualified namespace name as though it declared a variable inside the module block, so the rule returns without reporting. ESLint 10.8.1 with `@typescript-eslint/parser` 8.58.2 reports the call and fixes it to `[]`, while rslint emits nothing; qualified namespace-name segments should not suppress the diagnostic unless the body has an actual matching scope declaration.

swwind: Fixed on the base branch in a66e63f3, since `no-object-constructor` misses `namespace N.Object { Object(); }` the same way; the regression cases for this rule are in 826d087c.

The parser nests one `ModuleDeclaration` per segment of a dotted name and the binder gives every segment a symbol, while scope-manager creates a variable only for a namespace named by a plain identifier — the whole dotted declaration opens a single `tsModule` scope with no variables in it. Both segments were affected, not only the last one.

| code | upstream |
| --- | --- |
| `namespace N.Array { Array(); }` | reports |
| `namespace Array.M { Array(); }` | reports |
| `namespace N { namespace M.Array { } Array(); }` | reports |
| `namespace N.Array { const Array = () => 1; Array(); }` | silent |
| `namespace N { const Array = () => 1; namespace M.Array { Array(); } }` | silent |

Two places carried the assumption. `utils.IsShadowed` read the outer segment's name straight off the statement, which is what hid `namespace Array.M { Array(); }`; the new `utils.IsQualifiedNamespaceSegment` excludes both segment shapes there. And `RefStore.resolveName` now steps over a symbol whose declarations are all dotted segments, resuming outside the scope whose table binds it, so the last two rows still find the real declaration.

elecmonkey: **[P2] Ignore declarations inside `declare global` augmentations**

When an external module contains `declare global { interface Array {} }` followed by `Array()`, ESLint 10.8.1 with `@typescript-eslint/parser` still reports and fixes the constructor call because a global augmentation creates no lexical `Array` variable. This all-meanings lookup instead treats the augmentation declaration as an in-file binding and suppresses the diagnostic; `const`, type, and namespace declarations inside `declare global` are affected too. Skip global-augmentation declarations when emulating `getVariableByName`, then continue resolving outer scopes.

swwind: Not reproducible — the rule already reports every one of these.

| code | ESLint 10.8.1 | this PR |
| --- | --- | --- |
| `export {}; declare global { interface Array {} } Array();` | reports, fixes to `[]` | same |
| `export {}; declare global { const Array: any; } Array();` | reports, fixes to `[]` | same |
| `export {}; declare global { type Array = any; } Array();` | reports, fixes to `[]` | same |
| `export {}; declare global { namespace Array {} } Array();` | reports, fixes to `[]` | same |
| `export {}; declare global { interface Array {} const a: any; Array(); }` | silent | same |

`IsNameDefinedInFileWithMeaning` runs the binder's scope walk, never the checker. A global augmentation's members are bound into the augmentation's own symbol table, so a walk starting outside the `declare global` block passes no table that carries them, and it ends with no result once it leaves the module's own locals — the resolver is handed no globals table for a module file. That an augmentation merges into the global scope is a checker-level fact, and the checker is not consulted here.

Locked in as invalid cases with the `[]` autofix, plus the last row as a valid case holding the other side of the boundary: a reference written inside the augmentation body does resolve against it.

internal/rules/no_array_constructor/no_array_constructor.go
elecmonkey: [P2] Ignore JSDoc-synthesized type parameters

In an `allowJs` file such as `/** @template Array */ function f(){ Array() }`, this guard sees ts-go's synthesized JSDoc type parameter and suppresses the diagnostic. ESLint 10.8.1 with `@typescript-eslint/parser` treats that JSDoc tag as a comment for scope purposes and reports/fixes the call to `[]`; the same miss occurs for JSDoc-generic arrows and classes. Exclude reparsed JSDoc type parameters when applying this shadow check.

swwind: Fixed in ae63aecc.

`HasEnclosingTypeParameter` now ignores ts-go type parameters reparsed from JSDoc `@template` tags. Class templates also step outside the synthetic class type-parameter environment before resolving authored outer bindings.

Added focused allowJs cases to the existing extras suite for JSDoc-generic functions, arrows, and classes, plus a boundary case confirming that a real outer `Array` binding still suppresses the rule. Verified against ESLint 10.8.1 with `@typescript-eslint/parser`.

internal/rules/no_array_constructor/no_array_constructor.go
```diff
@@ -0,0 +1,216 @@
...
+				ast.SymbolFlagsValue|ast.SymbolFlagsType|ast.SymbolFlagsNamespace|ast.SymbolFlagsAlias,
+			) {
+				return
+			}
+			// A config `/* global Array: off */` / `languageOptions.globals`
+			// entry un-declares the builtin, so `Array` no longer resolves to
+			// a known global — ESLint's `getVariableByName` would return
+			// undefined and the rule stays silent.
+			if !ctx.Globals.Access("Array").IsDeclared() {
+				return
```
elecmonkey, line 151: When linting a `.ts`/`.tsx` file with `languageOptions.globals: { Array: "off" }` or `/* global Array:off */`, `@typescript-eslint/parser` 8.58.2 still installs its library `Array` variable, so ESLint 10.8.1 reports `Array()` and fixes it to `[]`. This early return instead makes rslint silent because `ctx.Globals` applies the override to its language global; the migrated valid case was copied from upstream's Espree-only suite but runs as `file.ts` here, so it locks in the wrong TypeScript behavior.

internal/rules/no_array_constructor/no_array_constructor.go
```diff
@@ -0,0 +1,216 @@
...
+			if callee == nil || callee.Kind != ast.KindIdentifier {
+				return
+			}
+			if callee.AsIdentifier().Text != "Array" {
+				return
+			}
+			// `Array<Foo>()` — explicit type arguments mean this isn't the
+			// bare-constructor call the rule targets.
+			if typeArgs != nil && len(typeArgs.Nodes) > 0 {
+				return
```
elecmonkey, line 106: For the malformed but recoverable TypeScript input `Array<>(1, 2)`, this nonempty check lets the call through, and rslint reports it and rewrites the source to `[1,2]`. ESLint 10.8.1 with `@typescript-eslint/parser` 8.58.2 emits only `Parsing error: Type argument list cannot be empty` and never runs or fixes this rule, so the presence of the type-argument list should exclude the call even when its parsed list is empty.

internal/rules/no_array_constructor/no_array_constructor.go
```diff
@@ -0,0 +1,216 @@
...
+			// A single non-spread argument (`Array(9)`, `Array(x)`) may be
+			// intentionally creating a sparse array of that length — leave
+			// it alone, since the argument's runtime type can't be known
+			// statically.
+			if args != nil && len(args.Nodes) == 1 && args.Nodes[0].Kind != ast.KindSpreadElement {
+				return
+			}
+
+			// A local declaration shadows the global constructor.
+			if utils.IsShadowed(callee, "Array") {
```
elecmonkey: In a scope containing many matching calls, each invocation of `IsShadowed` recursively scans the enclosing body through `HasHoistedVarDeclaration`, making this rule quadratic in the number of calls/statements. For `function f(){ void Array(1,2); ... }`, measured rule time at the current head grew from 37 ms for 1,000 calls to 155 ms, 633 ms, and 2,554 ms for 2,000, 4,000, and 8,000 calls; cache shadowing by scope/name or use the binder/ref-store lookup so each report does not rewalk the body.

internal/rules/no_array_constructor/no_array_constructor.go
swwind: Fixed in b2d566a6.

The `globals: { Array: "off" }` early return now applies only to Espree-flavoured JavaScript files. TypeScript-flavoured files continue to report because typescript-eslint/parser installs its library `Array` variable independently of the authored override.

Added focused config and inline-comment cases to `TestNoArrayConstructorExtras`, and moved the migrated Espree case to an allowJs `.js` fixture so both parser behaviours remain covered.

internal/rules/no_array_constructor/no_array_constructor.go
swwind: Fixed in b2d566a6.

The rule now treats the presence of `TypeArguments` as sufficient to exclude the call, including a recoverable empty list. `Array<>(1, 2)` is covered as a valid rule case, so rslint no longer emits a rule diagnostic or autofix on top of the parser error.

internal/rules/no_array_constructor/no_array_constructor.go
swwind: Fixed in b2d566a6.

Normal local-binding checks now use the existing binder-backed RefStore lookup instead of recursively scanning every enclosing body for each matching call. The scope-manager-only parameter, type-parameter, and class-decorator boundaries remain handled explicitly.

`BenchmarkNoArrayConstructorRepeatedCalls` covers 1,000, 2,000, and 4,000 calls. On this machine the one-iteration timings changed from approximately 53/207/812 ms to 1.8/2.4/5.6 ms, removing the quadratic growth.

internal/rules/no_array_constructor/no_array_constructor.go
```diff
@@ -0,0 +1,216 @@
...
+			if ctx.Refs == nil {
+				if utils.IsShadowed(callee, "Array") {
+					return
+				}
+			} else if ctx.Refs.IsNameDefinedInFileWithMeaning(
+				callee,
+				"Array",
+				ast.SymbolFlagsValue|ast.SymbolFlagsType|ast.SymbolFlagsNamespace|ast.SymbolFlagsAlias,
+			) {
+				return
```
elecmonkey, line 129: [P2] Preserve parameter bindings in decorator calls

When `Array()` appears directly in a method parameter decorator and any method parameter is named `Array`—for example `class C { m(@dec(Array()) x: number, Array: any) {} }`—typescript-eslint’s function scope contains that parameter, so ESLint 10.8.1 emits nothing. This binder-backed lookup misses that scope boundary, while the later fallback only checks body declarations, causing rslint to report and autofix the call to `[]`; add an explicit parameter-decorator parameter lookup before treating it as the global constructor.

packages/rslint-test-tools/tests/eslint/rules/no-array-constructor.test.ts
```diff
@@ -0,0 +1,135 @@
...
+    'Array.foo()',
+    'new globalThis.Array',
+    'const createArray = Array => new Array()',
+    'var Array; new Array;',
+    {
+      code: 'new Array()',
+      languageOptions: {
+        globals: { Array: 'off' },
+      },
+    },
```
elecmonkey, line 24: [P1] Run the globals-off fixture as JavaScript

This object-form case defaults to `src/virtual.ts`, but the implementation deliberately reports `Array()` in TypeScript files even when `globals.Array` is `off`, matching `@typescript-eslint/parser`; therefore this valid case fails the newly registered npm test suite. Give it a `.js` filename, as the Go upstream test does, or move it to the invalid cases with the TypeScript expectation.

internal/rules/no_array_constructor/no_array_constructor.go
swwind: Fixed in cd1ae81f.

Added `utils.HasEnclosingParameter` to model scope-manager’s parameter visibility at the reference site. A reference directly inside a parameter decorator now sees the decorated method’s parameters, while the existing function-scope escape boundary ensures that an intervening arrow, function, or class still drops the method scope.

Added the reported sibling-parameter case to `TestNoArrayConstructorExtras` and covered the helper boundary in `TestShadowingScopeModels`.

packages/rslint-test-tools/tests/eslint/rules/no-array-constructor.test.ts
swwind: Fixed in cd1ae81f.

The migrated `globals: { Array: "off" }` valid fixture now explicitly uses `src/virtual.js`, matching the Espree behavior and the Go upstream fixture. TypeScript-flavoured cases continue to report as intended.

The focused `no-array-constructor.test.ts` suite now passes both its valid and invalid groups.

internal/utils/shadowing.go
```diff
@@ -215,6 +219,58 @@ func HasEnclosingTypeParameter(node *ast.Node, name string) bool {
...
+		if ast.IsFunctionLikeDeclaration(current) &&
+			!escapesFunctionScope(current, prevChild, inParameterDecorator, crossedScope) {
+			for _, parameter := range current.Parameters() {
+				if parameter != nil && parameter.Name() != nil && HasNameInBindingPattern(parameter.Name(), name) {
+					return true
+				}
+			}
+		}
+		if ast.IsFunctionLikeDeclaration(current) || ast.IsClassLike(current) {
+			crossedScope = true
```
elecmonkey, line 328: [P2] Keep outer bindings visible in object method keys

When a call is in the computed key of an object-literal method or accessor inside a parameter decorator, for example `class C { m(@dec({ [Array()]() {} }) x: number, Array: any) {} }`, typescript-eslint keeps the decorated method's function scope active because evaluating that key does not enter the object's method scope, so ESLint 10.8.1 emits nothing. This sets `crossedScope` merely upon encountering the ts-go method node; `escapesFunctionScope` then drops the actual decorated method, causing rslint to report and autofix the local call to `[[]]`. Only scopes actually entered while evaluating the reference should trigger this boundary.

internal/rules/no_array_constructor/no_array_constructor.go
```diff
@@ -0,0 +1,218 @@
...
+			}
+			// scope-manager keeps function parameters visible directly inside
+			// parameter decorators and class type parameters visible inside
+			// static members, where TypeScript's resolver hides them. It also
+			// keeps a function's parameter initializers in the same scope as its
+			// body declarations, where both TypeScript's resolver and IsShadowed
+			// follow runtime lexical semantics instead.
+			if utils.HasEnclosingParameter(callee, "Array") ||
```
elecmonkey: [P2] Avoid rescanning every parameter for each Array call

When a large or generated function contains P parameters and C matching `Array` calls but no `Array` parameter, this unconditional helper walks all P parameters for every call. At this head, a focused one-iteration benchmark with equal parameter/call counts took about 2 ms at 500, 5.5 ms at 1,000, 24 ms at 2,000, and 96 ms at 4,000, while the existing zero-parameter benchmark stayed around 0.8/1.3/2.4 ms for 1,000/2,000/4,000 calls. This reintroduces quadratic rule time; restrict the scan to parameter-decorator paths or cache the binding result per enclosing function.

internal/utils/shadowing.go
swwind: Fixed in ac1d05f8.

The scope-boundary state now advances only when evaluation actually enters a function or class scope. An object-literal method or accessor's computed name is outside that member's function scope, so encountering the ts-go member node no longer drops the decorated method's scope. The sibling `Array` parameter therefore remains visible and the rule stays silent.

Added focused valid cases for object-method and getter computed names inside a parameter decorator, plus helper coverage. `go test ./internal/utils` and `go test ./internal/rules/no_array_constructor` pass.

internal/rules/no_array_constructor/no_array_constructor.go
swwind: Fixed in ac1d05f8.

`HasEnclosingParameter` now scans bindings only when the reference is directly inside a parameter decorator. Ordinary function-body `Array()` calls no longer walk the enclosing function's parameter list.

Added `BenchmarkNoArrayConstructorRepeatedCallsWithParameters` with equal parameter/call counts. One-iteration results after the fix were approximately 1.05/1.92/3.83/9.23 ms at 500/1,000/2,000/4,000, showing near-linear scaling. `go test ./internal/utils` and `go test ./internal/rules/no_array_constructor` pass.

internal/rules/no_array_constructor/no_array_constructor.go
```diff
@@ -0,0 +1,218 @@
...
+				utils.HasEnclosingClassExpressionName(callee, "Array") ||
+				utils.IsShadowedFromParameterInitializer(callee, "Array") {
+				return
+			}
+			// A config `/* global Array: off */` / `languageOptions.globals`
+			// entry un-declares Espree's builtin. typescript-eslint/parser also
+			// installs its library `Array` variable, which remains visible after
+			// the override, so TypeScript-flavoured files still report.
+			if ast.IsInJSFile(callee) && !ctx.Globals.Access("Array").IsDeclared() {
+				return
```
elecmonkey, line 151: [P2] Respect parserOptions.lib before treating Array as predefined

When a TypeScript config sets `parserOptions.lib: []` together with `globals: { Array: "off" }`, `@typescript-eslint/parser` 8.58.2 has no predefined `Array` variable, so ESLint 10.8.1 reports nothing. This extension-only condition bypasses the undeclared-global check for every TypeScript file; a real CLI run at this head emitted `preferLiteral`, and `--fix` rewrote `Array()` to `[]`. Determine whether the effective TypeScript library supplies `Array` instead of relying only on `ast.IsInJSFile`.

internal/rules/no_array_constructor/no_array_constructor.go
```diff
@@ -0,0 +1,218 @@
...
+			// scope-manager keeps function parameters visible directly inside
+			// parameter decorators and class type parameters visible inside
+			// static members, where TypeScript's resolver hides them. It also
+			// keeps a function's parameter initializers in the same scope as its
+			// body declarations, where both TypeScript's resolver and IsShadowed
+			// follow runtime lexical semantics instead.
+			if utils.HasEnclosingParameter(callee, "Array") ||
+				utils.HasEnclosingTypeParameter(callee, "Array") ||
+				utils.HasEnclosingClassExpressionName(callee, "Array") ||
+				utils.IsShadowedFromParameterInitializer(callee, "Array") {
```
elecmonkey: [P2] Cache the scope-manager compatibility lookups

Each matching `Array` call executes uncached compatibility walkers: `HasEnclosingTypeParameter` scans every type parameter, `HasEnclosingParameter` scans every parameter in a parameter decorator, and `IsShadowedFromParameterInitializer` rescans the function body. Current-head benchmarks with equal numbers of calls and bindings grew from roughly 2–3 ms at 500 to 83–99 ms at 4,000 for type parameters, 86–89 ms for decorator parameters, and 165–168 ms for parameter initializers, while the existing 4,000-call no-binding baseline stayed near 2.4 ms. Cache these results per enclosing scope/name or derive them from the binder-backed store to avoid quadratic rule time on generated files.

internal/rules/no_array_constructor/no_array_constructor.go
swwind: Kept as a documented rslint divergence in e22daca4.

Matching this behavior exactly would require projecting `parserOptions.lib` into native rule contexts. That parser-specific setting is not part of the current native `RuleContext` / `LanguageOptions` contract, and I avoided widening the shared API solely for this rule.

The rule documentation now explicitly states that TypeScript files still treat `Array` as predefined when `parserOptions.lib: []` and `globals.Array: "off"` are configured. Added a focused configuration regression to the existing extras suite that verifies the diagnostic and `[]` autofix.

internal/rules/no_array_constructor/no_array_constructor.go
swwind: Fixed in e22daca4.

Added a per-rule `ScopeManagerShadowCache` keyed by enclosing declaration node and name. It caches type-parameter sets, parameter binding sets, and function-body declarations, while every reference still follows the existing evaluation-scope boundaries. This removes repeated O(bindings) scans without changing decorator, computed-key, or parameter-initializer behavior.

Added focused benchmarks for type parameters, parameter decorators, and parameter initializers. At 500/1,000/2,000/4,000 items, the latest one-iteration timings were approximately 1.16/1.53/3.95/7.39 ms, 1.11/1.76/3.58/5.61 ms, and 1.21/1.96/3.12/5.92 ms respectively. `go test ./internal/utils` and `go test ./internal/rules/no_array_constructor` pass.

elecmonkey (dismissed):
Good Good Good

elecmonkey (approved):

PR #1760, opened 2026-08-16
feat: port rule no-implicit-globals, add inline /* exported */ API

swwind:
## Summary

Port the `no-implicit-globals` rule from ESLint to rslint.

Disallows declarations (`var`/`function`, and optionally `const`/`let`/`class` via the `lexicalBindings` option) at the global scope, implicit global variable leaks from undeclared assignments, and redeclarations of or assignments to read-only global variables.

### Inline `/* exported */` API

The directive is parsed in `internal/rule` rather than inside the rule, so the other rules ESLint applies it to (`no-unused-vars`, `prefer-const`, `no-useless-assignment`) can consume it later without a second parser:

- `rule.ParseInlineExported(sourceFile, comments)` returns the declared name set plus per-comment name ranges, mirroring `ParseInlineGlobals`.
- `RuleContext.Exported` is the per-file view, with `Has(name)` and `Declarations()`. ESLint resolves the directive against the global scope and marks the variable it finds; this view answers for the raw name, so each rule applies its own scope gating.
- `internal/rule/directive_comment.go` holds the scanning both directives share: label match, `-- justification` split, and ESLint's exact whitespace class.

### Notes

`utils.IsInStrictMode` gained a CommonJS carve-out. TypeScript gives `.cjs` an `ExternalModuleIndicator` so it gets its own top-level scope, but it stays sloppy mode until it uses module syntax itself. `.cts` is excluded: ESLint's default language selection picks CommonJS for `.cjs` alone, so a `.cts` follows the module default and is strict. Besides deciding whether `globalVariableLeak` is reportable, this is shared with `no-lone-blocks`, `no-inner-declarations` and `no-eval`.

TypeScript declaration forms that bind no runtime variable of their own — ambient declarations and overload signatures — are not reported. ESLint reports them; the rule doc records the divergence.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/no-implicit-globals
- Source code: https://github.com/eslint/eslint/blob/v10.8.1/lib/rules/no-implicit-globals.js

## Checklist


elecmonkey (dismissed):
no issues found

internal/rule/directive_comment.go
```diff
@@ -0,0 +1,126 @@
...
+		end -= size
+	}
+	return start, end
+}
+
+// ECMAScript's \s set is Unicode Zs plus ASCII spacing/line terminators,
+// U+2028/U+2029, and BOM. unicode.IsSpace is not exact: it includes U+0085 and
+// excludes BOM. TypeScript's internal stringutil helper also accepts U+0085
+// and U+200B, so it cannot model ESLint's JavaScript regexp semantics here.
+func isECMAScriptWhitespace(r rune) bool {
```
swwind, line 116: This duplicates the ecmascript.IsWhitespace function

internal/rules/no_implicit_globals/no_implicit_globals.go
```diff
@@ -0,0 +1,343 @@
...
+		switch parent.Kind {
+		case ast.KindBinaryExpression:
+			binary := parent.AsBinaryExpression()
+			if binary == nil || binary.OperatorToken == nil ||
+				binary.OperatorToken.Kind != ast.KindEqualsToken || binary.Left != current {
+				return nil
+			}
+			if utils.IsDefaultValueInDestructuringAssignment(parent) {
+				current = parent
+				continue
```
elecmonkey, line 303: **[P2] Emit both diagnostics for destructuring defaults**

For `[foo = 1] = arr;`, ESLint 10.8.1's scope analysis represents the default initializer and the containing destructuring assignment as two writes, so `no-implicit-globals` emits two `globalVariableLeak` diagnostics over the outer assignment. The same happens for a read-only global: `/*global foo:readonly*/ [foo = 1] = arr;` produces two `assignmentToReadonlyGlobal` diagnostics. This implementation visits the single `foo` identifier once and only carries it through the default-value assignment to the outer root, so `checkImplicitGlobalWrite` reports once. Please preserve both scope writes (or otherwise emit the equivalent two diagnostics) so the diagnostic count matches upstream.

internal/rule/directive_comment.go
swwind: Therefore blocked by #1766

swwind: #1766 has landed, so this is unblocked. Merged main and dropped the local helper — the directive scanning now calls `ecmascript.IsWhiteSpaceOrLineTerminator`, which is the same set (Zs plus `\t`, `\v`, `\f`, `\n`, `\r`, U+2028, U+2029, U+FEFF).

internal/rules/no_implicit_globals/no_implicit_globals.go
swwind: Fixed. eslint-scope's `referencingDefaultValue` records one extra write per enclosing `AssignmentPattern` before the write for the assignment itself, so the diagnostic count is 1 + the number of destructuring defaults the identifier sits under. Verified against ESLint 10.8.1: `[foo = 1] = arr` reports twice, `[[foo = 1] = []] = arr` three times, `({a: {foo = 1} = {}} = obj)` three times, `({...foo} = obj)` once.

`findPureAssignmentRoot` now returns that write count alongside the root, and `checkImplicitGlobalWrite` reports the root once per write. The count also picks up `ShorthandPropertyAssignment`'s `ObjectAssignmentInitializer`, since `{foo = 1} = obj` keeps its default on the property rather than in a nested assignment. Tests cover the nested, shorthand and read-only-global shapes.

elecmonkey (commented):
Found several parity issues with the new no-implicit-globals port and exported-directive parser.

internal/rules/no_implicit_globals/no_implicit_globals.go
```diff
@@ -0,0 +1,388 @@
...
+// variable before reaching its reference loop, so the readonly assignment goes
+// unreported, while the leak is collected from the global scope's implicit
+// variables, which the directive never touches.
+func checkImplicitGlobalWrite(ctx rule.RuleContext, node *ast.Node, strictTopLevel bool) {
+	root, writes := findPureAssignmentRoot(node)
+	if root == nil {
+		return
+	}
+	if ctx.Refs != nil && ctx.Refs.IsDefinedInFile(node) {
+		return
```
elecmonkey: **[P2] Count top-level type-only declarations as definitions**

In a global TypeScript script such as `foo = 1; interface foo {}`, typescript-eslint's global scope already contains `foo`, so ESLint 10.8.1 creates neither an implicit global nor a read-only-global assignment. `IsDefinedInFile` resolves only the identifier's value meaning, so this implementation instead reports `globalVariableLeak` (and reports `assignmentToReadonlyGlobal` for `Array = 1; interface Array {}`). Use the global-script declaration semantics exposed by `IsGlobalNameReference`, which intentionally treats any authored top-level declaration as defining the scope variable.

internal/rules/no_implicit_globals/no_implicit_globals.go
```diff
@@ -0,0 +1,388 @@
...
+		if ctx.Exported.Has(name) {
+			return
+		}
+		for range writes {
+			ctx.ReportNode(root, assignmentToReadonlyGlobalMessage)
+		}
+	case utils.GlobalAccessWritable:
+		// Writable globals may be freely assigned.
+	default:
+		if !strictTopLevel && !utils.IsInStrictMode(node, ctx.SourceFile) {
```
elecmonkey: **[P2] Derive strictness from the actual evaluation scope**

For TypeScript syntax, this ancestor-based strictness check does not match scope-manager: `namespace N { foo = 1 }` and `enum E { A = (foo = 1) }` execute in strict TS module/enum scopes and ESLint reports nothing, while a top-level class decorator such as `@dec(foo = 1) class C {}` is evaluated outside the class scope and ESLint reports `globalVariableLeak`. `IsInStrictMode` currently misses the former scopes and treats every class descendant as strict, producing both false positives and false negatives.

internal/rules/no_implicit_globals/no_implicit_globals.go
```diff
@@ -0,0 +1,388 @@
...
+			wrappers = 0
+			current = parent
+
+		case ast.KindSpreadElement, ast.KindSpreadAssignment:
+			wrappers = 0
+			current = parent
+
+		case ast.KindNonNullExpression, ast.KindAsExpression, ast.KindTypeAssertionExpression:
+			wrappers++
+			current = parent
```
elecmonkey, line 398: **[P2] Traverse `satisfies` inside assignment patterns**

When `satisfies` occurs beneath a real assignment pattern, such as `[foo satisfies any] = arr` or `for ((foo satisfies any) of arr) {}`, @typescript-eslint/parser with ESLint 10.8.1 treats `foo` as a pure write and reports `globalVariableLeak`. This switch stops at `KindSatisfiesExpression`, so rslint misses the diagnostic; `satisfies` should be transparent after entering a destructuring or for-in/of pattern while remaining excluded for the direct target `(foo satisfies any) = 1`.

internal/rules/no_implicit_globals/no_implicit_globals.go
```diff
@@ -0,0 +1,388 @@
...
+		// binding at all. Upstream counts one declaration per signature and
+		// reports ambient forms too; see the rule doc's differences list.
+		return
+	}
+	nameNode := fn.Name()
+	if nameNode == nil || nameNode.Kind != ast.KindIdentifier {
+		return
+	}
+	if ast.GetEnclosingBlockScopeContainer(node) != sourceFileNode {
+		return
```
elecmonkey: **[P2] Honor ES5 hoisting for block functions**

When `languageOptions.ecmaVersion` is 3 or 5, eslint-scope hoists an ordinary block-level function into the surrounding variable scope, so ESLint 10.8.1 reports `{ function foo() {} }` in a script. `GetEnclosingBlockScopeContainer` always applies modern lexical block semantics here and returns early, causing rslint to miss that global declaration for supported legacy ECMAScript configurations.

internal/rule/inline_exported.go
```diff
@@ -0,0 +1,164 @@
...
+		}
+	}
+
+	if len(exported) == 0 {
+		return nil, nil
+	}
+	names := make(map[string]bool, len(exported))
+	for _, entry := range exported {
+		names[entry.Name] = true
+	}
```
elecmonkey, line 88: **[P3] Preserve `__proto__` behavior in exported directives**

For `/* exported __proto__ */ var __proto__;`, ESLint 10.8.1 still reports `globalNonLexicalBinding` because plugin-kit's `parseListConfig` uses an ordinary object and assigning the `__proto__` key does not create an own enumerable exported entry. Storing every parsed name in a Go map makes `Exported.Has("__proto__")` true and incorrectly suppresses the declaration, contrary to the stated parser-compatibility contract.

internal/rules/no_implicit_globals/no_implicit_globals.go
swwind: Fixed. The write check now asks `ctx.Refs.IsGlobalNameReference(node, name, ast.SymbolFlagsValue)` instead of `IsDefinedInFile`.

Both diagnostics need the global-script answer for the same reason: `GlobalScope#close` creates the implicit variable only when `!this.set.has(name)`, which ignores value/type-ness, and the read-only branch requires `variable.defs.length === 0`. A top-level `interface` satisfies both. Verified against ESLint 10.8.1: `foo = 1; interface foo {}`, `foo = 1; type foo = number;`, `Array = 1; interface Array {}`, `Array = 1; type Array<T> = T;` and `foo = 1; namespace foo {}` all report nothing. An inner scope keeps the old answer, since `#isValidResolution` rejects a type-only variable for a value reference: `function f() { foo = 1; interface foo {} }` still reports `globalVariableLeak` and `{ Array = 1; interface Array {} }` still reports `assignmentToReadonlyGlobal`.

internal/rules/no_implicit_globals/no_implicit_globals.go
swwind: Fixed in `utils.IsInStrictMode`. Enum bodies are strict scopes now, and a class's own decorators no longer take their strictness from the class — they are evaluated before the class scope exists. (The namespace scope arrived separately in main and came in with the merge.)

Verified against ESLint 10.8.1: `namespace N { foo = 1; }`, `namespace N { function f() { foo = 1; } }`, `enum E { A = (foo = 1) }`, `class C { p = (foo = 1); }`, `class C { [foo = 1]: number; }`, `class C { static { foo = 1; } }`, `class C extends (foo = 1) {}` and `class C { @dec(foo = 1) m() {} }` report nothing, while `@dec(foo = 1) class C {}` and `const C = @dec(foo = 1) class {};` report `globalVariableLeak`. The decorator carve-out is scoped to a decorator whose parent is the class itself, so a member decorator stays inside the class scope.

internal/rules/no_implicit_globals/no_implicit_globals.go
swwind: Fixed. `satisfies` is now transparent once the walk is inside a pattern and still opaque as the direct assignment target, which is the split the single unwrap an `AssignmentExpression`'s `Left` gets produces — it sheds a `TSAsExpression`, `TSTypeAssertion` or `TSNonNullExpression`, never a `TSSatisfiesExpression`.

Verified against ESLint 10.8.1: `[foo satisfies any] = arr`, `[[foo satisfies any]] = arr`, `({a: foo satisfies any} = obj)`, `[(foo satisfies any) as any] = arr`, `for ((foo satisfies any) of arr) {}` and `for ((Array satisfies any) in obj) {}` all report, while `(foo satisfies any) = 1`, `(foo satisfies any as any) = 1` and `((foo as any) satisfies any) = 1` do not.

internal/rules/no_implicit_globals/no_implicit_globals.go
swwind: Fixed. `ctx.LanguageOptions.EffectiveECMAVersion()` arrived in main and came in with the merge, so the function-declaration check now resolves the binding scope through `utils.FindEnclosingScope` below ES2015 and `ast.GetEnclosingBlockScopeContainer` from ES2015 on.

Verified against ESLint 10.8.1: at `ecmaVersion` 3 and 5, `{ function foo() {} }` and `if (true) { function foo() {} }` report `globalNonLexicalBinding` and stop reporting from ES2015 on. It is the enclosing variable scope rather than the global one, so `function outer() { { function foo() {} } }` reports only `outer` at ES5.

internal/utils/strict_mode.go
```diff
@@ -22,17 +32,20 @@ func IsInStrictMode(node *ast.Node, sourceFile *ast.SourceFile) bool {
...
+		if current.Kind == ast.KindModuleDeclaration || current.Kind == ast.KindEnumDeclaration {
 			return true
 		}
 
-		// Class bodies are always strict in ES2015+
-		if ast.IsClassLike(current) {
+		// Class bodies are always strict in ES2015+. A decorator applied to the
+		// class itself is evaluated before the class scope exists, so it is not
+		// part of that body; a decorator on a member is.
+		if ast.IsClassLike(current) && !(child.Kind == ast.KindDecorator && child.Parent == current) {
```
elecmonkey: [P1] Rewrite this condition so the lint gate passes

The current PR head fails the required `Lint&Check` job because staticcheck reports QF1001 on this condition (`could apply De Morgan's law`). Since this blocks CI for every build of the patch, rewrite the negated conjunction into the equivalent accepted form.

elecmonkey (commented):
The implementation still diverges from the targeted ESLint behavior for multiple parser-accepted assignment-pattern forms and for the `__proto__` exported directive. The current head also fails the repository's required staticcheck gate.

internal/rules/no_implicit_globals/no_implicit_globals.go
```diff
@@ -0,0 +1,420 @@
...
+			current = parent
+
+		case ast.KindSatisfiesExpression:
+			satisfies = true
+			current = parent
+
+		case ast.KindParenthesizedExpression:
+			current = parent
+
+		default:
```
elecmonkey: [P2] Traverse expression nodes inside assignment patterns

For parser-accepted TypeScript such as `[foo + bar] = arr`, ESLint 10.8.1 with `@typescript-eslint/parser` 8.67.0 emits two `globalVariableLeak` diagnostics because its pattern visitor treats both identifiers as writes. This default arm stops at the inner `BinaryExpression` before reaching the enclosing array assignment, so rslint emits none; the walk needs to preserve the scope-manager behavior for expression-shaped pattern elements.

internal/rules/no_implicit_globals/no_implicit_globals.go
```diff
@@ -0,0 +1,420 @@
...
+	for current != nil && current.Parent != nil {
+		parent := current.Parent
+		switch parent.Kind {
+		case ast.KindBinaryExpression:
+			binary := parent.AsBinaryExpression()
+			if binary == nil || binary.OperatorToken == nil ||
+				binary.OperatorToken.Kind != ast.KindEqualsToken || binary.Left != current {
+				return nil, 0
+			}
+			if utils.IsDefaultValueInDestructuringAssignment(parent) {
```
elecmonkey: [P2] Count defaults nested under object rest

When a default is nested under object rest, for example `[{...foo = bar}] = arr`, ESLint 10.8.1 with `@typescript-eslint/parser` 8.67.0 records two writes and emits two `globalVariableLeak` diagnostics over the outer assignment. Here `IsDefaultValueInDestructuringAssignment` does not recognize a `SpreadAssignment` parent, so rslint treats `foo = bar` as the root and emits only one diagnostic at the inner range.

swwind:
Fixed the latest review findings in 21ef43ac1:

- Rewrote the strict-mode condition so QF1001 passes.
- Matched PatternVisitor traversal for parser-accepted expression-shaped assignment targets, including call-callee versus argument behavior.
- Counted destructuring defaults nested under object rest and reported them at the outer assignment range.

The regression cases are included in no_implicit_globals_extras_test.go. `go test ./internal/rules/no_implicit_globals -count=1` passes, and the focused QF1001 staticcheck passes.

swwind:
Follow-up on the `__proto__` review comment: this appears to be an upstream ESLint bug rather than intended directive behavior.

`parseListConfig` stores entries in `{}`, so assigning the `__proto__` key invokes the inherited legacy setter instead of creating an own property. Even changing that object alone would not be enough, because `Object.assign` later copies the result into another `{}` and would drop the key again. Both maps should likely use `Object.create(null)` (or an equivalent safe representation).

Rslint intentionally does not reproduce this accidental object-prototype behavior: `/* exported __proto__ */` is treated like any other exported name. I’ll report the issue upstream and we can revisit this divergence if upstream defines different intended semantics.

internal/utils/strict_mode.go
```diff
@@ -2,17 +2,27 @@ package utils
...
 func IsInStrictMode(node *ast.Node, sourceFile *ast.SourceFile) bool {
-	// ES modules are always strict
-	if ast.IsExternalModule(sourceFile) {
+	// ES modules are always strict. ast.IsExternalModule also reports true for
+	// a .cjs file with no import/export at all: TS forces its
+	// ExternalModuleIndicator so it gets its own (non-global) top-level scope,
+	// but it is still CommonJS — sloppy mode by default — not a real ES module.
+	// Such a file is strict only once it uses module syntax itself.
+	if ast.IsExternalModule(sourceFile) &&
+		!(!HasModuleSyntax(sourceFile) && IsCommonJSFileExtension(sourceFile.FileName())) {
```
elecmonkey: [P1] Rewrite both negated conjunctions so lint passes

The required `Lint&Check` job fails on the current head with QF1001 at this condition and at line 48 (`could apply De Morgan's law`). The patch cannot pass CI until both negated conjunctions are rewritten into the staticcheck-approved form.

internal/rules/no_implicit_globals/no_implicit_globals.go
```diff
@@ -0,0 +1,455 @@
...
+		switch parent.Kind {
+		case ast.KindBinaryExpression:
+			binary := parent.AsBinaryExpression()
+			if binary == nil || binary.OperatorToken == nil {
+				return nil, 0
+			}
+			if binary.OperatorToken.Kind != ast.KindEqualsToken {
+				if utils.IsInDestructuringAssignment(parent) {
+					current = parent
+					continue
```
elecmonkey: [P2] Preserve compound-assignment roles inside patterns

With `@typescript-eslint/parser` 8.67.0, `[foo += bar] = arr;` is accepted and PatternVisitor records two writes for `foo` while treating `bar` as the right-hand read. ESLint 10.8.1 therefore emits two leaks for `foo`; this branch instead climbs from either child without incrementing `writes`, so making `foo` writable leaves a false-positive leak for `bar`, while making `bar` writable leaves only one diagnostic for `foo`. Restrict this traversal to the compound assignment's left side and preserve its extra write.

internal/rules/no_implicit_globals/no_implicit_globals.go
```diff
@@ -0,0 +1,455 @@
...
+			}
+			if utils.IsDefaultValueInDestructuringAssignment(parent) {
+				writes++
+				enterPattern(parent)
+				continue
+			}
+			if wrappers > 1 || satisfies {
+				return nil, 0
+			}
+			return parent, writes
```
elecmonkey, line 358: [P2] Count nested equals expressions as pattern assignments

For `[(foo = bar) + baz] = arr;` with `baz` configured writable, ESLint 10.8.1 emits two `globalVariableLeak` diagnostics for `foo`, both over the outer assignment. `IsDefaultValueInDestructuringAssignment` only recognizes directly contained defaults here, so this path returns the inner `foo = bar` with one write, producing the wrong count and range.

elecmonkey (commented):
The current head fails the required lint gate and still diverges from ESLint for several parser-accepted assignment-pattern forms.

internal/rules/no_implicit_globals/no_implicit_globals.go
```diff
@@ -0,0 +1,455 @@
...
+// to, so the same declaration inside a function or namespace leaves the name
+// global — the distinction IsGlobalNameReference draws.
+func checkImplicitGlobalWrite(ctx rule.RuleContext, node *ast.Node, strictTopLevel bool) {
+	root, writes := findPureAssignmentRoot(node)
+	if root == nil {
+		return
+	}
+	name := node.Text()
+	if ctx.Refs != nil && !ctx.Refs.IsGlobalNameReference(node, name, ast.SymbolFlagsValue) {
+		return
```
elecmonkey, line 272: [P2] Treat synthetic pattern names as write targets

For parser-accepted patterns such as `[function foo(){}] = arr;` and `[class C {}] = arr;`, typescript-eslint's PatternVisitor visits `foo` or `C` as an assignment target, so ESLint 10.8.1 emits one leak. This lookup instead recognizes those identifiers as function/class self-bindings and returns early, causing Rslint to emit nothing; identifiers reached through the synthetic expression-pattern traversal need parity-specific handling rather than ordinary lexical-reference filtering.

internal/utils/strict_mode.go
swwind: Fixed in 338fd1c8. Both negated conjunctions in `IsInStrictMode` now use their equivalent De Morgan forms, removing the QF1001 findings while preserving the CommonJS and class-decorator behavior.

internal/rules/no_implicit_globals/no_implicit_globals.go
swwind: Fixed in 338fd1c8. Compound assignments inside expression-shaped patterns now traverse only their left side and add the compound write before the containing destructuring write. The focused cases verify that `[foo += bar] = arr` reports twice for `foo` and never treats `bar` as a write.

internal/rules/no_implicit_globals/no_implicit_globals.go
swwind: Fixed in 338fd1c8. An `=` expression anywhere inside a parser-accepted destructuring pattern now contributes its additional assignment-pattern write and continues to the outer assignment, so `[(foo = bar) + baz] = arr` reports twice over the outer range.

internal/rules/no_implicit_globals/no_implicit_globals.go
swwind: Fixed in 338fd1c8. Function and class expression names reached through PatternVisitor's synthetic expression traversal now bypass their ordinary expression-local self-binding and use the containing assignment's strictness. Focused cases cover both `[function foo(){}] = arr` and `[class C {}] = arr`.

internal/rule/inline_exported.go
swwind: Kept as an intentional rslint divergence and documented it in 338fd1c8. `/* exported __proto__ */` treats `__proto__` like every other exported name; rslint does not reproduce the accidental behavior caused by plugin-kit's ordinary-object map invoking JavaScript's inherited `__proto__` setter. A focused valid case now locks in that behavior.

internal/rules/no_implicit_globals/no_implicit_globals.go
```diff
@@ -0,0 +1,487 @@
...
+// global — the distinction IsGlobalNameReference draws.
+func checkImplicitGlobalWrite(ctx rule.RuleContext, node *ast.Node, strictTopLevel bool) {
+	root, writes := findPureAssignmentRoot(node)
+	if root == nil {
+		return
+	}
+	name := node.Text()
+	syntheticPatternName := isSyntheticPatternName(node)
+	if ctx.Refs != nil && !syntheticPatternName &&
+		!ctx.Refs.IsGlobalNameReference(node, name, ast.SymbolFlagsValue) {
```
elecmonkey: [P2] Resolve synthetic pattern writes in the outer scope

With parser-accepted expression-shaped targets, PatternVisitor records each identifier it visits as a write in the scope containing the outer assignment. This special case covers only direct function/class names: `[(foo => 0)] = arr;` produces no diagnostic here instead of ESLint 10.8.1/@typescript-eslint/parser 8.67.0's one leak because `foo` resolves as a parameter, while `var C; [class C {}] = arr;` bypasses the real outer `C` binding and adds a spurious leak. Synthetic writes need to be resolved from the outer assignment scope rather than using the identifier's ordinary scope or skipping resolution entirely.

internal/rules/no_implicit_globals/no_implicit_globals.go
```diff
@@ -0,0 +1,487 @@
...
+			// inside the expression. PatternVisitor records both as reads.
+			return nil, 0
+
+		default:
+			// @typescript-eslint's PatternVisitor falls back to ordinary child
+			// traversal for parser-accepted expression-shaped pattern elements.
+			// Keep climbing while this expression remains inside the assignment
+			// pattern; IsInDestructuringAssignment excludes default RHS values
+			// and computed property keys.
+			if !utils.IsInDestructuringAssignment(parent) {
```
elecmonkey: [P2] Stop traversing decorators as assignment targets

When a parser-accepted destructuring target contains a decorator, PatternVisitor's `Decorator()` override deliberately skips the decorator subtree. For `[@foo class {}] = arr;`, upstream reports no leak, but this generic ancestry climb accepts `foo` merely because the decorator is inside the array target and Rslint reports one; `[@dec(foo) class C {}] = arr;` similarly gains two spurious decorator writes. Treat decorators as a traversal boundary when emulating PatternVisitor.

internal/rules/no_implicit_globals/no_implicit_globals.go
```diff
@@ -0,0 +1,487 @@
...
+			(ctx.Refs != nil && ctx.Refs.HasNonGlobalTopLevelScope())
+
+		// ESLint records an implicit global only for a write in sloppy-mode
+		// code, so top-level strictness decides whether a leak is reportable.
+		// It comes from the same module-ness the declaration checks use, minus
+		// CommonJS: a .cjs file has its own top-level scope without being an ES
+		// module, so its top level stays sloppy until it uses module syntax
+		// itself.
+		strictTopLevel := hasNonGlobalTopLevelScope &&
+			(utils.HasModuleSyntax(ctx.SourceFile) || !utils.IsCommonJSFileExtension(ctx.SourceFile.FileName()))
```
elecmonkey: [P2] Keep CommonJS files sloppy when they contain module syntax

For a `.cjs` file, ESLint's default language remains CommonJS and sloppy even when the TypeScript parser accepts import/export syntax. With `export {}; foo = 1;` in `x.cjs`, ESLint 10.8.1 reports `globalVariableLeak`, whereas `HasModuleSyntax` makes `strictTopLevel` true here and Rslint reports nothing. Extension-selected CommonJS should remain sloppy unless an explicitly supported source type changes it; the same predicate in `utils.IsInStrictMode` also needs the corresponding correction.

elecmonkey (commented):
The current implementation still produces both false positives and false negatives for parser-accepted assignment patterns, and it incorrectly suppresses leaks in CommonJS files containing module syntax. These are observable parity regressions against the targeted ESLint behavior.

internal/rules/no_implicit_globals/no_implicit_globals.go
swwind: Fixed in `9afd6320`. Pattern writes are now resolved from the outer assignment node rather than the identifier's syntactic scope. This makes `[(foo => 0)] = arr` report one leak, while `var C; [class C {}] = arr` reports only the existing global declaration and no spurious synthetic leak. Focused regression cases cover both directions.

internal/rules/no_implicit_globals/no_implicit_globals.go
swwind: Fixed in `9afd6320`. The PatternVisitor emulation now stops at `Decorator` nodes. `[@foo class {}] = arr` reports nothing, while `[@dec(foo) class C {}] = arr` skips the decorator callee and argument and reports only the class name's synthetic write. Focused cases lock in both shapes.

internal/rules/no_implicit_globals/no_implicit_globals.go
swwind: Fixed in `9afd6320`. Top-level strictness now follows the normalized `sourceType`, so a default `.cjs` file remains CommonJS and sloppy even when the parser accepts `export {}`. The shared strict-mode helper now also supports an explicit source goal. Focused rule and utility tests cover `.cjs` with module syntax.

packages/rslint-test-tools/tests/eslint/rules/no-implicit-globals.test.ts
```diff
@@ -0,0 +1,176 @@
...
+    '/* exported foo */ function foo() {}',
+    {
+      code: '/* exported a */ const a = 1;',
+      options: { lexicalBindings: true },
+    },
+  ],
+  invalid: [
+    // `var` and function declarations
+    {
+      code: 'var foo = 1;',
```
elecmonkey, line 118: [P1] Run this suite with script semantics

These cases use `src/virtual.ts` by default, so the repository normalizes the file to `sourceType: "module"` unless each case overrides it. Consequently, both current npm CI jobs fail on the first invalid case, `var foo = 1;`, because no global declaration diagnostic is produced. Please give this script-oriented suite an explicit `sourceType: "script"` or a filename whose defaults provide script semantics.

elecmonkey (commented):
The current head still has a CI-blocking integration-test configuration issue and several observable parity mismatches in assignment-pattern traversal, source-type strictness, and TypeScript block-function scope handling.

internal/rules/no_implicit_globals/no_implicit_globals.go
```diff
@@ -0,0 +1,459 @@
...
+// be a pattern, so `(foo as any) = 1` is a write while `(foo as any)! = 1`,
+// `foo!! = 1` and `(foo satisfies any) = 1` are not. A wrapper around a
+// destructuring target does not survive that test either: `[foo]` is no longer a
+// destructuring pattern once an assertion wraps it, so `([foo] as any) = arr` is
+// no write at all. Every other path — a for-in/for-of head, anything nested
+// inside a pattern — visits the pattern directly and accepts a wrapped target
+// however deeply it is nested, `satisfies` included: `[foo satisfies any] = arr`
+// is a write.
+func findPureAssignmentRoot(node *ast.Node) (*ast.Node, int) {
+	if ast.IsPartOfTypeNode(node) {
```
elecmonkey: [P2] Preserve PatternVisitor traversal through type nodes

This early return drops identifiers that upstream PatternVisitor visits inside parser-accepted expression-shaped assignment patterns. For example, with `foo` writable and `T` readonly, ESLint 10.8.1 plus `@typescript-eslint/parser` 8.65.0 reports `assignmentToReadonlyGlobal` for `[foo as T] = arr;`, while Rslint reports nothing; `[foo<T>]` and `[foo satisfies T]` show the same mismatch. Please distinguish ordinary type positions from identifiers reached through this synthetic pattern traversal.

internal/rules/no_implicit_globals/no_implicit_globals.go
```diff
@@ -0,0 +1,459 @@
...
+			// inside the expression. PatternVisitor records both as reads.
+			return nil, 0
+
+		default:
+			// @typescript-eslint's PatternVisitor falls back to ordinary child
+			// traversal for parser-accepted expression-shaped pattern elements.
+			// Keep climbing while this expression remains inside the assignment
+			// pattern; IsInDestructuringAssignment excludes default RHS values
+			// and computed property keys.
+			if !utils.IsInDestructuringAssignment(parent) {
```
elecmonkey: [P2] Respect PatternVisitor boundaries for object methods

The generic ancestor walk treats object-method keys and bodies as assignment targets, whereas upstream PatternVisitor's `Property` visitor does not visit a method key and visits the method body according to its visitor keys. For `[({foo(){}})] = arr;`, ESLint reports nothing but Rslint reports a leak for `foo`; for `[({foo(){return bar}})] = arr;`, ESLint reports only `bar` while Rslint reports both `foo` and `bar`. Please model these node roles explicitly instead of allowing every descendant to climb to the outer destructuring assignment.

internal/utils/strict_mode.go
```diff
@@ -2,17 +2,35 @@ package utils
...
+//   - The node is inside a TypeScript namespace/module or enum scope
+//
+// A class's own decorators are the one part of a class that is evaluated
+// outside the class scope, so they take their strictness from whatever encloses
+// the class.
 func IsInStrictMode(node *ast.Node, sourceFile *ast.SourceFile) bool {
-	// ES modules are always strict
-	if ast.IsExternalModule(sourceFile) {
+	sourceType := "script"
+	if tspath.GetAnyExtensionFromPath(sourceFile.FileName(), nil, false) == tspath.ExtensionCjs {
```
elecmonkey, line 21: [P2] Pass the normalized sourceType to strict-mode consumers

This wrapper re-derives the source goal from the extension and module syntax, so existing callers in `no-inner-declarations`, `no-lone-blocks`, and `no-eval` cannot observe an explicit `languageOptions.sourceType`. A `.cjs` file configured as `module` is therefore treated as sloppy by Rslint although ESLint treats it as strict; conversely, a file containing `export {}` but configured as `script` is treated as strict here. Please have these callers use the normalized `RuleContext.LanguageOptions.SourceType`.

internal/rules/no_implicit_globals/no_implicit_globals.go
```diff
@@ -0,0 +1,459 @@
...
+	reportDeclaration(ctx, node, nameNode.Text(), "function", false)
+}
+
+// functionDeclarationScope returns the scope a function declaration binds its
+// name in. eslint-scope only creates block scopes from ES2015 on, so under
+// `languageOptions.ecmaVersion` 3 or 5 a block-level function declaration binds
+// in the enclosing variable scope instead, and `{ function foo() {} }` is a
+// global declaration.
+func functionDeclarationScope(ctx rule.RuleContext, node *ast.Node) *ast.Node {
+	if ctx.LanguageOptions.EffectiveECMAVersion() < 2015 {
```
elecmonkey: [P2] Follow the configured TypeScript parser for ES5 block functions

This applies Espree/eslint-scope's ES5 hoisting rule even though Rslint always uses the TypeScript parser. For a TypeScript-parsed script with `ecmaVersion: 5`, ESLint 10.8.1 plus the repository's `@typescript-eslint/parser` 8.65.0 reports nothing for `{ function foo() {} }` because its scope manager retains the block scope, while Rslint reports `globalNonLexicalBinding`. Please base this decision on the actual parser semantics and update the added regression cases accordingly.

packages/rslint-test-tools/tests/eslint/rules/no-implicit-globals.test.ts
swwind: Fixed in `a36eadf7`. The npm RuleTester suite now applies explicit `sourceType: "script"` semantics to every case while preserving case-specific language options. After rebuilding the current native CLI, the focused `no-implicit-globals.test.ts` run passes both valid and invalid suites.

internal/rules/no_implicit_globals/no_implicit_globals.go
swwind: Fixed in `a36eadf7`. Synthetic assignment-pattern traversal no longer rejects an identifier solely because it is part of a type node; it can continue through type syntax only while the ancestor remains inside the outer destructuring pattern. Focused cases cover `[foo as T]`, `[foo<T>]`, and `[foo satisfies T]` with `T` read-only.

internal/rules/no_implicit_globals/no_implicit_globals.go
swwind: Fixed in `a36eadf7`. The PatternVisitor emulation now treats `MethodDeclaration` explicitly: the method key is skipped, while parameters, types, and the body retain their ordinary visitor traversal inside a synthetic assignment pattern. Focused cases verify that `[({foo(){}})] = arr` reports nothing and `[({foo(){return bar}})] = arr` reports only `bar`.

internal/utils/strict_mode.go
swwind: Fixed in `a36eadf7`. `no-inner-declarations`, `no-lone-blocks`, and `no-eval` now pass `ctx.LanguageOptions.EffectiveSourceType()` to the shared strict-mode helper instead of re-deriving the source goal from syntax or extension. The regression coverage exercises both an explicit module goal without module syntax and an explicit script goal with export syntax.

internal/rules/no_implicit_globals/no_implicit_globals.go
swwind: Fixed in `a36eadf7`. Function-declaration scope now always follows the TypeScript parser’s block-scope structure, independent of the configured ECMAScript edition. The ES5 cases for `{ function foo() {} }` and `if (...) { function foo() {} }` are now valid, while the enclosing top-level function remains reportable.

internal/utils/strict_mode.go
```diff
@@ -2,17 +2,35 @@ package utils
...
+	}
+	return IsInStrictModeWithSourceType(node, sourceFile, sourceType)
+}
+
+// IsInStrictModeWithSourceType is IsInStrictMode with the normalized source
+// goal selected by the caller. CommonJS stays sloppy even when the parser
+// accepts import/export syntax, while an explicitly selected module is strict
+// regardless of syntax or file extension.
+func IsInStrictModeWithSourceType(node *ast.Node, sourceFile *ast.SourceFile, sourceType string) bool {
+	if sourceType == "module" {
```
elecmonkey, line 34: [P1] Restore script semantics in the affected JS suites

Making normalized `module` sources strict changes existing RuleTester cases because their default `src/virtual.ts` filename resolves to module. At head `27258a2b`, both npm CI jobs fail: the script-oriented invalid cases in `no-eval`, `no-inner-declarations`, and `no-lone-blocks` now produce zero diagnostics. The Go suites were updated with explicit script defaults, but the corresponding TypeScript suites also need that adjustment before this can merge.

internal/rules/no_implicit_globals/no_implicit_globals.go
```diff
@@ -0,0 +1,461 @@
...
+func checkImplicitGlobalWrite(ctx rule.RuleContext, node *ast.Node) {
+	root, writes := findPureAssignmentRoot(node)
+	if root == nil {
+		return
+	}
+	name := node.Text()
+	// PatternVisitor records every target write in the scope containing the
+	// outer assignment, even when the identifier is syntactically inside an
+	// arrow, function, or class expression. Resolve from that same location.
+	if ctx.Refs != nil && !ctx.Refs.IsGlobalNameReference(root, name, ast.SymbolFlagsValue) {
```
elecmonkey: [P2] Resolve imported aliases before consulting global access

This lookup requests only `SymbolFlagsValue`, so import aliases are not recognized as local bindings. For `import { Array } from "foo"; Array = 1;` in module mode, ESLint 10.8.1 with `@typescript-eslint/parser` 8.65.0 reports nothing, while Rslint reports `assignmentToReadonlyGlobal` by falling through to the built-in `Array` global; `import foo = require("foo")` and configured read-only globals exhibit the same false positive. Include alias semantics when resolving these write targets.

internal/rules/no_implicit_globals/no_implicit_globals.go
```diff
@@ -0,0 +1,461 @@
...
+		case ast.KindDecorator:
+			// PatternVisitor deliberately skips decorator subtrees.
+			return nil, 0
+
+		case ast.KindMethodDeclaration:
+			// PatternVisitor's object-property handling skips a method's key,
+			// while its ordinary visitor keys still traverse parameters, return
+			// types, and the body of parser-accepted expression patterns.
+			method := parent.AsMethodDeclaration()
+			if method == nil || method.Name() == current || !utils.IsInDestructuringAssignment(parent) {
```
elecmonkey: [P2] Distinguish class-member keys from object-method keys

The same ts-go `MethodDeclaration` handling is applied to class and object members, while upstream reaches them through different ESTree nodes. For `[class { foo() {} }] = arr;`, upstream reports one `globalVariableLeak` for the class method key and Rslint reports none; conversely, `[({get foo(){}})] = arr;` produces no upstream diagnostic but one Rslint leak because accessors fall through the generic walk. Class member keys must remain traversable while object method/accessor keys must be skipped.

internal/rules/no_implicit_globals/no_implicit_globals.go
```diff
@@ -0,0 +1,461 @@
...
+			// inside the expression. PatternVisitor records both as reads.
+			return nil, 0
+
+		default:
+			// @typescript-eslint's PatternVisitor falls back to ordinary child
+			// traversal for parser-accepted expression-shaped pattern elements.
+			// Keep climbing while this expression remains inside the assignment
+			// pattern; IsInDestructuringAssignment excludes default RHS values
+			// and computed property keys.
+			if !utils.IsInDestructuringAssignment(parent) {
```
elecmonkey: [P2] Stop synthetic traversal at TSTypeAnnotation nodes

The generic ancestry path traverses parameter, return, and local-declaration type annotations even though upstream `PatternVisitor` has an explicit `TSTypeAnnotation()` boundary. For `[(foo: T) => 0] = arr;`, upstream emits one leak for `foo`, whereas Rslint emits two by also treating `T` as a write; `[function foo(x: T): U { let z: V }] = arr;` similarly produces six Rslint diagnostics versus three upstream. Preserve direct assertion traversal such as `[foo as T]`, but stop when the identifier is under a real type-annotation wrapper.

internal/rules/no_implicit_globals/no_implicit_globals.go
elecmonkey: [P2] Use configured script scope for top-level declarations

The selected global-reference helper recognizes type-only top-level declarations only when `ast.IsGlobalSourceFile` is true, which follows parsed module syntax instead of the configured source goal. With `sourceType: "script"`, `export {}; foo = 1; interface foo {}` produces no diagnostic in ESLint 10.8.1 because `foo` is defined in the global scope, but Rslint reports `globalVariableLeak`; the same mismatch occurs for type aliases and namespaces. The global-declaration check needs to honor the normalized script scope even when import/export syntax is present.

elecmonkey (commented):
The PR head builds and its focused Go suites pass, but real differential execution found multiple false positives and false negatives against ESLint 10.8.1 with @typescript-eslint/parser 8.65.0. The required npm CI jobs also fail in three existing rule suites.

internal/utils/strict_mode.go
swwind: Fixed in `a2b086750`. The affected RuleTester suites now preserve explicit script semantics. I reran the focused `no-implicit-globals`, `no-eval`, `no-inner-declarations`, and `no-lone-blocks` Rstest files: all 8 valid/invalid suites pass.

internal/rules/no_implicit_globals/no_implicit_globals.go
swwind: Fixed. The write-target lookup now includes `SymbolFlagsAlias`, so imported aliases resolve as local definitions before the rule consults configured or built-in globals. Focused cases cover a named import shadowing `Array`, an import-equals binding, and an imported name that also exists as a configured read-only global.

internal/rules/no_implicit_globals/no_implicit_globals.go
swwind: I did not implement this recovery-only behavior. After TypeScript syntax is erased, both `[class { foo() {} }] = arr` and `[({ get foo() {} })] = arr` still produce JavaScript that the engine rejects as an invalid destructuring assignment target. I removed the earlier recovery-only method/decorator/expression traversal and its invalid test cases, and documented that this rule limits assignment-target parity to code that transpiles to executable JavaScript.

internal/rules/no_implicit_globals/no_implicit_globals.go
swwind: I did not implement this recovery-only behavior. `[(foo: T) => 0] = arr` transpiles to `[(foo) => 0] = arr`, which the JavaScript engine rejects as an invalid destructuring assignment target. I removed the earlier invalid expression-pattern cases and their dedicated traversal logic. Runnable TypeScript assertion targets such as `[foo as T] = arr` remain covered.

internal/rules/no_implicit_globals/no_implicit_globals.go
swwind: Fixed. `RefStore.IsGlobalNameReference` now uses the normalized `globalTopLevelScope` supplied by language defaults instead of deriving the answer from parsed module syntax. Focused cases cover interfaces, type aliases, and namespaces in a configured script containing `export {}`.

elecmonkey (commented):
Found two issues in the current head.

internal/rule/ref_store.go
```diff
@@ -516,17 +516,19 @@ func functionExpressionSelfBinding(fn *ast.Node, name string, meaning ast.Symbol
...
+// authoritative even when the parser sees module syntax. Inner scopes hold
+// separate variables, and a value reference only resolves to one declared in a
+// requested space, so a type-only declaration inside a namespace or function
+// leaves it global.
 func (s *RefStore) IsGlobalNameReference(location *ast.Node, name string, meaning ast.SymbolFlags) bool {
 	if s == nil || location == nil || name == "" {
 		return false
 	}
-	if s.sourceFile != nil && ast.IsGlobalSourceFile(s.sourceFile.AsNode()) &&
+	if s.sourceFile != nil && s.init.globalTopLevelScope &&
```
elecmonkey: [P1] Making the normalized source type authoritative here causes `TestNoPromiseExecutorReturnExtras/valid-21` and `valid-22` to fail because their default `file.ts` inputs now have module semantics, while the cases expect script-scope `interface/type Promise` declarations to shadow the global. Both Linux and Windows Go CI fail, and `go test ./internal/rules/no_promise_executor_return -count=1` reproduces locally. Mark these cases with `SourceType: "script"` (which makes the package pass locally) or otherwise update the affected expectations before merging.

internal/rules/no_implicit_globals/no_implicit_globals.go
```diff
@@ -0,0 +1,425 @@
...
+			// A member expression writes its property, not either identifier
+			// inside the expression. PatternVisitor records both as reads.
+			return nil, 0
+
+		default:
+			// TypeScript assertion and satisfies wrappers expose their type syntax
+			// to PatternVisitor. Preserve that traversal for otherwise valid
+			// assignment targets, but do not emulate recovery-only expression
+			// targets whose emitted JavaScript is not executable.
+			if (!ast.IsPartOfTypeNode(current) && !ast.IsPartOfTypeNode(parent)) ||
```
elecmonkey, line 438: [P2] This fallback misses the runtime expression side of a TS instantiation expression used inside a destructuring pattern. For `[foo<string>] = arr;`, TypeScript 5.9.3 erases the target to `[(foo)] = arr;`, which is valid JavaScript; ESLint 10.8.1 with `@typescript-eslint/parser` 8.65.0 reports one `globalVariableLeak`, while Rslint reports nothing. Treat the `ExpressionWithTypeArguments.Expression` child as a transparent pattern target so this remains outside the documented recovery-only divergence.

internal/rules/no_eval/no_eval.go
```diff
@@ -373,12 +373,12 @@ func isThisReferringToGlobal(thisNode *ast.Node, sourceFile *ast.SourceFile) boo
 	switch container.Kind {
 	case ast.KindSourceFile:
 		// Top level of script — 'this' is always global (even in strict mode).
-		// In modules, 'this' is undefined.
-		return !ast.IsExternalModule(sourceFile)
+		// In modules it is undefined, and in CommonJS it is module.exports.
+		return sourceType == "script"
```
elecmonkey: [P2] Treat CommonJS top-level `this` like ESLint no-eval

For `sourceType: "commonjs"` or the default `.cjs` source goal, ESLint 10.8.1 reports `this.eval("x")` and `() => this.eval("x")` as `no-eval` because its `Program` handler treats every non-module top-level scope as `isTopLevelOfScript`. This new check returns true only for `sourceType == "script"`, so the PR build reports no diagnostic for those CommonJS cases. This is a false negative in the existing `no-eval` rule introduced by this shared strictness/source-type change.

internal/rules/no_implicit_globals/no_implicit_globals.go
```diff
@@ -0,0 +1,425 @@
...
+
+		case ast.KindNonNullExpression, ast.KindAsExpression, ast.KindTypeAssertionExpression:
+			wrappers++
+			current = parent
+
+		case ast.KindSatisfiesExpression:
+			satisfies = true
+			current = parent
+
+		case ast.KindParenthesizedExpression:
```
elecmonkey, line 404: [P2] Stop reporting parenthesized destructuring targets

The unconditional parenthesis climb still reports recovered destructuring targets such as `([foo]) = arr;`, `(({foo})) = obj;`, and `(([foo, bar])) = arr;`. ESLint 10.8.1 with `@typescript-eslint/parser` emits no `no-implicit-globals` diagnostics for these parser-accepted but runtime-invalid assignment targets, and the PR docs say rslint intentionally does not emulate those recovery-only diagnostics. This branch makes those cases false positives by reaching the inner array/object pattern and then the outer assignment.

internal/rule/ref_store.go
swwind: Fixed in 70029565e. The two file-scope `interface Promise` and `type Promise` cases now explicitly use `SourceType: "script"`, preserving their intended global-script shadowing semantics after the default `.ts` source goal became module. `go test ./internal/rules/no_promise_executor_return -count=1` passes.

internal/rules/no_implicit_globals/no_implicit_globals.go
swwind: Fixed in 70029565e. `findPureAssignmentRoot` now treats the runtime `Expression` child of `ExpressionWithTypeArguments` as transparent inside a real assignment pattern, while preserving the existing traversal of type arguments. A focused `[foo<string>] = arr` regression case now reports `globalVariableLeak`, and `go test ./internal/rules/no_implicit_globals -count=1` passes.

internal/rules/no_eval/no_eval.go
swwind: Fixed in 70029565e. The `no-eval` top-level `this` check now follows ESLint's Program-scope behavior and accepts every non-module source goal, including CommonJS. Focused cases cover both explicit `sourceType: "commonjs"` and an arrow inheriting top-level `this` in a `.cjs` file. `go test ./internal/rules/no_eval -count=1` passes.

internal/rules/no_implicit_globals/no_implicit_globals.go
swwind: Fixed in 70029565e. The assignment-target walk now stops when parentheses wrap the complete array or object destructuring container, while parentheses around a bare identifier remain transparent. Focused valid cases cover `([foo]) = arr`, `(({foo})) = obj`, and `(([foo, bar])) = arr`. `go test ./internal/rules/no_implicit_globals -count=1` passes.

elecmonkey (dismissed):
no more issues

elecmonkey:
merge conflict~~~ plz resolve~

elecmonkey (dismissed):
LGTM

elecmonkey (approved):

PR #1761, opened 2026-08-16
feat: port rule no-restricted-exports

swwind:
## Summary

Port the `no-restricted-exports` rule from ESLint to rslint.

This rule disallows specified names from being used as exported names, with options to restrict named exports by exact name (`restrictedNamedExports`) or regex pattern (`restrictedNamedExportsPattern`), and to restrict various forms of `export default` (`restrictDefaultExports`).

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/no-restricted-exports
- Source code: https://github.com/eslint/eslint/blob/v10.8.1/lib/rules/no-restricted-exports.js

## Checklist


swwind:
Blocked by #1705

internal/rules/no_restricted_exports/no_restricted_exports.schema.json
```diff
@@ -0,0 +1,51 @@
...
+          "type": "object",
+          "properties": {
+            "restrictedNamedExports": {
+              "type": "array",
+              "items": {
+                "type": "string"
+              },
+              "uniqueItems": true
+            },
+            "restrictedNamedExportsPattern": { "type": "string" }
```
elecmonkey: **[P2] Reject invalid restricted-name patterns**

When `restrictedNamedExportsPattern` is an invalid JavaScript regex such as `"["`, this schema accepts it and `parseOptions` silently discards the compilation error, so `export const a = 1;` produces no diagnostic and linting succeeds. ESLint 10.8.1 instead throws `Invalid regular expression` when the rule encounters an exported name; add `"format": "regex"` here, as the other regex-valued rule schemas do, so invalid configurations fail rather than silently disabling the restriction.

swwind: Fixed in `fef4d8897`. Both schema branches now mark `restrictedNamedExportsPattern` with `"format": "regex"`, so invalid ECMAScript patterns are rejected during config validation instead of being silently ignored.

internal/rules/no_restricted_exports/no_restricted_exports.schema.json
```diff
@@ -0,0 +1,57 @@
...
+            "restrictedNamedExports": {
+              "type": "array",
+              "items": {
+                "type": "string"
+              },
+              "uniqueItems": true
+            },
+            "restrictedNamedExportsPattern": {
+              "type": "string",
+              "format": "regex"
```
elecmonkey, line 18: Validate restricted export patterns in Unicode mode

When the pattern is valid without `u` but invalid with `u` (for example, `"\\a"`), this `format` check succeeds because the schema regex engine compiles values without flags, while `parseOptions` later compiles with `"u"` and silently discards the error. For `export const a = 1;`, ESLint 10.8.1 throws `Invalid regular expression`, but Rslint accepts the configuration and emits no diagnostic; validate this option with the same Unicode semantics or propagate the runtime compilation error.

elecmonkey (commented):
Reviewed one issue.

internal/rules/no_restricted_exports/no_restricted_exports.schema.json
swwind: This is not possible to fix here.

elecmonkey (approved):
✅

PR #1766, opened 2026-08-17
refactor(utils): port JavaScript's string utility functions on Unicode 17

swwind:
## Summary

Ports JavaScript's string utility functions into `internal/utils/ecmascript`, reads them on Unicode 17 rather than the toolchain's Unicode 15, and routes every character question a rule asks through the packages that answer it the way JavaScript does.

`StringToUpperCase`, `StringToLowerCase` and the two locale variants apply Unicode's full case mapping rather than the single-character one `unicode.ToUpper` stops at: `ß` uppercases to `SS`, `ﬁ` to `FI`, `İ` lowercases to `i` plus a combining dot, and a capital sigma lowercases to `ς` when it ends a word. The locale pair answers with the locale-insensitive mapping, since reading the host locale would make diagnostics depend on the machine.

The edition gap gets a package of its own. Go 1.26 is on Unicode 15.0 and Node 26 on 17.0, so `internal/utils/unicode17` carries the difference: the case mappings the two editions added, the three pairs they gave a simple folding onto one another where before only their full folding said anything, the `Cased` and `Case_Ignorable` characters that decide where a sigma is final, and the `Lu`, `Ll`, `L` and `M` categories. It deletes itself — `TestDeltaStillNeeded` fails the moment the toolchain leaves 15.0, which on Go 1.27 is the signal to drop the package rather than maintain it.

`Canonicalize`, `CaseEquivalents` and `CaseEquivalenceGroups` move to `internal/utils/ecmascript/regexp`, the only caller, which now has no dependency on its parent.

`IsWhiteSpace` splits in two. ECMAScript has two productions — WhiteSpace, which is a tab, a vertical tab, a form feed, U+FEFF and a Space_Separator, and LineTerminator, which is the other four — and the old predicate answered for the pair, so a rule that wanted only the first had to spell it `IsWhiteSpace(r) && !IsLineTerminator(r)`. The pair is now `IsWhiteSpaceOrLineTerminator`, which `\s` and `String.prototype.trim` read, and `IsWhiteSpace` is the production it is named after.

In the rules, close to 200 calls that asked Go's standard library a JavaScript question now ask the port: `unicode.ToUpper` / `IsUpper` where upstream writes `c === c.toUpperCase()`, `unicode.Is(unicode.M, …)` where upstream reads `\p{M}`, `strings.ToLower` / `ToUpper` / `TrimSpace` throughout, and the identifier checks that belong to tsgo's scanner. Two guards hold the line under `internal/rules/**`, `internal/plugins/**` and `internal/utils/**` — a helper a rule calls answers the rule's question, with the ports themselves excepted: `depguard` denies the standard library's `unicode`, `forbidigo` denies the `strings` calls.

`strings.EqualFold` is one of those calls, and it is neither comparison JavaScript makes: it folds U+017F ſ onto `s` in both directions, and never folds U+0131 ı onto `i` although uppercasing does. Its call sites move to `ecmascript.EqualsWhenUppercased` where jsx-ast-utils and the rules built on it write `toUpperCase`, and to `EqualsWhenLowercased` where upstream writes `toLowerCase`.

`utils.NaturalCompare` becomes a port of natural-compare, the package `member-ordering` hands its names to, rather than an approximation of it. That package gives each printable-ASCII character a rank of its own rather than comparing code units, walks UTF-16 code units, does not read a digit run starting with `0` as a number, reads the ones it does as JavaScript numbers, and reads a NUL as the end of the string.

## Behavior changes

- `utils.IsConstructorName` calls `٠Foo` and `中Foo` constructors, as ESLint does: a character with no case of its own passes `c === c.toUpperCase()`.
- `explicit-member-accessibility`'s `requiresQuoting` answers no for `a·b` — U+00B7 has `Other_ID_Continue`, which is what `ts.isIdentifierPart` reads — and yes for a name holding a supplementary-plane character, since upstream indexes UTF-16 code units.
- `react/self-closing-comp` treats `<_foo />` and `<中文 />` as components: upstream tests `/^[a-z]/`, not the first character's case.
- `react/jsx-curly-brace-presence` counts U+2028 and U+2029 as `\s`.
- `member-ordering`'s `natural` and `natural-case-insensitive` orders answer as natural-compare does: `_` sorts before `B` rather than after `Z`, `a01` before `a1`, `a007` reads as seven, a character outside the basic plane sorts by its leading surrogate, and two runs too large to tell apart as JavaScript numbers sort alike.
- A name a `jsx-a11y` rule looks up is matched by uppercasing it, as jsx-ast-utils does, so `<div tabındex="1" />` carries a `tabIndex`. Where upstream lowercases instead — `role-has-required-aria-props`, `media-has-caption`, `react/jsx-no-target-blank`, the two `valid-title` rules — U+017F ſ no longer matches `s`.

## Verification

Every code point run through `toUpperCase`, `toLowerCase` and four Final_Sigma contexts against Node 26: no differences, where the tsgo-backed version this replaces had 820.

Every code point checked against Unicode 17.0's database for its `Lu`, `Ll`, `L` and `M` category and its `Cased` and `Case_Ignorable` property, and grouped by `Canonicalize` under both readings against the partition CaseFolding.txt defines by simple case folding and the one `toUpperCase` gives in Node 26: no differences.

`NaturalCompare` compared against natural-compare 1.4.0 over two corpora — 9,855,625 pairs covering digits of other scripts, punctuation, the supplementary plane, leading zeros, runs past 2^53 and NUL: no differences, where a tenth of the pairs disagreed before.

## Related Links

Stacked on #1705.

## Checklist


fansenze (dismissed):

internal/utils/unicode17/unicode17.go
```diff
@@ -0,0 +1,372 @@
...
+
+// Fold returns the other member of the pair r folds together with under
+// Unicode 16 and 17, reporting false when Go's own tables already hold the
+// answer. Each pair is a lowercase character and its uppercase, which simple
+// case folding puts in one orbit.
+func Fold(r rune) (rune, bool) {
+	if upper, ok := ToUpper(r); ok {
+		return upper, true
+	}
+	return ToLower(r)
```
elecmonkey, line 85: **[P2] Add the simple-fold-only Unicode 17 pairs**

When a user regexp has both `i` and `u`/`v`, Unicode 17 has three new simple-fold pairs with no simple upper/lower mapping: U+1FD3↔U+0390, U+1FE3↔U+03B0, and U+FB05↔U+FB06. `Fold` only delegates to `ToUpper`/`ToLower`, so these return `ok == false`; consequently `Canonicalize` and the case tables leave each pair separate, making patterns such as `/\u1FD3/iu` fail to match U+0390 even though Node does. Carry these simple-fold mappings separately and include their members when building the case tables.

swwind: Fixed in a482eb7b.

You are right. The delta between Unicode 15.0 and 17.0 in CaseFolding.txt's C and S entries is exactly the pairs `casePairs` and `caseRuns` already carried, plus three S entries — `1FD3; S; 0390`, `1FE3; S; 03B0` and `FB05; S; FB06`. Each of those three had only a full folding onto a sequence of characters before, and neither member of a pair is the other's case, so `Fold` could not reach them through `ToUpper` and `ToLower` and `buildCaseTables` could not reach them through `unicode.CaseRanges` either.

The package now carries them in a `foldPairs` table that `Fold` consults first, with a `FoldAdditions` beside `CaseAdditions` so the case tables record their members. `TestDeltaStillNeeded` checks that `unicode.SimpleFold` still leaves each pair apart, so these entries retire along with the rest of the delta.

Verification: grouping every code point by `Canonicalize(r, true)` and comparing the resulting partition against the one Unicode 17.0's CaseFolding.txt defines by simple case folding gives no differences.

fansenze (approved):

PR #1770, opened 2026-08-17
feat: honor /* exported */ in no-unused-vars, prefer-const and no-useless-assignment

swwind:
## Summary

Consumes the inline `/* exported */` API added in #1760 in the four remaining rules ESLint applies the directive to: `no-unused-vars` (core and `@typescript-eslint`), `prefer-const`, and `no-useless-assignment`.

`RuleContext.IsExportedGlobalBinding(decl, name)` performs the lookup ESLint does before marking a variable: the name is listed by a comment, the file's top level is the global scope, and `decl` binds the name in that scope. A `var` reaches the global scope from inside any block, while `let`/`const`/`class`, a function declaration, and a catch parameter belong to their nearest block scope.

Each rule then reads the flag the way its upstream counterpart does:

- `no-unused-vars`, core and TypeScript, counts the directive as a use. `reportUsedIgnorePattern` still sees the binding as used, which is what turns a directive on an ignored name into a `usedIgnoredVar` report. The TypeScript scope manager puts type-only declarations in the same global scope as value ones, so the directive also reaches `interface`, `type`, `enum`, `namespace`, and `import x = require()`.
- `prefer-const` declines to judge such a global, since a separately loaded file may reassign it. Under `destructuring: "all"` that holds the whole destructuring group back.
- `no-useless-assignment` treats it like an `export`: the value can still be read from outside the file, so no assignment to it is reported.

Six `/* exported */` skips are removed from the core `no-unused-vars` upstream suite and one from `no-useless-assignment`'s.

## Checklist


elecmonkey:
The shared scope-gating helper and all four rule integrations match ESLint's exported-directive semantics across global, block, function, module, CommonJS, destructuring, and TypeScript-only declaration shapes. No actionable correctness, regression, performance, or maintainability issue was found.

fansenze (approved):

internal/rule/exported.go
```diff
@@ -38,3 +45,55 @@ func (e Exported) Has(name string) bool {
...
+func bindingScopeContainer(decl *ast.Node) *ast.Node {
+	root := ast.GetRootDeclaration(decl)
+	if root == nil {
+		root = decl
+	}
+	if root.Kind == ast.KindVariableDeclaration && root.Parent != nil &&
+		root.Parent.Kind == ast.KindVariableDeclarationList && utils.IsVarKeyword(root.Parent) {
+		return utils.FindEnclosingScope(root)
+	}
+	return ast.GetEnclosingBlockScopeContainer(root)
```
elecmonkey, line 98: [P2] Restrict directives to the actual global binding

When a directive name collides with a nested TypeScript binding, `bindingScopeContainer` incorrectly returns the source file for declarations whose scopes are not represented by `GetEnclosingBlockScopeContainer`, including class/interface/type-alias type parameters, mapped/infer parameters, and enum members. For `/* exported T */ class C<T> {}
void C;`, ESLint 10.7.0 with `@typescript-eslint/parser` 8.67.0 and `@typescript-eslint/no-unused-vars` 8.67.0 both report `T` at 1:26, while both Rslint implementations emit nothing; with `_T`, `varsIgnorePattern: '^_'`, and `reportUsedIgnorePattern: true`, Rslint additionally emits a false `usedIgnoredVar`. Verify that `decl` belongs to the actual global-scope variable by scope or symbol identity instead of relying only on its nearest ts-go block-scope container.

internal/plugins/typescript/rules/no_unused_vars/no_unused_vars_typescript_test.go
```diff
@@ -517,3 +518,47 @@ export interface Constructable {
 
 	rule_tester.RunRuleTester(fixtures.GetRootDir(), "tsconfig.json", t, &NoUnusedVarsRule, validTestCases, invalidTestCases)
 }
+
+// TestNoUnusedVarsExportedDirective covers the `/* exported */` comment, which
+// marks a global-scope binding used for a separately loaded file's sake. The
+// TypeScript scope manager puts type-only declarations in the same global scope
+// as value ones, so the directive reaches every declaration form a script-mode
+// file can bind at its top level.
+func TestNoUnusedVarsExportedDirective(t *testing.T) {
```
elecmonkey, line 527: [P2] Unskip the TypeScript upstream exported cases

The new TypeScript smoke suite does not re-enable the six exact upstream `/* exported */` cases: `no-unused-vars-eslint-basic.test.ts` still marks the two invalid and four valid cases as `skip: true` with the obsolete ‘script-mode code incompatible’ reason. Those cases from typescript-eslint 8.67.0 include partial declarations and destructuring assertions that are not all preserved here. Set `languageOptions.sourceType` to `script`, remove the skips, and refresh their snapshots so the upstream migration remains complete.

PR #1775, opened 2026-08-17
feat(rstest): add no-import-node-test rule

elecmonkey:
## Summary

Report imports of Node's test runner — `import` declarations and `require()` calls alike, `node:test` and its sub-paths — which editors frequently insert by accident when auto-importing `describe` / `test` / `it`. Named imports of `node:test` containing only `test`, `it` and `describe` are fixed by rewriting the module to `@rstest/core`, preserving the original quote style; every other form is reported without a fix, because the two modules do not have equivalent export surfaces. Added to the rstest `recommended` preset at `error`.

## Credits

Port from [`eslint-plugin-vitest`](https://github.com/vitest-dev/eslint-plugin-vitest)'s `no-import-node-test`.

- **Every form that reaches Node's test runner is reported, not only a bare `node:test` import.** Upstream matches an `ImportDeclaration` whose specifier is exactly `node:test`. Sub-paths such as `node:test/reporters`, and `require('node:test')`, pull in the same runner and are reported too. Neither is fixed, so the wider reporting introduces no new rewrites.

- **The fix is withheld unless the rewrite is provably safe.** Upstream rewrites the specifier for any import shape, which can turn working code into code that does not resolve: the two modules do not have equivalent export surfaces, so a default import, a namespace import, a side-effect import, or a named import of something like `mock` has nothing to be rewritten to. Only a declaration whose bindings are all named imports drawn from `test`, `it` and `describe` is fixed. The reported set matches upstream's; only the fixed set narrows.

- **The message states the prohibition instead of prescribing a replacement module.** Upstream's text tells the reader to import from `vitest` instead. This rule reports forms that have no `@rstest/core` counterpart — sub-paths, `require()`, and imports of bindings the two modules do not share — so naming one replacement in the message would be advice the reader often cannot act on. The message is `Do not import the Node test runner in Rstest test files`, and the rewrite to `@rstest/core` is offered as a fix wherever one is safe.

## Related Links

Related #935

## Checklist

### DIFF
--- a/internal/plugins/rstest/rules/no_import_node_test/no_import_node_test_test.go
+++ b/internal/plugins/rstest/rules/no_import_node_test/no_import_node_test_test.go
@@ -20,6 +20,7 @@ func TestNoImportNodeTestRule(t *testing.T) {
 			{Code: `import { test } from 'node:testing'`},
 			{Code: `const assert = require('node:assert')`},
 			{Code: `const t = import('node:test')`},
+			{Code: `import assert = require('node:assert')`},
 		},
 		[]rule_tester.InvalidTestCase{
 			{
@@ -87,6 +88,22 @@ func TestNoImportNodeTestRule(t *testing.T) {
 			},
 			{
 				Code:   `const { tap } = require('node:test/reporters')`,
+				Errors: []rule_tester.InvalidTestCaseError{{MessageId: "noImportNodeTest"}},
+			},
+			{
+				Code:   `import nodeTest = require('node:test')`,
+				Errors: []rule_tester.InvalidTestCaseError{{MessageId: "noImportNodeTest"}},
+			},
+			{
+				Code:   `import reporters = require('node:test/reporters')`,
+				Errors: []rule_tester.InvalidTestCaseError{{MessageId: "noImportNodeTest"}},
+			},
+			{
+				Code:   `const nodeTest = (require)('node:test')`,
+				Errors: []rule_tester.InvalidTestCaseError{{MessageId: "noImportNodeTest"}},
+			},
+			{
+				Code:   `const nodeTest = require(('node:test'))`,
 				Errors: []rule_tester.InvalidTestCaseError{{MessageId: "noImportNodeTest"}},
 			},
 			{
--- a/internal/plugins/rstest/rules/no_import_node_test/no_import_node_test_test.go
+++ b/internal/plugins/rstest/rules/no_import_node_test/no_import_node_test_test.go
@@ -17,8 +17,9 @@ func TestNoImportNodeTestRule(t *testing.T) {
 		[]rule_tester.ValidTestCase{
 			{Code: `import { test } from '@rstest/core'`},
 			{Code: `import { test } from 'vitest'`},
-			{Code: `import { test } from 'node:test/reporters'`},
-			{Code: `const t = require('node:test')`},
+			{Code: `import { test } from 'node:testing'`},
+			{Code: `const assert = require('node:assert')`},
+			{Code: `const t = import('node:test')`},
 		},
 		[]rule_tester.InvalidTestCase{
 			{
@@ -26,7 +27,7 @@ func TestNoImportNodeTestRule(t *testing.T) {
 				Output: []string{`import { test } from '@rstest/core'`},
 				Errors: []rule_tester.InvalidTestCaseError{{
 					MessageId: "noImportNodeTest",
-					Message:   "Do not import from `node:test` in Rstest test files",
+					Message:   "Do not import the Node test runner in Rstest test files",
 					Line:      1,
 					Column:    1,
 					EndLine:   1,
@@ -70,6 +71,22 @@ func TestNoImportNodeTestRule(t *testing.T) {
 			},
 			{
 				Code:   `import nodeTest, { test } from 'node:test'`,
+				Errors: []rule_tester.InvalidTestCaseError{{MessageId: "noImportNodeTest"}},
+			},
+			{
+				Code:   `import { tap } from 'node:test/reporters'`,
+				Errors: []rule_tester.InvalidTestCaseError{{MessageId: "noImportNodeTest"}},
+			},
+			{
+				Code:   `import { test } from 'node:test/reporters'`,
+				Errors: []rule_tester.InvalidTestCaseError{{MessageId: "noImportNodeTest"}},
+			},
+			{
+				Code:   `const nodeTest = require('node:test')`,
+				Errors: []rule_tester.InvalidTestCaseError{{MessageId: "noImportNodeTest"}},
+			},
+			{
+				Code:   `const { tap } = require('node:test/reporters')`,
 				Errors: []rule_tester.InvalidTestCaseError{{MessageId: "noImportNodeTest"}},
 			},
 			{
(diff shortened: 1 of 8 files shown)


internal/plugins/rstest/rules/no_import_node_test/rule.go
```diff
@@ -0,0 +1,139 @@
...
+					rule.RuleFixReplaceRange(
+						internalUtils.TrimNodeTextRange(ctx.SourceFile, declaration.ModuleSpecifier),
+						replacement,
+					),
+				)
+			},
+			// `require('node:test')` is reported as well, but never fixed: the
+			// binding forms a require can take do not map onto a specifier
+			// rewrite the way a static import does.
+			ast.KindCallExpression: func(node *ast.Node) {
```
elecmonkey, line 151: [P2] Handle TypeScript import-equals declarations

When TypeScript uses `import nodeTest = require('node:test')`, ts-go emits a `KindImportEqualsDeclaration`, not a `KindCallExpression`, so this listener never runs and the Node test runner import is silently allowed. This is a static runner import covered by the rule's stated import/require scope, so add a listener for external import-equals references.

internal/plugins/rstest/rules/no_import_node_test/rule.go
```diff
@@ -0,0 +1,139 @@
...
+				callee := node.AsCallExpression().Expression
```
elecmonkey: [P2] Unwrap parentheses in require detection

Parentheses are transparent in JavaScript, but the direct kind checks miss equivalent static calls such as `(require)('node:test')` and `require(('node:test'))`. Both execute the same Node runner import yet produce no diagnostic because ts-go preserves the parenthesized AST nodes; unwrap the callee and first argument before checking them.

internal/plugins/rstest/rules/no_import_node_test/rule.go
```diff
@@ -0,0 +1,139 @@
...
+				if callee.Kind != ast.KindIdentifier || callee.AsIdentifier().Text != "require" {
```
elecmonkey: [P2] Exclude shadowed require identifiers

Whenever a local binding named `require` is invoked with this string—for example `function f(require) { require('node:test') }`—the rule reports it even though no Node module is imported. Because the rule is enabled as an error in the recommended preset, ordinary user-defined functions with this name become false positives; verify that the identifier is not locally shadowed before treating it as CommonJS `require`.

elecmonkey (commented):
The rule misses valid Node test runner imports expressed through TypeScript import-equals syntax or parenthesized require expressions, and it reports locally shadowed functions as CommonJS imports. These correctness gaps affect a rule enabled by the recommended preset.

internal/plugins/rstest/rules/no_import_node_test/rule.go
elecmonkey: Fixed in f562759b.

internal/plugins/rstest/rules/no_import_node_test/rule.go
elecmonkey: Fixed in f562759b.

internal/plugins/rstest/rules/no_import_node_test/rule.go
elecmonkey: Not changing this here: no rule in the repo checks whether `require` is shadowed — `typescript/no-require-imports`, `typescript/no-var-requires`, `unicorn/prefer-node-protocol` and `import`'s shared module visitor all match on the identifier text alone, so this is worth solving once in a shared helper rather than in this rule.

internal/plugins/rstest/rules/no_import_node_test/rule.go
```diff
@@ -0,0 +1,161 @@
...
+
+var NoImportNodeTestRule = rule.Rule{
+	Name:   "rstest/no-import-node-test",
+	Schema: rule.EmptyArraySchema,
+	Run: func(ctx rule.RuleContext, _ []any) rule.RuleListeners {
+		return rule.RuleListeners{
+			ast.KindImportDeclaration: func(node *ast.Node) {
+				declaration := node.AsImportDeclaration()
+				if declaration == nil || declaration.ModuleSpecifier == nil ||
+					!isNodeTestModule(declaration.ModuleSpecifier.Text()) {
```
swwind: [P1] Guard the module specifier before reading its text

`parseModuleSpecifier` accepts arbitrary expressions ("We allow arbitrary expressions here, even though the grammar only allows string literals. We check to ensure that it is only a string literal later in the grammar check pass"), and `ast.Node.Text()` panics on kinds it does not handle. Recoverable-but-invalid source such as `import x from a.b;` therefore takes down the whole lint run:

```
panic: Unhandled case in Node.Text: *ast.PropertyAccessExpression
	internal/ast/ast.go:307
	internal/plugins/rstest/rules/no_import_node_test/rule.go:97
```

Reproduced by adding ``{Code: `import x from a.b;`}`` as a valid case. Line 101's second `Text()` call needs the same protection.

The other two listeners in this file already call `isStringNode` first, and elsewhere in the repo `import/no-webpack-loader-syntax` early-returns on `specifier.Kind != ast.KindStringLiteral` while `unicorn/prefer-node-protocol` goes through `ast.GetExternalModuleName`. An `isStringNode(declaration.ModuleSpecifier)` check here would match.

swwind (commented):
Only one blocking issue left; the rest of the rule checks out.

| Code | ESLint | This PR |
| --- | --- | --- |
| `import x from 'node:test'` | report | report + fix |
| `import x from a.b;` | no report | panic |

Verified as fine: the fixer's safe-name check (`PropertyName` is used, so `import { test as mock }` is fixed and `import { mock }` is not), quote preservation, sub-path handling, and the `all.go` / `presets/rstest.ts` / `presets.test.ts` / `rstest.config.mts` / doc-path wiring.

`import('node:test')` and `export { test } from 'node:test'` going unreported reads as a deliberate scope choice rather than a gap, so I left it alone.

internal/plugins/rstest/rules/no_import_node_test/rule.go
elecmonkey: Fixed in 6b3b40e7. The `ImportDeclaration` listener now narrows the module specifier with `isStringNode` before reading it, matching the other two listeners in the file, and the specifier text is read once into a local instead of twice.

No regression case added.

internal/plugins/rstest/rules/no_import_node_test/no_import_node_test_test.go
```diff
@@ -0,0 +1,116 @@
...
+		"tsconfig.json",
+		t,
+		&no_import_node.NoImportNodeTestRule,
+		[]rule_tester.ValidTestCase{
+			{Code: `import { test } from '@rstest/core'`},
+			{Code: `import { test } from 'vitest'`},
+			{Code: `import { test } from 'node:testing'`},
+			{Code: `const assert = require('node:assert')`},
+			{Code: `const t = import('node:test')`},
+			{Code: `import assert = require('node:assert')`},
```
swwind, line 23: [P3] Lock the guard with a regression case

`import x from a.b;` produces no syntactic error: the parser accepts an arbitrary expression as a module specifier, and the "String literal expected" diagnostic only comes from the later grammar check, which linting does not gate on. So it reaches the listener, and before 6b3b40e7 it panicked the whole run rather than producing a diagnostic on that file. `import x from (a);` and `import x from f();` panic the same way.

It passes cleanly as a valid case:

```suggestion
			{Code: `import assert = require('node:assert')`},
			{Code: `import x from a.b;`},
```

swwind (commented):
Re-checked at 6b3b40e7 — the guard is correct.

| Code | ESLint | This PR |
| --- | --- | --- |
| `import x from a.b;` | no report | no report |

Reverting rule.go to f562759b still panics on that input, so the guard is doing the work. One suggestion inline.

swwind (approved):

PR #1776, opened 2026-08-17
feat(rstest): add prefer-called-exactly-once-with rule

elecmonkey:
## Summary

Report a `toHaveBeenCalledOnce` and a `toHaveBeenCalledWith` on the same target within one block, and merge them into `toHaveBeenCalledExactlyOnceWith`. `toHaveBeenCalledWith` passes when any recorded call matches, so the pair states "called exactly once, with these arguments" across two statements that can drift apart; the combined matcher states it in one and cannot silently weaken if the call-count half is later removed. All three matchers ship with the `@vitest/expect` version Rstest depends on.

Expect recognition goes through the shared `ParseExpectCall`, so every Rstest expect source — global, import, alias, namespace, `import.meta.rstest`, and the test-context `expect` — is covered without re-deriving provenance, and an `expect` imported from another framework is left alone. Added to the rstest `recommended` preset at `error`, matching upstream.

## Credits

Port from [`eslint-plugin-vitest`](https://github.com/vitest-dev/eslint-plugin-vitest)'s `prefer-called-exactly-once-with`.

- **Chai-style spellings are recognised.** Rstest installs `@vitest/expect`'s `ChaiStyleAssertions` alongside `JestChaiExpect`, registering `calledOnce`, `calledWith` and `calledOnceWith` as delegates to the jest matchers on the same Chai `Assertion` prototype. Both spellings assert on the same spy, so they pair with each other and with the jest spellings; the merged assertion keeps the spelling of the surviving argument-side call.

- **Assertions pair on the assertion factory call, its modifiers and whether it is awaited**, rather than on the source text preceding the matcher. That text happens to carry the modifiers, but it also carries Chai's language chains, which assert nothing and would keep `expect(x).to.have.been.calledOnce` from pairing with `expect(x).calledWith('a')`. Composing the key from the parse keeps what changes meaning and drops the sugar. It also adds `await`, which upstream never looks through: an unawaited `resolves` assertion does not throw where it is written, so deleting an awaited half and keeping an unawaited one would leave a promise floating.

- **Only inert statements may sit between the two assertions.** The merge claims both halves describe one call history, so nothing in between may call the target or rebind it — and no enumeration of dangerous statements can be closed, since an arbitrary call reaches anywhere. Instead of listing what is dangerous, the rule lists what runs nothing of the author's: an assertion whose `expect(...)` and matcher calls are its only calls, where no matcher executes what it asserts on, no modifier awaits it, and the arguments hold no other call, no `await` and no assignment; a call into TypeScript's default library such as `console.log('checkpoint')` or `JSON.parse(text)`, provided nothing callable is handed to it, so `items.forEach(cb)` and a `console` the author declared themselves are excluded; and a declaration whose initializers meet the same bar, where a hoisted `var` must also declare names neither assertion reads. Assertions on other mocks therefore keep their place between the two halves, while a mock reset, a reassignment such as `x = createMock()`, a further call to the target, and any call the rule cannot resolve leave the pair alone. Resolving a library call needs type information, so with no type checker every call blocks the merge. Upstream asks only whether a `mockClear` on a plain identifier sits between them, and merges across everything else.

- **The fix rewrites the surviving matcher accessor in place** instead of rebuilding the call from its parts, so arguments, type arguments, comments and formatting survive unchanged. The folded statement is removed together with its line only when it owns that line; upstream always deletes whole lines and can therefore swallow a neighbouring statement or a trailing comment.

- **A chain that asserts more than once is reported without a fix.** `expect(x).to.have.been.calledOnce.and.to.be.ok` states the call-count half plus something else; folding the statement away would delete that other assertion, and rewriting the chain in place would have to splice a segment out of it.

- **The message names the matchers it found**, so a Chai chain is not described with jest-style names. For the jest spellings the text is unchanged.

## Related Links

Related #935

## Checklist


internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
```diff
@@ -0,0 +1,517 @@
...
+		return nil
+	}
+	targetText, ok := nodeText(sourceFile, arguments[0])
+	if !ok {
+		return nil
+	}
+
+	candidate.statement = statement
+	candidate.position = internalUtils.TrimNodeTextRange(sourceFile, statement).Pos()
+	candidate.pairKey = pairKey(headText, parsed.Modifiers, awaited)
```
elecmonkey, line 261: **[P1] Avoid fixing repeated side-effectful expect arguments**

When the repeated factory argument is an expression such as `getMock()`, textual equality of `headText` does not establish that both evaluations have the same target or are side-effect free. The current fix turns two evaluations into one—for example, it changes a counter increment from twice to once—and can also merge assertions on different mocks if the call returns a different value each time. Restrict autofixing to stable target expressions or otherwise preserve both evaluations.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
```diff
@@ -0,0 +1,517 @@
...
+		minPosition, maxPosition = maxPosition, minPosition
+	}
+
+	found := false
+	var visit func(node *ast.Node) bool
+	visit = func(node *ast.Node) bool {
+		if node == nil {
+			return false
+		}
+		if receiver, ok := mockResetReceiver(node); ok {
```
elecmonkey: **[P1] Normalize reset receivers before matching them**

For an equivalent receiver written with comments or parentheses, such as `obj /* keep */ .fn.mockClear()` between two `expect(obj.fn)` assertions, `nodeText(receiver)` differs from `targetText`, so the reset is missed and the rule offers a behavior-changing merge across two call histories. Compare the receiver and target structurally, or at least normalize transparent parentheses and trivia before comparing.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
```diff
@@ -0,0 +1,517 @@
...
+	}
+	rootCall, awaited := assertionRootCall(expressionStatement.Expression)
+	if rootCall == nil {
+		return nil
+	}
+	parsed := analysis.ParseExpectCall(rootCall)
+	if parsed == nil {
+		return nil
+	}
+	// `not` never merges, even on both assertions: "not called once" and "not
```
elecmonkey, line 204: **[P2] Reject negation after the first Chai matcher**

When `not` occurs after an earlier Chai assertion, as in `expect(x).to.have.been.calledOnce.and.not.calledWith('a')`, `parsed.Modifiers` is empty because it only contains modifiers preceding the first matcher, while `parsed.Matchers` still includes both matchers. The rule consequently recommends the positive `calledOnceWith`, although the original can pass for a once-called spy with different arguments and the recommendation fails. Post-first-matcher modifiers need to disqualify or semantically qualify the affected hit.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
```diff
@@ -0,0 +1,517 @@
...
+		if node == nil {
+			return false
+		}
+		if receiver, ok := mockResetReceiver(node); ok {
+			if text, textOK := nodeText(sourceFile, receiver); textOK && text == first.targetText {
+				found = true
+				return true
+			}
+		}
+		return node.ForEachChild(visit)
```
elecmonkey, line 602: **[P2] Do not treat deferred reset bodies as executed**

The unrestricted recursive walk treats a reset inside a merely declared callback or function as if it occurred between the assertions; for example, `expect(x)...Once(); const clearLater = () => x.mockClear(); expect(x)...With('a');` produces no diagnostic even though `clearLater` is never invoked and the call history is unchanged. Stop traversal at nested function-like scopes unless their body is actually executed at this point.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
```diff
@@ -0,0 +1,517 @@
...
+	}
+	expressionStatement := statement.AsExpressionStatement()
+	if expressionStatement == nil {
+		return nil
+	}
+	rootCall, awaited := assertionRootCall(expressionStatement.Expression)
+	if rootCall == nil {
+		return nil
+	}
+	parsed := analysis.ParseExpectCall(rootCall)
```
elecmonkey, line 200: **[P2] Exclude local expect bindings without a type checker**

When this rule runs on a source-only Program with no `TypeChecker`, `ParseExpectCall` falls back to treating every identifier named `expect` as the Rstest global. Consequently `const expect = makeExpect(); expect(x).toHaveBeenCalledOnce(); expect(x).toHaveBeenCalledWith('a');` is reported and autofixed even though the tests classify this source as non-Rstest. This needs a binder/reference check or another no-checker guard before accepting the global fallback.

elecmonkey (commented):
The rule can emit behavior-changing fixes for side-effectful targets and miss reset barriers written with equivalent syntax. It also produces incorrect recommendations for negated Chai chains and has false-positive/false-negative cases in source-only and nested-function scenarios.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
elecmonkey: Fixed in 897320a0. Autofix now requires every argument of the surviving `expect(...)` call to be stable under a second evaluation — identifiers, property accesses, and element accesses with a literal key; the pair is still reported otherwise.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
elecmonkey: Fixed in 897320a0. The reset receiver and the assertion target are now compared structurally after stripping parentheses and type assertions, instead of by source text.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
elecmonkey: Fixed in 897320a0. The check now scans the whole member chain for `not` instead of only the modifiers preceding the first matcher.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
elecmonkey: Leaving this as is. Pruning every function-like scope is cheaper than the current walk but turns a safe false negative into a false positive: a reset inside an executed callback such as `arr.forEach(() => m.mockClear())` or an IIFE would stop blocking the merge, and the rule would then autofix across a real reset. Deciding whether the body actually runs between the two assertions costs more analysis than the over-blocking it would remove, so the conservative walk stays.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
elecmonkey: Not changing this here: the no-checker fallback lives in the shared `ParseExpectCall` (`internal/plugins/rstest/utils/parse_rstest_expect.go`), where every rstest rule inherits it, so it is worth fixing once there rather than in this rule.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
```diff
@@ -0,0 +1,623 @@
...
+}
+
+// onceMatchers lists the "called exactly one time" matchers, whose assertion
+// the merge folds into the argument-side call and then deletes.
+var onceMatchers = map[string]bool{
+	"toHaveBeenCalledOnce": true,
+	"calledOnce":           true,
+}
+
+var mockResetMethods = map[string]bool{
```
swwind: **[P2] Let the global reset APIs block the merge too**

`clearAllMocks`, `resetAllMocks` and `restoreAllMocks` reset every mock, so they split the two call histories exactly as `x.mockClear()` does. `mockResetReceiver` only recognises a reset whose receiver is the assertion target, so `expect(x).toHaveBeenCalledOnce(); rstest.clearAllMocks(); expect(x).toHaveBeenCalledWith('a');` is reported and autofixed. With further calls to `x` after the reset the original passes and the merged assertion fails. These three need no receiver match — one call to any of them between the pair is enough to block the merge.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
```diff
@@ -0,0 +1,623 @@
...
+	}
+	if len(hits) == 0 {
+		return nil
+	}
+	return hits
+}
+
+// mockResetReceiver returns the object a mockClear / mockReset / mockRestore
+// call resets, or false for any other call.
+func mockResetReceiver(callNode *ast.Node) (*ast.Node, bool) {
```
swwind: **[P2] Look through `mocked()` when matching the reset receiver**

`rstest.mocked` is identity at runtime — `mocked: ((item: any) => item)` in rstest's `runtime/api/utilities.ts` — and only narrows the type, so `rstest.mocked(x).mockClear()` really does reset `x`. The receiver is a call expression, which `sameTargetExpression` rejects, so `expect(x).toHaveBeenCalledOnce(); rstest.mocked(x).mockClear(); expect(x).toHaveBeenCalledWith('a');` is reported and autofixed across a real reset. Unwrapping a `mocked(...)` / `rstest.mocked(...)` receiver before the comparison covers the idiomatic typed-mock spelling.

swwind (commented):
The reset barrier does not cover two spellings that really do reset the target, and both of them still get an autofix.

| Code | eslint-plugin-vitest | This PR |
| --- | --- | --- |
| `expect(x).toHaveBeenCalledOnce(); x.mockClear(); expect(x).toHaveBeenCalledWith('a');` | no report | no report |
| `expect(x).toHaveBeenCalledOnce(); rstest.clearAllMocks(); expect(x).toHaveBeenCalledWith('a');` | report + fix | report + fix |
| `expect(x).toHaveBeenCalledOnce(); rstest.mocked(x).mockClear(); expect(x).toHaveBeenCalledWith('a');` | report + fix | report + fix |
| `expect(getMock()).toHaveBeenCalledOnce(); getMock().mockClear(); expect(getMock()).toHaveBeenCalledWith('a');` | report + fix | report, no fix |

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
```diff
@@ -0,0 +1,623 @@
...
+// histories. The search descends into the statements that sit between the
+// pair, so a reset nested in an if or a loop still blocks the merge; upstream
+// only scans the sibling statements and merges across such a reset.
+func hasMockResetBetween(
+	sourceFile *ast.SourceFile,
+	statements []*ast.Node,
+	first, second *mergeCandidate,
+) bool {
+	if !sameTargetExpression(first.targetNode, second.targetNode) {
+		return false
```
swwind, line 551: **[P3] An unstable target disables the barrier as well as the fix**

`sameTargetExpression` returns false for a call expression, so `hasMockResetBetween` reports no reset for `expect(getMock()).toHaveBeenCalledOnce(); getMock().mockClear(); expect(getMock()).toHaveBeenCalledWith('a');` and the pair is still reported. The pair keys on identical head text, which already assumes both spellings denote the same mock; the barrier should make the same assumption rather than silently not applying. Returning true when the targets are not structurally comparable keeps the report honest at the cost of a false negative.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
elecmonkey: Fixed in e64cf01c. `clearAllMocks`, `resetAllMocks` and `restoreAllMocks` now block the merge on the method name alone, with no receiver match required.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
elecmonkey: Fixed in e64cf01c. A `mocked(...)` / `rstest.mocked(...)` receiver is unwrapped before it is compared with the assertion target.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
elecmonkey: Fixed in e64cf01c, though not by returning true. That would also drop the report for a pair with no reset between it at all, so `expect(getMock()).toHaveBeenCalledOnce(); expect(getMock()).toHaveBeenCalledWith('a');` would stop being reported. Instead, when the target cannot be compared structurally, any reset between the two assertions blocks the merge, whatever its receiver. `getMock().mockClear()` in between is covered, and the pair without a reset is still reported.

swwind (commented):
All three reset-barrier gaps from the last round are fixed — verified locally. Two things left:

| Code | eslint-plugin-vitest | This PR |
| --- | --- | --- |
| `expect(x).toHaveBeenCalledOnce(); rstest.clearAllMocks(); expect(x).toHaveBeenCalledWith('a');` | report + fix | no report |
| `expect(x).toHaveBeenCalledOnce(); rstest.mocked(x).mockClear(); expect(x).toHaveBeenCalledWith('a');` | report + fix | no report |
| `expect(getMock()).toHaveBeenCalledOnce(); getMock().mockClear(); expect(getMock()).toHaveBeenCalledWith('a');` | report + fix | no report |
| `expect(x).toHaveBeenCalledOnce(); x = createMock(); expect(x).toHaveBeenCalledWith('a');` | report + fix | report + fix |

The barrier cases added in e64cf01c only exist in the TS test under `packages/rslint-test-tools/tests/rstest/rules/`; `prefer_called_exactly_once_with_test.go` has no counterpart. Worth mirroring them there.

In the docs, the `// →` in the "Reported without a fix" section uses the same notation as the applied-fix examples above it, so it reads as if this case is autofixed too.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.md
```diff
@@ -0,0 +1,68 @@
+# prefer-called-exactly-once-with
```
swwind, line 1: The doc file needs to be named `prefer_called_exactly_once_with.md`. `getDocPath()` in `scripts/gen-rule-manifest.js` builds the path from the directory name (`rules/<dir>/<dir>.md`), so a kebab-case name does not match: this rule comes out as `"docPath": null` in the generated `website/generated/rule-manifest.json`, and `website/plugin-rule-manifest.ts` then drops it via `rules.filter((r) => r.docPath)`. The rule ships without a docs page on the website. It is also the only kebab-case name among the 341 rule docs in the repo.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
```diff
@@ -0,0 +1,695 @@
...
+	}
+}
+
+// hasMockResetBetween reports a reset between the two assertions, which splits
+// them into claims about two different call histories: either a reset of the
+// shared target or a suite-wide reset, which clears that target along with
+// everything else. The search descends into the statements that sit between
+// the pair, so a reset nested in an if or a loop still blocks the merge;
+// upstream only scans the sibling statements and merges across such a reset.
+func hasMockResetBetween(
```
swwind: Reassigning the target between the two assertions still merges, with a fix. It is the same class as a mock reset — the two assertions no longer talk about one call history:

```ts
expect(x).toHaveBeenCalledOnce();
x = createMock();
expect(x).toHaveBeenCalledWith('a');
```

fixes to

```ts
x = createMock();
expect(x).toHaveBeenCalledExactlyOnceWith('a');
```

The `toHaveBeenCalledOnce` assertion about the original `x` is silently dropped, and "called exactly once" is transferred to the replacement mock. Upstream has the same hole because it pairs on the `expect(x)` text, but this barrier already walks the statements between the pair, so picking up an assignment or update to the target on the way is cheap.

elecmonkey: Fixed in 1bd1f8c0, by replacing the barrier rather than extending it. The set of dangerous statements cannot be closed: a reassignment breaks the merge, so does a further call to the target, and so does any call that reaches it, which is not decidable at all.

The rule now enumerates what is safe instead. A statement may sit between the two halves only if it is an assertion that transfers control nowhere — its expect and matcher calls are the only calls in it, no matcher executes what it asserts on, no modifier awaits it, and its arguments hold no other call, no await and no assignment — or a const / let declaration whose initializers meet the same bar. `x = createMock()`, `x('b')`, `doMoreWork()` and every mock reset are all blocked by the same condition, and the reset barrier is deleted along with the target comparison it needed.

The cost is one report lost where an unrelated statement sits between the two halves: `expect(x).toHaveBeenCalledOnce(); y.mockClear(); expect(x).toHaveBeenCalledWith('hoge');` moved from invalid to valid, since nothing here proves y is not the target. Assertions on other mocks between the two halves still merge, which is the grouping this mostly appears in.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.md
elecmonkey: Fixed in ebd8fa87. Renamed to `prefer_called_exactly_once_with.md`; the regenerated manifest now carries a docPath for the rule instead of null.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
```diff
@@ -0,0 +1,826 @@
...
+		return nil
+	}
+	return hits
+}
+
+// invokingMatchers execute the value they assert on, so an assertion using one
+// can run arbitrary code — `expect(callTheMock).toThrow()` calls the mock. The
+// set is closed: it is fixed by the matchers @vitest/expect installs, unlike
+// the open question of which function eventually reaches the mock.
+var invokingMatchers = map[string]bool{
```
swwind, line 330: **[P2] The invoking-matcher set is not closed by `@vitest/expect`**

Rstest installs Chai's core assertions and the snapshot matchers alongside `JestChaiExpect` / `ChaiStyleAssertions`, and several of those run the author's code:

- `toThrowErrorMatchingSnapshot` and `toThrowErrorMatchingInlineSnapshot` call the subject, exactly as `toThrow` does;
- `satisfy` / `satisfies` call their argument with the subject — `let result = matcher(obj)` in the Chai bundled into `@rstest/core`;
- `change` / `changes` / `increase` / `increases` / `decrease` / `decreases` call the subject — `assertChanges` does `initial = subject(); fn();`.

So `expect(x).toHaveBeenCalledOnce(); expect(callX).toThrowErrorMatchingInlineSnapshot(); expect(x).toHaveBeenCalledWith('a');` is reported with a fix, and `callX` is free to call `x`. Adding those names keeps the set closed against the matchers Rstest actually installs, rather than the ones `@vitest/expect` installs alone.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
```diff
@@ -0,0 +1,826 @@
...
+
+// isLibraryCall reports a call that runs nothing of the author's: its callee is
+// declared in TypeScript's default library, and nothing callable is handed to
+// it, so the library function has nothing of the author's to invoke.
+// `console.log('checkpoint')` and `JSON.stringify(value)` qualify;
+// `items.forEach(cb)`, `setTimeout(cb)` and `promise.then(cb)` do not, and
+// neither does a `console` the author declared themselves, which resolves to
+// their own binding rather than the library's. Without a checker nothing
+// qualifies, which leaves the pair unreported — the safe direction.
+func (scanner *betweenScanner) isLibraryCall(node *ast.Node) bool {
```
swwind, line 498: **[P3] `eval` is a default-library call that runs the author's code**

`eval('x("b")')` resolves to the default library and is handed nothing callable, so it counts as inert and the pair merges across it. It is the one library function whose entire job is running the author's source; excluding it by name closes it.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.md
```diff
@@ -0,0 +1,68 @@
...
+```
+
+## Reported without a fix
+
+A chain that asserts more than once is reported, but the rewrite is left to the author, because applying it would mean deleting or splicing assertions the rule does not model — and getting that wrong removes a live assertion silently:
+
+```typescript
+expect(handler).to.have.been.calledOnce.and.to.be.ok;
+expect(handler).to.have.been.calledWith('ready');
+// → expect(handler).to.have.been.calledOnceWith('ready').and.to.be.ok;
```
swwind: **[P3] The `// →` here reads as an applied fix**

This is the same notation the fixed examples above use, in the section that says the rewrite is left to the author. Something like `// merge by hand into:` would keep the two apart.

swwind (commented):
The inert-statement barrier holds for everything I tried except an intervening assertion whose matcher runs the author's code: the invoking-matcher set misses spellings Rstest ships.

| Statement between the two halves | eslint-plugin-vitest | This PR |
| --- | --- | --- |
| `expect(callX).toThrow();` | report + fix | no report |
| `expect(callX).toThrowErrorMatchingInlineSnapshot();` | report + fix | report + fix |
| `expect(v).to.satisfy(pred);` | report + fix | report + fix |
| `expect(callX).to.change(obj, 'value');` | report + fix | report + fix |
| `eval('x("b")');` | report + fix | report + fix |

Upstream merges across all of these, so none of it is a regression against it — but each one runs code that can reach the target, which is what the barrier is there to stop.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
elecmonkey: Fixed in 2b99cf18. Added `toThrowErrorMatchingSnapshot`, `toThrowErrorMatchingInlineSnapshot`, `satisfy`, `satisfies`, `change`, `changes`, `increase`, `increases`, `decrease`, `decreases`. Confirmed against the `@rstest/core` bundle: `assertChanges` does `initial = subject()`, `satisfy` does `let result = matcher(obj)`, and both snapshot matchers are exported there. The bundle has no `toSatisfy` at all — the jest spelling I had was the one that never fires, and the Chai spellings that do were the ones missing.

The comment above the set claimed it was closed by the matchers `@vitest/expect` installs, which was the wrong boundary; it now names what Rstest actually installs.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
elecmonkey: Fixed in 2b99cf18. `isLibraryCall` now rejects a callee named `eval` before it looks the symbol up, so the pair is not reported at all rather than reported without a fix — same treatment as a reset or a reassignment in between. The distinction the rule draws is that "report without a fix" means the merge is right and only the rewrite is unsafe to automate, whereas here the merge itself is wrong: `expect(x).toHaveBeenCalledOnce(); eval('x("b")'); expect(x).toHaveBeenCalledWith('a')` passes today and fails however it is merged, by hand or not.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.md
elecmonkey: Fixed in 2b99cf18. That line now reads `// merge by hand into:`, so the `// →` notation is left to the examples the rule actually applies.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
```diff
@@ -0,0 +1,848 @@
...
+	if node.Kind != ast.KindCallExpression || scanner.ctx.TypeChecker == nil {
+		return false
+	}
+	call := node.AsCallExpression()
+	if call == nil {
+		return false
+	}
+	callee := ast.SkipParentheses(call.Expression)
+	if ast.IsIdentifier(callee) && callee.Text() == "eval" {
```
swwind: **[P3] Exclude `eval` by symbol, not by callee text**

`globalThis.eval` is the same default-library `eval`, but the callee is a property access rather than an identifier, so this guard does not fire: the symbol resolves into lib.es5, the string argument is not callable, and the call counts as inert. `expect(x).toHaveBeenCalledOnce(); globalThis.eval('x("b")'); expect(x).toHaveBeenCalledWith('a');` is reported with a fix.

Moving the check below the symbol lookup and asking for `symbol.Name == "eval"` covers both spellings with one condition, and the symbol is already known to come from the default library at that point.

swwind (commented):
One case left: the `eval` exclusion is name-only, so the member form slips through.

| Statement between the two halves | eslint-plugin-vitest | This PR |
| --- | --- | --- |
| `eval('x("b")');` | report + fix | no report |
| `globalThis.eval('x("b")');` | report + fix | report + fix |

Everything else from the last round checks out. Detail inline.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
elecmonkey: Fixed in 6ac0c517, exactly as suggested — the check moved below the symbol lookup and now reads `symbol.Name == "eval"`, so the two spellings are covered by one condition and the default-library check is no longer duplicated by a name match on the callee.

Added both as valid cases. Confirmed the guard is what carries them: with that condition removed, `eval('x("b")')` and `globalThis.eval('x("b")')` between the pair both go back to being reported with a fix.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
```diff
@@ -0,0 +1,848 @@
...
+// Upstream gets the first two from the text preceding the matcher, and misses
+// the third because it never looks through an await. That text also carries
+// Chai's language chains, which assert nothing, so it would keep
+// `expect(x).to.have.been.calledOnce` from pairing with
+// `expect(x).calledWith('a')`. Composing the key from the parse instead keeps
+// what changes meaning and drops the sugar.
+//
+// The separator cannot occur in source text, so no factory text can collide
+// with a modified chain.
+func pairKey(headText string, modifiers []string, awaited bool) string {
```
swwind, line 281: **[P2] Two `expect.poll` halves do not describe one instant**

The key captures the factory text, the modifiers and the `await` — everything that makes two statements halves of one claim for `expect(x)`, where both are evaluated at the same point. `expect.poll(fn)` adds a retry window the key cannot express: each half polls on its own until it passes.

```ts
await expect.poll(getSpy).toHaveBeenCalledOnce();
await expect.poll(getSpy).toHaveBeenCalledWith('a');
```

A spy called once with `'b'` settles the first assertion, and a later call with `'a'` settles the second, since `toHaveBeenCalledWith` matches any recorded call. The pair passes; `toHaveBeenCalledExactlyOnceWith('a')` polled once never does, because by the time `'a'` arrives the count is 2. Both matchers are allowed under poll — its `unsupported` list is the snapshot and throw matchers — and `expect.element(locator)` returns the same proxy.

`parsed.Entry` already names the factory, so restricting the pair to `RstestExpectEntryCall` and `RstestExpectEntrySoft` covers both. An unawaited poll between the two halves needs nothing: the returned thenable builds its promise only in `then`/`catch`/`finally`, so `fn` is never called, and treating that statement as inert is right.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
```diff
@@ -0,0 +1,848 @@
...
+	return hits
+}
+
+// invokingMatchers execute something the assertion was handed, so an assertion
+// using one can run arbitrary code — `expect(callTheMock).toThrow()` calls the
+// mock. The set is enumerable because Rstest fixes which matchers it installs:
+// the jest-style matchers, the snapshot matchers, and Chai's core assertions,
+// which `@rstest/core` bundles alongside them. That is unlike the open question
+// of which function eventually reaches the mock.
+var invokingMatchers = map[string]bool{
```
swwind, line 330: **[P3] `expect.extend` reopens this set**

The comment argues the set is enumerable because Rstest fixes which matchers it installs, but a matcher registered through `expect.extend` is installed by the author, and a custom `toThrow`-alike invokes its subject: `expect(x).toHaveBeenCalledOnce(); expect(callX).toRejectWithMyError(); expect(x).toHaveBeenCalledWith('a');` is reported with a fix.

I do not think it is worth closing in code: inverting the test to "a matcher Rstest ships" needs a full matcher table the repo does not have, and requiring a non-callable subject would drop `expect(otherMock).toHaveBeenCalledWith(...)`, which is the shape the barrier exists to allow. Narrowing the claim in the comment to the matchers Rstest installs itself would be enough.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
```diff
@@ -0,0 +1,848 @@
...
+		return ""
+	}
+	return entries[0].Name
+}
+
+// isExpectHelperCall reports a call such as expect.any(String) or
+// expect.objectContaining({}), reached through the same root as the assertion
+// itself. These are framework functions with known semantics: they build a
+// value to compare against and run nothing of the author's.
+func isExpectHelperCall(node *ast.Node, rootName string) bool {
```
swwind, line 410: **[P3] The helper exemption matches the root name, not the helper**

Any two-segment call under the assertion's root name is accepted, which covers more than the value constructors described above:

- `expect(y).toEqual(expect.toSatisfy(pred))` — the static form calls `pred(actual)` when the comparison runs, so the pair merges across it. The instance spelling is already blocked through `invokingMatchers`, so the two spellings of one matcher disagree.
- `ctx.expect(y).toEqual(ctx.myFixture())` — for the test-context expect the root is `ctx`, whose surface the author extends through `test.extend`, so an arbitrary fixture call is exempt.
- `expect(y).toEqual(expect.myAsymmetric())` for anything registered through `expect.extend`, which is the same premise as the note on `invokingMatchers` above.

`RSTEST_ASYMMETRIC_MATCHERS` is the table this wants: accepting only a name in it, minus `toSatisfy` for the reason above, keeps `expect.any(String)` and `expect.objectContaining({})` exempt and closes the rest.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.md
```diff
@@ -0,0 +1,68 @@
...
+expect(handler).to.have.been.calledOnce;
+expect(handler).to.have.been.calledWith('ready');
+// → expect(handler).to.have.been.calledOnceWith('ready');
+```
+
+A Chai chain can also state both halves by itself, and is reported on its own:
+
+```typescript
+expect(handler).to.have.been.calledOnce.and.calledWith('ready');
+// → expect(handler).to.have.been.calledOnceWith('ready');
```
swwind: **[P3] `// →` here is the applied-fix notation again**

A chain that states both halves is reported without a fix — `reportSelfMerge` passes no fixer, and its test expects no output — but this example uses the same `// →` as the fixed examples above it, and the "Reported without a fix" section below lists only the multi-assertion chain and the unstable target. Either `// merge by hand into:` here, or a line in that section.

swwind (commented):
The `eval` guard from the last round holds for both spellings. Four cases left — one in the pair key, two in what the barrier treats as running nothing of the author's, one in the docs.

| Code | eslint-plugin-vitest | This PR |
| --- | --- | --- |
| `await expect.poll(s).toHaveBeenCalledOnce(); await expect.poll(s).toHaveBeenCalledWith('a');` | no report | report + fix |
| `expect(y).toEqual(expect.toSatisfy(pred));` between the two halves | report + fix | report + fix |
| `ctx.expect(y).toEqual(ctx.myFixture());` between the two halves | no report | report + fix |
| `expect(callX).toRejectWithMyError();` between the two halves, registered through `expect.extend` | report + fix | report + fix |

Detail inline.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
elecmonkey: Fixed in a31a9dbe. `reportCandidate` now requires `parsed.Entry` to be `RstestExpectEntryCall` or `RstestExpectEntrySoft`, so a polled half never enters the map and the key keeps the meaning it was built with.

Both spellings are covered by valid cases, and the unawaited poll in between is an invalid case with a fix, per your last paragraph — the statement runs nothing, so the pair still merges across it.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
elecmonkey: Fixed in a31a9dbe, using `RSTEST_ASYMMETRIC_MATCHERS` minus `toSatisfy` as you suggested. `expect.any(String)` and `expect.objectContaining({})` stay exempt; `expect.toSatisfy(pred)`, an unknown member of a test-context `expect`, and anything from `expect.extend` no longer are. The two spellings of `toSatisfy` now agree.

`expect.not.arrayContaining([])` is three segments, so it was already outside this path and is unaffected.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
elecmonkey: Agreed on both counts, and the comment was mine to overclaim — narrowed in a31a9dbe. It now says the set enumerates the matchers Rstest installs itself, that a matcher from `expect.extend` is the author's and invokes its subject just the same, and why neither way of closing it is worth taking: a full matcher table the repo does not have, or refusing a callable subject, which would drop the `expect(otherMock).toHaveBeenCalledWith(...)` the barrier exists to allow. The docs carry the same note under Limits so it is not only visible in the source.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.md
elecmonkey: Fixed in a31a9dbe — I changed the one below last round and left this one. It now reads `// merge by hand into:`, and the sentence introducing the example says the chain is reported without a fix and why, so the reader does not have to reach the section below to find out.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
```diff
@@ -0,0 +1,875 @@
...
+		return false
+	}
+	// The root name alone says nothing: for the test-context `expect` the root
+	// is the context object, whose surface the author extends through
+	// `test.extend`, and `expect` itself takes new members through
+	// `expect.extend`. Only the constructors Rstest ships are known to build a
+	// value and run nothing, so the name has to be one of those — minus
+	// `toSatisfy`, which calls the predicate it is given when the comparison
+	// runs, exactly as the instance matcher of the same name does.
+	return rstestUtils.RSTEST_ASYMMETRIC_MATCHERS[entries[1].Name] && entries[1].Name != "toSatisfy"
```
swwind: **[P2] `schemaMatching` calls the schema it is handed**

The exemption is `RSTEST_ASYMMETRIC_MATCHERS` minus `toSatisfy`, but `schemaMatching` is the second constructor in that table that runs something of the author's: `SchemaMatching.asymmetricMatch` calls `this.sample['~standard'].validate(other)` when the comparison runs (@vitest/expect 4.1.11, `dist/index.js:1028-1037`), exactly as `toSatisfy` calls its predicate.

```ts
expect(x).toHaveBeenCalledOnce();
expect(y).toEqual(expect.schemaMatching(schema));
expect(x).toHaveBeenCalledWith('a');
```

is reported with a fix, and the schema's `validate` runs between the two halves. Excluding it closes the table: the remaining names compare values, and `toBeOneOf`'s only escape is a custom equality tester, which is the `expect.extend` limit you already document.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
```diff
@@ -0,0 +1,875 @@
...
+	"rejects":  true,
+}
+
+// controlTransfer reports whether a node hands control to code this rule
+// cannot see. Anything that runs user code qualifies, as does any write: a
+// write needs no target check, because a statement between the assertions has
+// no business assigning at all.
+func controlTransfer(node *ast.Node) bool {
+	switch node.Kind {
+	case ast.KindCallExpression, ast.KindNewExpression, ast.KindTaggedTemplateExpression,
```
swwind, line 372: **[P2] A spread runs the iterator protocol**

The switch models a tagged template as the implicit call it is, but a spread is the same: `[...gen]` calls `gen[Symbol.iterator]()` and then `next()` per element, and both are the author's when the operand is a generator or a custom iterable.

```ts
expect(x).toHaveBeenCalledOnce();
expect(y).toEqual([...gen]);
expect(x).toHaveBeenCalledWith('a');
```

merges, while `items.forEach(cb)` — the same "hands control to something of the author's" shape — does not. `ast.KindSpreadElement` in this switch covers it; the object form `{ ...obj }` only triggers getters, which the rule accepts by design.

An array binding pattern is the same protocol on the declaration side, and `isInertDeclaration` walks initializers only, so `const [a] = items;` merges too — rejecting a non-identifier binding name there closes both.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
```diff
@@ -0,0 +1,875 @@
...
+// Upstream gets the first two from the text preceding the matcher, and misses
+// the third because it never looks through an await. That text also carries
+// Chai's language chains, which assert nothing, so it would keep
+// `expect(x).to.have.been.calledOnce` from pairing with
+// `expect(x).calledWith('a')`. Composing the key from the parse instead keeps
+// what changes meaning and drops the sugar.
+//
+// The separator cannot occur in source text, so no factory text can collide
+// with a modified chain.
+func pairKey(headText string, modifiers []string, awaited bool) string {
```
swwind, line 281: **[P3] An awaited pair keeps two suspension points the merge removes**

`isInertStatement` rejects an intervening assertion carrying `resolves`/`rejects` because it hands control to the microtask queue and lets what is already pending run. The pair itself is exempt from that reasoning: with both halves awaited, the second half's own suspension is a window between the count check and the argument check, and the merge closes it. A mock called once with `'b'` before the first await and once with `'a'` in that window satisfies both halves — one call at the first, one matching call among two at the second — and fails `toHaveBeenCalledExactlyOnceWith('a')`.

I do not think it is worth closing in code: refusing awaited pairs would drop the shape the key was extended to support. A line under Limits saying the merge collapses the suspension between two awaited halves would be enough.

swwind (commented):
Three cases left, all in what the barrier treats as running nothing of the author's. Upstream merges across all of them, so none is a regression against it — but each one runs code that can reach the target, which is what the barrier is there to stop.

| Statement between the two halves | eslint-plugin-vitest | This PR |
| --- | --- | --- |
| `expect(y).toEqual(expect.schemaMatching(schema));` | report + fix | report + fix |
| `expect(y).toEqual([...gen]);` | report + fix | report + fix |
| `const [a] = items;` | report + fix | report + fix |

Detail inline.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
elecmonkey: Fixed in 82692218. `schemaMatching` is now excluded alongside `toSatisfy`, so the exemption is the table minus the two constructors that run what they are handed.

One note on how live this is today: the bundled `@rstest/core` here is 0.11.3, and its bundle has no `schemaMatching` at all — the table was built from `@vitest/expect@4.1.10` type declarations, which are ahead of what Rstest ships. So the exemption was accepting a name that does not exist yet rather than letting a validator run. That does not change the conclusion: the table is what the check consults, and the name becomes live the moment Rstest upgrades.

Confirmed the guard carries it: with the `schemaMatching` condition removed, `expect(y).toEqual(expect.schemaMatching(schema))` between the two halves goes back to being reported with a fix.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
elecmonkey: Fixed in 82692218, both halves. `ast.KindSpreadElement` joins the switch, and `isInertDeclaration` now requires every declared name to be a plain identifier.

The declaration side is checked on the shape rather than on `KindArrayBindingPattern` alone: an object pattern only reads properties, which the rule accepts wherever it reads one, but it can nest an array pattern (`const { a: [b] } = obj;`), so rejecting any binding pattern is the check that actually closes it. The cost is `const { b } = obj;` between the two halves, which is a report lost.

`{ ...obj }` is `KindSpreadAssignment`, so it stays inert as you say, and there is now an invalid case pinning that.

Confirmed both guards carry their cases: without the spread kind, `expect(y).toEqual([...gen])` between the halves is reported again; without the binding check, `const [a] = items;` is.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
elecmonkey: Agreed, and documented in 82692218 rather than closed in code, for the reason you give — refusing awaited pairs would drop the shape the key was extended to support.

The Limits section now says that with both halves awaited the second `await` is itself a suspension between the call-count check and the argument check, that the merge closes it, and spells out the failing history: one call before the pair and one more while the second half is suspended satisfies both halves separately and fails the merged assertion. It also names the inconsistency you point at — the barrier rejects an intervening assertion that awaits for exactly this reason, and the pair's own `await` escapes only because it is part of a half rather than a statement between them.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
```diff
@@ -0,0 +1,895 @@
...
+	}
+	call := node.AsCallExpression()
+	if call == nil {
+		return false
+	}
+	symbol := scanner.ctx.TypeChecker.GetSymbolAtLocation(ast.SkipParentheses(call.Expression))
+	if symbol == nil || !internalUtils.IsSymbolFromDefaultLibrary(scanner.ctx.Program(), symbol) {
+		return false
+	}
+	if symbol.Name == "eval" {
```
swwind: **[P1] `x.call(null, 'b')` reaches the target through a default-library symbol**

The symbol test asks where the callee is declared and never what it is bound to. `Function.prototype.call` and `apply` are declared in `lib.es5.d.ts`, so `x.call` resolves to a default-library symbol for any callable `x`, and `null` / `['b']` carry no call signature:

```ts
declare const x: { (a: string): void };
expect(x).toHaveBeenCalledOnce();
x.call(null, 'b');
expect(x).toHaveBeenCalledWith('a');
```

is reported with a fix that deletes the count half, and the target is called between the two halves. `obj.fn.apply(null, ['b'])` is the same shape. `x.reset()` on a method the author declared does block, so it is the library-symbol test alone that lets these through, and `Reflect.apply(x, null, [])` blocks because the target is an argument there.

`call` and `apply` are the same kind of exception as `eval`: the criterion cannot cover them, because what runs is the receiver rather than anything handed over. Excluding them by symbol here closes it.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
```diff
@@ -0,0 +1,895 @@
...
+		return false
+	}
+	for _, argument := range node.Arguments() {
```
swwind, line 514: **[P2] A library call can run the iterator protocol on a non-callable argument**

`Array.from(gen)` calls `gen[Symbol.iterator]()` and then `next()` per element — the same protocol 82692218 rejected on the spread — and a generator object carries no call signature, so the argument test reports that nothing was handed over:

```ts
declare const gen: Iterable<number>;
expect(x).toHaveBeenCalledOnce();
Array.from(gen);
expect(x).toHaveBeenCalledWith('a');
```

merges, while `[...gen]` does not. `Object.fromEntries(gen)` and `Promise.all(gen)` are the same; `new Set(gen)` blocks only because `new` is a control transfer by kind, not because the argument was looked at.

I do not think this one is worth closing in code. Testing the argument for iterability catches strings and arrays too, so `console.log(items)` would leave the inert set — a real cost for the shape the exemption exists for — and naming the library functions that iterate is the kind of enumeration the rule avoids elsewhere. A line under Limits saying a default-library call may still iterate what it is handed would square the docs with the behaviour.

swwind (commented):
Two cases left, both in the default-library exemption. Upstream merges across both, so neither is a regression against it — but each one runs code that can reach the target, which is what the barrier is there to stop.

| Statement between the two halves | eslint-plugin-vitest | This PR |
| --- | --- | --- |
| `x.call(null, 'b');` | report + fix | report + fix |
| `Array.from(gen);` | report + fix | report + fix |

Detail inline.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
```diff
@@ -0,0 +1,901 @@
...
+		return true
+	}
+	switch node.Kind {
+	case ast.KindFunctionExpression, ast.KindArrowFunction, ast.KindClassExpression:
+		return true
+	}
+	if scanner.ctx.TypeChecker == nil {
+		return true
+	}
+	argumentType := scanner.ctx.TypeChecker.GetTypeAtLocation(node)
```
swwind: **[P1] A callable argument reads as non-callable when it is a union, or when an assertion is written on it**

Two ways an argument that does carry a call signature gets past this test, and either one lets a default-library call that invokes it count as inert.

A union: many library callbacks are optional parameters, so `| undefined` is assignable and the code type-checks.

```ts
declare const cmp: ((a: number, b: number) => number) | undefined;
expect(x).toHaveBeenCalledOnce();
[2, 1].sort(cmp);
expect(x).toHaveBeenCalledWith('a');
```

`getSignaturesOfType` answers nothing for the union itself, so `sort` is inert and the pair is reported with a fix that deletes the count half — while `cmp` runs synchronously between the two halves. `JSON.stringify({}, replacer)` is the same shape.

An assertion: `unwrapExpression` strips `as` and `!` before the type query, so the declared type answers instead of the asserted one.

```ts
[1, 2].every(pred!);                      // pred: ((v: number) => boolean) | undefined
[1].forEach(t as unknown as () => void);  // t: T
```

Both merge. The second is the general form — a cast to a function type is a common spelling in tests, and it hides any callable from the barrier.

Querying `GetTypeAtLocation` on the written node, and testing each `UnionTypeParts` constituent for call and construct signatures, closes both; `CollectAllCallSignatures` already does the union half. With that change the four cases above stop being reported and the rule's existing cases are unaffected.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
```diff
@@ -0,0 +1,901 @@
...
+	}
+	var parsed *rstestUtils.ParsedRstestExpectCall
+	if rootCall != nil {
+		parsed = scanner.analysis.ParseExpectCall(rootCall)
+	}
+	if parsed == nil || parsed.Head == nil {
+		// Not an assertion. It can still be inert on its own terms — a
+		// `console.log('checkpoint')` calls nothing of the author's — with no
+		// framework call to exempt.
+		return scanner.inertSubtree(expressionStatement.Expression, nil, "")
```
swwind, line 651: **[P3] A static expect call between the halves blocks the merge**

```ts
expect(x).toHaveBeenCalledOnce();
expect.hasAssertions();
expect(x).toHaveBeenCalledWith('a');
```

is not reported. `ParseExpectCall` recognises `expect.hasAssertions()` and returns a static entry with a nil `Head`, so this branch treats the statement as a non-assertion and exempts no framework call, leaving the call itself a control transfer. `expect.assertions(2)` is the same. Exempting the root call when the parse succeeded with a static entry would let it through, and arguments stay walked, so `expect.extend({ m() { x('b') } })` still blocks.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
```diff
@@ -0,0 +1,901 @@
...
+// isExpectHelperCall reports a call such as expect.any(String) or
+// expect.objectContaining({}), reached through the same root as the assertion
+// itself. These are framework functions with known semantics: they build a
+// value to compare against and run nothing of the author's.
+func isExpectHelperCall(node *ast.Node, rootName string) bool {
+	if rootName == "" || node.Kind != ast.KindCallExpression {
+		return false
+	}
+	entries := test_framework.GetMemberEntries(ast.SkipParentheses(node.AsCallExpression().Expression))
+	if len(entries) != 2 || entries[0].Name != rootName || entries[0].Call != nil {
```
swwind, line 415: **[P3] The helper exemption keys on the chain root, so it misses the member-chain spellings of `expect`**

```ts
test('t', ctx => {
  ctx.expect(handler).toHaveBeenCalledOnce();
  ctx.expect(payload).toEqual(ctx.expect.any(String));
  ctx.expect(handler).toHaveBeenCalledWith('ready');
});
```

is not reported: `expectRootName` returns `ctx`, and `ctx.expect.any(String)` is three segments, so the helper is not exempt and the statement is not inert. `import.meta.rstest.expect.any(String)`, and `rstest.expect.any(String)` behind a namespace import, are the same; the pairs themselves report in all three spellings. The inverse holds too — `ctx.any(String)`, which is not an API, is exempt.

Taking the head callee's member chain minus a trailing `soft` / `poll` / `element` as the prefix, and requiring the helper to be that prefix plus one name from `RSTEST_ASYMMETRIC_MATCHERS`, covers every spelling with one test.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
```diff
@@ -0,0 +1,901 @@
...
+}
+
+// inertSubtree reports whether a subtree transfers control nowhere, treating
+// the given framework calls, `<root>.helper(...)` calls and default-library
+// calls as known-safe. Their arguments are still walked, so a call nested in
+// one of them still counts.
+func (scanner *betweenScanner) inertSubtree(node *ast.Node, frameworkCalls map[*ast.Node]bool, rootName string) bool {
+	inert := true
+	var visit func(node *ast.Node) bool
+	visit = func(node *ast.Node) bool {
```
swwind, line 529: **[P3] A function that is defined but not called blocks the merge**

```ts
expect(x).toHaveBeenCalledOnce();
const later = () => x('b');
expect(x).toHaveBeenCalledWith('a');
```

is not reported: the walk descends into the arrow body and counts `x('b')` as a call that happens here, though the statement only builds a closure. A `function helper() {}` between the halves is refused for the same reason. Stopping at a function-like node's body would fix both and stays safe, because an immediately invoked one carries its own call node at the top.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
```diff
@@ -0,0 +1,901 @@
...
+
+	for _, statement := range statements {
+		if statement == nil {
+			continue
+		}
+		position := internalUtils.TrimNodeTextRange(ctx.SourceFile, statement).Pos()
+		if position <= minPosition || position >= maxPosition {
+			continue
+		}
+		if !scanner.isInertStatement(statement) && !scanner.isInertDeclaration(statement) {
```
swwind, line 715: **[P3] Type-only statements and empty statements block the merge**

`type T = string;` and `interface I { a: string }` between the two halves block it, as do `{}` and `;`. A type-only declaration runs nothing at all, and `declare const y: number;` already passes through `isInertDeclaration`, so these kinds are the inconsistent ones. A `void`-prefixed assertion is not recognised as a candidate either, since `assertionRootCall` stops at the `void`.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.md
```diff
@@ -0,0 +1,75 @@
...
+
+The second: a matcher registered through `expect.extend` can invoke its subject the way `toThrow` does, and the rule has no way to know it. Recognising only the matchers Rstest ships would need a matcher table this repo does not have, and refusing an intervening assertion whose subject is callable would reject `expect(otherMock).toHaveBeenCalledWith(...)` — the assertion the barrier exists to let through.
+
+The last: an exempt default-library call may still run the iterator protocol on what it is handed. `Array.from(gen)` and `Promise.all(gen)` iterate a generator the way `[...gen]` does, but a generator object carries no call signature, so the test for a callable argument reports that nothing was handed over. Testing instead whether an argument can be iterated would take arrays and strings with it, and with those `console.log(items)` — the shape the exemption exists for — so this one is left open.
```
swwind: **[P3] The uncovered case is broader than the iterator protocol**

An exempt default-library call can also reach the author's code through coercion or a write: `JSON.stringify(obj)` runs `obj.toJSON()`, `` console.log(`${obj}`) `` runs `obj.toString()`, and `Object.assign(obj, src)` runs the setters on `obj`. All three merge today. The last is the one worth naming, since an assignment written in the source is a disqualifier while the same write through `Object.assign` is not. Generalising the sentence to "may still run code the author wrote on the value it is handed", with these next to `Array.from(gen)`, would square the paragraph with the behaviour.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.md
```diff
@@ -0,0 +1,75 @@
...
...y if it runs nothing of the author's. That covers an assertion whose `expect(...)` and matcher calls are its only calls, where no matcher executes what it asserts on and the arguments hold no other call, no `await`, no assignment and no spread, since `[...gen]` runs the iterator protocol the way a call runs a function; a call into TypeScript's default library, such as `console.log('checkpoint')` or `JSON.parse(text)`, provided it is none of `eval`, whose source arrives as a string, and `call` and `apply`, which run their receiver rather than anything handed over — all three are declared in that library themselves — and nothing callable is handed to it; and a declaration whose initializers meet the same bar, where a hoisted `var` must also declare names neither assertion reads and every declared name is a plain identifier, since a destructuring pattern runs that same protocol. Assertions on other mocks therefore keep their place between the two halves, while a reset, a reassignment, a further call to the target, and any call the rule cannot resolve leave the pair alone. Resolving the library call needs type information, so without a type checker such a call blocks the merge as well;
```
swwind: **[P3] The inert enumeration omits the plain expression statement**

A statement that is neither an assertion, a library call, nor a declaration is inert when it transfers control nowhere, so `obj.trigger;` between the two halves merges. That follows the getter argument the source makes, but it is not derivable from this list, which names only assertions, default-library calls and declarations.

swwind (commented):
One false positive that ships a fix, plus a set of missed reports and two doc lines. Details inline.

| Statement between the two halves | This PR |
| --- | --- |
| `[2, 1].sort(cmp)`, `cmp: ((a, b) => number) \| undefined` | merges, and `cmp` runs between the halves |
| `JSON.stringify({}, replacer)`, replacer likewise optional-typed | merges |
| `[1, 2].every(pred!)` / `[1].forEach(t as unknown as () => void)` | merges |
| `expect.assertions(2)`, `expect.hasAssertions()` | no report |
| `ctx.expect(y).toEqual(ctx.expect.any(String))` | no report |
| `type T = string;`, `interface I { a: string }` | no report |
| `const later = () => x('b');`, `function helper() {}` | no report |
| `{}`, `;` | no report |

The first three rows are one defect and the only one that turns a passing test into a failing one; the rest are missed reports, all in the safe direction.

I checked the fix output under CRLF, tabs, ASI, a missing trailing newline, a comment line of its own, `?.`, string- and template-literal accessors, two interleaved pairs and two straddling pairs, class methods, static blocks, a `this.spy` target and a namespace import — all correct.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
elecmonkey: Fixed in 542c49e4. `call` and `apply` join `eval` in the symbol exclusion, after the default-library check, so `x.call(null, 'b')` and `obj.fn.apply(null, ['b'])` block the merge. Both spellings are now valid cases in `prefer_called_exactly_once_with_test.go` and the TS test; reverting the exclusion makes them fail with the report and the count-deleting fix.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
elecmonkey: Agreed, not worth closing in code. Documented instead: Limits now ends with a third uncovered case saying an exempt default-library call may still run the iterator protocol on what it is handed, with `Array.from(gen)` against `[...gen]`, and why testing arguments for that would cost the `console.log(items)` shape the exemption exists for.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
elecmonkey: Fixed in 78c785ef. The type now comes from the node as written, with only parentheses stripped, and each `UnionTypeParts` constituent is asked for call and construct signatures. `[2, 1].sort(cmp)`, `JSON.stringify({}, replacer)`, `[1, 2].every(pred!)` and `[1].forEach(t as unknown as () => void)` are valid cases in both suites now; each fails against the previous test if the change is reverted.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
elecmonkey: Accurate, and left as is. These are missed reports rather than merges the rule should not make, and the barrier only reaches them by admitting more statements, which is the direction that turns into a merge over a call the rule cannot see. The false positive above was the one worth spending a change on.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
elecmonkey: Accurate, and left as is. These are missed reports rather than merges the rule should not make, and the barrier only reaches them by admitting more statements, which is the direction that turns into a merge over a call the rule cannot see. The false positive above was the one worth spending a change on.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
elecmonkey: Accurate, and left as is. These are missed reports rather than merges the rule should not make, and the barrier only reaches them by admitting more statements, which is the direction that turns into a merge over a call the rule cannot see. The false positive above was the one worth spending a change on.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
elecmonkey: Accurate, and left as is. These are missed reports rather than merges the rule should not make, and the barrier only reaches them by admitting more statements, which is the direction that turns into a merge over a call the rule cannot see. The false positive above was the one worth spending a change on.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.md
elecmonkey: Generalised in 78c785ef: the case now reads as an exempt default-library call running any code the author wrote on the value it is handed, with `Array.from(gen)`, `JSON.stringify(obj)` reaching `toJSON`, a template literal reaching `toString`, and `Object.assign(obj, src)` reaching a setter — noting that a write spelled out in the source disqualifies while the same write through the library does not.

internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.md
elecmonkey: Added in 78c785ef: the enumeration now ends with any other statement that transfers control nowhere, naming `obj.trigger;` and the getter tradeoff it shares with a property read inside an assertion.

swwind (approved):
LGTM

PR #1778, opened 2026-08-18
feat(rstest): add require-local-test-context-for-concurrent-snapshots rule

elecmonkey:
## Summary

Report snapshot assertions made through a global or imported `expect` inside a concurrent test. Concurrent tests share the global `expect`'s snapshot state, so only the `expect` handed to the test callback keeps each test's counter to itself.

Effective concurrency follows the runtime: a suite propagates its mode only to children that have not declared one. Registrations resolve through the shared `ParseFnCall`, and Test Context provenance through the shared expect parser. Added to the rstest `recommended` preset at `error`, matching upstream.

## Credits

Port from [`eslint-plugin-vitest`](https://github.com/vitest-dev/eslint-plugin-vitest)'s `require-local-test-context-for-concurrent-snapshots`.

- **An explicit inner mode overrides the mode inherited from a suite.** Upstream reports whenever any enclosing `describe` or `test` carries `.concurrent`, so `describe.concurrent('s', () => it.sequential('x', …))` is reported even though that test runs sequentially, and `describe.sequential('s', () => it.concurrent('x', …))` is missed even though that one runs concurrently. Rstest only pushes a suite's mode onto a child that has not declared its own, so the port walks outward to the first registration with an explicit mode instead of asking whether any ancestor is concurrent.

- **`concurrent` wins when both modifiers are chained.** Each modifier sets its own flag and the flags accumulate, so `test.concurrent.sequential(…)` carries both; the runner only ever consults the concurrent flag, and `sequential` acts purely as a veto on inheritance. Source order does not matter, so the port does not treat the last modifier as the winner.

- **Concurrency is read from the resolved registration, not from the members at the call site.** `const t = test.concurrent; t('x', …)` and `import.meta.rstest.test.concurrent(…)` carry the mode through the shared parser, and `@rstest/playwright` registrations resolve the same way. Upstream matches `.concurrent` on the callee of the ancestor call, so it sees none of these.

- **Callbacks passed by name are attributed to their registration.** `test.concurrent('x', callback)` with `function callback() { … }` declared at file scope has no syntactic ancestor linking the two, so upstream reports nothing inside `callback`. The port indexes both test and describe callbacks by the registrations that run them, which also propagates a `describe.concurrent('s', suite)` mode into the tests that `suite` registers. A function registered by more than one call is concurrent if any of those registrations is, and self-referential registrations terminate instead of recursing. Attribution by name is deliberately conservative: the index is file-wide and has no scope information, so a name that is redeclared, or declared inside another callback, is left unattributed rather than blamed for a test that never runs it.

- **The matcher is taken from the parsed assertion chain**, so `expect(x)['toMatchSnapshot']()` is reported. Upstream reads `node.callee.property` and requires it to be an identifier, which skips computed access.

- **`matchSnapshot` is treated as a snapshot matcher.** Rstest exposes this Chai-style short alias for `toMatchSnapshot`; it writes to the same snapshot state, so leaving it out would let the concurrent hazard through under a different name. Vitest has no such alias, which is why upstream's list omits it.

## Related Links

Related #935

## Checklist


internal/plugins/rstest/utils/parse_rstest_fn.go
```diff
@@ -603,6 +605,10 @@ func applyResolvedRstestChainPart(resolved *rstestResolvedAPI, part rstestChainP
 	// across aliases — which is why the member names themselves are collected
 	// by the caller from parts[consumed:] instead.
 	switch part.name {
+	case "concurrent":
+		// Runtime option merging gives concurrent priority when both modifiers
+		// are present, regardless of their source order.
+		resolved.executionMode = RstestExecutionConcurrent
```
swwind, line 611: Execution mode is only ever read from chained members; the `(name, options, fn)` options object is never inspected, so `test('x', { concurrent: true }, …)` and `describe('s', { concurrent: true }, …)` resolve as sequential and report nothing. `resolveRstestTestCallback` already treats `(name, options, fn)` as a real registration shape (`test_callback.go:113`), and `concurrent`/`sequential` are keys on that object. Upstream misses this too, but since this PR deliberately resolves the mode from the parsed registration rather than from the call-site members, it is worth either handling or listing under the intentional differences.

internal/plugins/rstest/utils/test_callback.go
```diff
@@ -25,35 +33,74 @@ func newRstestTestCallbacks() RstestTestCallbacks {
...
+		if info.functionNode != nil {
+			visit(info.functionNode, registration)
+		} else if info.name != "" {
+			pending[info.name] = append(pending[info.name], registration)
 		}
 	}
 
-	for name, parsedCalls := range pending {
+	for name, registrations := range pending {
 		function := analysis.functions[name]
```
swwind: `analysis.functions` is a file-wide name→first-declaration map with no scope awareness (`call_analysis.go:420`), so a registration whose callback identifier does not resolve gets attributed to an unrelated same-named function: `describe('s', () => { function cb() { expect(1).toMatchSnapshot(); } }); test.concurrent('x', cb);` reports inside the nested `cb`, which `test.concurrent` never runs. Triggering it needs an unresolvable identifier (already a TS error) and the pending path predates this PR, but the new ownership index is the first consumer that turns it into a user-visible report.

swwind (commented):
Two issues, both low severity; the mode-resolution core itself checks out.

| Code | eslint-plugin-vitest | This PR |
| --- | --- | --- |
| `test('x', { concurrent: true }, () => expect(1).toMatchSnapshot())` | no report | no report |
| `describe('s', { concurrent: true }, () => it('x', () => expect(1).toMatchSnapshot()))` | no report | no report |
| `describe('s', () => { function cb() { expect(1).toMatchSnapshot(); } }); test.concurrent('x', cb);` | no report | report inside `cb` |

internal/plugins/rstest/utils/parse_rstest_fn.go
elecmonkey: Handled rather than documented. `parseRstestFnCall` now reads the mode from the options object of a `(name, options, fn)` registration, so `test('x', { concurrent: true }, …)` and `describe('s', { concurrent: true }, …)` are reported. The runtime merges that object into the same option bag the modifiers write to, so an options object overrides the chain it is written on: `concurrent: false` and `sequential: true` opt back out of a mode the chain or an enclosing suite would otherwise supply, `concurrent` still beats `sequential` within one object, and a value that cannot be read statically declares nothing. A spread withdraws whatever the keys before it said, so `{ concurrent: true, ...options }` is not reported while `{ ...options, concurrent: true }` is. Covered by new Go and JS cases, and listed under the intentional differences in the PR body.

internal/plugins/rstest/utils/test_callback.go
elecmonkey: Fixed. `analysis.functions` entries now record when a name is declared more than once, and the pending lookup accepts an entry only when the name is unambiguous and declared at module scope — the one scope a file-wide lookup is visible from everywhere in the file. Your `describe('s', () => { function cb() … }); test.concurrent('x', cb)` case is now left unattributed instead of reported, along with the `const cb = () => …` variant. Reaching that path at all means the checker could not resolve the identifier, so declining to guess only costs coverage on code that already fails to compile. The parent walk runs only for those unresolved callbacks, and the ownership benchmark is unchanged. Covered by a table-driven test in `concurrent_context_test.go`, which exercises the index directly since a checker-less context sends every named callback through it.

internal/plugins/rstest/utils/parse_rstest_fn.go
```diff
@@ -620,6 +641,98 @@ func applyResolvedRstestChainPart(resolved *rstestResolvedAPI, part rstestChainP
...
+// the same option bag the chained modifiers write to, so `test('x', {
+// concurrent: true }, fn)` runs concurrently exactly as `test.concurrent` does,
+// and an explicit `concurrent: false` opts back out of a mode the chain or an
+// enclosing describe would otherwise supply.
+//
+// Only statically known booleans are read. A shorthand property, a computed
+// key, or a spread leaves the merged value unknowable, and a spread also
+// invalidates whatever the keys before it said, so the object stops declaring
+// anything rather than guessing a mode that would be reported on.
+func rstestOptionsExecutionMode(call *ast.CallExpression) RstestExecutionMode {
```
swwind: The premise here does not hold: the runtime does not merge this object into the option bag. `testFn` / `describeFn` pull only `{ timeout, retry, repeats, meta }` out of it before spreading the chain's `...options` on top (`@rstest/core@0.11.4`, `dist/api~0.js:6324`, `:6399`), so `concurrent` and `sequential` written here never reach `it()` / `describe()`. That makes the two new invalid cases false positives, and the `{ concurrent: false }` / `{ sequential: true }` valid cases false negatives — those tests do run concurrently. Reverting to reading the mode from the chain alone matches the runtime.

swwind (commented):
Final pass. The callback-name guard holds up.

The options-object change does not, and that is on me: my earlier comment asserted `concurrent` / `sequential` are keys on the `(name, options, fn)` object, and that is wrong. The runtime never reads them, so the mode should keep coming from the chain alone, as it did before this commit.

`testFn` and `describeFn` both destructure only `{ timeout, retry, repeats, meta }` out of the resolved options object, and the chain's own `...options` is spread after it, so `concurrent` and `sequential` written in an options object are discarded (`@rstest/core@0.11.4`, `dist/api~0.js:6324`, `:6399`). Measured on 0.11.4 by interleaving two async bodies: `test('a', { concurrent: true }, …)` / `test('b', { concurrent: true }, …)` logs `a-start,a-end,b-start,b-end` (sequential), while the same two bodies under `test.concurrent` log `a-start,b-start,b-end,a-end`; `describe.concurrent('outer', () => { test('a', { concurrent: false }, …); test('b', { sequential: true }, …) })` logs `a-start,b-start,b-end,a-end`, so an options object does not opt back out either.

| Code | Rstest runtime | This PR |
| --- | --- | --- |
| `test('x', { concurrent: true }, () => expect(1).toMatchSnapshot())` | sequential | report |
| `describe('s', { concurrent: true }, () => it('x', () => expect(1).toMatchSnapshot()))` | sequential | report |
| `test.concurrent('x', { concurrent: false }, () => expect(1).toMatchSnapshot())` | concurrent | no report |
| `describe.concurrent('s', () => it('x', { sequential: true }, () => expect(1).toMatchSnapshot()))` | concurrent | no report |

Dropping `rstestOptionsExecutionMode` along with its tests, the doc paragraph and the PR-body bullet restores the behavior the runtime actually has.

internal/plugins/rstest/utils/parse_rstest_fn.go
elecmonkey: You are right, and the runtime source agrees — checked against `@rstest/core@0.11.3` locally: `testFn` (`dist/0~api.js:9336`) and `describeFn` (`:9411`) destructure only `timeout`, `retry`, `repeats` and `meta` out of the resolved options, and the chain's own options are spread after that, so `concurrent` / `sequential` written in an options object never reach the runner.

Reverted in 9198783c: `rstestOptionsExecutionMode` is gone, the mode comes from the chain alone again, and the doc paragraph and PR-body bullet went with it. The cases are kept as tests, inverted to lock the runtime behavior in: `test('x', { concurrent: true }, …)` and `describe('s', { concurrent: true }, …)` are valid, while `test.concurrent('x', { concurrent: false }, …)` and `describe.concurrent('s', () => it('x', { sequential: true }, …))` still report, since an options object cannot opt back out either. The callback-name guard is untouched.

internal/plugins/rstest/utils/parse_rstest_fn.go
elecmonkey: Confirmed and reverted in 9198783c. Verified the same against `@rstest/core@0.11.3`: `testFn` (`dist/0~api.js:9336`) and `describeFn` (`:9411`) take only `timeout`, `retry`, `repeats` and `meta` out of the resolved options before spreading the chain's own options on top, so those two keys never reach `it()` / `describe()`.

`rstestOptionsExecutionMode` and its helpers are gone and the mode comes from the chain alone again; the doc paragraph and the PR-body bullet went with it. The cases stay as tests, inverted to lock the runtime behavior in: `test('x', { concurrent: true }, …)` and `describe('s', { concurrent: true }, …)` are valid, while `test.concurrent('x', { concurrent: false }, …)` and `describe.concurrent('s', () => it('x', { sequential: true }, …))` still report, since an options object cannot opt back out either.

internal/plugins/rstest/rules/require_local_test_context_for_concurrent_snapshots/require_local_test_context_for_concurrent_snapshots.md
swwind, line 1: Rename this file into snake_name, please. Nothing else from here now.

elecmonkey: Done.

internal/plugins/rstest/rules/require_local_test_context_for_concurrent_snapshots/require_local_test_context_for_concurrent_snapshots.go
```diff
@@ -0,0 +1,36 @@
...
+func requireLocalTestContextMessage() rule.RuleMessage {
+	return rule.RuleMessage{
+		Id:          "requireLocalTestContext",
+		Description: "Use local Test Context instead",
+	}
+}
+
+var RequireLocalTestContextForConcurrentSnapshotsRule = rule.Rule{
+	Name:   "rstest/require-local-test-context-for-concurrent-snapshots",
+	Schema: rule.EmptyArraySchema,
```
swwind: Without a `TypeChecker`, `FromTestContext` is always false and `expect` is matched by name alone, so `test.concurrent('x', ({ expect }) => expect(1).toMatchSnapshot())` is reported. Rules that do not declare `RequiresTypeInfo: true` are scheduled on source-only gap programs with a nil checker, so please add it here.

swwind (commented):
One issue left; the rest checks out — two-level describe inheritance, `describe.sequential` inside `describe.concurrent`, `test.extend`, `expect.soft`, aliased registrations and the `.each` / `.for` context positions all match the runtime.

Expect provenance is read from symbols, so on a file with a nil `TypeChecker` the rule reports the local Test Context form it asks for.

| Code | With type info | No type info |
| --- | --- | --- |
| `test.concurrent('x', ({ expect }) => expect(1).toMatchSnapshot())` | no report | report |

internal/plugins/rstest/rules/require_local_test_context_for_concurrent_snapshots/require_local_test_context_for_concurrent_snapshots.go
elecmonkey: You're right. Without type-checker, we will throw error to correct case. It's unacceptable.

elecmonkey: Fixed in c608b72e!

internal/plugins/rstest/rules/require_local_test_context_for_concurrent_snapshots/require_local_test_context_for_concurrent_snapshots.go
```diff
@@ -0,0 +1,41 @@
...
+	Run: func(ctx rule.RuleContext, _ []any) rule.RuleListeners {
+		analysis := rstestUtils.GetRstestCallAnalysis(ctx)
+		// The ownership index behind this context is built on the first query,
+		// so obtaining it here costs nothing on files without snapshots.
+		concurrentContext := rstestUtils.GetRstestConcurrentContext(ctx, analysis)
+		return rule.RuleListeners{
+			ast.KindCallExpression: func(node *ast.Node) {
+				parsed := analysis.ParseExpectCall(node)
+				if parsed == nil || parsed.FromTestContext ||
+					!rstestUtils.RSTEST_SNAPSHOT_MATCHERS[parsed.Matcher] ||
```
swwind: `parsed.Matcher` is only the first matcher in the chain, so a Chai chain that reaches `matchSnapshot` after another assertion is missed: `test.concurrent('x', () => expect('a').to.be.a('string').and.matchSnapshot())` reports nothing, while the single-assertion `expect(1).to.matchSnapshot()` does. `matchSnapshot` is on the list precisely because Chai style is supported, and Chai style is where multi-assertion chains appear. `no-interpolation-in-snapshots` already loops over `parsed.Matchers` for the same reason.

Iterating `parsed.Matchers` and reporting once when any of them is a snapshot matcher covers both forms.

swwind (commented):
One issue left; everything else from the earlier rounds checks out. The type-info gate resolves the `({ expect })` false positive, and `expect.soft`, `resolves`, computed matcher access, `not`, multi-level describe inheritance and the scope-blind callback-name guard all behave as documented.

| Code in a concurrent test | Snapshot state | This PR |
| --- | --- | --- |
| `expect(1).to.matchSnapshot()` | shared | report |
| `expect('a').to.be.a('string').and.matchSnapshot()` | shared | no report |

internal/plugins/rstest/rules/require_local_test_context_for_concurrent_snapshots/require_local_test_context_for_concurrent_snapshots.go
elecmonkey: Fixed in 93a79f83. The listener now takes only the outermost call of a chain and reports when any entry of `parsed.Matchers` is a snapshot matcher, so `expect('a').to.be.a('string').and.matchSnapshot()` is reported and a chain carrying two snapshot assertions still reports once. Tests cover both, plus a multi-assertion chain with no snapshot matcher as a valid case; the first of them fails against the previous `parsed.Matcher` check.

internal/plugins/rstest/rules/require_local_test_context_for_concurrent_snapshots/require_local_test_context_for_concurrent_snapshots.go
```diff
@@ -0,0 +1,58 @@
...
+	Run: func(ctx rule.RuleContext, _ []any) rule.RuleListeners {
+		analysis := rstestUtils.GetRstestCallAnalysis(ctx)
+		// The ownership index behind this context is built on the first query,
+		// so obtaining it here costs nothing on files without snapshots.
+		concurrentContext := rstestUtils.GetRstestConcurrentContext(ctx, analysis)
+		return rule.RuleListeners{
+			ast.KindCallExpression: func(node *ast.Node) {
+				// Only the outermost call of a chain is parsed, so a chain
+				// carrying two snapshot assertions is answered once.
+				if rstestUtils.FindTopMostCallExpression(node) != node {
```
swwind: `parseRstestExpectCall` already returns nil for any call that is not the top of its own chain (`parse_rstest_expect.go:234`), so this guard never changes an answer — the two-snapshot chain reports once because of that check, not this one. Dropping lines 37-41 keeps every case in the suite passing.

The ordering is the part worth changing. `IsInConcurrentTest` now runs ahead of the matcher test, so every non-context `expect` call pays for the callback ownership index even in a file with no snapshot assertion anywhere, which is the common case. `BenchmarkRstestCallAnalysisConcurrentOwnership` puts the index plus the queries at roughly half of what `ParseExpectCall` costs over a file of the same size, and that is without a checker — with one, every registration also costs a symbol lookup. Testing the matcher first restores the short-circuit the previous revision had:

```go
parsed := analysis.ParseExpectCall(node)
if parsed == nil || parsed.FromTestContext ||
	!slices.ContainsFunc(parsed.Matchers, isSnapshotMatcher) ||
	!concurrentContext.IsInConcurrentTest(node) {
	return
}
```

swwind (commented):
The Chai chain fix behaves as described: `expect('a').to.be.a('string').and.matchSnapshot()` is reported, a multi-assertion chain with no snapshot matcher is left alone, and a chain carrying two snapshot assertions still answers once. Everything from the earlier rounds continues to hold.

One item left, on the shape of the listener rather than on what it reports.

internal/plugins/rstest/rules/require_local_test_context_for_concurrent_snapshots/require_local_test_context_for_concurrent_snapshots.go
elecmonkey: Both correct, done in 4336f844. The guard is gone — and with it the comment above it, which credited the guard for the single report on a two-snapshot chain when `parseRstestExpectCall` is what does that. The three conditions are one `if` again with the matcher test ahead of `IsInConcurrentTest`, so a file with no snapshot assertion never builds the ownership index. Full `go test ./internal/plugins/rstest/...` stays green.

internal/plugins/rstest/utils/concurrent_context_test.go
```diff
@@ -0,0 +1,263 @@
...
+	first := rstestUtils.GetRstestConcurrentContext(ctx, analysis)
+	second := rstestUtils.GetRstestConcurrentContext(ctx, analysis)
+	if first != second {
+		t.Fatal("contexts for one file did not share concurrent ownership")
+	}
+	if len(analysis.Callbacks().Functions) == 0 {
+		t.Fatal("test callback fixture was not collected")
+	}
+
+	var markerCall *ast.Node
```
swwind: `findCallByCalleeName` at the bottom of this file is this same walk, so this test can take `findCallByCalleeName(sourceFile, "marker")` instead of repeating it.

The name promises more than the test checks, too: laziness is asserted by `TestRstestConcurrentContextBuildsOwnershipLazily`, which can see the `ownership` field from inside the package, and the `Callbacks().Functions` assertion belongs to the callback collector. Sharing is the part only this test covers.

internal/plugins/rstest/utils/call_analysis_benchmark_test.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+}
+
+func BenchmarkRstestCallAnalysisConcurrentOwnership(b *testing.B) {
+	sourceFile, calls := benchmarkRstestSourceFile(b, true)
+	b.ReportAllocs()
+	b.ResetTimer()
+	for range b.N {
+		analysis := newRstestCallAnalysis(rule.RuleContext{SourceFile: sourceFile})
+		context := &RstestConcurrentContext{
```
swwind: The literal leaves `analysis` nil, which holds only while `ownership` is pre-set: any path that reaches `ownershipIndex()` with a nil index calls `collectRstestCallbackOwnership(nil)` and dereferences it. `&RstestConcurrentContext{analysis: analysis, modes: ...}`, letting the first query build the index, measures the same work without the trap.

swwind (commented):
Behavior is settled: the mode resolution, the listener order, the callback-name guard and the Chai chain handling all hold, and the cycle guard in `callbackRunsConcurrently` terminates — self-referential and mutually referential registrations resolve without recursing, and an explicit `concurrent` anywhere in the cycle still wins. Two test-hygiene items left.

internal/plugins/rstest/utils/concurrent_context_test.go
elecmonkey: Both fixed in aecaa25b. The inline walk is gone in favour of `findCallByCalleeName(sourceFile, "marker")`, the `Callbacks().Functions` assertion is gone since the collector has its own coverage, and the test is now `TestRstestConcurrentContextIsSharedPerFile` — it claims only the per-file sharing it is the sole cover for, with a comment pointing at `TestRstestConcurrentContextBuildsOwnershipLazily` for the laziness half.

internal/plugins/rstest/utils/call_analysis_benchmark_test.go
elecmonkey: The nil `analysis` is real — it survives only because `ownership` is pre-set, so `ownershipIndex()` never takes the branch that would pass nil to `collectRstestCallbackOwnership`. Rather than repair the literal, the benchmarks are dropped altogether in 1fd98b8f: they do not belong in the repository. The trap goes with them.

swwind (dismissed):
LGTM

swwind (approved):
LGTM again!

PR #1786, opened 2026-08-19
feat(unicorn): port rule no-unsafe-string-replacement

elecmonkey:
## Summary

Port the `unicorn/no-unsafe-string-replacement` rule from ESLint to rslint.

This rule uses type information only, where the upstream rule classifies the
receiver syntactically before consulting type information. It declares
`RequiresTypeInfo: true`, so it is skipped rather than guessing on files where
type information is unavailable. It reports unless the type shows the receiver
is not a string, so where the two classifiers disagree this rule follows the
type:

- A receiver typed as a no-substitution template literal type (`` `x` ``) is
  reported. Upstream's `isStringTypeAnnotation` uses `isStringLiteral(node.literal)`,
  which a `TemplateLiteral` fails, so `TSLiteralType` lands in
  `nonTargetTypeAnnotations` and the call is skipped. Such a type is a string,
  so reporting is the correct outcome.
- A receiver whose type reduces to `never` (`string & number`, `'x' & 1`) is not
  reported, where upstream reports it. The receiver is uninhabited, so staying
  quiet is fine.
- A receiver whose numeric or bigint literal type is inferred rather than
  annotated (`const value = 1`, `(1)`, `1 | 2` after narrowing) is not reported,
  where upstream reports it. Upstream has no annotation to classify there and
  leaves symbol-less literal types unknown; the inferred type is the same type
  as the annotated one, which is provably not a string, and calling `replace` on
  such a receiver is a TypeScript error in the first place.

All three are recorded in the rule doc.

## Related Links

- ESLint rule: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v73.0.0/docs/rules/no-unsafe-string-replacement.md
- Source code: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v73.0.0/rules/no-unsafe-string-replacement.js

## Checklist


internal/plugins/unicorn/rules/no_unsafe_string_replacement/no_unsafe_string_replacement.go
```diff
@@ -0,0 +1,351 @@
...
+			return unicornutil.TypeUnknown
+		}
+		symbol := ctx.Refs.Resolve(node)
+		if symbol == nil || visiting[symbol] {
+			return unicornutil.TypeUnknown
+		}
+		visiting[symbol] = true
+		defer delete(visiting, symbol)
+
+		for _, declaration := range symbol.Declarations {
```
swwind: This loop takes the first non-`Unknown` classification across *all* declarations of the symbol, but upstream's `getTypeFromVariable` bails out with `unknown` as soon as `variable.defs.length !== 1`. When a type alias merges with a same-named value, the alias annotation wins here and the receiver is classified `TypeNonTarget`, which then short-circuits `isKnownNonStringReceiver` before the TypeChecker ever runs:

```ts
type Sep = number; const Sep = '-'; Sep.replace('a', r);   // ESLint reports, this rule is silent
type T = {a: 1}; const T = 'x';     T.replace('a', r);     // ESLint reports, this rule is silent
```

Adding `|| len(symbol.Declarations) != 1` to the guard above makes both cases match ESLint.

Separately, `declaration.Type()` falls back to `funcLike.Type` for function-like declarations, i.e. the *return type* annotation, so `function f(): number {}` classifies the identifier `f` itself as non-string. Worth restricting this loop to declaration kinds that actually carry a type annotation for the binding.

internal/plugins/unicorn/unicornutil/type_classifier.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	constraint := checker.Checker_getBaseConstraintOfType(ctx.TypeChecker, t)
+	if constraint != nil && constraint != t {
+		return ClassifyType(ctx, constraint, options)
+	}
+
+	name, ok := TypeSymbolName(t)
+	if !ok {
+		if utils.IsTypeFlagSet(t, checker.TypeFlagsPrimitive|checker.TypeFlagsIntrinsic) {
```
swwind, line 78: Upstream's `getTypeScriptType` reaches `nonTarget` only through `type.intrinsicName`, which is set on `string` / `number` / `boolean` but *not* on literal types, and returns `unknown` when the type has no symbol. `TypeFlagsPrimitive` covers `NumberLiteral` / `BooleanLiteral` / `BigIntLiteral`, so symbol-less literal types land on `NonTarget` here where upstream says `unknown`:

```ts
const v = 1; v.replace('a', r);   // ESLint reports, this rule is silent in .ts
(1).replace('a', r);              // same
```

The heuristic is inherited unchanged from `prefer_array_some`, where it was fine for that rule's semantics — it just doesn't hold for `isKnownNonString`. Note the extras test `invalid((1).replace("x", replacement))` only passes because it uses `FileName: "file.js"` and so takes the checker-skipping path; the `.ts` behaviour is currently uncovered.

internal/plugins/unicorn/rules/no_unsafe_string_replacement/no_unsafe_string_replacement.go
```diff
@@ -0,0 +1,351 @@
...
+}
+
+func isKnownNonStringReceiver(ctx rule.RuleContext, node *ast.Node) bool {
+	if classifyStringReceiverSyntax(ctx, node, map[*ast.Symbol]bool{}) == unicornutil.TypeNonTarget {
+		return true
+	}
+
+	// ESLint's default JavaScript parser has no TypeScript Program. Preserve
+	// that behavior even though rslint may have a checker for a JS file.
+	if ctx.TypeChecker == nil || tspath.HasJSFileExtension(ctx.SourceFile.FileName()) {
```
swwind: `prefer_array_some`'s `classifyReceiver` gates only on `ctx.TypeChecker == nil`; this is the only unicorn rule that additionally refuses type information for `.js` files. Since the rule is enabled in the `recommended` preset, the same `router.replace(pathname, opts)` is silent in `app.ts` and reported in `app.js` of the same project. If matching ESLint's default parser is meant to be the plugin's convention it should be applied consistently across these rules; otherwise this is a JS-only false-positive source.

internal/plugins/unicorn/rules/no_unsafe_string_replacement/no_unsafe_string_replacement.go
```diff
@@ -0,0 +1,351 @@
...
+		return false
+	}
+	name := property.Name()
+	if name == nil || name.Kind == ast.KindComputedPropertyName {
+		return false
+	}
+	// ESTree represents these non-computed object keys as Literal nodes. They
+	// cannot name one of the three coercion hooks, so they are plain keys even
+	// though the shared static-name helper only handles them in computed form.
+	if name.Kind == ast.KindNullKeyword || name.Kind == ast.KindTrueKeyword ||
```
swwind: This branch is unreachable: TypeScript parses `{null: v}` / `{true: v}` / `{false: v}` property names as `Identifier`, never as `KindNullKeyword` / `KindTrueKeyword` / `KindFalseKeyword`. Replacing the body with a `panic` keeps the whole suite green, including the three tests that use exactly those keys — `GetStaticPropertyName` already handles them through the identifier path. The comment about ESTree `Literal` keys also doesn't describe what this code sees.

internal/plugins/unicorn/rules/no_unsafe_string_replacement/no_unsafe_string_replacement.go
```diff
@@ -0,0 +1,351 @@
...
+	switch node.Kind {
+	case ast.KindStringKeyword:
+		return unicornutil.TypeTarget
+	case ast.KindLiteralType:
+		literal := node.AsLiteralTypeNode().Literal
+		if literal != nil && ast.IsStringLiteralLike(literal) {
+			return unicornutil.TypeTarget
+		}
+		return unicornutil.TypeNonTarget
+	case ast.KindBigIntKeyword, ast.KindBooleanKeyword, ast.KindNeverKeyword,
```
swwind, line 252: Two gaps versus upstream in the syntax-only path, which is exactly the path this function exists to serve:

- no `KindUndefinedKeyword` case here — upstream's `nullishTypeAnnotations` maps `undefined` to nullish, i.e. non-target;
- no `KindTypeParameter` case in `classifyStringTypeReference` — upstream follows `definition.node.constraint`, so `function f<T extends number>(v: T) { v.replace('a', r); }` is silent there and reported here on a source-only Program.

Both are masked by the TypeChecker during normal `.ts` linting.

swwind (commented):
I ran ~70 differential cases against `eslint-plugin-unicorn@73.0.0` (wrappers, optional chains, spread, `String.raw` shapes, plain-object key classes, unions / intersections / heritage / recursive aliases, `this` / `super`, literal types). Report ranges match everywhere; the divergences I found are below.

| Code | ESLint | This PR |
| --- | --- | --- |
| `type Sep = number; const Sep = '-'; Sep.replace('a', r);` | reports | no report |
| `type T = {a: 1}; const T = 'x'; T.replace('a', r);` | reports | no report |
| `const v = 1; v.replace('a', r);` (`.ts`) | reports | no report |
| `(1).replace('a', r);` (`.ts`) | reports | no report |
| `function f<T extends number>(v: T) { v.replace('a', r); }` (source-only Program) | no report | reports |

The first two are the ones worth fixing before merge: a syntax-level `TypeNonTarget` verdict skips the TypeChecker entirely, so a misjudgment there is an unconditional false negative rather than a conservative fallback.

The refactor of `prefer_array_some` onto `unicornutil.ClassifyType` looks behaviour-preserving: the old `targetNames.Has("Array")` heritage gate was true for both call sites, the new `NonTargetTypeNames` step sits between two `NonTarget` returns, and nullish-union filtering is preserved via `AllowNullishInMixedUnion: true`.

internal/plugins/unicorn/rules/no_unsafe_string_replacement/no_unsafe_string_replacement.go
elecmonkey: Fixed in de81404b. The syntax path now returns unknown unless the resolved symbol has exactly one declaration, and it only reads binding annotations from variable/parameter declarations. This keeps merged type/value symbols conservative and no longer treats a function return annotation as the function binding annotation. Added the two merged-symbol cases and a source-only function-return regression.

internal/plugins/unicorn/unicornutil/type_classifier.go
elecmonkey: Fixed in de81404b. `no-unsafe-string-replacement` now leaves symbol-less number/boolean/bigint literal types unknown, matching the Unicorn `intrinsicName`/symbol path, without changing shared `ClassifyType` behavior used by `prefer_array_some`. Added a `.ts` literal receiver regression.

internal/plugins/unicorn/rules/no_unsafe_string_replacement/no_unsafe_string_replacement.go
elecmonkey: Fixed in de81404b. Removed the `.js` filename gate: type information is now used whenever `ctx.TypeChecker` is available, matching the upstream parser-services Program gate. Added a JSDoc-typed JavaScript receiver regression.

internal/plugins/unicorn/rules/no_unsafe_string_replacement/no_unsafe_string_replacement.go
elecmonkey: Fixed in de81404b. Removed the unreachable keyword-key branch and inaccurate ESTree comment; `GetStaticPropertyName` handles those parsed identifier keys.

internal/plugins/unicorn/rules/no_unsafe_string_replacement/no_unsafe_string_replacement.go
elecmonkey: Fixed in de81404b. Added `undefined` to syntax non-target annotations and follows `TypeParameter` constraints in type references. Expanded source-only Program coverage for both, where TypeChecker cannot mask a difference.

swwind (commented):
Re-checked the five earlier findings against `eslint-plugin-unicorn@73.0.0`: all five now match, including the source-only Program path (25 syntax-only cases run through both a `project`-less ESLint config and `NewFromBoundSources`, identical verdicts). Two divergences remain, both introduced by the new literal-type gate.

| Code (`.ts`) | ESLint | This PR |
| --- | --- | --- |
| `declare const o: { flag: true }; o.flag.replace("x", r)` | no report | reports |
| `declare function get(): true; get().replace("x", r)` | no report | reports |
| `function f(v: 1 \| 2 \| string) { if (typeof v === "number") { v.replace("x", r) } }` | reports | no report |

internal/plugins/unicorn/rules/no_unsafe_string_replacement/no_unsafe_string_replacement.go
```diff
@@ -0,0 +1,371 @@
...
+	if ctx.TypeChecker == nil {
+		return false
+	}
+
+	t := ctx.TypeChecker.GetTypeAtLocation(node)
+	// Unicorn only treats primitive types as non-target through intrinsicName.
+	// TypeScript literal types have neither that name nor a symbol upstream, so
+	// keep them unknown rather than suppressing this rule's diagnostic.
+	if checker.Type_symbol(t) == nil && checker.Type_alias(t) == nil &&
+		utils.IsTypeFlagSet(t, checker.TypeFlagsNumberLiteral|checker.TypeFlagsBooleanLiteral|checker.TypeFlagsBigIntLiteral) {
```
swwind: `true` and `false` are intrinsic types in TypeScript, so `intrinsicName` is set on them and upstream's `getTypeScriptType` returns `nonTarget` before reaching the symbol-less `unknown` return:

    flag: true | intrinsicName = "true"    | symbol = (none)
    n: 1       | intrinsicName = undefined | symbol = (none)
    b: 1n      | intrinsicName = undefined | symbol = (none)

Only number / bigint literal types land on `unknown`. Dropping `checker.TypeFlagsBooleanLiteral` from the mask silences `declare const o: { flag: true }; o.flag.replace("x", r)` and `declare function get(): true; get().replace("x", r)` again, with the rest of the package suite still green.

Separately, upstream applies this `intrinsicName` / symbol test at every level of `getTypeScriptType`, while this gate only sees the top-level type. Inside a union, `ClassifyType` still falls through to `TypeFlagsPrimitive` and classifies a symbol-less number literal as `NonTarget`, so `function f(v: 1 | 2 | string) { if (typeof v === "number") { v.replace("x", r) } }` is silent here and reported by ESLint.

elecmonkey: Fixed in 4d7432d2. Boolean literals now remain known non-string, while symbol-less number/bigint literals stay unknown through a rule-specific classifier option that is applied recursively inside unions, intersections, and constraints. Added regressions for property/return boolean literals, bigint literals, and the narrowed `1 | 2 | string` case; all match eslint-plugin-unicorn@73.0.0.

elecmonkey:
Addressed the latest review in 4d7432d2: both boolean-literal false positives are gone, and the narrowed number-literal union now reports. I re-ran the focused cases against eslint-plugin-unicorn@73.0.0, the full Unicorn Go suite (including prefer_array_some), the JS integration test, build/vet, native build, format check, and lint:go; all pass. The five earlier inline findings remain covered by de81404b and their regression tests.

internal/plugins/unicorn/rules/no_unsafe_string_replacement/no_unsafe_string_replacement.go
```diff
@@ -0,0 +1,365 @@
...
+	if node == nil {
+		return unicornutil.TypeUnknown
+	}
+
+	switch node.Kind {
+	case ast.KindStringKeyword:
+		return unicornutil.TypeTarget
+	case ast.KindLiteralType:
+		literal := node.AsLiteralTypeNode().Literal
+		if literal != nil && ast.IsStringLiteralLike(literal) {
```
swwind: `ast.IsStringLiteralLike` also accepts `NoSubstitutionTemplateLiteral`, so ``type: `x` `` classifies as `TypeTarget` here and the receiver falls through to the TypeChecker, which reports it. Upstream's `isStringTypeAnnotation` uses `isStringLiteral(node.literal)`, which is `Literal` + `typeof value === 'string'`; a `TemplateLiteral` fails it, so `TSLiteralType` lands in `nonTargetTypeAnnotations` and the call is skipped. `literal.Kind == ast.KindStringLiteral` restores parity, including inside unions and intersections (``` `x` | 1 ```, ``` `x` & {} ```), while ``` `a${string}` ``` stays a `TemplateLiteralType` node and keeps reporting.

internal/plugins/unicorn/rules/no_unsafe_string_replacement/no_unsafe_string_replacement.go
```diff
@@ -0,0 +1,365 @@
...
+			return unicornutil.TypeTarget
+		}
+		return unicornutil.TypeNonTarget
+	case ast.KindBigIntKeyword, ast.KindBooleanKeyword, ast.KindNeverKeyword,
+		ast.KindNumberKeyword, ast.KindObjectKeyword, ast.KindSymbolKeyword,
```
swwind: `ast.KindObjectKeyword` has no counterpart in upstream's `nonTargetTypeAnnotations` set (`TSObjectKeyword` is absent), so `getTypeAnnotationType` returns `unknown` for `object` and the call is reported when no type information is available. With a TypeChecker the two agree anyway — `object` has `intrinsicName` — so dropping this kind only affects the source-only path, where it currently suppresses `declare const v: object`, `object | number`, and `object & {a: 1}`.

swwind (commented):
Re-checked the three findings from the previous round against `eslint-plugin-unicorn@73.0.0`: all three now match, and the `UnknownSymbolLessTypeFlags` gate lines up with upstream's `intrinsicName`-before-`getTypeSymbol` ordering in `getTypeScriptType`. Two divergences remain, both in `classifyStringTypeSyntax`.

| Code (`.ts`) | ESLint | This PR |
| --- | --- | --- |
| ``declare const v: `x`; v.replace("x", r)`` | no report | reports |
| ``declare const v: `x` \| 1; v.replace("x", r)`` | no report | reports |
| ``declare const v: `x` & {}; v.replace("x", r)`` | no report | reports |
| `declare const v: object; v.replace("x", r)` (no `project`) | reports | no report |
| `declare const v: object \| number; v.replace("x", r)` (no `project`) | reports | no report |

The first group is a false positive in every configuration; the second only shows up on a source-only Program, where the syntax classifier is the whole answer.

internal/plugins/unicorn/rules/no_unsafe_string_replacement/no_unsafe_string_replacement.go
elecmonkey: Fixed in db19be09. Removed KindObjectKeyword from the syntax-only non-target set, leaving object annotations unknown without a TypeChecker. Added source-only regressions for object, object | number, and object & {a: 1}.

internal/plugins/unicorn/rules/no_unsafe_string_replacement/no_unsafe_string_replacement.go
elecmonkey: Fixed in db19be09. TSLiteralType is now a string target only when its literal is KindStringLiteral, so no-substitution template literal types are known non-string while interpolated template literal types remain reportable. Added direct, union, intersection, and interpolated regressions.

swwind (commented):
Let's make this a type-information-only rule rather than keep maintaining the syntax-level classifier. This is a bigger refactor than the remaining nits, so no inline comments — the whole request is below.

I measured what the classifier buys once a TypeChecker is present: dropping the `classifyStringReceiverSyntax` short-circuit and re-running this package leaves the upstream suite fully green, and only the three `` `x` `` cases in the extras suite change verdict. The classifier is also inert on JavaScript — it only ever returns a non-`unknown` verdict through TypeScript type annotations and aliases — so its entire domain is `.ts` files that happen to have no type information. That is not worth the ~190 lines and the divergences they keep producing.

What I'd like instead:

1. `isKnownNonStringReceiver` keeps only the `unicornutil.ClassifyType` path. `classifyStringReceiverSyntax`, `classifyStringTypeSyntax`, `combineStringTypeSyntax`, `classifyStringTypeReference`, `combineStringTypeExpressions` and `bindingTypeAnnotation` all go away. `unicornutil/type_classifier.go` stays as is — `prefer-array-some` uses it.
2. Declare `RequiresTypeInfo: true` on the rule, so it is skipped rather than guessing when no type information is available. This does not disturb the rule tester: the fixtures Program supplies type information for the `.js` cases too.
3. Delete `TestNoUnsafeStringReplacementSourceOnlyTypeSyntax`, which exists only to cover the syntax path.
4. Flip the cases that held only through the syntax path: the three `` `x` `` valid cases become invalid. ``type T = `x`; declare const v: T`` and ``function f<T extends `x`>(v: T)`` are worth adding next to them — same divergence, different route to the literal type.
5. Record the divergence in the rule doc.

This is a deliberate divergence from upstream, in both directions:

| Code (`.ts`) | ESLint | This PR after the change |
| --- | --- | --- |
| ``declare const v: `x`; v.replace("x", zz)`` | no report | reports |
| ``declare const v: `x` \| 1; …`` | no report | reports |
| ``declare const v: `x` & {}; …`` | no report | reports |
| ``type T = `x`; declare const v: T; …`` | no report | reports |
| ``function f<T extends `x`>(v: T) { … }`` | no report | reports |
| `declare const v: string & number; …` | reports | no report |
| `type S = string; declare const v: S & 1; …` | reports | no report |
| `function f(v: 'x' & 1) { … }` | reports | no report |

The first group is upstream treating every `TSLiteralType` whose literal is not a plain string literal as a non-string; a no-substitution template literal type is a string, so reporting is the correct outcome. The second group is intersections that reduce to `never` — the receiver is uninhabited, so staying quiet is fine.

Suggested wording for the doc:

> This rule uses type information to decide whether the receiver is a string, and is skipped on files where type information is unavailable.
>
> ## Differences from the original rule
>
> The original rule classifies the receiver syntactically before it consults type information. This rule always uses type information, so it reports receivers typed as a no-substitution template literal type (`` `x` ``), which are strings, and it stays quiet on receivers whose type reduces to `never`.

elecmonkey:
Done in 844805ef.

- `isKnownNonStringReceiver` keeps only the `unicornutil.ClassifyType` path.
  `classifyStringReceiverSyntax`, `classifyStringTypeSyntax`,
  `combineStringTypeSyntax`, `classifyStringTypeReference`,
  `combineStringTypeExpressions` and `bindingTypeAnnotation` are gone;
  `unicornutil/type_classifier.go` is untouched.
- The rule declares `RequiresTypeInfo: true`.
- `TestNoUnsafeStringReplacementSourceOnlyTypeSyntax` is deleted.
- The three `` `x` `` cases moved to invalid, plus ``type T = `x`; declare const v: T``
  and ``function run<T extends `x`>(value: T)``.
- The doc carries your wording, and the PR body now has the divergence table's
  content as a section.

Net -259 lines in the rule file. Upstream and extras suites are green, as are
the `RequiresTypeInfo` contract tests in `internal/config`. The JS-side rule
tester passes unchanged — the fixtures Program does supply type information for
the `.js` cases, as you said.

internal/plugins/unicorn/unicornutil/type_classifier.go
```diff
@@ -0,0 +1,175 @@
...
+// createTypeCheckers helper. Syntactic target/non-target recognition remains
+// with each rule because those shapes are domain-specific.
+type TypeClassifierOptions struct {
+	TargetTypeNames            *utils.Set[string]
+	NonTargetTypeNames         *utils.Set[string]
+	IsTargetType               func(*checker.Type) bool
+	HeritageSymbolFlags        ast.SymbolFlags
+	UnknownSymbolLessTypeFlags checker.TypeFlags
+	AllowNullishInMixedUnion   bool
+	TreatMixedUnionAsTarget    bool
```
swwind: `TreatMixedUnionAsTarget` and `TreatMixedUnionAsNonTarget` currently have no caller and no test coverage. They appear to be unused generalization left from extracting the classifier. Could we remove these two options and the corresponding branches in `classifyUnion` for now? The current consumers only need the default all-target / all-non-target behavior.

elecmonkey: Fixed in f268c19f. Removed both unused mixed-union options and simplified `classifyUnion` to retain the existing all-target / all-non-target behavior. The targeted tests for `unicornutil`, `no_unsafe_string_replacement`, and `prefer_array_some` all pass.

internal/plugins/unicorn/rules/no_unsafe_string_replacement/no_unsafe_string_replacement.go
```diff
@@ -0,0 +1,174 @@
...
+}
+
+// isKnownNonStringReceiver mirrors the type-information half of Unicorn's
+// createTypeCheckers. Unlike upstream there is no syntax-level fallback: the
+// rule declares RequiresTypeInfo, so a TypeChecker is always present.
+func isKnownNonStringReceiver(ctx rule.RuleContext, node *ast.Node) bool {
+	t := ctx.TypeChecker.GetTypeAtLocation(node)
+	return unicornutil.ClassifyType(ctx, t, unicornutil.TypeClassifierOptions{
+		HeritageSymbolFlags:        ast.SymbolFlagsClass | ast.SymbolFlagsInterface,
+		UnknownSymbolLessTypeFlags: checker.TypeFlagsNumberLiteral | checker.TypeFlagsBigIntLiteral,
```
swwind: Making this rule type-information-only introduces more divergences than the two documented ones. Upstream v73 classifies numeric/bigint literal annotations and tuple annotations as non-target before consulting the type checker, so all of these are skipped upstream but reported here:

```ts
declare const a: 1; a.replace("x", replacement);
declare const b: 1n; b.replace("x", replacement);
declare const c: [string]; c.replace("x", replacement);
declare const d: readonly [string]; d.replace("x", replacement);
```

The same mismatch occurs through aliases, generic constraints, unions, and named/optional tuples. `UnknownSymbolLessTypeFlags` makes the numeric literal types unknown, while symbol-less tuple types fall through `ClassifyType` as unknown. These receivers are provably non-string, so reporting them is also a false positive rather than only a parity discrepancy. Please either keep these categories non-target or explicitly document and cover the additional intentional divergences; the PR currently says dropping the syntax classifier diverges only for no-substitution template literal types and types reduced to `never`.

swwind: Please classify numeric/bigint literal types and tuple types as definitely non-string in the type-information-only classifier, and add regression tests; there is no need to restore the deleted syntax classifier.

elecmonkey: Done in ac0321e4. Numeric/bigint literal types and tuple types are now decided as non-string in the type-information-only classifier, and the deleted syntax classifier stays deleted.

- Dropped `UnknownSymbolLessTypeFlags` from the shared classifier. It had no other caller, so symbol-less number and bigint literal types fall through to the primitive/intrinsic arm and reach `TypeNonTarget` again.
- Added `NonTargetSymbolLessTypeFlags`, which this rule sets to `TypeFlagsObject`. Symbol-less object types — tuples among them, including `readonly`, named and optional elements — were reaching `TypeUnknown`, and this rule reports on undecided types. It is rule-scoped rather than the shared default because a tuple is an array receiver for `prefer-array-some` and a non-string receiver here.
- Regressions for `1`, `1n`, an alias, a generic constraint, `1 | 2`, `1 & {brand: true}`, the four tuple shapes, a tuple alias, a tuple constraint and `[string] | 1`; `1 | string` is added on the invalid side so the mixed union stays covered.

One consequence worth flagging, because it reverses the earlier rounds on literal types: a type-information-only classifier cannot tell `declare const value: 1` from `const value = 1`. They are the same type, and only the annotation upstream reads separates them. So `const value = 1`, `(1).replace()`, `(1n).replace()` and the narrowed `1 | 2 | string` case are now silent where ESLint reports them. Those are false negatives on receivers where calling `replace` is a TypeScript error in the first place, which seems the right side to err on given the alternative is the false positives above. All four moved to the valid list with a comment, and the rule doc now records this as a third deliberate divergence.

swwind (approved):
LGTM

PR #1787, opened 2026-08-19
feat: port rule yoda

swwind:
## Summary

Ports the ESLint core rule `yoda`, covering the `"never"` / `"always"` modes, the `exceptRange` and `onlyEquality` options, and the autofix that flips the comparison.

Range-test bounds are compared with JavaScript's abstract relational comparison semantics, so a BigInt bound compares exactly against a Number or String counterpart instead of going through a float64 round-trip.

`utils.NormalizeNumericLiteral` now normalizes hexadecimal literals, which its doc comment already promised (`"0x1" -> "1"`). Go's `ParseFloat` only reads hexadecimal floats written with a `p` exponent, so `0x1` previously came back unchanged and two literals spelling the same number did not compare equal.

The `exceptRange` examples in the rule doc differ from the ones in the upstream ESLint doc: a range test is exempt only when it is directly wrapped in parentheses, and two of the upstream "correct" examples are in fact reported by ESLint.

## Checklist


swwind:
Blocked by #1766

internal/rules/yoda/yoda.go
```diff
@@ -0,0 +1,500 @@
...
+
+// lessOrEqual reports whether v <= other, matching JS's abstract relational
+// comparison: two strings compare lexicographically, everything else
+// compares numerically (ToNumber-like coercion). A non-numeric string
+// bound (e.g. `'a'` compared against a numeric bound) coerces to NaN, and
+// any comparison against NaN is false — Go's float64 already has that
+// behavior built in, so no special case is needed here.
+func (v literalValue) lessOrEqual(other literalValue) bool {
+	if v.kind == "string" && other.kind == "string" {
+		return v.str <= other.str
```
elecmonkey: [P2] Use UTF-16 ordering for string range bounds

With `{ exceptRange: true }`, ESLint 10.8.0 exempts `if ("\ud83d\ude00" <= x && x < "\ue000") {}` because JavaScript compares strings by UTF-16 code units. Go's native string comparison orders UTF-8 bytes instead, so Rslint reports this ascending range and incorrectly exempts the reverse ordering; use ECMAScript code-unit comparison here.

internal/rules/yoda/yoda.go
```diff
@@ -0,0 +1,500 @@
...
+	rightRange := utils.TrimNodeTextRange(sf, bin.Right)
+
+	leftText := text[leftRange.Pos():leftRange.End()]
+	rightText := text[rightRange.Pos():rightRange.End()]
+	textBeforeOperator := text[leftRange.End():opRange.Pos()]
+	textAfterOperator := text[opRange.End():rightRange.Pos()]
+	flippedOperator := scanner.TokenToString(flipOperatorKind(bin.OperatorToken.Kind))
+
+	flipped := rightText + textBeforeOperator + flippedOperator + textAfterOperator + leftText
+	return utils.SafeReplacementText(sf, node, flipped)
```
elecmonkey, line 447: [P1] Preserve separation after a swapped regexp operand

On valid input `/a/ < x++instanceof C`, ESLint's fix is `x++ > /a/ instanceof C`, but this path produces `x++ > /a/instanceof C`, which is invalid on the next parse; `/a/ < f()in obj` similarly becomes `f() > /a/in obj`. `SafeReplacementText` does not recognize that a regexp token followed by `in` or `instanceof` needs whitespace, so the fix must preserve that token boundary.

internal/rules/yoda/yoda.go
```diff
@@ -0,0 +1,500 @@
...
+	case ast.KindNullKeyword:
+		return literalValue{kind: "number", num: 0}, true
+	case ast.KindRegularExpressionLiteral:
+		// ESLint's getNormalizedLiteral returns any `Literal` node verbatim,
+		// including regex literals — it only checks node.type, not the
+		// runtime type of node.value. JS's abstract relational comparison
+		// then coerces a RegExp operand via ToPrimitive, which falls back to
+		// RegExp.prototype.toString() (there is no valueOf override), i.e.
+		// its source text. Treat it as a string bound for the same reason.
+		return literalValue{kind: "string", str: node.Text()}, true
```
elecmonkey: [P2] Canonicalize RegExp values before ordering range bounds

For `{ exceptRange: true }`, ESLint 10.8.0 exempts `if (/a/mi <= x && x < /a/im) {}` because RegExp coercion uses `RegExp.prototype.toString()`, which canonicalizes both flag lists to `/a/im`. Keeping `node.Text()` preserves the source flag order, so Rslint treats these equivalent regexps as descending and reports the first comparison.

internal/rules/yoda/yoda.go
swwind: Fixed in 309bb402. Range-bound string comparisons now use ecmascript.CompareStrings, matching JavaScript's UTF-16 code-unit ordering. I added regression coverage for both the ascending astral-to-U+E000 case and the reversed descending case.

internal/rules/yoda/yoda.go
swwind: Fixed in 309bb402. The flipped-text builder now preserves a space when a moved regexp literal would otherwise touch a following `in` or `instanceof` token. I added regression coverage for both examples and verified the expected fixed output.

internal/rules/yoda/yoda.go
swwind: Fixed in 309bb402. RegExp range bounds now rebuild their string value with canonical flag ordering before comparison. I added a regression covering `/a/mi` and `/a/im` as equivalent bounds.

internal/rules/yoda/yoda.go
```diff
@@ -0,0 +1,514 @@
...
+	rightRange := utils.TrimNodeTextRange(sf, bin.Right)
+
+	leftText := text[leftRange.Pos():leftRange.End()]
+	rightText := text[rightRange.Pos():rightRange.End()]
+	textBeforeOperator := text[leftRange.End():opRange.Pos()]
+	textAfterOperator := text[opRange.End():rightRange.Pos()]
+	flippedOperator := scanner.TokenToString(flipOperatorKind(bin.OperatorToken.Kind))
+
+	flipped := rightText + textBeforeOperator + flippedOperator + textAfterOperator + leftText
+	flipped = utils.SafeReplacementText(sf, node, flipped)
```
elecmonkey: [P1] Preserve keyword separation after earlier template expressions

With the default `"never"` option, the valid input `` `${seed}`; function *f(){yield(1)<a} `` is fixed by ESLint 10.8.0 to `` `${seed}`; function *f(){yield a>(1)} ``, but Rslint produces `` `${seed}`; function *f(){yielda>(1)} ``, silently changing the generator's semantics. `SafeReplacementText` relies on `TokenBeforePosition`, whose scanner loses lexical state after a preceding interpolated template; the same issue also makes `` `${seed}`; if (0 <= x && x < 1) {}`` report despite `exceptRange: true`. Use parser-backed surrounding tokens or preserve the scanner's lexical state.

internal/rules/yoda/yoda.go
```diff
@@ -0,0 +1,514 @@
...
+	if bin.Left.Kind == ast.KindRegularExpressionLiteral {
+		nodeRange := utils.TrimNodeTextRange(sf, node)
+		if after, ok := utils.TokenAtOrAfter(sf, nodeRange.End()); ok &&
+			after.Start == nodeRange.End() &&
+			(after.Kind == ast.KindInKeyword || after.Kind == ast.KindInstanceOfKeyword) &&
```
elecmonkey: [P1] Separate moved regexps from TypeScript operators

For valid TypeScript such as `/a/ < (x)as number`, ESLint 10.8.0 with `@typescript-eslint/parser` 8.65.0 fixes the expression to `(x) > /a/ as number`, whereas Rslint emits `(x) > /a/as number`, which fails to parse; `satisfies` has the same problem. The added suffix guard handles only `in` and `instanceof`, but a moved regexp must also be separated from adjacent TypeScript word operators.

internal/rules/yoda/yoda.go
```diff
@@ -0,0 +1,514 @@
...
+		n.Neg(n)
+	}
+	return literalValue{kind: "bigint", big: n}, true
+}
+
+// isBetweenTest reports whether leftBin/rightBin form a "between" range test
+// like `0 <= x && x < 1`: the shared operand is leftBin.Right/rightBin.Left,
+// and the literal bounds (when both present) must be ascending.
+func isBetweenTest(leftBin, rightBin *ast.BinaryExpression) bool {
+	if !utils.IsSameReference(leftBin.Right, rightBin.Left, false) {
```
elecmonkey, line 341: [P2] Canonicalize regexp computed keys before comparison

With `["never", { exceptRange: true }]`, ESLint 10.8.0 accepts `if (0 <= a[/a/mi] && a[/a/im] < 2) {}` because converting each RegExp key to a property name canonicalizes both flag lists to `/a/im`; Rslint reports the first comparison because `IsSameReference` compares the raw computed-key spellings. The canonicalization used for range bounds does not cover member keys, so the reference comparison also needs the ECMAScript RegExp string value.

elecmonkey, line 341: [P2] Distinguish private fields from string-keyed properties

With `["never", { exceptRange: true }]`, ESLint 10.8.0 reports the first comparison in `class C { #x=0; m(){ if (0 <= this.#x && this['#x'] < 1) {} } }` because `this.#x` and `this['#x']` are different references, while Rslint suppresses it as a range. `IsSameReference`'s static-name fast path collapses both accesses to `#x`; private-property access must bypass that cross-syntax equivalence.

internal/rules/yoda/yoda.go
swwind: Fixed in b27d33a9. `TokenBeforePosition` now reads the parser-backed token stream, preserving lexical context after interpolated templates. Added regressions for both the `yield` autofix boundary and `exceptRange` recognition after a preceding template expression.

internal/rules/yoda/yoda.go
swwind: Fixed in b27d33a9. The moved-RegExp suffix guard now covers the TypeScript `as` and `satisfies` operators in addition to `in` and `instanceof`. Added focused regressions verifying that both fixed outputs retain the required space and parse successfully.

internal/rules/yoda/yoda.go
swwind: Fixed in b27d33a9. RegExp string coercion is now centralized in `RegExpLiteralStringValue`, which canonicalizes flag order for both range bounds and statically computed property keys. Added a regression for `/a/mi` and `/a/im` as equivalent computed keys.

swwind: Fixed in b27d33a9. `AccessExpressionStaticName` now excludes `PrivateIdentifier` property accesses from the ordinary static-name fast path, so `this.#x` and `this["#x"]` remain distinct while identical private accesses still match. Added focused regression coverage.

internal/rules/yoda/yoda.go
```diff
@@ -0,0 +1,511 @@
...
```
elecmonkey, line 341: [P2] Preserve JS numeric-key stringification for radix keys

With `{ exceptRange: true }`, `if (0 <= a[0x10000000000000801] && a["18446744073709556000"] < 1) {}` is reported by ESLint 10.8.0 because the numeric key stringifies to `18446744073709552000`, so the two member expressions are not the same reference. Rslint currently emits no diagnostic, since this range check delegates to `utils.IsSameReference`, whose explicit-radix key path accumulates into `float64` and rounds this value differently. That incorrectly suppresses yoda errors for large hex/octal/binary computed keys that are near IEEE-754 rounding boundaries.

swwind: Fixed after syncing with `origin/main` and covered in a3498676. The shared explicit-radix parser now performs multiplication and addition as separate `float64` operations, matching Acorn's per-operation JavaScript Number rounding. The reported hex key therefore stringifies to `18446744073709552000`, remains distinct from `"18446744073709556000"`, and Yoda reports and fixes the first comparison. I added the exact `exceptRange` regression case.

internal/rules/yoda/yoda.go
```diff
@@ -0,0 +1,511 @@
...
+	rightRange := utils.TrimNodeTextRange(sf, bin.Right)
+
+	leftText := text[leftRange.Pos():leftRange.End()]
+	rightText := text[rightRange.Pos():rightRange.End()]
+	textBeforeOperator := text[leftRange.End():opRange.Pos()]
+	textAfterOperator := text[opRange.End():rightRange.Pos()]
+	flippedOperator := scanner.TokenToString(flipOperatorKind(bin.OperatorToken.Kind))
+
+	flipped := rightText + textBeforeOperator + flippedOperator + textAfterOperator + leftText
+	flipped = utils.SafeReplacementText(sf, node, flipped)
```
elecmonkey: [P1] Preserve separators before escaped identifiers

When a comparison directly follows `yield`, `return`, or `throw` and the moved operand begins with an escaped identifier—for example `function *f(){yield(1) < \u0061}`—ESLint 10.8.0 fixes it to `yield \u0061 > (1)`, but this path emits `yield\u0061 > (1)`. The latter tokenizes as the identifier `yielda`, changing the program's semantics; the same failure occurs after `return` and `throw`. Add token-aware separation instead of relying only on the replacement's first raw rune.

internal/utils/tokens.go
```diff
@@ -68,67 +68,10 @@ func TokenAtOrAfter(sourceFile *ast.SourceFile, pos int) (SourceToken, bool) {
...
-
-	var found *ast.Node
-	node.ForEachChild(func(child *ast.Node) bool {
-		if child.Pos() <= pos && pos < child.End() {
-			found = parsedTokenContainingPosition(child, pos)
-		}
-		return found != nil
-	})
-	return found
+	return PreviousTokenBefore(sourceFile, sourceFile.AsNode(), pos)
```
elecmonkey: [P1] Avoid rebuilding the full token stream for every lookup

`TokenBeforePosition` now calls `PreviousTokenBefore` on the entire source file, which materializes every parser token before searching it. Repeated calls therefore allocate and traverse the whole file each time; on a 61,780-byte fixture with 2,000 comparisons, 2,000 `SafeReplacementText` calls took about 9.72 seconds and allocated 39.3 GB, versus 0.27 seconds and 2.6 MB with the previous implementation. Since yoda invokes this per fix and `exceptRange` invokes it while checking matching comparisons, large files become quadratic and can consume extreme memory; use a cached token index or a targeted parser-backed lookup.

internal/rules/yoda/yoda.go
```diff
@@ -0,0 +1,511 @@
...
+// normalizedLiteralValue extracts a comparable bound from a range-test
+// operand, mirroring ESLint's getNormalizedLiteral. Every literal is a bound,
+// including a boolean, null or regex one, because ESLint compares bounds with
+// `<=`, which coerces whatever value the literal carries. ok is false for
+// anything that isn't a literal (e.g. a variable or a call).
+func normalizedLiteralValue(node *ast.Node) (literalValue, bool) {
+	node = ast.SkipParentheses(node)
+	switch node.Kind {
+	case ast.KindNumericLiteral:
+		return numericValue(node.AsNumericLiteral().Text, false)
```
elecmonkey: [P2] Recover raw radix literals when comparing range bounds

With `["never", { "exceptRange": true }]`, ESLint 10.8.0 emits no diagnostic for `if (0x10000000000000801 <= x && x < 18446744073709552000) {}`, because its two parsed Number bounds are equal. Rslint reports and fixes the first comparison because ts-go's normalized `NumericLiteral.Text` is reparsed through `NormalizeNumericLiteral`, producing a different rounding result. The raw-token recovery already used for computed numeric property names should also be used when constructing numeric range bounds.

internal/rules/yoda/yoda.go
swwind: Fixed in a474a7d5f. `CanTokenTextsBeAdjacent` now uses the TypeScript scanner to recognize when a leading Unicode escape starts an identifier token, so fixes retain the required separator after `yield`, `return`, and `throw`. Added focused regressions for all three statement keywords.

internal/utils/tokens.go
swwind: Fixed in a474a7d5f. `TokenBeforePosition` now stores the parser-backed token index on the `SourceFile`, builds it only once, and uses binary search for subsequent lookups. Repeated lookups therefore reuse the same index instead of rebuilding and traversing the full token stream each time.

internal/rules/yoda/yoda.go
swwind: Fixed in a474a7d5f. Numeric range bounds now reuse `GetStaticExpressionValue`, recovering the raw explicit-radix token before applying the Acorn-compatible JavaScript Number rounding path. Added the exact `0x10000000000000801` range regression, which is now correctly exempt.

internal/rules/yoda/yoda.go
```diff
@@ -0,0 +1,519 @@
...
+
+	leftText := text[leftRange.Pos():leftRange.End()]
+	rightText := text[rightRange.Pos():rightRange.End()]
+	textBeforeOperator := text[leftRange.End():opRange.Pos()]
+	textAfterOperator := text[opRange.End():rightRange.Pos()]
+	flippedOperator := scanner.TokenToString(flipOperatorKind(bin.OperatorToken.Kind))
+
+	flipped := rightText + textBeforeOperator + flippedOperator + textAfterOperator + leftText
+	flipped = utils.SafeReplacementText(sf, node, flipped)
+	if bin.Left.Kind == ast.KindRegularExpressionLiteral {
```
elecmonkey: [P1] Separate trailing-dot literals from following word operators

With the default `never` mode, valid input `1. > x++instanceof C` is fixed by ESLint 10.8.0 to `x++ < 1. instanceof C`, while Rslint emits `x++ < 1.instanceof C`; the next fix pass reports TS1351 because a keyword cannot immediately follow the numeric literal. The same failure occurs before `in` and before TypeScript `as`/`satisfies`, since this suffix guard only considers regexp operands instead of checking the moved token generically.

internal/rules/yoda/yoda.go
```diff
@@ -0,0 +1,519 @@
...
+		n.Neg(n)
+	}
+	return literalValue{kind: "bigint", big: n}, true
+}
+
+// isBetweenTest reports whether leftBin/rightBin form a "between" range test
+// like `0 <= x && x < 1`: the shared operand is leftBin.Right/rightBin.Left,
+// and the literal bounds (when both present) must be ascending.
+func isBetweenTest(leftBin, rightBin *ast.BinaryExpression) bool {
+	if !utils.IsSameReference(leftBin.Right, rightBin.Left, false) {
```
elecmonkey, line 341: [P2] Compare shared radix literals by their parser Number value

With `["never", { "exceptRange": true }]`, ESLint 10.8.0 emits no diagnostic for `if (0 < 0x10000000000000801 && 18446744073709552000 < x) {}` because Espree parses the two shared literals to the same Number value. Rslint reports and fixes the second comparison because `IsSameReference` compares ts-go's normalized literal text without recovering the explicit-radix source; this path needs the same raw-token, Acorn-compatible conversion already used for range bounds and computed keys.

internal/rules/yoda/yoda.go
```diff
@@ -0,0 +1,519 @@
...
+	rightRange := utils.TrimNodeTextRange(sf, bin.Right)
+
+	leftText := text[leftRange.Pos():leftRange.End()]
+	rightText := text[rightRange.Pos():rightRange.End()]
+	textBeforeOperator := text[leftRange.End():opRange.Pos()]
+	textAfterOperator := text[opRange.End():rightRange.Pos()]
+	flippedOperator := scanner.TokenToString(flipOperatorKind(bin.OperatorToken.Kind))
+
+	flipped := rightText + textBeforeOperator + flippedOperator + textAfterOperator + leftText
+	flipped = utils.SafeReplacementText(sf, node, flipped)
```
elecmonkey: [P3] Preserve upstream spacing before a moved regexp literal

In `always` mode, ESLint 10.8.0 fixes `function f(){return(x) < /a/}` to `function f(){return /a/ > (x)}`, while this path produces `function f(){return/a/ > (x)}`; `yield` and `throw` behave the same way. Both forms parse equivalently, but the autofix output does not match upstream because the generic adjacency check treats a keyword followed by a regexp as safely adjacent.

internal/rules/yoda/yoda.go
swwind: Fixed. `SafeReplacementText` now treats a trailing decimal point followed by a word-like token as a required separator, so moved `1.` literals remain separated from `in`, `instanceof`, `as`, and `satisfies`. Focused regressions cover all four operators, and the complete fixed sources parse successfully.

internal/rules/yoda/yoda.go
swwind: Fixed. `IsSameReference` now compares numeric literals through `GetStaticExpressionValue`, which recovers explicit-radix source tokens and applies JavaScript Number rounding before comparison. The reported shared bounds are now recognized as equal, with focused rule and utility regressions.

internal/rules/yoda/yoda.go
swwind: Fixed. The shared token-adjacency check now preserves separation between word-like tokens and regexp delimiters, so `return`, `yield`, and `throw` retain the same spacing as ESLint when a regexp is moved after them. Focused regressions cover all three statements.

elecmonkey:
lgtm. plz fix ci fail, seems that other rules are effected

internal/utils/tokens.go
```diff
@@ -187,6 +142,18 @@ func CanTokenTextsBeAdjacent(left string, right string) bool {
...
 	}
+	if scanner.IsIdentifierPart(leftRune) && startsWithEscapedIdentifier(right) {
+		return false
+	}
+	// Keep a regexp delimiter distinct from a word-like token, and keep a
+	// trailing decimal point from swallowing the following word operator.
+	// These boundaries occur when fixes move regexp and `1.` literals next to
+	// statement keywords or operators such as `in`, `as`, and `satisfies`.
+	if (scanner.IsIdentifierPart(leftRune) && rightRune == '/') ||
+		(leftRune == '/' && scanner.IsIdentifierPart(rightRune)) ||
```
elecmonkey: [P1] Distinguish division punctuators from regexp literals

When `SafeReplacementText` receives a replacement or neighbor whose final token is division, this raw `leftRune == '/'` check treats the slash as a regexp delimiter. As a result, the existing `operator-assignment` case `foo/=bar` now fixes to `foo= foo/ bar` instead of ESLint 10.8.0's `foo= foo/bar`, and `prefer-numeric-literals` fixes `foo/parseInt("111", 2)` to `foo/ 0b111` instead of `foo/0b111`; both existing tests fail on Linux, macOS, and Windows. Please distinguish a complete regexp token from a division punctuator (or make this adjacency check token-aware) before inserting the space.

swwind: Fixed in 7f49b80b9. `CanTokenTextsBeAdjacent` now applies regexp-specific spacing only when parser-backed boundary tokens are complete valid regexp literals, so division punctuators remain adjacent. Added direct coverage for `/` on both sides and retained whole-expression regexp boundary cases. The affected `operator_assignment`, `prefer_numeric_literals`, `yoda`, and `internal/utils` packages pass on Linux with Go 1.26.0.

elecmonkey (approved):
よだ

PR #1788, opened 2026-08-19
feat: port rule no-underscore-dangle

swwind:
## Summary

Port the `no-underscore-dangle` rule from ESLint to rslint.

The rule disallows identifiers with a leading or trailing underscore. Variable declarations, function declaration names, and member access are checked by default; function parameters, destructured names, method names, and class field names are each governed by an option.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/no-underscore-dangle
- Source code: https://github.com/eslint/eslint/blob/v10.8.1/lib/rules/no-underscore-dangle.js

## Checklist


internal/rules/no_underscore_dangle/no_underscore_dangle.go
```diff
@@ -0,0 +1,442 @@
...
+// counts while a quoted `this['constructor']` does not.
+func isThisConstructorReference(object *ast.Node) bool {
+	if object.Kind != ast.KindPropertyAccessExpression && object.Kind != ast.KindElementAccessExpression {
+		return false
+	}
+	name, _, ok := estreeName(memberPropertyNode(object))
+	if !ok || name != "constructor" {
+		return false
+	}
+	return ast.SkipParentheses(memberObjectNode(object)).Kind == ast.KindThisKeyword
```
elecmonkey: **[P2] Do not unwrap optional `this?.constructor` receivers**

When `allowAfterThisConstructor` is enabled, `(this?.constructor)._bar` and `(this?.constructor)?._bar` are incorrectly exempted. ESLint v10.8.1 preserves the optional chain as the receiver and reports both accesses, but this helper strips the parentheses and treats the inner optional property access exactly like `this.constructor`, without checking its question-dot token. This causes valid diagnostics to disappear whenever parentheses terminate an optional `this?.constructor` chain.

swwind: Fixed in 3dc1412f.

Confirmed against ESLint v10.8.1: parentheses terminate the optional chain, so `(this?.constructor)._bar` gives the outer access a `ChainExpression` receiver rather than the `MemberExpression` that `isThisConstructorReference` matches, and upstream reports it. `ast.SkipParentheses` was stripping that boundary unconditionally.

Added `unwrapReceiver`, which stops at the parentheses when the node inside them is part of an optional chain, and routed both `checkMemberAccess` and `isThisConstructorReference` through it. Regression cases cover `(this?.constructor)._bar`, `(this?.constructor)?._bar`, `((this?.constructor))._bar` and `(this?.[constructor])._bar`, plus a valid case for `(this.constructor)?._bar`, which upstream still exempts.

elecmonkey (approved):
The implementation matches ESLint 10.8.1 across the upstream suite and targeted differential cases covering all options, TypeScript-specific syntax, optional chains, computed/private members, destructuring, comments, and diagnostic ranges. The previously reported optional-chain discrepancy is fixed, targeted Go tests and CI pass, and no remaining correctness issues were found.

PR #1789, opened 2026-08-19
feat(unicorn): port rule no-invalid-fetch-options

elecmonkey:
## Summary

Port the `unicorn/no-invalid-fetch-options` rule from ESLint to Rslint.

## Related Links

- ESLint rule: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v73.0.0/docs/rules/no-invalid-fetch-options.md
- Source code: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v73.0.0/rules/no-invalid-fetch-options.js

## Checklist

### DIFF
--- a/internal/plugins/unicorn/rules/no_invalid_fetch_options/no_invalid_fetch_options.go
+++ b/internal/plugins/unicorn/rules/no_invalid_fetch_options/no_invalid_fetch_options.go
@@ -1,0 +1,188 @@
+package no_invalid_fetch_options
+
+import (
+	"strings"
+
+	"github.com/microsoft/typescript-go/shim/ast"
+	"github.com/web-infra-dev/rslint/internal/rule"
+	"github.com/web-infra-dev/rslint/internal/utils"
+)
+
+const messageID = "no-invalid-fetch-options"
+
+func invalidBodyMessage(method string) rule.RuleMessage {
+	return rule.RuleMessage{
+		Id:          messageID,
+		Description: "\"body\" is not allowed when method is \"" + method + "\".",
+	}
+}
+
+func isIdentifierPropertyNamed(property *ast.Node, name string) bool {
+	if property == nil {
+		return false
+	}
+
+	switch property.Kind {
+	case ast.KindPropertyAssignment,
+		ast.KindShorthandPropertyAssignment,
+		ast.KindMethodDeclaration,
+		ast.KindGetAccessor,
+		ast.KindSetAccessor:
+		propertyName := property.Name()
+		return propertyName != nil && propertyName.Kind == ast.KindIdentifier &&
+			propertyName.AsIdentifier().Text == name
+	default:
+		return false
+	}
+}
+
+func findLastIdentifierProperty(properties []*ast.Node, name string) *ast.Node {
+	for i := len(properties) - 1; i >= 0; i-- {
+		if isIdentifierPropertyNamed(properties[i], name) {
+			return properties[i]
+		}
+	}
+	return nil
+}
+
+func propertyValue(property *ast.Node) *ast.Node {
+	if property == nil {
+		return nil
+	}
+
+	switch property.Kind {
+	case ast.KindPropertyAssignment:
+		return property.AsPropertyAssignment().Initializer
+	case ast.KindShorthandPropertyAssignment:
+		return property.Name()
+	default:
+		// ESTree represents object methods and accessors as Property nodes whose
+		// values are functions. Neither can be absent-body values or static strings.
+		return nil
+	}
+}
+
+func isEffectivelyAbsentBody(property *ast.Node) bool {
+	value := propertyValue(property)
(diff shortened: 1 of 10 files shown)


swwind:
This seems conflicts with #1766, hold short of merge queue.

elecmonkey:
Rebased this PR onto the latest `main` after #1766 was merged. The conflicts are resolved, and the new static-evaluation paths now use the ECMAScript string case and numeric-conversion helpers introduced by #1766. I also added regressions for ECMAScript whitespace handling and UTF-16 diagnostic columns. Local Go/JS tests and all pre-commit checks pass; CI is running now.

internal/plugins/unicorn/rules/no_invalid_fetch_options/no_invalid_fetch_options.go
```diff
@@ -0,0 +1,187 @@
...
+			if property.Kind == ast.KindSpreadAssignment {
+				return
+			}
+		}
+
+		ctx.ReportNode(bodyProperty.Name(), invalidBodyMessage("GET"))
+		return
+	}
+
+	method, valueKind := staticStrings.EvalStringValue(propertyValue(methodProperty))
```
swwind, line 152: [P2] Preserve upstream static-value coverage

Upstream uses `@eslint-community/eslint-utils`'s `getStaticValue`, but `EvalStringValue` currently supports only a subset of the constant expressions that it evaluates. For example, eslint-plugin-unicorn@73.0.0 reports `fetch(url, {method: "G".concat("ET"), body})` at `body`, while this implementation emits no diagnostic. The same false negative occurs for `"xGETy".substr(1, 3)`, `"GET".charAt(0) + "ET"`, `"GETx".slice(NaN, 3)`, and `+"71" === 71 ? "GET" : "POST"`.

Please extend the evaluator coverage needed by this rule and add these parity cases to the existing extras test.

elecmonkey: Fixed in 7d0bf019. `EvalStringValue` now folds the cases you listed:

- `String#concat`, `String#charAt` and `String#substr` join `slice` / `substring` in the static-call table, so `"G".concat("ET")`, `"GET".charAt(0) + "ET"` and `"xGETy".substr(1, 3)` fold. `substr` follows the Annex B length semantics, and all three index UTF-16 code units like the existing helpers.
- Unary `+`, `-` and `~` fold over any static operand. tsgo's evaluator folds them only when the operand is already a number, so `+"71"` reached the rule unfolded.
- `===` and `!==` fold over string, number, boolean, null and undefined operands. Values that compare by identity and bigints stay unknown, but comparing across kinds is still false, so `1n === 1` is not folded to `true`.

Coverage: `TestStaticStringEvaluator` gets the folding cases and the negative ones (unknown operand, `{} === {}`, cross-kind comparison), and the extras suite gets your four `fetch` cases.

Two things I left out deliberately:

- Full parity with `getStaticValue` is not a goal for this PR. It is a general constant folder over the whole language, and following it (`Array#map`, `Object.entries`, `RegExp`, ...) has no natural boundary. What this evaluator targets is the string-producing subset rules actually consume; the operators and methods above are that subset for the cases you found.
- `"GETx".slice(NaN, 3)` still does not fold, because `NaN` and `Infinity` are not resolved as global identifier references at all. That is a gap in identifier resolution rather than in the folding table, and I would rather fix it separately, where it can be covered once for every rule that uses the evaluator.

swwind, line 152: **[P2] Preserve the remaining upstream static-value coverage**

This still produces false negatives for many expressions that upstream evaluates through `getStaticValue`. For example, eslint-plugin-unicorn v73.0.0 reports all of the following, while rslint reports none:

```js
fetch(url, {method: " GET ".trim(), body})
fetch(url, {method: String.fromCodePoint(71, 69, 84), body})
fetch(url, {method: "GETx".slice(NaN, 3), body})
fetch(url, {method: Math.max(1, 2) === 2 ? "GET" : "POST", body})
```

A 121-case differential matrix against the actual pinned rule found 29 false negatives, covering additional upstream-supported paths such as `padStart`, `at`, `normalize`, `Object.values`, `decodeURI`, `parseInt`, `Number`, `charCodeAt`, and the global `Infinity` value. The evaluator added here only implements a small subset of `getStaticValue`'s supported globals and built-in calls, so the previous static-evaluation issue is not fully resolved.

Please preserve upstream evaluation semantics for these method expressions, or document and test every intentional divergence instead of silently treating them as unknown.

elecmonkey: should expand utils in new PR.

swwind (approved):
LGTM

PR #1792, opened 2026-08-19
feat(unicorn): port rule no-array-concat-in-loop

elecmonkey:
## Summary

Port the unicorn/no-array-concat-in-loop rule from eslint-plugin-unicorn to rslint.

The rule reports result = result.concat(…) inside a loop when result is a let or var accumulator initialized to an empty array literal and declared outside the loop body. Accumulators that live in the global scope of a script file are excluded.

## Credits

Port from eslint-plugin-unicorn's no-array-concat-in-loop.

## Related Links

- ESLint rule: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v73.0.0/docs/rules/no-array-concat-in-loop.md
- Source code: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v73.0.0/rules/no-array-concat-in-loop.js

## Checklist


swwind:
Duplicated implementation to #1605

internal/plugins/unicorn/rules/no_array_concat_in_loop/no_array_concat_in_loop.go
```diff
@@ -0,0 +1,146 @@
...
+	Id:          messageID,
+	Description: "Do not use `Array#concat()` to accumulate an array in a loop.",
+}
+
+var NoArrayConcatInLoopRule = rule.Rule{
+	Name:   "unicorn/no-array-concat-in-loop",
+	Schema: rule.EmptyArraySchema,
+	Run: func(ctx rule.RuleContext, options []any) rule.RuleListeners {
+		return rule.RuleListeners{
+			ast.KindBinaryExpression: func(node *ast.Node) {
```
swwind, line 22: ts-go represents destructuring defaults as `BinaryExpression` nodes with an `=` token, while ESTree represents them as `AssignmentPattern`. Listening to every equals binary therefore reports cases that the upstream rule never visits.

For example, eslint-plugin-unicorn v73.0.0 emits no rule diagnostic for:

```js
let result = [];
for (const x of xs) {
  [result = result.concat(x)] = source;
}
```

This implementation reports `concat`. The same false positive occurs for nested array patterns and renamed object properties such as:

```js
[[result = result.concat(x)]] = source;
({value: result = result.concat(x)} = source);
```

Please gate `checkAssignment` on this being an ESTree-equivalent `AssignmentExpression`, rather than a destructuring default, and add focused regression cases for these shapes.

internal/plugins/unicorn/rules/no_array_concat_in_loop/no_array_concat_in_loop.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+func checkAssignment(ctx rule.RuleContext, node *ast.Node) {
+	binary := node.AsBinaryExpression()
+	if binary == nil || binary.OperatorToken == nil ||
+		binary.OperatorToken.Kind != ast.KindEqualsToken {
+		return
+	}
+
+	left := ast.SkipParentheses(binary.Left)
```
swwind, line 47: The PR description says TypeScript assertions are skipped when matching the assignment target, but the left side currently uses only `ast.SkipParentheses`.

I checked all four claimed wrapper forms:

```ts
(result as string[]) = result.concat(x);
(result satisfies string[]) = result.concat(x);
result! = result.concat(x);
(<string[]>result) = result.concat(x);
```

Upstream intentionally reports none of them, and the current Rslint implementation also reports none, so the documented intentional difference is not present. If this extension is intended, please unwrap assertion wrappers on the assignment target after applying the AssignmentExpression/AssignmentPattern distinction above and add regression tests. Otherwise, please remove or correct the claim in the PR description.

internal/plugins/unicorn/rules/no_array_concat_in_loop/no_array_concat_in_loop.go
elecmonkey: Good catch — fixed. `checkAssignment` now bails on any equals expression that is itself an assignment target (`ast.IsAssignmentTarget`), which walks out through array literals, spread elements, property assignments and parentheses, so destructuring defaults no longer reach the `concat` match. Regression cases cover `[result = result.concat(x)] = source`, the nested array and renamed property shapes, `[...[result = ...]] = source`, and `for ([result = ...] of sources)`. An array literal that is *not* a destructuring target still holds a real assignment expression, so `sink = [result = result.concat(x)]` stays reported, and that is covered too.

internal/plugins/unicorn/rules/no_array_concat_in_loop/no_array_concat_in_loop.go
elecmonkey: You're right, and I've corrected the description rather than extending the implementation. Assertion wrappers on the left-hand side are not a valid assignment target in TypeScript to begin with, so unwrapping them would only affect code that already fails to compile. The bullet now claims assertion skipping only where it is actually implemented: the `concat()` receiver and the empty array initializer.

internal/plugins/unicorn/rules/no_array_concat_in_loop/no_array_concat_in_loop.go
```diff
@@ -0,0 +1,157 @@
...
+	initializer := utils.SkipAssertionsAndParens(declaration.Initializer)
+	if initializer == nil || !ast.IsEmptyArrayLiteral(initializer) {
+		return nil
+	}
+	return declarationNode
+}
+
+func nearestLoop(node *ast.Node) *ast.Node {
+	for ancestor := node.Parent; ancestor != nil; ancestor = ancestor.Parent {
+		if ast.IsFunctionLike(ancestor) {
```
swwind: `ast.IsFunctionLike` treats a ts-go `MethodDeclaration`, getter, or setter as a function boundary even when the assignment is inside its computed name or decorator. In ESTree those metadata expressions are outside the method's `FunctionExpression`, so eslint-plugin-unicorn v73 reports this case:

```js
let result = [];
for (const x of xs) {
  class C {
    [result = result.concat(x)]() {}
  }
}
```

Rslint emits no diagnostic. The same false negative affects object methods, getters/setters, static/async/generator methods, method decorators, and computed methods whose parameter shadows `result`.

Please stop only at the ESTree-equivalent function boundary: method bodies and parameters should remain boundaries, while computed names and member decorators should continue toward the enclosing loop and resolve in their outer evaluation scope. Please add focused cases for these adjacent forms.

swwind (commented):
The remaining “Intentional differences from upstream” bullet for TypeScript assertion wrappers is inaccurate. eslint-plugin-unicorn v73 already unwraps `TSAsExpression`, `TSSatisfiesExpression`, `TSNonNullExpression`, and `TSTypeAssertion` in its shared helper, and its upstream test suite includes these TypeScript cases. Rslint's wrapper handling is parity with upstream, not an intentional divergence. Please remove that bullet or describe it as an implementation detail instead.

internal/plugins/unicorn/rules/no_array_concat_in_loop/no_array_concat_in_loop.go
elecmonkey: Fixed in 8e56cb0d8b. `nearestLoop` now treats a method or accessor as a boundary only when the assignment is in its body or parameter list; computed names and member decorators continue to the enclosing scope. Added regression coverage for class/object computed methods, a static async generator, a getter, a member decorator, and a parameter-shadowed computed method. `go test ./internal/plugins/unicorn/rules/no_array_concat_in_loop` passes.

fansenze (approved):

PR #1793, opened 2026-08-19
feat(unicorn): port rule no-array-fill-with-reference-type

elecmonkey:
## Summary

Port the `unicorn/no-array-fill-with-reference-type` rule from ESLint to rslint.

## Type information

Reporting is decided syntactically: the fill value being an object literal,
array literal, class expression or `new` expression. Type information is only
consulted, when available, to skip receivers that are known non-arrays, so this
rule does not declare `RequiresTypeInfo` — without a TypeChecker it reports
more, not less, exactly like upstream with no parser services.
`unicorn/prefer-array-some`, which shares
`internal/plugins/unicorn/unicornutil/array_receiver.go`, documents the same
arrangement.

## Related Links

- ESLint rule: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v73.0.0/docs/rules/no-array-fill-with-reference-type.md
- Source code: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v73.0.0/rules/no-array-fill-with-reference-type.js

## Checklist


internal/plugins/unicorn/unicornutil/array_receiver.go
```diff
@@ -0,0 +1,580 @@
...
+	if symbol == nil {
+		if utils.IsTypeFlagSet(t, checker.TypeFlagsPrimitive|checker.TypeFlagsIntrinsic) {
+			return arrayClassNonTarget
+		}
+		return arrayClassUnknown
+	}
+	if targetNames.Has(symbol.Name) && isDefaultLibraryArrayTarget(ctx, symbol) {
+		return arrayClassTarget
+	}
+	if targetNames.Has("Array") && classifyArrayHeritage(ctx, t, targetNames, nonTargetNames) == arrayClassTarget {
```
swwind: Upstream's array checkers set `checkClassHeritage: false`, and `getTypeScriptType` gates its `getBaseTypes` walk behind that flag, so the type-information path never widens through heritage. The `extends` walk upstream does have is on the annotation path only (`getInterfaceType`), which this port already covers in `classifyArrayInterfaceNode` — and that one agrees: `declare const i: Items; i.fill({})` reports on both sides. It's receivers that only have type information (`get()`, `o.items`, `N.Items`) where this block makes the port report while ESLint is silent. Dropping it keeps the annotation cases and fixes the three false positives.

internal/plugins/unicorn/unicornutil/array_receiver.go
```diff
@@ -0,0 +1,580 @@
...
+	}
+
+	symbol := arrayTypeSymbol(t)
+	if symbol == nil {
+		if utils.IsTypeFlagSet(t, checker.TypeFlagsPrimitive|checker.TypeFlagsIntrinsic) {
+			return arrayClassNonTarget
+		}
+		return arrayClassUnknown
+	}
+	if targetNames.Has(symbol.Name) && isDefaultLibraryArrayTarget(ctx, symbol) {
```
swwind: `isDefaultLibraryArrayTarget` has no upstream counterpart — `getTypeScriptType` ends with a plain `targetTypeNames.has(typeName) ? target : nonTarget`. And it isn't merely a no-op when the gate fails: control falls through to `return arrayClassNonTarget` below, so a name-matched `Array` whose declaration isn't from a default lib flips from "definitely an array" to "definitely not an array". With a module-scope `interface Array<T>` ESLint reports and this port stays silent.

swwind (commented):
The syntactic paths look faithful; the divergences I found are all in the new type-information path of `classifyArrayType`. Checked against `eslint-plugin-unicorn@73.0.0` with `@typescript-eslint/parser` on a real project, and reproduced on this branch with `rule_tester` cases.

| Code | ESLint | This PR |
| --- | --- | --- |
| `interface Items extends Array<object> {} declare function get(): Items; get().fill({})` | no report | reports |
| `interface Items extends Array<object> {} declare const o: {items: Items}; o.items.fill({})` | no report | reports |
| `namespace N { export interface Items extends Array<object> {} } declare const i: N.Items; i.fill({})` | no report | reports |
| `interface Items extends Array<object> {} declare const i: Items; i.fill({})` | reports | reports |
| `export {}; interface Array<T> { fill(v: object): void } declare function get(): Array<object>; get().fill({})` | reports | no report |
| `function items(): object[] {} items.fill({})` | no report | reports |

Root causes inline. The same helpers now gate `prefer-array-some`, where `IsKnownNonIndexedCollection` guards a fix and `IsArray` guards the variable rewrite, so both directions matter there too.

internal/plugins/unicorn/unicornutil/array_receiver.go
```diff
@@ -0,0 +1,580 @@
...
+		return arrayClassUnknown
+	}
+	visitedSymbols[symbol] = true
+	defer delete(visitedSymbols, symbol)
+
+	declaration := symbol.Declarations[0]
+	if declaration == nil {
+		return arrayClassUnknown
+	}
+	if class := classifyArrayTypeNode(ctx, declaration.Type(), targetNames, nonTargetNames, map[*ast.Symbol]bool{}); class != arrayClassUnknown {
```
swwind: `Node.Type()` falls back to `FunctionLikeData().Type` for function-like declarations, i.e. the return type annotation. Upstream reads `definition.name?.typeAnnotation`, which a function declaration has none of, so it falls through to type information and classifies `items` by its function type (non-target). Here `function items(): object[] {} items.fill({})` is classified by what the function returns and reports. Restricting this to declaration kinds that actually carry a binding annotation (variable / parameter / property) would match.

internal/plugins/unicorn/rules/no_array_fill_with_reference_type/no_array_fill_with_reference_type.go
```diff
@@ -0,0 +1,127 @@
...
+// than the broader shared utility: only a plain identifier with exactly one
+// direct const VariableDeclaration definition is followed, and aliases are not
+// chased recursively.
+func getConstVariableInitializer(ctx rule.RuleContext, node *ast.Node) *ast.Node {
+	node = utils.SkipAssertionsAndParens(node)
+	if node == nil || !ast.IsIdentifier(node) || ctx.Refs == nil {
+		return nil
+	}
+
+	symbol := ctx.Refs.Resolve(node)
```
swwind: `Resolve` falls back to the TypeChecker and can return a symbol declared in another file, while upstream's `findVariable` is per-file scope — `array_receiver.go` uses `ResolveInFile` for the same job. Two non-module scripts, `const value = {}` in one and `array.fill(value)` in the other, would report here and not in ESLint. `ResolveInFile` would keep this consistent with the rest of the port.

internal/plugins/unicorn/unicornutil/array_receiver.go
```diff
@@ -0,0 +1,580 @@
...
+			return arrayClassTarget
+		}
+	}
+	return arrayClassUnknown
+}
+
+func combineArrayUnion(ctx rule.RuleContext, parts []*checker.Type, targetNames, nonTargetNames *utils.Set[string]) arrayClass {
+	classes := make([]arrayClass, 0, len(parts))
+	for _, part := range parts {
+		if utils.IsTypeFlagSet(part, checker.TypeFlagsNull|checker.TypeFlagsUndefined) {
```
swwind: Upstream's `combineUnionTypes` only filters nullish parts under `allowNullishInMixedUnion`, which the array checkers don't set; they take the `types.map(normalizeType)` branch, so nullish becomes non-target and still contributes to the vote. `object[] | undefined` is therefore `unknown` upstream but `arrayClassTarget` here. No visible difference for this rule (both report), but `IsArray` is what gates the variable rewrite in `prefer-array-some`.

elecmonkey: Fixed in 4f09d33. Nullish parts now normalize to non-target and participate in the union vote. Added a prefer-array-some regression for the variable-rewrite gate.

internal/plugins/unicorn/unicornutil/array_receiver.go
elecmonkey: Fixed in 4f09d33. Binding annotations are now limited to variable, parameter, and property declarations, so a function return type is no longer treated as an annotation on the function binding. Added regressions for both affected rules.

internal/plugins/unicorn/unicornutil/array_receiver.go
elecmonkey: Fixed in 4f09d33. Type-information classification now matches upstream's symbol-name check without requiring a default-library declaration. Added regressions here and in prefer-array-some.

internal/plugins/unicorn/unicornutil/array_receiver.go
elecmonkey: Fixed in 4f09d33. The TypeChecker path no longer walks interface heritage; explicit binding annotations still use classifyArrayInterfaceNode. Added regressions for return, property, qualified-name, and explicit-annotation receivers.

internal/plugins/unicorn/rules/no_array_fill_with_reference_type/no_array_fill_with_reference_type.go
elecmonkey: Fixed in 4f09d33. The rule-local const initializer lookup now uses ResolveInFile, with a two-file regression proving a TypeChecker-resolved script-global declaration is not followed.

internal/plugins/unicorn/unicornutil/array_receiver.go
```diff
@@ -0,0 +1,572 @@
...
+		ast.KindArrowFunction,
+		ast.KindClassExpression,
+		ast.KindTemplateExpression,
+		ast.KindNoSubstitutionTemplateLiteral:
+		return true
+	}
+	return false
+}
+
+func classifyArrayType(ctx rule.RuleContext, t *checker.Type, targetNames, nonTargetNames *utils.Set[string]) arrayClass {
```
swwind, line 470: **`nonTargetNames` is dead on the type-information path.** With the default-library gate gone, `classifyArrayType` only reads `targetNames` (`targetNames.Has("Array")` and `targetNames.Has(symbol.Name)`); `nonTargetNames` is threaded through `combineArrayUnion` / `combineArrayIntersection` and never read again. Dropping the parameter from those three signatures and their call sites is mechanical. The syntactic path still reads it in `classifyKnownArrayTypeName`, so only the type-checker side changes.

internal/plugins/unicorn/unicornutil/array_receiver.go
```diff
@@ -0,0 +1,572 @@
...
+// @typescript-eslint/scope-manager. Function-like declarations are excluded:
+// Node.Type() exposes their return type, but that is not an annotation on the
+// identifier binding.
+func arrayBindingTypeAnnotation(declaration *ast.Node) *ast.Node {
+	if declaration == nil {
+		return nil
+	}
+	switch declaration.Kind {
+	case ast.KindVariableDeclaration, ast.KindParameter,
+		ast.KindPropertyDeclaration, ast.KindPropertySignature:
```
swwind, line 216: **These two kinds are unreachable.** `arrayBindingTypeAnnotation` is only called with `symbol.Declarations[0]` for an identifier resolved through `ResolveInFile`, and a bare identifier never resolves to a class property or an interface member, so `KindPropertyDeclaration` / `KindPropertySignature` can't be hit. Upstream's `definition.name?.typeAnnotation` covers variable and parameter definitions only, which `case ast.KindVariableDeclaration, ast.KindParameter:` matches exactly.

swwind (commented):
Re-checked the fix commit against eslint-plugin-unicorn@73.0.0 with @typescript-eslint/parser: all five divergences now match ESLint on report, column and suggestion output, including the shapes the new cases don't cover (`filter().length > 0` and `findIndex(...) !== -1` on an interface-heritage receiver, a nullish union under `!`, an imported interface named `Array`, and `Coll as Array`). Reverting `ResolveInFile` to `Resolve` makes the new cross-file test fail, so it pins what it claims.

Nothing blocking. Two mechanical leftovers of the fix are inline — both are no-behavior-change cleanups, together +11/-12 in one file.

swwind (commented):
Same here, re-invent the TypeChecker is not a good idea for unicorn rules in rslint.

internal/plugins/unicorn/unicornutil/array_receiver.go
elecmonkey: Fixed in 06abff48. `classifyArrayType`, `combineArrayUnion` and
`combineArrayIntersection` no longer take a non-target name set;
`classifyKnownArrayTypeName` on the syntactic path still does. Added a comment
on `classifyArrayType` recording why it takes only `targetNames`, so it does not
come back.

internal/plugins/unicorn/unicornutil/array_receiver.go
elecmonkey: Fixed in 06abff48. Only `case ast.KindVariableDeclaration, ast.KindParameter:`
remains, and the doc comment now records why a bare identifier resolved through
`ResolveInFile` can never bind to a class property or an interface member.

elecmonkey:
On making this rule type-information-only as well: I would rather not, and I
have recorded the reasoning in the rule instead.

The two rules are not the same shape. In `no-unsafe-string-replacement` the
syntactic classifier ran before the checker and could veto it, it was inert on
JavaScript, and its only domain was TypeScript files with no type information —
dropping it was a clear win. Here the syntactic path does not veto the checker:
`classifyArrayInterfaceNode` is what makes `declare const i: Items; i.fill({})`
agree with ESLint, which is what we settled on earlier in this PR, and
`array_receiver.go` is shared with `prefer-array-some`.

`RequiresTypeInfo` would also cost more than it buys here. Type information only
ever suppresses in this rule: the report is decided by the fill value being an
object literal, array literal, class expression or `new` expression, and
`IsKnownNonArray` just removes non-array receivers. Without a checker the rule
reports more, not less, so the flag would disable it on gap files and LSP
inferred-project files in order to remove a false-positive class that barely
exists — nothing in the platform has a `fill()` that takes a reference value
(typed arrays take numbers, `CanvasRenderingContext2D#fill()` takes no reference
value). `prefer-array-some` documents the same arrangement and reaches the same
conclusion, and the two share `array_receiver.go`, so splitting them would be
inconsistent.

What did land is 06abff48: the two cleanups above, plus a comment on the rule
recording why it must not declare `RequiresTypeInfo`.

swwind (approved):
LGTM

PR #1794, opened 2026-08-19
feat: port rule logical-assignment-operators

swwind:
## Summary

Port the `logical-assignment-operators` rule from ESLint to rslint.

The rule reports assignments and logical expressions that can be written with `&&=`, `||=`, or `??=`, and — with `enforceForIfStatements` — `if` statements whose only purpose is to assign to the value they test. The `never` mode reports the shorthand and expands it back. A rewrite is applied automatically only where it cannot change how many times a getter or a setter runs; otherwise it is offered as a suggestion.

Also narrows `utils.AccessExpressionStaticName` so that a private name has no static property name, matching ESLint's `getStaticPropertyName`. Without that, `obj.#x` and `obj['#x']` compared equal as references.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/logical-assignment-operators
- Source code: https://github.com/eslint/eslint/blob/v10.8.1/lib/rules/logical-assignment-operators.js

## Checklist


internal/rules/logical_assignment_operators/logical_assignment_operators.go
```diff
@@ -0,0 +1,899 @@
...
+}
+
+// requiresOuterParenthesis reports whether an assignment written in place of
+// the logical expression at node would bind too loosely for its surroundings.
+func requiresOuterParenthesis(node *ast.Node) bool {
+	parent := estreeParent(node)
+	if parent == nil || parent.Kind == ast.KindExpressionStatement {
+		return false
+	}
+	parentPrecedence := utils.EslintLikePrecedence(parent)
```
elecmonkey, line 453: [P2] Preserve parentheses in destructuring defaults

When the logical expression is a destructuring default, such as `[x = a || (a = z)] = arr`, ts-go represents the surrounding assignment pattern as a binary assignment, so this precedence check returns false and Rslint fixes it to `[x = a ||= z] = arr`. ESLint v10.8.1 treats the parent as an `AssignmentPattern` and produces `[x = (a ||= z)] = arr`; distinguish this context so the autofix remains upstream-compatible.

swwind: Fixed in d7cf2048. I added a focused regression case for `[x = a || (a = z)] = arr` and now treat tsgo's binary assignment parent as ESTree's `AssignmentPattern` when it is a destructuring default, preserving the required parentheses. `go test ./internal/rules/logical_assignment_operators` passes.

internal/rules/logical_assignment_operators/logical_assignment_operators.go
```diff
@@ -0,0 +1,906 @@
...
+
+// isUndefined reports whether node is the global `undefined` or a void
+// expression on a zero-valued numeric literal (`void 0`).
+func isUndefined(ctx *rule.RuleContext, node *ast.Node) bool {
+	node = skipParens(node)
+	if node == nil {
+		return false
+	}
+	if node.Kind == ast.KindIdentifier && node.AsIdentifier().Text == "undefined" {
+		return ctx.Refs.IsGlobalNameReference(node, "undefined", globalValueMeaning)
```
elecmonkey: When `languageOptions.globals` or an inline `/* global undefined:off */` disables `undefined`, ESLint stops treating `undefined` as the built-in. This helper still asks only `RefStore`, which considers source declarations/wrapper bindings but not `ctx.Globals`, so `if (a == undefined) a = b` reports and autofixes to `a ??= b` even with `{ undefined: "off" }`. The `Boolean` path in `booleanCastArgument` has the same issue. Please gate these built-in lookups on the effective global being declared/enabled before calling `IsGlobalNameReference`.

internal/rules/logical_assignment_operators/logical_assignment_operators.go
```diff
@@ -0,0 +1,906 @@
...
+	}
+	call := node.AsCallExpression()
+	callee := skipParens(call.Expression)
+	if callee == nil || callee.Kind != ast.KindIdentifier || callee.AsIdentifier().Text != "Boolean" {
+		return nil, false
+	}
+	if call.Arguments == nil || len(call.Arguments.Nodes) != 1 {
+		return nil, false
+	}
+	if !ctx.Refs.IsGlobalNameReference(callee, "Boolean", globalValueMeaning) {
```
elecmonkey, line 354: [P2] Respect globals disabled by configuration

When `enforceForIfStatements` is enabled for a `.js` file and `languageOptions.globals` or a `/* global */` directive sets `Boolean` or `undefined` to `"off"`, ESLint v10.8.1 no longer treats that identifier as a global and does not report cases such as `if (Boolean(a)) a = b` or `if (a == undefined) a = b`. `IsGlobalNameReference` only checks source declarations, so this call and the equivalent one in `isUndefined` still report and can autofix those cases; also require `ctx.Globals.Access(name).IsDeclared()`.

internal/rules/logical_assignment_operators/logical_assignment_operators.go
```diff
@@ -0,0 +1,906 @@
...
+var LogicalAssignmentOperatorsRule = rule.Rule{
+	Name:   "logical-assignment-operators",
+	Schema: rule.NewSchema(schemaJSON),
+	Run: func(ctx rule.RuleContext, options []any) rule.RuleListeners {
+		opts := parseOptions(options)
+		state := &ruleState{
+			ctx: &ctx,
+			// Upstream reads the global scope's strictness, which a `"use
+			// strict"` directive inside a `with` body does not reach.
+			isStrict: ast.IsExternalModule(ctx.SourceFile) ||
```
elecmonkey: [P2] Derive strictness from ESLint's global scope

For TypeScript parsed as a module that also contains a recovered `with` statement, such as `export {}; with (object) a = a || b`, ESLint v10.8.1 with `@typescript-eslint/parser` treats the Program/global scope as non-strict and offers only a suggestion because `a` may resolve through the `with` object. `ast.IsExternalModule` forces `isStrict` to true here, so Rslint exposes an autofix instead; derive this flag from the effective ESLint global-scope semantics rather than tsgo's external-module marker.

internal/rules/logical_assignment_operators/logical_assignment_operators.go
swwind: Fixed in 4d71ab60. `isUndefined` now requires `undefined` to be enabled in `ctx.Globals` before checking whether source syntax shadows the global. I also added focused coverage for `languageOptions.globals` disabling `undefined` and an inline directive disabling `Boolean`. `go test ./internal/rules/logical_assignment_operators` passes.

internal/rules/logical_assignment_operators/logical_assignment_operators.go
swwind: Fixed in 4d71ab60. Built-in detection for both `Boolean` and `undefined` now checks `ctx.Globals.Access(name).IsDeclared()` before consulting `RefStore`, so configured and inline `off` values no longer report or autofix. Focused regressions cover both effective-global input paths, and `go test ./internal/rules/logical_assignment_operators` passes.

internal/rules/logical_assignment_operators/logical_assignment_operators.go
swwind: Fixed in 4d71ab60. Global-scope strictness now follows the top-level `"use strict"` directive instead of tsgo's external-module marker. A focused `export {}; with (object) a = a || b` regression verifies that the diagnostic carries only the upstream-compatible suggestion, while the existing `"use strict"` control still autofixes. `go test ./internal/rules/logical_assignment_operators` passes.

internal/rules/logical_assignment_operators/logical_assignment_operators.go
```diff
@@ -0,0 +1,907 @@
...
+					state.checkNever(node, node.AsBinaryExpression())
+				},
+			}
+		}
+
+		listeners := rule.RuleListeners{
+			ast.KindBinaryExpression: func(node *ast.Node) {
+				binary := node.AsBinaryExpression()
+				if binary.OperatorToken.Kind == ast.KindEqualsToken {
+					state.checkAssignment(node, binary)
```
elecmonkey, line 896: [P2] Skip destructuring defaults before checking assignments

With `[a = a || b] = array`, this listener treats ts-go’s default-value `BinaryExpression(=)` as an ESTree `AssignmentExpression`. ESLint 10.8.1 reports nothing because that node is an `AssignmentPattern`, but Rslint emits `assignment` and autofixes it to `[a ||= b] = array`, which is invalid syntax; nested object defaults and `for...of` patterns fail similarly. Please exclude `utils.IsDefaultValueInDestructuringAssignment(node)` before calling `checkAssignment`.

swwind: Fixed in `bc42668b`. The binary-expression listener now skips `utils.IsDefaultValueInDestructuringAssignment(node)` before calling `checkAssignment`. Focused valid cases cover array defaults, nested object defaults, and `for...of` destructuring; `go test ./internal/rules/logical_assignment_operators` passes.

internal/rules/logical_assignment_operators/logical_assignment_operators.go
```diff
@@ -0,0 +1,910 @@
...
+// isUndefined reports whether node is the global `undefined` or a void
+// expression on a zero-valued numeric literal (`void 0`).
+func isUndefined(ctx *rule.RuleContext, node *ast.Node) bool {
+	node = skipParens(node)
+	if node == nil {
+		return false
+	}
+	if node.Kind == ast.KindIdentifier && node.AsIdentifier().Text == "undefined" {
+		return ctx.Globals.Access("undefined").IsDeclared() &&
+			ctx.Refs.IsGlobalNameReference(node, "undefined", globalValueMeaning)
```
elecmonkey, line 237: [P2] Preserve built-in globals alongside module-scoped type declarations

In a TypeScript module, a top-level type-only declaration does not shadow the value-space built-in. For example, `type undefined = string; if (a === null || a === undefined) a = b` is reported and autofixed by ESLint 10.8.1 with `@typescript-eslint/parser` 8.65.0, but Rslint emits no diagnostic because `IsGlobalNameReference` treats any authored top-level declaration as hiding the global. `type Boolean = string; if (Boolean(a)) a = b` and interface declarations fail similarly. Please make this check respect value-space lookup in non-global Program scopes while retaining ESLint’s global-script behavior.

elecmonkey:
Plz resolve conflict.

internal/rules/logical_assignment_operators/logical_assignment_operators.go
swwind: Fixed in aceda4e1e. IsGlobalNameReference now uses the RefStore’s effective program-scope semantics rather than import/export syntax alone. A source configured as module therefore resolves value references past top-level type aliases and interfaces to the built-in undefined/Boolean, while sourceType: "script" retains ESLint’s merged global-scope behavior and value-producing namespace declarations still shadow the built-in. Focused regressions cover both built-ins, both source goals, and the namespace control; go test ./internal/rule and go test ./internal/rules/logical_assignment_operators pass.

swwind:
Resolved the conflict with origin/main in cf7fff3f6, retaining both the new default-param-last and logical-assignment-operators Rstest entries.

internal/rules/logical_assignment_operators/logical_assignment_operators.go
```diff
@@ -0,0 +1,910 @@
...
+	}
+	call := node.AsCallExpression()
+	callee := skipParens(call.Expression)
+	if callee == nil || callee.Kind != ast.KindIdentifier || callee.AsIdentifier().Text != "Boolean" {
+		return nil, false
+	}
+	if call.Arguments == nil || len(call.Arguments.Nodes) != 1 {
+		return nil, false
+	}
+	if !ctx.Globals.Access("Boolean").IsDeclared() ||
```
elecmonkey: [P2] Don’t gate Boolean casts on effective globals

For TypeScript-parsed files, ESLint 10.8.1 with `@typescript-eslint/parser` 8.65.0 still treats `Boolean(a)` as the global cast even when `Boolean` is disabled via `languageOptions.globals` or `/* global Boolean:off */`. For example, `/* global Boolean:off */ if (Boolean(a)) a = b` reports `messageId: "if"` and fixes to `/* global Boolean:off */ a &&= b`, but this check makes Rslint emit no diagnostic. Please avoid applying the effective-global `off` gate to the `Boolean` path, or otherwise mirror the TS-parser scope behavior.

swwind: Fixed in 6a4d406b. `booleanCastArgument` no longer gates the TypeScript scope-manager’s seeded `Boolean` binding on effective global access, while retaining the source value-space shadowing check and the separate effective-global behavior for `undefined`. Focused regressions cover both `languageOptions.globals` and `/* global Boolean:off */`. `go test ./internal/rules/logical_assignment_operators` passes.

internal/rule/ref_store.go
```diff
@@ -534,7 +534,7 @@ func (s *RefStore) IsGlobalNameReference(location *ast.Node, name string, meanin
 	if s == nil || location == nil || name == "" {
 		return false
 	}
-	if s.sourceFile != nil && ast.IsGlobalSourceFile(s.sourceFile.AsNode()) &&
+	if s.sourceFile != nil && !s.HasNonGlobalProgramScope() &&
```
elecmonkey, line 537: [P1] Update dependent tests for the new module-scope semantics

This shared behavior change makes `TestNoPromiseExecutorReturnExtras/valid-21` and `valid-22` fail because the default `file.ts` is treated as a module, where top-level type-only `Promise` declarations no longer hide the value-space global. `go test ./internal/rules/...` fails locally, and all three Go CI jobs fail on this head. ESLint 10.8.1 with `@typescript-eslint/parser` 8.65.0 confirms the new module behavior, so please update those existing cases to use explicit `sourceType: "script"` and add corresponding module invalid cases, or otherwise reconcile the shared helper's callers before merging.

elecmonkey:
conflict. plz resolve.

swwind:
> conflict. plz resolve.

fixed

elecmonkey (approved):
a

PR #1801, opened 2026-08-20
feat: port rule prefer-object-has-own

swwind:
## Summary

Port the `prefer-object-has-own` rule from ESLint to rslint.

The rule reports `Object.prototype.hasOwnProperty.call(object, property)` — along with the `Object.hasOwnProperty.call()` and `({}).hasOwnProperty.call()` spellings of it — and autofixes the callee to `Object.hasOwn`.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/prefer-object-has-own
- Source code: https://github.com/eslint/eslint/blob/v10.8.1/lib/rules/prefer-object-has-own.js

## Checklist


elecmonkey (approved):
The implementation matches ESLint 10.8.1 across the upstream suite and targeted differential cases covering optional chains, computed properties, shadowing, comments, diagnostics, and fixes. The only observed behavioral difference is the explicitly documented handling of TypeScript type-only bindings.

PR #1802, opened 2026-08-20
feat: port rule id-denylist

swwind:
## Summary

Port the `id-denylist` rule from ESLint to rslint.

The rule reports identifiers whose names appear in a configured deny list — variable and function declarations, object properties assigned during object creation, class fields and class methods. Function calls, property reads, and references to global variables the code does not declare are left alone.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/id-denylist
- Source code: https://github.com/eslint/eslint/blob/v10.8.1/lib/rules/id-denylist.js

## Checklist


elecmonkey (commented):
The rule diverges from ESLint 10.8.1 on parenthesized call children, non-null-wrapped destructuring targets, JSDoc syntax, and synthesized CommonJS/JSDoc bindings. These produce both false positives and incorrect global classification on valid inputs.

internal/rules/id_denylist/id_denylist.go
```diff
@@ -0,0 +1,333 @@
...
+	// it potentially creates a new property with a restricted name.
+	if parent.Kind == ast.KindPropertyAccessExpression &&
+		parent.AsPropertyAccessExpression().Name() == node {
+		return isAssignmentTarget(parent)
+	}
+
+	// A callee or an argument of a call or a `new` is never checked. tsgo
+	// spells `import(specifier, options)` as a CallExpression too, where
+	// upstream has a distinct ImportExpression that this exclusion misses.
+	switch parent.Kind {
```
elecmonkey: **[P2] Skip through parentheses before applying call exclusions**

When a direct call/new child is parenthesized, its immediate ts-go parent is `ParenthesizedExpression`, so this switch no longer applies. For example, ESLint 10.8.1 emits no diagnostics for `(foo)()` or `foo((bar))` when those names are denied, while this implementation reports `foo` and `bar`; this breaks the rule's documented exemption for callees and arguments whenever users add parentheses.

internal/rules/id_denylist/id_denylist.go
```diff
@@ -0,0 +1,333 @@
...
+// has to be found by walking out. The walk climbs only the nodes a pattern is
+// built from and stops at the first parent that consumes the literal as a value
+// — `[{ a: obj.b }.c] = d` destructures the result of a member access, not the
+// literal the member access reads from. An update expression is not one of
+// upstream's arms, so it does not count as an enclosing assignment here.
+func isDestructuredFrom(node *ast.Node) bool {
+	if node == nil || node.Parent == nil {
+		return false
+	}
+	assignment := ast.GetAssignmentTarget(node)
```
elecmonkey, line 208: **[P2] Stop destructuring walks at TypeScript non-null wrappers**

For TypeScript assignment targets such as `[foo.bar]! = source`, `ast.GetAssignmentTarget` climbs through the `NonNullExpression`, so this classifies `bar` as a destructured write and reports it. `@typescript-eslint/parser` retains the non-null wrapper, so ESLint 10.8.1 reports only `foo` here; the property remains a read rather than one of `isAssignmentTarget`'s pattern positions.

internal/rules/id_denylist/id_denylist.go
```diff
@@ -0,0 +1,333 @@
...
+	switch parent.Kind {
+	case ast.KindCallExpression:
+		if !ast.IsImportCall(parent) {
+			return false
+		}
+	case ast.KindNewExpression:
+		return false
+	}
+
+	return !isRenamedImport(node) &&
```
elecmonkey, line 151: **[P2] Exclude identifiers parsed from JSDoc comments**

On JavaScript files, ts-go exposes identifiers parsed from JSDoc to the traversal, but ESLint treats the same text only as a comment. With `Foo` denied, `/** @type {Foo} */ let x;` produces no ESLint 10.8.1 diagnostic, whereas this final filter reports `Foo` inside the comment; `utils.IsInJSDocSyntax(node)` needs to be excluded before reporting.

internal/rules/id_denylist/id_denylist.go
```diff
@@ -0,0 +1,333 @@
...
+	if ctx.Refs == nil || !isVariableReference(node) {
+		return false
+	}
+	name := node.Text()
+	if !ctx.Globals.Access(name).IsDeclared() {
+		return false
+	}
+	// A name this file declares in the reference's own declaration space
+	// resolves to that declaration instead of to the global.
+	if ctx.Refs.ResolveInFile(node) != nil {
```
elecmonkey: **[P2] Do not treat CommonJS wrapper bindings as declarations**

When linting a `.cjs` file, ts-go can resolve CommonJS wrapper names such as `exports` to synthesized bindings, although ESLint's commonjs scope gives those variables zero source definitions. Consequently, with `exports` denied, ESLint 10.8.1 accepts `exports.foo = 1`, while this code treats the resolved wrapper as source-declared and reports `exports`; the test must distinguish authored declarations from wrapper bindings.

internal/rules/id_denylist/id_denylist.go
```diff
@@ -0,0 +1,333 @@
...
+	}
+	// In a script every top-level declaration shares the one global scope, so a
+	// single one of them takes the name away from the global for the whole
+	// file: `interface Number {}` at the top of a script also claims a later
+	// `Number` used as a value, and a `Number` read from inside a function. A
+	// module keeps its top level to itself, and a declaration nested in a scope
+	// of its own reaches only what that scope encloses; the resolution above
+	// already settles both.
+	if ctx.SourceFile != nil && ast.IsGlobalSourceFile(ctx.SourceFile.AsNode()) &&
+		ctx.SourceFile.Locals[name] != nil {
```
elecmonkey: **[P2] Ignore JSDoc-only symbols in the script-global check**

In a script such as `/** @typedef {number} Number */ Number;`, the `Number` entry in `SourceFile.Locals` is synthesized from JSDoc rather than declared by executable source. ESLint 10.8.1 therefore still recognizes the final `Number` as the built-in global and emits nothing, while this unconditional locals check disables the global exemption and reports it; only authored declarations should count here.

internal/rules/id_denylist/id_denylist.go
swwind: This seems to be conflicts with #1877

swwind: blocked by #1877

internal/rules/id_denylist/id_denylist.go
swwind: Fixed in `be4d9d65`. `shouldCheck` now walks through parenthesized wrappers before determining whether an identifier is a direct child of a call or `new` expression. Expanded regressions cover deeply parenthesized and optional calls while confirming that member expressions, TypeScript assertion wrappers, and dynamic imports remain checked.

internal/rules/id_denylist/id_denylist.go
swwind: Fixed in `be4d9d65`. Destructuring detection now rejects paths that cross a TypeScript `NonNullExpression`, so the property in `[foo.bar]! = source` remains a read and is not reported. Expanded coverage checks non-null wrappers around array and object patterns, around members, and inside member receivers.

internal/rules/id_denylist/id_denylist.go
swwind: Fixed in `be4d9d65`. `shouldCheck` now excludes identifiers for which `utils.IsInJSDocSyntax` is true before applying ordinary ESTree identifier classification. Expanded checked-JavaScript coverage includes `@type`, `@param`, `@template`, qualified `@typedef`, and real authored declarations following JSDoc.

internal/rules/id_denylist/id_denylist.go
swwind: Fixed in `be4d9d65`. `RefStore` now distinguishes authored declarations from synthesized `SourceFile` wrapper bindings, so implicit CommonJS `exports` remains global while an authored `const exports` still shadows it. Expanded coverage confirms the corresponding `module`, `require`, `global`, `arguments`, `__filename`, and `__dirname` boundaries.

internal/rules/id_denylist/id_denylist.go
swwind: Fixed in `be4d9d65`. Global-reference classification now ignores declarations synthesized only from JSDoc while retaining authored source declarations as shadows. Expanded coverage includes JSDoc-only built-in names, `@import`, and an authored `@enum` declaration counterexample.

internal/rule/ref_store.go
```diff
@@ -538,20 +538,33 @@ func (s *RefStore) IsGlobalNameReference(location *ast.Node, name string, meanin
 		hasAuthoredDeclaration(s.sourceFile.Locals[name]) {
 		return false
 	}
-	return !s.IsNameDefinedInFileWithMeaning(location, name, meaning)
+	if hasAuthoredDeclaration(s.resolveName(location, name, meaning)) {
```
elecmonkey, line 547: [P2] Preserve authored imports during global-reference checks

When a JavaScript file contains a JSDoc import followed by a real import of the same name, `resolveName` can return the synthesized JSDoc symbol even though `hasAuthoredImportBinding` confirms an authored binding. The new `hasAuthoredDeclaration` check then discards that symbol and classifies later references as global. For `/** @import { Promise } from "x" */
import { Promise } from "y";
Promise;`, ESLint 10.8.1 reports both the real import and final reference, while Rslint reports only the import. This also regresses the existing `no-promise-executor-return` test for the same scope shape, which passes on the merge base and fails on this PR.

internal/rules/id_denylist/id_denylist.go
```diff
@@ -0,0 +1,333 @@
...
+// variables.
+func isReferenceToGlobalVariable(ctx rule.RuleContext, node *ast.Node) bool {
+	if ctx.Refs == nil || !isVariableReference(node) {
+		return false
+	}
+	name := node.Text()
+	if !ctx.Globals.Access(name).IsDeclared() {
+		return false
+	}
+	return ctx.Refs.IsGlobalReference(node)
```
elecmonkey, line 265: [P2] Honor configured module scope in global-reference checks

The new rule delegates global classification to `RefStore`, but that helper derives the top-level scope only from module syntax and cannot see `ctx.LanguageOptions.SourceType`. With `sourceType: "module"`, `interface Promise {} Promise;`, and `Promise` denied, ESLint 10.8.1 reports only the interface declaration because the value reference still resolves to the outer language global; Rslint also reports the final reference because ts-go considers the syntax-only file global. Authored module mode must affect this classification even when the file contains no import or export.

elecmonkey, line 265: [P2] Retain dotted namespace definitions in script global scope

For a script containing `namespace Promise.Inner {} Promise;`, ESLint 10.8.1 reports only the authored `Promise` namespace name; its global-scope definition prevents the final value reference from being treated as an uncontrollable built-in. Rslint reports both occurrences because the shared resolver intentionally steps over qualified namespace segments and the global-source locals check does not recover this declaration shape. The rule therefore produces a false positive for a valid TypeScript namespace form.

internal/rule/ref_store.go
swwind: Fixed in 6dfe721e. `IsGlobalNameReference` now treats the authored-import source scan as authoritative when ts-go retains only the same-name JSDoc-synthesized declaration on the resolved symbol. Focused `RefStore` and `id-denylist` regressions cover the JSDoc/import collision, and the existing `no-promise-executor-return` regression passes. `go test ./internal/rule`, `go test ./internal/rules/id_denylist`, and `go test ./internal/rules/no_promise_executor_return` pass.

internal/rules/id_denylist/id_denylist.go
swwind: Fixed in 6dfe721e. The global-reference fast path now uses `RefStore.HasNonGlobalProgramScope()`, so configured `sourceType: "module"` takes effect even without import/export syntax. A focused `interface Promise {} Promise;` regression verifies that only the interface declaration is reported, while the existing script-global cases now state their source goal explicitly. `go test ./internal/rule` and `go test ./internal/rules/id_denylist` pass.

swwind: Fixed in 6dfe721e. The global-program locals fast path now excludes symbols made only from qualified namespace segments, allowing the normal resolver to preserve the upstream global classification for the final `Promise` reference. A focused `sourceType: "script"` regression verifies that `namespace Promise.Inner {} Promise;` reports only the authored namespace name. `go test ./internal/rule` and `go test ./internal/rules/id_denylist` pass.

internal/rule/ref_store.go
```diff
@@ -524,34 +524,65 @@ func functionExpressionSelfBinding(fn *ast.Node, name string, meaning ast.Symbol
...
+	return hasAuthoredDeclaration(symbol) && !isQualifiedNamespaceNameOnly(symbol)
+}
+
+// IsGlobalReference is IsGlobalNameReference using the declaration-space
+// meaning implied by node's syntactic reference position.
+func (s *RefStore) IsGlobalReference(node *ast.Node) bool {
+	if s == nil || node == nil || node.Kind != ast.KindIdentifier || !isReferencePosition(node) {
+		return false
+	}
+	return s.IsGlobalNameReference(node, node.Text(), referenceMeaning(node))
```
elecmonkey: [P2] Preserve namespace bindings for value references

In module scope, `referenceMeaning(node)` omits `SymbolFlagsNamespace`, so a namespace containing no value member is ignored and the subsequent reference is mistaken for the language global. With `sourceType: "module"`, `namespace Number {} Number;`, and `Number` denied, ESLint 10.8.1 reports both identifiers, while Rslint reports only the namespace declaration. Include namespace meaning when resolving these value references, as the existing `no-promise-executor-return` caller already does.

internal/rule/ref_store.go
```diff
@@ -811,7 +842,9 @@ func referenceMeaning(n *ast.Node) ast.SymbolFlags {
 		return ast.SymbolFlagsNamespace | ast.SymbolFlagsAlias
 	}
 	if ast.IsExpressionNode(n) {
-		return ast.SymbolFlagsValue | ast.SymbolFlagsAlias
+		// A namespace declaration can be referenced as a value expression
+		// (`namespace N {}; N`), even when it contains no value member.
+		return ast.SymbolFlagsValue | ast.SymbolFlagsNamespace | ast.SymbolFlagsAlias
```
elecmonkey: [P1] Keep namespace meaning scoped to global-reference checks

Adding `SymbolFlagsNamespace` to every expression lookup fixes the `id-denylist` namespace case, but it changes shared `RefStore.Resolve` semantics for every consumer. The existing `no_misleading_character_class` `invalid-90` case now reports both `RegExp` calls instead of only the final call, while ESLint 10.8.1 reports one; all three Go CI jobs fail on this regression. Reverting this line restores that test, and passing `Value | Namespace | Alias` only from `IsGlobalReference` makes the `id-denylist` namespace cases and the existing rule pass together.

internal/rules/id_denylist/id_denylist.go
```diff
@@ -0,0 +1,333 @@
...
+		return false
+	}
+
+	// Member access has special rules for checking property names. Read access
+	// to a property with a restricted name is allowed, because it can be on an
+	// object that user has no control over. Write access isn't allowed, because
+	// it potentially creates a new property with a restricted name.
+	if parent.Kind == ast.KindPropertyAccessExpression &&
+		parent.AsPropertyAccessExpression().Name() == node {
+		return isAssignmentTarget(parent)
```
elecmonkey, line 131: [P2] Preserve the optional-chain boundary for property writes

With `foo?.bar = baz` and all three names denied, ESLint 10.8.1 reports `foo` and `baz` but not `bar` because the member remains inside a `ChainExpression`. ts-go flattens that wrapper, so `isAssignmentTarget` returns true here and Rslint emits an extra `bar` diagnostic; the same mismatch occurs for `+=` and `foo?.bar.baz = qux`. Return false while this property access remains part of an optional chain, while retaining the current behavior when parentheses terminate the chain.

internal/rules/id_denylist/id_denylist.go
```diff
@@ -0,0 +1,333 @@
...
+			}
+			ctx.ReportNode(node, rule.RuleMessage{
+				Id:          "restricted",
+				Description: fmt.Sprintf("Identifier '%s' is restricted.", name),
+			})
+		}
+
+		return rule.RuleListeners{
+			ast.KindIdentifier:        check,
+			ast.KindPrivateIdentifier: check,
```
elecmonkey, line 67: [P2] Visit class constructors

With `constructor` denied, ESLint 10.8.1 reports the identifier in `class C { constructor() {} }`, but Rslint reports nothing because ts-go represents it as `KindConstructor` and this listener set only visits `Identifier` and `PrivateIdentifier`. The same false negative occurs for a private constructor. Add a constructor listener that reports the keyword range while continuing to ignore string-literal-named methods.

elecmonkey (approved):

PR #1804, opened 2026-08-20
feat: port rule id-match

swwind:
## Summary

Port the `id-match` rule from ESLint to rslint.

The rule requires every identifier to match a configured regular expression, with options to extend the check to property names (`properties`), class fields (`classFields`), and to narrow it to declarations (`onlyDeclarations`) or to skip names a destructuring pattern merely repeats (`ignoreDestructuring`).

TypeScript names are checked as identifiers; names the file does not declare — globals and standard-library types such as `Record` — are left alone.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/id-match
- Source code: https://github.com/eslint/eslint/blob/v10.8.1/lib/rules/id-match.js

## Checklist


internal/rules/id_match/id_match.schema.json
```diff
@@ -0,0 +1,28 @@
+{
+  "type": "array",
+  "items": [
+    {
+      "type": "string"
```
elecmonkey: **Reject invalid regular expressions during config validation**

When the configured pattern is invalid, such as `"("`, ESLint fails while loading `id-match`, but this schema accepts it and `Run` silently returns no listeners after compilation fails. This makes a typo disable the rule without any error; declare the string with `"format": "regex"` so normal configuration validation rejects it.

internal/rules/id_match/id_match.go
```diff
@@ -0,0 +1,795 @@
...
+//
+// NOTE: Unlike ESLint, whose scope model has no types in it at all, rslint has
+// to answer this to stay usable on TypeScript: without it every `Record<K, V>`
+// in the project is reported against a pattern its author never chose. A name
+// the file itself declares or imports still counts as the author's.
+func (r *idMatch) isExternallyDeclaredType(node *ast.Node) bool {
+	if r.ctx.Refs == nil || r.ctx.SourceFile == nil || !ast.IsPartOfTypeNode(node) {
+		return false
+	}
+	symbol := r.ctx.Refs.Resolve(node)
```
elecmonkey, line 363: **Preserve built-in type exemptions without a checker**

For a TypeScript file loaded through a source-only Program, `Refs.Resolve` has no checker fallback, so this returns `false` for standard-library types. Consequently `let x: Record<string, number>;` with pattern `^x$` reports `Record` when the file is outside a tsconfig, while the same file under a compiler-backed Program is clean, contradicting the documented exemption and making diagnostics depend on project ownership.

internal/rules/id_match/id_match.go
```diff
@@ -0,0 +1,795 @@
...
+	start := node.Pos()
+	if modifiers := node.ModifierNodes(); len(modifiers) > 0 {
+		start = modifiers[len(modifiers)-1].End()
+	}
+	text := r.ctx.SourceFile.Text()
+	start = scanner.SkipTrivia(text, start)
+	end := start + len(constructorKeyword)
+	// `class A { 'constructor'() {} }` is a constructor too, but its name is a
+	// string literal, which ESLint's AST holds as a `Literal` and never visits.
+	if end > len(text) || text[start:end] != constructorKeyword {
```
elecmonkey, line 403: **Report escaped constructor spellings**

For valid JavaScript such as `class foo { constr\u0075ctor() {} }` with pattern `^foo$`, ts-go still produces `KindConstructor`, but this raw-text comparison rejects it and Rslint emits nothing. ESLint v10.8.1 treats the decoded key as the identifier `constructor` and reports it, so constructor detection must not require the source spelling to equal the literal ASCII keyword.

internal/rules/id_match/id_match.schema.json
swwind: Fixed in 6cdb9bf1 — the pattern now declares `"format": "regex"`, matching what every other rule taking a regex string option does.

One caveat worth recording: schema validation compiles the value with no flags while the rule compiles it with `u`, so a pattern that is legal without `u` and illegal with it still reaches `Run` and still checks nothing. The branch test that used to pin that guard with `(` now uses `\p` instead, which is exactly that configuration.

internal/rules/id_match/id_match.go
swwind: Confirmed and fixed in 6cdb9bf1.

Reproduced first: a lint target no tsconfig owns is linted through a parsed-only Program, which supplies no TypeChecker, and a lib declaration reaches `Refs.Resolve` only through its checker fallback. So `let x: Record<string, number>;` with `^x$` reported `Record` there and stayed clean under a tsconfig.

`isExternallyDeclaredType` now answers the standard library without asking for a symbol at all: a name in a type position that is in `rule.IsDefaultTypeScriptTypeGlobal` and that the file does not declare itself is exempt. That list is the type-capable global scope typescript-eslint's scope manager seeds, which is where its own model finds these names, so the answer no longer depends on how rslint happened to build the Program. The checker-backed path stays for everything else; where a checker is present it already gave the same answer for these names.

`id_match_extras_noproject_test.go` pins this against a real parsed-only Program: the lib type, a lib type as a type argument, a `Record` the file redeclares (reported), `Record` in a value position (reported), and a type nothing declares (reported).

internal/rules/id_match/id_match.go
swwind: This one does not reproduce — ts-go does not produce a `KindConstructor` for the escaped spelling.

`constr\u0075ctor` is TS1260 "Keywords cannot contain escape characters", in `.js` as well as `.ts`, and `shouldSkipRulesForSyntax` in internal/linter/linter.go skips every rule for a file that has syntactic diagnostics. Linting that source through a parsed-only Program with listeners on `KindConstructor`, `KindMethodDeclaration` and `KindClassDeclaration` fires none of them.

The raw-text comparison is still load-bearing for the case it was written for: ts-go does give `class foo { 'constructor'() {} }` a `KindConstructor`, and dropping the comparison makes both the `'constructor'()` and `"constructor"()` cases report a name ESLint's AST holds as a `Literal` and never visits.

internal/rules/id_match/id_match.go
```diff
@@ -0,0 +1,807 @@
...
+	if !r.opts.properties {
+		return
+	}
+	if object := ast.SkipParentheses(member.Expression()); object.Kind == ast.KindIdentifier &&
+		object.Text() == name {
+		r.reportIfInvalid(node, name)
+		return
+	}
+
+	effective := estreeParent(member)
```
elecmonkey, line 227: When TypeScript's optional-chain assignment syntax is accepted, e.g. `a_1?.b_1 = a_2` with `{ properties: true }`, typescript-eslint wraps the member in a `ChainExpression`, so ESLint reports `a_1` and `a_2` but not the property `b_1`. `estreeParent(member)` skips the corresponding ts-go wrapper and makes the member look directly assigned, causing an extra diagnostic for `b_1`. Preserve the chain boundary when deciding whether the property is the assignment target.

internal/rules/id_match/id_match.go
```diff
@@ -0,0 +1,807 @@
...
+	switch {
+	case isImportBinding(parent):
+		// Only the local imported identifier is checked; the name the module
+		// exports it under is not the reader's to choose. Upstream compares the
+		// two by name, so `import { a as a }` reports both halves.
+		if local := parent.Name(); local != nil && local.Kind == ast.KindIdentifier &&
+			local.Text() == name {
+			r.reportIfInvalid(node, name)
+		}
+	case parent.Kind == ast.KindPropertyDeclaration && !ast.HasAccessorModifier(parent):
```
elecmonkey: For a TypeScript abstract property such as `abstract class C { abstract a_1: string; }`, typescript-eslint emits `TSAbstractPropertyDefinition`, not `PropertyDefinition`, so upstream reports `a_1` even when `classFields` is false. This branch treats every non-accessor `PropertyDeclaration` as a class field and suppresses that diagnostic unless `classFields` is enabled; private abstract properties are suppressed similarly in `checkPrivateIdentifier`. Distinguish abstract property definitions from ordinary fields.

internal/rules/id_match/id_match.go
```diff
@@ -0,0 +1,807 @@
...
+	case ast.KindNewExpression:
+		return parentNew
+	case ast.KindVariableDeclaration:
+		// tsgo writes a catch parameter as a variable declaration; ESTree
+		// hangs it off the catch clause, which is no declaration at all.
+		if parent.Parent != nil && parent.Parent.Kind == ast.KindCatchClause {
+			return parentOther
+		}
+		return parentVariableDeclarator
+	case ast.KindFunctionDeclaration:
```
elecmonkey, line 495: With `onlyDeclarations: true`, a bodyless TypeScript signature such as `function f_1(): void;` is represented by typescript-eslint as `TSDeclareFunction`, so ESLint does not report `f_1`; this classifier maps every ts-go `KindFunctionDeclaration` to the allowed `FunctionDeclaration` category and reports it. The same creates an extra diagnostic for each overload signature before an implementation. Check that the declaration has a body before returning `parentFunctionDeclaration`.

internal/rules/id_match/id_match.go
swwind: Confirmed and fixed in 454dc372.

`checkMemberAccess` now preserves ts-go's optional-chain marker as the ESTree `ChainExpression` boundary after checking the receiver, so an optional member is no longer treated as the direct assignment target. `TestIdMatchExtrasDim4` covers the reported direct assignment, a nested optional chain, and the parenthesized boundary where the chain ends.

`go test ./internal/rules/id_match` passes.

internal/rules/id_match/id_match.go
swwind: Confirmed and fixed in 454dc372.

Abstract property declarations are now excluded from the ordinary class-field gate, matching typescript-eslint's `TSAbstractPropertyDefinition` behavior even when `classFields` is false. `TestIdMatchExtrasBranches` covers both public and private abstract properties.

`go test ./internal/rules/id_match` passes.

internal/rules/id_match/id_match.go
swwind: Confirmed and fixed in 454dc372.

`classifyParent` now classifies a ts-go function declaration as an ESTree `FunctionDeclaration` only when it has a body. Bodyless signatures therefore remain `TSDeclareFunction`-equivalent and are skipped by `onlyDeclarations`, while the overload implementation is still reported. `TestIdMatchExtrasDim4` covers both the standalone signature and mixed overload case.

`go test ./internal/rules/id_match` passes.

internal/rules/id_match/id_match.go
```diff
@@ -0,0 +1,818 @@
...
+	if r.ctx.SourceFile == nil || !ast.IsPartOfTypeNode(node) {
+		return false
+	}
+	// The standard library's own names are answered without asking for a
+	// symbol, because a lib declaration reaches Resolve only through its
+	// TypeChecker fallback: asking there alone reports `Record` in a file no
+	// tsconfig owns and stays quiet in one a tsconfig does. This list is the
+	// type-capable global scope typescript-eslint seeds, which is where its
+	// own scope model finds these names.
+	if rule.IsDefaultTypeScriptTypeGlobal(name) && !r.isDeclaredInFile(node, name) {
```
elecmonkey: **Restrict built-in exemptions to standalone type names**

When a qualified type member happens to share a standard-library name, this unconditional name-based exemption suppresses an authored identifier. For example, with `type X = Foo.Record<string>;` and pattern `^(X|Foo|string)$`, ESLint 10.8.1 with `@typescript-eslint/parser` reports `Record`, while Rslint reports nothing; the same occurs for members such as `Foo.Array` and `Foo.Partial`. Restrict the built-in fallback to standalone global type references rather than the right-hand member of a qualified name.

swwind: Confirmed and fixed in 04d6c3cc.

`isExternallyDeclaredType` now applies the built-in TypeScript type fallback only when the identifier is not the right-hand member of a qualified type name. This keeps standalone global types such as `Record` exempt in a source-only Program, while `Foo.Record`, `Foo.Array`, and `Foo.Partial` remain authored identifiers and are reported.

`TestIdMatchExtrasNoProject` covers the standalone and qualified cases, and `go test ./internal/rules/id_match` passes.

internal/rules/id_match/id_match.go
```diff
@@ -0,0 +1,829 @@
...
+var IdMatchRule = rule.Rule{
+	Name:   "id-match",
+	Schema: rule.NewSchema(schemaJSON),
+	Run: func(ctx rule.RuleContext, options []any) rule.RuleListeners {
+		opts := parseOptions(options)
+		pattern, err := esregexp.Compile(opts.pattern, "u")
+		if err != nil {
+			// The configured pattern is not a regexp; nothing can be checked
+			// against it.
+			return rule.RuleListeners{}
```
elecmonkey, line 33: [P2] Reject regexes that are invalid with the `u` flag

The schema's `"format": "regex"` check compiles without flags, so a pattern such as `"\\p"` reaches this branch even though the rule compiles it with `u`. ESLint 10.8.1 throws `Invalid regular expression: /\p/u` while loading `id-match`, whereas Rslint silently returns no listeners and disables the rule for the file. Please validate with the same `u` semantics or surface the compilation error instead of swallowing it.

internal/rules/id_match/id_match.go
```diff
@@ -0,0 +1,829 @@
...
+	}
+	// The standard library's own names are answered without asking for a
+	// symbol, because a lib declaration reaches Resolve only through its
+	// TypeChecker fallback: asking there alone reports `Record` in a file no
+	// tsconfig owns and stays quiet in one a tsconfig does. This list is the
+	// type-capable global scope typescript-eslint seeds, which is where its
+	// own scope model finds these names.
+	if rule.IsDefaultTypeScriptTypeGlobal(name) && !isQualifiedTypeMember(node) &&
+		!r.isDeclaredInFile(node, name) {
+		return true
```
elecmonkey, line 361: [P2] Restrict the type-global exemption to actual global references

This name-based fallback also exempts authored identifiers for which `ResolveInFile` necessarily returns nil. With `type X = [Record: string]` and pattern `^(X|string)$`, or `type X = import('./m').Record` and pattern `^X$`, ESLint 10.8.1 with `@typescript-eslint/parser` 8.65.0 reports `Record`, while Rslint reports nothing in both source-only and checker-backed runs. Please distinguish standalone library references from tuple labels and import-type members rather than relying only on the spelling and qualified-name right side.

internal/rules/id_match/id_match.go
swwind: This is an old problem, ignore here.

internal/rules/id_match/id_match.go
swwind: Confirmed and fixed in f9f93966, with expanded import-type coverage in f8bab65c. The default TypeScript type-global fallback now requires an actual reference position: declaration and property names are rejected by `isNonReferenceIdentifier`, and complete import-type qualifiers are treated as imported-module metadata rather than globals. The same boundary now also applies to the configured/runtime-global exemption, so `import("./m").Array.Record` reports both authored members while a real type argument such as `Partial<string>` remains exempt. Source-only and checker-backed regressions cover the reported tuple-label and import-member cases plus the chained qualifier boundary, and `go test -count=1 ./internal/rules/id_match` passes.

internal/rules/id_match/id_match.go
```diff
@@ -0,0 +1,832 @@
...
+	// symbol, because a lib declaration reaches Resolve only through its
+	// TypeChecker fallback: asking there alone reports `Record` in a file no
+	// tsconfig owns and stays quiet in one a tsconfig does. This list is the
+	// type-capable global scope typescript-eslint seeds, which is where its
+	// own scope model finds these names.
+	if rule.IsDefaultTypeScriptTypeGlobal(name) && isTypeGlobalReference(node) &&
+		!r.isDeclaredInFile(node, name) {
+		return true
+	}
+	symbol := r.ctx.Refs.Resolve(node)
```
elecmonkey, line 363: [P2] Report resolved import-type members

When the imported module resolves, `type X = import("./foo").T;` with pattern `^X$` produces no diagnostic because this checker-backed branch resolves `T` to `foo.ts` and treats it as an externally declared type. ESLint 10.8.1 reports `T`, and Rslint also reports it when the module is unresolved or the file is linted through a source-only Program, so the result incorrectly depends on module resolution. Please exclude import-type qualifiers from this external-type exemption.

swwind: Confirmed and fixed in 8a390d449 after merging the latest `origin/main`.

`isExternallyDeclaredType` now requires an actual type-global reference before either the built-in fallback or checker-backed external-declaration exemption runs. A resolved import-type qualifier such as `import("./foo").T` therefore remains an authored module member and is reported consistently with unresolved and source-only runs, while genuine global types such as `Record` remain exempt.

`TestIdMatchExtrasTypescript` now covers the resolved-module case using the existing `foo.ts` fixture, and `go test -count=1 ./internal/rules/id_match` passes.

internal/rules/id_match/id_match.go
```diff
@@ -0,0 +1,831 @@
...
+	if r.ctx.SourceFile == nil || !ast.IsPartOfTypeNode(node) || !isTypeGlobalReference(node) {
+		return false
+	}
+	// The standard library's own names are answered without asking for a
+	// symbol, because a lib declaration reaches Resolve only through its
+	// TypeChecker fallback: asking there alone reports `Record` in a file no
+	// tsconfig owns and stays quiet in one a tsconfig does. This list is the
+	// type-capable global scope typescript-eslint seeds, which is where its
+	// own scope model finds these names.
+	if rule.IsDefaultTypeScriptTypeGlobal(name) && !r.isDeclaredInFile(node, name) {
```
elecmonkey: **[P2] Respect the configured TypeScript libs when exempting globals**

When `parserOptions.lib` is restricted, this unconditional `esnext` fallback suppresses identifiers that upstream reports. For example, with `lib: ["es5"]`, `type X = Disposable;` and pattern `^X$`, ESLint 10.8.1 with `@typescript-eslint/parser` 8.65.0 reports `Disposable`, while Rslint reports nothing; the same false negative occurs with `noLib`. Since `Disposable` is not provided by those configurations, please derive this exemption from the active lib or an actually resolved external symbol instead of the fixed default `esnext` name set.

swwind: fixed

internal/rules/id_match/id_match.go
```diff
@@ -0,0 +1,838 @@
...
+const constructorKeyword = "constructor"
+
+// IdMatchRule requires identifiers to match a specified regular expression.
+// https://eslint.org/docs/latest/rules/id-match
+var IdMatchRule = rule.Rule{
+	Name:   "id-match",
+	Schema: rule.NewSchema(schemaJSON),
+	Run: func(ctx rule.RuleContext, options []any) rule.RuleListeners {
+		opts := parseOptions(options)
+		pattern, err := esregexp.Compile(opts.pattern, "u")
```
elecmonkey, line 29: [P2] Do not silently disable invalid Unicode regexes

The schema's `"format": "regex"` rejects a pattern such as `"("`, but it accepts `"\\p"` because validation does not apply the `u` flag. `Run` then compiles the same pattern with `u`, fails, and returns no listeners. ESLint 10.8.1 throws while loading `id-match`, whereas Rslint accepts the configuration and emits no rule diagnostics. Please validate with the same Unicode semantics or surface the compilation failure instead of silently disabling the rule.

internal/rules/id_match/id_match.go
```diff
@@ -0,0 +1,838 @@
...
+	start := node.Pos()
+	if modifiers := node.ModifierNodes(); len(modifiers) > 0 {
+		start = modifiers[len(modifiers)-1].End()
+	}
+	text := r.ctx.SourceFile.Text()
+	start = scanner.SkipTrivia(text, start)
+	end := start + len(constructorKeyword)
+	// `class A { 'constructor'() {} }` is a constructor too, but its name is a
+	// string literal, which ESLint's AST holds as a `Literal` and never visits.
+	if end > len(text) || text[start:end] != constructorKeyword {
```
elecmonkey, line 403: [P2] Preserve escaped constructor identifiers

`class foo { constr\u0075ctor() {} }` is valid ECMAScript, and ESLint 10.8.1 reports the decoded `constructor` identifier. At the current head, ts-go emits TS1260 and Rslint suppresses native rule execution; this source-spelling guard would also return without reporting if traversal reached it. The earlier constructor review is therefore not fixed. Please ensure this valid spelling reaches the rule and is checked as `constructor`, or explicitly retain and document it as an accepted parser divergence.

internal/rules/id_match/id_match.go
swwind: omg, please skip this issue. otherwise we can only invent a new "format" to make it compile with "u" flag.

elecmonkey (dismissed):

elecmonkey (approved):

PR #1811, opened 2026-08-20
fix: report no-global-assign writes through a single type assertion

swwind:
## Summary

`(Object as any) = 1` and `(<any>Object) = 1` assign to a read-only global, but the rule did not report them.

ESLint's scope analysis unwraps a plain `=` assignment target exactly once — an `AsExpression`, `TypeAssertionExpression` or `NonNullExpression` — before asking whether what remains is an assignment pattern. One wrapper therefore still leaves a write, while `((Object as any) as any) = 1` and `Object!! = 1` do not. Compound assignments, update expressions and destructuring patterns never take that path and recognize a wrapped target however deeply it is nested.

The old guard rejected an `as`/`<T>` wrapper at any depth instead. It missed the single-wrapper writes, and `[Object as any] = arr` and `(Object as any) += 1` with them, while accepting `Object!! = 1`. `satisfies` is still never a write at any depth, which `IsWriteReference` already answers.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/no-global-assign

## Checklist


internal/rules/no_global_assign/no_global_assign.go
```diff
@@ -35,21 +35,32 @@ func parseOptions(rawOptions []any) options {
...
+	for parent := current.Parent; parent != nil; parent = current.Parent {
+		switch parent.Kind {
+		case ast.KindAsExpression, ast.KindTypeAssertionExpression, ast.KindNonNullExpression:
+			wrappers++
+		case ast.KindParenthesizedExpression:
+		case ast.KindBinaryExpression:
+			binary := parent.AsBinaryExpression()
+			return wrappers > 1 && binary != nil && binary.OperatorToken != nil &&
+				binary.OperatorToken.Kind == ast.KindEqualsToken && binary.Left == current
 		default:
```
elecmonkey, line 73: [P2] Preserve suppression for stacked compound and update targets

When two or more TypeScript wrappers surround a compound-assignment or update target, such as `((Object as any) as any) += 1` or `((Object as any) as any)++`, ESLint 10.8.1 with `@typescript-eslint/parser` 8.65.0 creates no write reference because only the outer wrapper is unwrapped. This helper instead returns false for every non-`=` binary expression and every unary parent, so Rslint newly reports these cases even though the previous guard suppressed them.

internal/rules/no_global_assign/no_global_assign.go
```diff
@@ -35,21 +35,32 @@ func parseOptions(rawOptions []any) options {
...
+	current := node
+	for parent := current.Parent; parent != nil; parent = current.Parent {
+		switch parent.Kind {
+		case ast.KindAsExpression, ast.KindTypeAssertionExpression, ast.KindNonNullExpression:
+			wrappers++
+		case ast.KindParenthesizedExpression:
+		case ast.KindBinaryExpression:
+			binary := parent.AsBinaryExpression()
+			return wrappers > 1 && binary != nil && binary.OperatorToken != nil &&
+				binary.OperatorToken.Kind == ast.KindEqualsToken && binary.Left == current
```
elecmonkey: [P2] Continue past destructuring default assignments

For a destructuring default whose target has stacked non-null wrappers, such as `[Object!! = 0] = arr`, ESLint reports the write because the inner `=` is an assignment pattern visited as part of the enclosing destructuring target. The helper mistakes that inner binary expression for the top-level plain assignment and returns true, causing Rslint to suppress a diagnostic that the previous implementation emitted; the walk needs to distinguish a destructuring default from the enclosing assignment.

internal/rules/no_global_assign/no_global_assign.go
swwind: Fixed in 106275f6. `hasStackedTypeWrappers` now applies the one-wrapper limit to all direct assignment operators and prefix/postfix updates. Added regressions for `((Object as any) as any) += 1` and `((Object as any) as any)++`; both remain unreported, matching ESLint.

internal/rules/no_global_assign/no_global_assign.go
swwind: Fixed in 106275f6. The helper now recognizes `=` nodes used as destructuring defaults via `utils.IsDefaultValueInDestructuringAssignment` and does not treat them as direct assignment boundaries. Added a regression for `[Object!! = 0] = arr`; it reports `Object` as ESLint does.

internal/rules/no_global_assign/no_global_assign.go
```diff
@@ -35,21 +35,45 @@ func parseOptions(rawOptions []any) options {
...
+		switch parent.Kind {
+		case ast.KindAsExpression, ast.KindTypeAssertionExpression, ast.KindNonNullExpression:
+			wrappers++
+		case ast.KindParenthesizedExpression:
+		case ast.KindBinaryExpression:
+			binary := parent.AsBinaryExpression()
+			if binary == nil || binary.OperatorToken == nil || binary.Left != current {
+				return false
+			}
+			if utils.IsDefaultValueInDestructuringAssignment(parent) {
```
elecmonkey: [P2] Keep walking past parenthesized destructuring defaults

When a destructuring default is parenthesized, `IsDefaultValueInDestructuringAssignment(parent)` returns false because the `=` node's direct parent is a `ParenthesizedExpression`; the subsequent `wrappers > 1` check therefore suppresses the write. For example, ESLint 10.8.x with `@typescript-eslint/parser` 8.65.0 reports `Object` in `[(Object!! = 0)] = arr;`, while this patch reports nothing. Parenthesized defaults in object patterns and `for` destructuring are affected as well, so the walk needs to recognize the enclosing destructuring context through parentheses.

swwind: Fixed in f675f23d. `IsDefaultValueInDestructuringAssignment` now skips parenthesized-expression wrappers before checking the enclosing array or object destructuring context, so `hasStackedTypeWrappers` continues past parenthesized defaults instead of suppressing the write. Added regressions for `[(Object!! = 0)] = arr;`, `({x: (Object!! = 0)} = src);`, and `for ([(Object!! = 0)] of rows) {}`; all report `Object` as ESLint does.

internal/utils/write_reference.go
```diff
@@ -251,7 +251,12 @@ func IsDefaultValueInDestructuringAssignment(node *ast.Node) bool {
 	if binary == nil || binary.OperatorToken == nil || binary.OperatorToken.Kind != ast.KindEqualsToken {
 		return false
 	}
-	parent := node.Parent
+	current := node
+	parent := current.Parent
+	for parent != nil && parent.Kind == ast.KindParenthesizedExpression {
```
elecmonkey: [P2] Recognize wrapped destructuring defaults

When a destructuring default is itself wrapped by a TypeScript expression, such as `[(Object!! = 0) as any] = arr;`, this loop stops at the `AsExpression`, so the helper returns false and `hasStackedTypeWrappers` mistakes the inner `=` for a direct assignment with stacked wrappers. ESLint 10.8.0 with `@typescript-eslint/parser` 8.65.0 reports `Object` at column 3, while the built PR head reports nothing; the pre-PR guard also reports it. The same regression occurs with `<any>`, non-null, object-pattern, and `for-of` variants. Skipping these transparent TS wrappers while locating the surrounding destructuring pattern fixes the reproducer and passes the focused tests for all callers of this helper.

internal/utils/write_reference.go
```diff
@@ -259,7 +259,22 @@ func IsDefaultValueInDestructuringAssignment(node *ast.Node) bool {
...
-	target := ast.GetAssignmentTarget(node)
+	assignmentTargetNode := node
+transparentWrappers:
+	for parent := assignmentTargetNode.Parent; parent != nil; parent = assignmentTargetNode.Parent {
+		switch parent.Kind {
+		case ast.KindParenthesizedExpression,
+			ast.KindAsExpression,
+			ast.KindTypeAssertionExpression,
+			ast.KindNonNullExpression,
+			ast.KindSatisfiesExpression:
```
elecmonkey: [P2] Keep expression wrappers out of AssignmentPattern detection

Skipping `AsExpression`, `TypeAssertionExpression`, `NonNullExpression`, and `SatisfiesExpression` here makes this shared predicate classify ordinary `AssignmentExpression` nodes as destructuring defaults. For `[(foo = foo || bar) as any] = arr`, ESLint 10.9.0 with `@typescript-eslint/parser` 8.65.0 reports and fixes `logical-assignment-operators`, while the PR-head Rslint reports nothing; `[(foo = this) as any] = arr` similarly loses the expected `unicorn/no-this-assignment` diagnostic. These wrappers remain expression nodes in ESTree, so this broadened traversal should be kept local to `no-global-assign` or the helper must distinguish actual `AssignmentPattern` contexts.

swwind: Fixed in b1fa0c8a9. Restored `IsDefaultValueInDestructuringAssignment` so it crosses only parenthesized expressions; TypeScript expression wrappers are now handled locally by `no-global-assign`’s scope-reference emulation. Added regressions confirming `[(foo = foo || bar) as any] = arr` is reported and fixed by `logical-assignment-operators`, and `[(foo = this) as any] = arr` is reported by `unicorn/no-this-assignment`.

internal/rules/no_global_assign/no_global_assign.go
```diff
@@ -35,25 +35,71 @@ func parseOptions(rawOptions []any) options {
...
+	for parent := assignmentTargetNode.Parent; parent != nil; parent = assignmentTargetNode.Parent {
+		switch parent.Kind {
+		case ast.KindParenthesizedExpression,
+			ast.KindAsExpression,
+			ast.KindTypeAssertionExpression,
+			ast.KindNonNullExpression,
+			ast.KindSatisfiesExpression:
+			assignmentTargetNode = parent
+		default:
+			return utils.IsDefaultValueInDestructuringAssignment(assignmentTargetNode)
```
elecmonkey: [P2] Keep wrapped defaults recognizable as destructuring defaults

When the default assignment is wrapped, this helper climbs from the inner `=` to the outer wrapper and then passes that wrapper into `utils.IsDefaultValueInDestructuringAssignment`, but the shared helper immediately returns false unless its input is a `BinaryExpression(=)`. As a result, PR-head Rslint reports no error for cases like `[(Object!! = 0)] = arr;`, `({x: (Object!! = 0)} = src);`, and `[(Object!! = 0) as any] = arr;`, while ESLint 10.8.0 with `@typescript-eslint/parser` 8.65.0 reports `no-global-assign` on `Object` for all of them; the newly added Go regressions fail with zero diagnostics for these same inputs.

swwind: Fixed in ba5b971a6. The rule-local helper now validates the original `BinaryExpression(=)` and uses the outer transparent wrapper only to locate the enclosing destructuring target. This restores diagnostics for parenthesized, `as`, type-assertion, non-null, and `satisfies` wrapped defaults without changing the shared predicate’s behavior for ordinary assignment expressions. The focused no-global-assign, logical-assignment-operators, and unicorn/no-this-assignment tests pass.

elecmonkey (approved):

PR #1823, opened 2026-08-21
ci: probe the Windows Go test runner and cap its GOMAXPROCS

swwind:
Same suspicion as #1796, now on Windows: the runner is a container on a shared host, so Go reads the host's core count and sizes both `-p` and its scheduler for cores the job does not get. The batching loop caps package concurrency at 32 but cannot bound the thread count inside each test binary.

This caps GOMAXPROCS at 16, replaces the batched loop with a single `go test -count=1 ./cmd/... ./internal/...`, and logs what the runner actually is: logical processors, memory and commit limit, process/thread counts, Go's own `NumCPU`/`GOMAXPROCS`, plus a CPU/queue/threads/commit sample every 10s during the run.

Every other job is disabled and the Go matrix is trimmed to Windows, so a run costs one runner. Both are marked `TEMP (probe)` and come out before this merges.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,47 +64,10 @@ jobs:
 
       - name: Unit Test (Windows)
         if: matrix.runner == 'rspack-windows-2022-large'
-        shell: pwsh
+        run: go test -count=1 ./cmd/... ./internal/...
         env:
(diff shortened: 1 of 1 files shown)


swwind:
Giving up, the windows runner always lost connection with full test suite, no matter how we adjust GOMAXPROCS.

PR #1824, opened 2026-08-21
feat: port rule capitalized-comments

swwind:
## Summary

Port the `capitalized-comments` rule from ESLint to rslint.

Enforces or disallows capitalization of the first letter of line/block comments, with `ignorePattern`, `ignoreInlineComments`, `ignoreConsecutiveComments`, and per-line/per-block option overrides.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/capitalized-comments
- Source code: https://github.com/eslint/eslint/blob/v10.8.1/lib/rules/capitalized-comments.js

## Checklist


internal/rules/all.go
```diff
@@ -183,6 +184,7 @@ func GetAllRules() []rule.Rule {
 		array_callback_return.ArrayCallbackReturnRule,
 		arrow_body_style.ArrowBodyStyleRule,
 		block_scoped_var.BlockScopedVarRule,
+		capitalized_comments.CapitalizedCommentsRule,
```
elecmonkey, line 230: In workspaces using a legacy `rslint.json`, `normalizeJSONConfig` automatically enables every entry returned by `GetCoreRules`, so this registration enables `capitalized-comments` even when the config never requested it. Both Linux and Windows VS Code CI jobs now fail 24 tests because fixture comments such as `// no lint errors` produce an unexpected diagnostic; update or explicitly disable the new rule in the affected legacy JSON fixtures before registering it.

internal/rules/capitalized_comments/capitalized_comments.go
```diff
@@ -0,0 +1,330 @@
...
+	"github.com/web-infra-dev/rslint/internal/utils/unicode17"
+)
+
+//go:embed capitalized_comments.schema.json
+var schemaJSON []byte
+
+// defaultIgnorePattern ports astUtils.COMMENTS_IGNORE_PATTERN: words this rule
+// always ignores at the start of a comment, regardless of any configured
+// ignorePattern.
+var defaultIgnorePattern = esregexp.MustCompile(`^\s*(?:eslint|jshint\s+|jslint\s+|istanbul\s+|globals?\s+|exported\s+|jscs)`, "u")
```
elecmonkey: When source uses Rslint's supported directive syntax, such as `// rslint-disable-next-line no-console`, this pattern does not exempt it even though it exempts the equivalent `eslint-...` directive. The rule reports it and `--fix` rewrites it to `Rslint-disable-next-line`, which `DisableManager` no longer recognizes, potentially reactivating the rule the directive was meant to suppress; include the `rslint` prefix in the built-in directive ignore pattern.

elecmonkey (changes requested):
The new registration breaks existing JSON-config integration suites, and the rule fix can invalidate Rslint disable directives. These correctness regressions need to be addressed before the patch is mergeable.

internal/rules/all.go
swwind: Fixed in 3046ce5e. `packages/rslint/fixtures/rslint.json` now sets `"capitalized-comments": "off"`, following the same pattern the fixture already uses for `prefer-const` and `one-var`. That workspace is the one behind all 24 failures — its sources carry comments like `// unnecessary type assertion`, and `extension.test.ts` / `fixall.test.ts` also write `// no lint errors` into documents at runtime.

The other legacy JSON fixtures (`fixtures-jsconfig`, `fixtures-import-cycle`, `fixtures-unicode-bom`) have no lowercase comments in their sources, so they need no change.

internal/rules/capitalized_comments/capitalized_comments.go
swwind: Fixed in be045aa4. `rslint` is now part of the built-in ignore pattern alongside `eslint`, so `// rslint-disable-next-line no-console` is exempt and `--fix` can no longer rewrite it into a spelling `DisableManager` does not recognize.

Added three valid cases in the extras test (line directive, block directive, `rslint-enable`) and updated the rule doc's always-ignored word list plus its Differences section.

internal/rules/capitalized_comments/capitalized_comments.go
```diff
@@ -0,0 +1,332 @@
...
+		return 0, false
+	}
+	return best, true
+}
+
+// nearestAfterStart is the mirror of nearestBeforeEnd for the position
+// immediately following pos.
+func nearestAfterStart(sourceFile *ast.SourceFile, comments []*ast.CommentRange, index int, pos int) (int, bool) {
+	best := -1
+	if tok, ok := utils.TokenAtOrAfter(sourceFile, pos); ok {
```
elecmonkey: Avoid rescanning the remaining file for every comment

`TokenAtOrAfter` calls `scanner.SkipTrivia` starting at `pos`, and this runs once for every comment whenever `ignoreInlineComments` is enabled. In a comment-only file, each call skips all subsequent comments again, making this path quadratic.

Through the public JS API, 16,000 `// Uppercase` lines took about 3.7 seconds with `ignoreInlineComments: true`, versus 0.35 seconds with it disabled. A direct scaling test took approximately 0.44 / 1.88 / 12.69 seconds for 8k / 16k / 32k lines, while ESLint remained around 40–50 ms.

Please compute the following real-token boundary for all comments in a single linear pass, similar to `precedingTokenEnds`, instead of rescanning trivia from each comment.

internal/rules/capitalized_comments/capitalized_comments.go
```diff
@@ -0,0 +1,332 @@
...
+func isCommentValid(value string, capitalize string, opts commentOptions, isInline func() bool, isConsecutive func() bool) bool {
+	// 1. Default ignore pattern.
+	if defaultIgnorePattern.Test(value) {
+		return true
+	}
+
+	withoutAsterisks := strings.ReplaceAll(value, "*", "")
+
+	// 2. Custom ignore pattern.
+	if opts.IgnorePattern != nil && opts.IgnorePattern.Test(withoutAsterisks) {
```
elecmonkey: Treat ignore-pattern timeouts as matches

Reproduction:

```js
// aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa!
```

```json
["always", { "ignorePattern": "(?:(a|aa)+$|a+!)" }]
```

ESLint 10.8.1 eventually matches the second alternative and reports nothing. Rslint reaches the regexp timeout in the first alternative, but `.Test()` converts that timeout into `false`, so the comment is incorrectly reported and autofixed.

This is the allow/ignore-pattern case that `TestOrTimeout` is designed for: its contract treats a timeout as a match to avoid a false positive. Please use it here.

internal/rules/capitalized_comments/capitalized_comments.go
```diff
@@ -0,0 +1,332 @@
...
+	}
+	if v, ok := source["ignoreConsecutiveComments"].(bool); ok {
+		result.IgnoreConsecutiveComments = v
+	}
+	if pattern, ok := source["ignorePattern"].(string); ok && pattern != "" {
+		// Upstream builds the RegExp when the rule is created, so a pattern the
+		// engine rejects surfaces as a configuration error. Rules here have no
+		// channel for reporting one, so a pattern that fails to compile is
+		// dropped and exempts nothing.
+		if re, err := esregexp.Compile(`^\s*(?:`+pattern+`)`, "u"); err == nil {
```
elecmonkey, line 118: Do not install patterns that JavaScript `/u` rejects

With:

```js
// lower] text
```

```json
["always", { "ignorePattern": "lower]" }]
```

ESLint 10.8.1 fails rule initialization with `Invalid regular expression ... Lone quantifier brackets`. Rslint accepts the pattern through `esregexp.Compile(..., "u")`, uses it as an active ignore matcher, and returns no diagnostic.

This also falls outside the documented divergence that invalid patterns “never match”: this invalid pattern does match and silently suppresses the expected report. Please validate against JavaScript `/u` syntax, or otherwise ensure syntax rejected by JavaScript cannot become an active ignore pattern.

internal/rules/capitalized_comments/capitalized_comments.go
swwind: Fixed in 55ee7d1d. The rule now lazily computes both preceding token ends and following token starts for all comments in one `ForEachToken` pass. `nearestAfterStart` consumes the precomputed boundary, so `ignoreInlineComments` no longer invokes `TokenAtOrAfter` once per comment.

Added a focused boundary regression and a 1k/2k/4k benchmark. On my machine the boundary pass measured 50.7 / 95.0 / 191.1 µs, respectively, which is linear.

internal/rules/capitalized_comments/capitalized_comments.go
swwind: Fixed in 55ee7d1d. Custom `ignorePattern` checks now use `TestOrTimeout`, so a regexp timeout fails open and cannot produce or autofix a false positive. Added the reported reproduction as a valid regression case.

internal/rules/capitalized_comments/capitalized_comments.go
swwind: Fixed in 55ee7d1d. `esregexp.Compile` now rejects unmatched `]`, `{`, and `}` under `u`, with shared compiler tests. The reported `lower]` option is therefore dropped and cannot become an active ignore matcher; the comment is reported and fixed by the focused rule regression.

This preserves rslint's documented divergence for invalid patterns: rule initialization does not throw, but an invalid pattern never matches.

swwind: Holy dude, this is non-related to this pr. I will revert it.

internal/rules/capitalized_comments/capitalized_comments.go
```diff
@@ -0,0 +1,346 @@
...
+			result.IgnorePattern = re
```
elecmonkey, line 119: When `ignorePattern` contains syntax JavaScript rejects under `/u`, such as `"lower]"`, `esregexp.Compile` still succeeds and installs a matcher, so `// lower] text` is silently exempted. ESLint 10.8.1 instead rejects the rule configuration, and even the documented Rslint divergence says invalid patterns should never match; validate these patterns before assigning `result.IgnorePattern` so malformed options cannot suppress diagnostics.

swwind: Sir, this is beyond this rule. That's the validator's job.

elecmonkey (approved):
🤔

PR #1825, opened 2026-08-21
feat: port rule func-style

swwind:
## Summary

Port the `func-style` rule from ESLint to rslint.

Enforces a particular function definition style — `function` declarations or expressions assigned to variables — across a codebase, with options for arrow function exceptions, typed variable declarations, and named-export overrides.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/func-style
- Source code: https://github.com/eslint/eslint/blob/v10.8.1/lib/rules/func-style.js

## Checklist


elecmonkey (dismissed):
The implementation matches ESLint v10.8.1 across the upstream suite and targeted differential cases covering options, TypeScript syntax, overloads, nested this/super frames, export overrides, and diagnostic ranges. No actionable correctness or regression issue was found.

internal/rules/func_style/func_style.go
```diff
@@ -0,0 +1,322 @@
...
+				}
+			},
+			rule.ListenerOnExit(ast.KindFunctionDeclaration): func(node *ast.Node) {
+				if node.AsFunctionDeclaration().Body == nil {
+					return
+				}
+				stack = stack[:len(stack)-1]
+			},
+
+			ast.KindThisKeyword: func(node *ast.Node) {
```
elecmonkey: **[P2] Recognize `this` identifiers inside type queries**

When an arrow’s only `this` reference appears in a TypeScript type query, such as `const foo = (): typeof this => ({});` with `["declaration"]`, ESLint v10.8.1 emits no diagnostic because `@typescript-eslint/parser` exposes a `ThisExpression`, but ts-go represents this occurrence as a `KindIdentifier` named `this`. Since these listeners only mark `KindThisKeyword`/`KindSuperKeyword`, Rslint incorrectly reports `declaration`; type-query identifiers need equivalent stack handling.

internal/rules/func_style/func_style.go
```diff
@@ -0,0 +1,322 @@
...
+			})
+		}
+
+		checkAssignedToVariable := func(node *ast.Node) {
+			// ESTree flattens parentheses, so `var foo = (function(){})`
+			// still has a VariableDeclarator parent for the FunctionExpression
+			// upstream visits. tsgo instead wraps it in an explicit
+			// ParenthesizedExpression, so the declarator has to be found by
+			// walking back up through any such wrapper(s) first.
+			declarator := ast.WalkUpParenthesizedExpressions(node.Parent)
```
elecmonkey: **[P2] Unwrap JavaScript JSDoc cast expressions**

For JavaScript such as `const foo = /** @type {Function} */ (() => 1);` with `["declaration"]`, ESLint v10.8.1 reports the variable declarator, while Rslint emits no diagnostic. ts-go inserts an `AsExpression` for this JSDoc cast, so walking up only parenthesized expressions stops before the `VariableDeclaration`; the assigned-function check must also traverse this transparent assertion wrapper.

internal/rules/func_style/func_style.go
swwind: Fixed in `75333dd4`. The rule now treats ts-go's `Identifier("this")` directly under `TypeQuery` like ESTree's `ThisExpression`, so `const foo = (): typeof this => ({})` preserves the arrow's `this` dependency. Added a focused valid regression case.

internal/rules/func_style/func_style.go
swwind: Fixed in `75333dd4`. `checkAssignedToVariable` now unwraps only ts-go's recognized JSDoc type assertion (`ast.IsJSDocTypeAssertion`) while leaving ordinary TypeScript assertions opaque. Added a checked-JavaScript regression for `const foo = /** @type {Function} */ (() => 1);`.

internal/rules/func_style/func_style.go
```diff
@@ -0,0 +1,338 @@
...
+			// In JavaScript, a leading JSDoc `@type` cast is represented by
+			// tsgo as a reparsed AsExpression inside a ParenthesizedExpression.
+			// ESTree keeps the annotation as a comment, so this wrapper is
+			// transparent to upstream's direct VariableDeclarator check. Do not
+			// unwrap ordinary TypeScript assertions: those remain real ESTree
+			// nodes and intentionally stop the check.
+			for declarator != nil && declarator.Kind == ast.KindAsExpression &&
+				declarator.Parent != nil && ast.IsJSDocTypeAssertion(declarator.Parent) {
```
swwind: blocked by #1877

elecmonkey: 1877 merged. plz go on.

internal/rules/func_style/func_style.go
```diff
@@ -0,0 +1,339 @@
...
+		enterFunctionExpressionLike := func(node *ast.Node) {
+			stack = append(stack, false)
+			checkAssignedToVariable(node)
+		}
+		exitFunctionExpressionLike := func(node *ast.Node) {
+			stack = stack[:len(stack)-1]
+		}
+		for _, kind := range []ast.Kind{
+			ast.KindFunctionExpression,
+			ast.KindMethodDeclaration,
```
elecmonkey, line 157: **[P2] Avoid frames for bodyless member declarations**

With `["declaration"]`, `const outer = () => { abstract class C { abstract method(): typeof this; } };` produces no diagnostic in ESLint 10.8.1 because `@typescript-eslint/parser` 8.58.2 exposes the bodyless member as `TSEmptyBodyFunctionExpression`, so `this` remains attributed to the outer arrow. This loop opens a frame for every ts-go method, accessor, and constructor even when it has no body, causing Rslint to consume `this` in an artificial frame and incorrectly report the outer declarator; only bodyful member implementations should open this frame.

internal/rules/func_style/func_style.go
```diff
@@ -0,0 +1,339 @@
...
+				stack = stack[:len(stack)-1]
+			},
+
+			ast.KindThisKeyword:  markThisOrSuper,
+			ast.KindSuperKeyword: markThisOrSuper,
+			ast.KindIdentifier: func(node *ast.Node) {
+				// typescript-eslint exposes `this` in `typeof this` as a
+				// ThisExpression. tsgo instead uses an Identifier under TypeQuery,
+				// so include that one AST-only spelling in the same frame tracking.
+				if node.Text() == "this" && node.Parent != nil && node.Parent.Kind == ast.KindTypeQuery {
```
elecmonkey: **[P2] Track qualified `typeof this` queries**

With `["declaration"]`, `const outer = (): typeof this.value => ({});` is accepted by ESLint 10.8.1 with `@typescript-eslint/parser` 8.58.2, while Rslint reports `declaration` over columns 7–44. ts-go places the `this` identifier beneath the qualified/property-access operand, so its immediate parent is not `TypeQuery` and this condition misses it; track `this` through the type-query operand while preserving the distinction that a plain `(): this` return type does not exempt the arrow.

internal/rules/func_style/func_style.go
```diff
@@ -0,0 +1,339 @@
...
+			// walking back up through any such wrapper(s) first.
+			declarator := ast.WalkUpParenthesizedExpressions(node.Parent)
+			// In JavaScript, a leading JSDoc `@type` cast is represented by
+			// tsgo as a reparsed AsExpression inside a ParenthesizedExpression.
+			// ESTree keeps the annotation as a comment, so this wrapper is
+			// transparent to upstream's direct VariableDeclarator check. Do not
+			// unwrap ordinary TypeScript assertions: those remain real ESTree
+			// nodes and intentionally stop the check.
+			for declarator != nil && declarator.Kind == ast.KindAsExpression &&
+				declarator.Parent != nil && ast.IsJSDocTypeAssertion(declarator.Parent) {
```
elecmonkey: **[P2] Unwrap JSDoc `@satisfies` casts**

For `const foo = /** @satisfies {Function} */ (() => 1);` with `["declaration"]`, ESLint 10.8.1 reports the variable declarator at columns 7–51, but Rslint emits no diagnostic. ts-go represents this annotation with a synthetic `SatisfiesExpression`, while this code unwraps only the `AsExpression` used for JSDoc `@type`; handle both transparent JSDoc cast forms, for example via the existing JSDoc cast helpers, without unwrapping authored TypeScript assertions.

internal/rules/func_style/func_style.go
swwind: Fixed in ba7565e35. Bodyless member declarations no longer open a function-expression frame, so `typeof this` in an abstract method remains attributed to the enclosing arrow. Added regression coverage for the reported case.

internal/rules/func_style/func_style.go
swwind: Fixed in ba7565e35. `this` tracking now follows only left/receiver edges through qualified-name and property-access chains up to the type query. Added regression coverage for `typeof this.value`; plain `this` types remain unaffected.

internal/rules/func_style/func_style.go
swwind: Fixed in ba7565e35. Variable assignment checks now unwrap only JSDoc-synthesized `@type` and `@satisfies` wrappers while preserving authored TypeScript assertions. Added JavaScript regression coverage for the reported case.

internal/rules/func_style/func_style.go
```diff
@@ -0,0 +1,353 @@
...
+func hasOverloadSignature(statements []*ast.Node, name string, exported bool) bool {
+	for _, stmt := range statements {
+		if stmt.Kind != ast.KindFunctionDeclaration {
+			continue
+		}
+		fd := stmt.AsFunctionDeclaration()
+		if fd.Body != nil {
+			continue
+		}
+		if isNamedExport(stmt) != exported {
```
elecmonkey: [P2] Distinguish default-export overload signatures

With `["expression"]`, `export default function alpha(value: string): string; function alpha(value: unknown) { return value; }` is reported on the second declaration by ESLint 10.8.1 (`@typescript-eslint/parser` 8.58.2), but Rslint emits no diagnostic. `isNamedExport` returns false for both default exports and unexported declarations, so this comparison incorrectly treats their wrappers as matching and `isOverloadedFunction` exempts the implementation. Please distinguish bare, named-export, and default-export wrappers when matching overload signatures.

swwind: Fixed in `7df29c492`. Overload matching now distinguishes bare, named-export, and default-export declaration wrappers, so a default-exported signature no longer exempts a bare implementation. Added a regression case for the reported default-export/bare pair.

elecmonkey (approved):
no more issues found.

elecmonkey:
plz fix ci... all fails...

elecmonkey (approved):

PR #1826, opened 2026-08-21
feat: port rule func-name-matching

swwind:
## Summary

Port the `func-name-matching` rule from ESLint to rslint.

Requires function names to match the name of the variable or property they're assigned to (or, with the `never` option, to not match). Supports `considerPropertyDescriptor` (checks `Object.defineProperty`/`defineProperties`/`create`, `Reflect.defineProperty`) and `includeCommonJSModuleExports`.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/func-name-matching
- Source code: https://github.com/eslint/eslint/blob/v10.8.1/lib/rules/func-name-matching.js

## Checklist


internal/rules/func_name_matching/func_name_matching.go
```diff
@@ -0,0 +1,396 @@
...
+
+				if opts.considerPropertyDescriptor && propertyName == "value" &&
+					node.Parent != nil && node.Parent.Kind == ast.KindObjectLiteralExpression {
+					switch {
+					case isPropertyCall(nthEstreeAncestor(node, 2), "Object", "defineProperty"),
+						isPropertyCall(nthEstreeAncestor(node, 2), "Reflect", "defineProperty"):
+						call := nthEstreeAncestor(node, 2).AsCallExpression()
+						if call.Arguments != nil && len(call.Arguments.Nodes) > 1 {
+							propArg := call.Arguments.Nodes[1]
+							if propArg.Kind == ast.KindStringLiteral {
```
elecmonkey: Parenthesized descriptor names are skipped

`propArg` can be a `ParenthesizedExpression` in ts-go, while ESTree drops these parentheses. With `["always", {"considerPropertyDescriptor": true}]`, `Object.defineProperty(foo, ("bar"), { value: function baz() {} })` produces `matchProperty` in ESLint 10.8.1, but Rslint reports nothing because `propArg.Kind` is not `StringLiteral`. The same issue affects `Reflect.defineProperty`. Please unwrap `call.Arguments.Nodes[1]` with `ast.SkipParentheses` before checking its kind, and add a regression case.

internal/rules/func_name_matching/func_name_matching.go
```diff
@@ -0,0 +1,396 @@
...
+//
+// Divergence: character-shape validity is answered by tsgo's own scanner
+// tables (github.com/microsoft/typescript-go/shim/scanner.IsValidIdentifier),
+// which track a current Unicode identifier snapshot rather than the frozen
+// ES5.1/Unicode-v9 tables esutils ships for its ES5 path. A small number of
+// code points added to Unicode's identifier properties after that snapshot
+// (e.g. U+1885) are treated as valid identifier characters here at every
+// ecmaVersion, where ESLint's ES5 mode rejects them.
+func isIdentifierName(name string, ecmaVersion int) bool {
+	if !scanner.IsValidIdentifier(name) {
```
elecmonkey, line 390: The Unicode divergence is broader than the documented ES5 case

`scanner.IsValidIdentifier` uses TypeScript's current Unicode identifier tables, whereas `esutils@2.0.3` uses its frozen Unicode 9 tables for both `isIdentifierES5` and `isIdentifierES6`. For example, with `ecmaVersion: 2022`, `({ "𐕰": function foo() {} })` is silent in ESLint 10.8.1 but reports `matchProperty` in Rslint. The assignment path also differs at every configured version because upstream calls `isIdentifier(name)` without an `ecmaVersion`: `obj["ᢅ"] = function foo() {};` is silent upstream but reports here. Please either align the identifier tables or expand the documented and tested divergence beyond ES5, including a post-Unicode-9 code point under a modern `ecmaVersion`.

internal/rules/func_name_matching/func_name_matching.md
```diff
@@ -0,0 +1,188 @@
+# func-name-matching
+
+## Rule Details
+
+This rule requires function names to match the name of the variable or property to which they are assigned. The rule will ignore property assignments where the property name is a literal that is not a valid identifier in the ECMAScript version specified in your configuration (default ES5).
```
elecmonkey: “default ES5” is no longer accurate

Rslint's zero-value `LanguageOptions.EffectiveECMAVersion()` returns `LatestECMAScriptVersion`, and ESLint v10 flat config also defaults to the latest version. This affects behavior: `({ "yield": function y() {} })` is ignored under the default/latest configuration but reports with an explicit `ecmaVersion: 5`. Please remove “(default ES5)” or state that the configured version defaults to latest.

internal/rules/func_name_matching/func_name_matching.go
swwind: Fixed. The descriptor-name argument is now passed through `ast.SkipParentheses` before the string-literal check, matching ESTree's transparent parentheses. Added focused regressions for both `Object.defineProperty` and `Reflect.defineProperty`.

internal/rules/func_name_matching/func_name_matching.go
swwind: Documented this as a broader intentional divergence. The rule documentation and implementation comment now explain that esutils' frozen tables affect both the ES5 and ES6 paths, and that the assignment path always selects upstream's default ES5 check. Added focused cases for U+10570 at `ecmaVersion: 2022` and a U+1885 assignment at `ecmaVersion: 2022`.

internal/rules/func_name_matching/func_name_matching.md
swwind: Fixed. The documentation now states that the configured ECMAScript version defaults to latest. Added a focused default/latest versus explicit ES5 regression using `yield`.

elecmonkey (approved):
👏

PR #1827, opened 2026-08-21
perf(rstest): optimize expect-expect rule

elecmonkey:
## Summary

- Compile the default assertion matchers once per rule instead of once per file.
- Add an ASCII-literal fast path for common assertion names such as `expect` and `assert`, while preserving the existing JavaScript `/iu` regexp behavior for Unicode, wildcard, escaped, and other regexp-compatible patterns.
- Add a conservative callee-root guard so unrelated calls skip full callee-chain construction and regexp matching.
- Build callee chains for `additionalTestBlockFunctions` only when the option is configured, and use a set for membership checks.
- Use the test-specific `ParseTestCall` path when classifying Rstest registrations.
- Add focused and differential tests covering prefix boundaries, case-insensitive matching, Unicode case folding, dotted assertion names, and regexp fallback behavior.

## Performance

Measured with `--timing all` while running the same 15 Rstest rules over all `*.test.*`, `*.spec.*`, and configured in-source test files in six Rstack repositories. Each result is the median `rstest/expect-expect` rule time from nine runs alternating with the same `7b4c4f95` baseline. The benchmark covers 1,972 files in total.

The complete JSONL diagnostics were identical before and after the optimization across all six repositories, covering 528 diagnostics.

| Rspack | 147 | 3.4ms | 0.9ms | 3.78× |
| Rsbuild | 629 | 36.4ms | 21.3ms | 1.71× |
| Rslib | 113 | 5.3ms | 2.3ms | 2.30× |
| Rspress | 134 | 10.8ms | 6.9ms | 1.57× |
| Rsdoctor | 89 | 3.7ms | 2.0ms | 1.85× |
| Rstest | 860 | 58.9ms | 21.7ms | 2.71× |
| **Combined** | **1,972** | **118.5ms** | **55.1ms** | **2.15×** |

`expect-expect` runs first and is therefore charged for constructing the shared file-scoped `RstestCallAnalysis`, so these figures include that one-time cost.

### DIFF
--- a/internal/utils/test_framework/rules/expect_expect/expect_expect.go
+++ b/internal/utils/test_framework/rules/expect_expect/expect_expect.go
@@ -436,7 +436,11 @@ func NewRule(config Config) rule.Rule {
 					calleeName := ""
 					if len(additionalTestBlockSet) > 0 {
 						calleeName = testFramework.CalleeChainName(callExpr.Expression)
-						if _, ok := additionalTestBlockSet[calleeName]; ok {
+						// CalleeChainName returns "" for every callee it cannot name
+						// (`arr[i]()`, `obj[key]()`, `(a ?? b)()`). The option's items
+						// carry no minLength, so a configured "" would otherwise turn
+						// each of those ordinary calls into a reported test block.
+						if _, ok := additionalTestBlockSet[calleeName]; ok && calleeName != "" {
 							if strings.HasSuffix(calleeName, ".todo") {
 								return
 							}
--- a/internal/utils/test_framework/rules/expect_expect/expect_expect.go
+++ b/internal/utils/test_framework/rules/expect_expect/expect_expect.go
@@ -76,6 +76,17 @@ type Config struct {
 	Prepare func(ctx rule.RuleContext) Runtime
 }
 
+type assertPattern struct {
+	re           *esregexp.RegExp
+	asciiLiteral string
+	asciiRoot    string
+}
+
+type assertMatcher struct {
+	patterns    []assertPattern
+	hasFallback bool
+}
+
 func buildErrorNoAssertionsMessage() rule.RuleMessage {
 	return rule.RuleMessage{
 		Id:          "noAssertions",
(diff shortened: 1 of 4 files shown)


internal/utils/test_framework/rules/expect_expect/expect_expect.go
```diff
@@ -315,11 +433,26 @@ func NewRule(config Config) rule.Rule {
 						return
 					}
 
+					calleeName := ""
+					if len(additionalTestBlockSet) > 0 {
+						calleeName = testFramework.CalleeChainName(callExpr.Expression)
+						if _, ok := additionalTestBlockSet[calleeName]; ok {
```
swwind: The `calleeName != ""` guard was dropped here (previously `isExtraBlock := calleeName != "" && slices.Contains(additionalTestBlocks, calleeName)`).

`additionalTestBlockFunctions` items have no `minLength` in the schema, so `{"additionalTestBlockFunctions": [""]}` is an accepted config and the set then contains `""`. `CalleeChainName` returns `""` for every callee it cannot name — `arr[i]()`, `obj[key]()`, `(a ?? b)()`, `(await import('x')).run()` — so each of those ordinary calls is now treated as an additional test block, pushed onto `unchecked`, and reported as "Test has no assertions" at end of file. They previously fell through to the assertion path and were ignored.

```go
if len(additionalTestBlockSet) > 0 {
	calleeName = testFramework.CalleeChainName(callExpr.Expression)
	if calleeName != "" {
		if _, ok := additionalTestBlockSet[calleeName]; ok {
			if strings.HasSuffix(calleeName, ".todo") {
				return
			}
			unchecked = append(unchecked, node)
			return
		}
	}
}
```

swwind (commented):
One behavioural regression from the restructuring; the rest of the fast-path work checks out.

internal/utils/test_framework/rules/expect_expect/expect_expect.go
elecmonkey: Confirmed and fixed in 889d4392. Reproduced against `7b4c4f95` with `{"additionalTestBlockFunctions": [""]}` over `arr[0](); obj[key](); (arr[0] ?? obj[key])();` — baseline reported nothing, the branch reported two `Test has no assertions`. Restored the `calleeName != ""` guard, placed before the set lookup so the empty name short-circuits, and added a valid case pinning those three call shapes under that config; it fails without the guard.

swwind (approved):
LGTM

PR #1828, opened 2026-08-21
fix: remove no-undef from JavaScript recommended preset

chenjiahan:
## Summary

The JavaScript recommended preset currently enables `no-undef`, which requires environment-specific globals configuration and can report valid runtime-provided globals. This PR removes the rule from `js.configs.recommended` and documents the difference from ESLint with an explicit opt-in example.

## Checklist

### DIFF
--- a/packages/rslint/src/config/presets/javascript.ts
+++ b/packages/rslint/src/config/presets/javascript.ts
@@ -1,6 +1,6 @@
 import type { RslintConfigEntry } from '../define-config.js';
 
-// Based on official eslint:recommended (@eslint/js@10.x), with `no-undef` intentionally omitted.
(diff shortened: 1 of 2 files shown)


packages/rslint/src/config/presets/javascript.ts
```diff
@@ -31,7 +31,6 @@ const recommended: RslintConfigEntry = {
     'no-setter-return': 'error',
     'no-shadow-restricted-names': 'error',
     'no-this-before-super': 'error',
-    'no-undef': 'error',
```
swwind, line 34: I personally suggest to comment out this line, not delete it.

chenjiahan: done

swwind (approved):

PR #1829, opened 2026-08-21
feat: port rule prefer-named-capture-group

swwind:
## Summary

Port the `prefer-named-capture-group` rule from ESLint to rslint.

Enforces using named capture groups instead of numbered capture groups in regular expressions, in regex literals and `RegExp()`/`new RegExp()` calls with statically-determinable patterns.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/prefer-named-capture-group
- Source code: https://github.com/eslint/eslint/blob/v10.8.1/lib/rules/prefer-named-capture-group.js

## Checklist


internal/utils/regex_capture_groups.go
```diff
@@ -0,0 +1,511 @@
...
+		save := *pos
+		i := *pos + 1
+		nStart := i
+		for i < len(pattern) && isRegexDigit(pattern[i]) {
+			i++
+		}
+		if i == nStart {
+			return
+		}
+		if i < len(pattern) && pattern[i] == ',' {
```
elecmonkey, line 659: [P2] Reject descending quantifier bounds

When a statically known constructor pattern contains an out-of-order quantifier, such as `new RegExp('(a){2,1}')`, this code consumes the quantifier without comparing its bounds and the rule reports `(a)`. ESLint 10.8.1 rejects the regex through regexpp and intentionally emits no rule diagnostic for invalid patterns, so this creates a false positive.

internal/utils/regex_capture_groups.go
```diff
@@ -0,0 +1,511 @@
...
+		groupQuantifiable, ok := scanRegexGroup(pattern, flags, pos, groups)
+		if !ok {
+			return false
+		}
+		quantifiable = groupQuantifiable
+	case '^', '$':
+		quantifiable = false
+		*pos++
+	case '[':
+		end, ok := ClassEnd(pattern, *pos, flags)
```
elecmonkey, line 489: [P2] Validate ranges before accepting character classes

For a constructor pattern with an invalid descending range, such as `new RegExp('(a)[z-a]')`, `ClassEnd` only finds the closing bracket and this scanner accepts the class, causing Rslint to report `(a)`. ESLint 10.8.1's regexpp rejects the range and emits no rule diagnostic, so character-class validity must be checked before returning groups.

internal/rules/prefer_named_capture_group/prefer_named_capture_group.go
```diff
@@ -0,0 +1,218 @@
...
+	return eval.EvalToString(node)
+}
+
+// checkRegex parses pattern (as it would be read under flags) and reports
+// every capturing group that isn't named. reportNode anchors the diagnostic
+// range (the literal, or the whole call/new expression for a constructor
+// call); patternSourceNode is the AST node the pattern text came from, used
+// to decide whether — and where — a fix-safe suggestion can be offered.
+func checkRegex(ctx rule.RuleContext, reportNode *ast.Node, patternSourceNode *ast.Node, pattern string, flags string) {
+	rxFlags := utils.ParseRegexFlags(flags)
```
elecmonkey, line 90: [P2] Reject patterns with both Unicode modes enabled

When the statically resolved flags contain both `u` and `v`, for example `new RegExp('(a)', 'uv')` or the equivalent concatenation, ESLint passes both modes to regexpp, which rejects the combination and emits no rule diagnostic. Rslint instead treats the flags as a generic Unicode mode and reports `(a)`, so this path needs to reject simultaneous `u` and `v` before scanning.

internal/utils/regex_capture_groups.go
```diff
@@ -0,0 +1,511 @@
...
+			}
+			i += w
+		}
+	}
+	return true
+}
+
+func hasGroupNamed(groups []RegexCapturingGroup, name string) bool {
+	for _, group := range groups {
+		if group.Name == name {
```
elecmonkey: [P2] Normalize escaped capture names before resolving references

For the valid literal `/(?<\u0061>a)\k<a>(b)/u`, regexpp decodes the capture name to `a` and ESLint reports the later `(b)`. This scanner stores the raw name `\u0061`, so `hasGroupNamed` fails to match the `\k<a>` reference and rejects the entire pattern, leaving Rslint with no diagnostic.

internal/rules/prefer_named_capture_group/prefer_named_capture_group.go
```diff
@@ -0,0 +1,218 @@
...
+// `RegExp` constructor: the bare identifier, or a property/element access on
+// one of the global-object aliases ESLint's ReferenceTracker recognizes
+// (`globalThis`, `window`, `self`, `global`). Each name must resolve to the
+// real global — not a local declaration or an `off`-disabled global.
+func isGlobalRegExpCallee(ctx rule.RuleContext, callee *ast.Node) bool {
+	if callee == nil {
+		return false
+	}
+
+	switch callee.Kind {
```
elecmonkey, line 181: [P2] Follow global RegExp references through expressions

For global references that ESLint's `ReferenceTracker` follows through expressions, such as `globalThis['Reg' + 'Exp']('(a)')` and `(0, RegExp)('(a)')`, ESLint 10.8.1 reports one diagnostic while this direct node-kind switch emits none. These cases are not local aliases covered by the documented divergence, so the callee matching needs equivalent expression tracing.

internal/utils/regex_capture_groups.go
swwind: Fixed. `skipRegexQuantifier` now compares the bounds after recognizing the complete closing `}`, and rejects an upper bound below the lower bound. I added regression coverage for `(a){2,1}` and the adjacent Annex B case `a{2,1x}(b)` so incomplete brace text remains literal outside `u`/`v`.

internal/utils/regex_capture_groups.go
swwind: Fixed. `RegexCapturingGroups` now checks decoded character-class range endpoints whenever the shared class parser can parse the class, rejecting ranges whose start exceeds their end while preserving unsupported Annex B forms. Regression coverage includes `(a)[z-a]` and an escaped endpoint under `u`.

internal/rules/prefer_named_capture_group/prefer_named_capture_group.go
swwind: Fixed. `RegexCapturingGroups` now rejects flags that enable both Unicode modes before scanning. The rule regression uses statically folded `'u' + 'v'` flags to cover the reported constructor path.

internal/utils/regex_capture_groups.go
swwind: Fixed. Named-backreference resolution now compares decoded IdentifierName values rather than raw spellings. The regression covers the reported `\u0061` case and an adjacent UTF-16 surrogate-pair escape matched against its astral spelling.

internal/rules/prefer_named_capture_group/prefer_named_capture_group.go
swwind: Fixed. Callee matching now evaluates statically resolvable computed keys and follows the final value of comma expressions, while retaining the configured-global and local-shadowing checks at the resolved root. Regression coverage includes both `globalThis['Reg' + 'Exp']('(a)')` and `(0, RegExp)('(a)')`.

internal/utils/regex_capture_groups.go
```diff
@@ -0,0 +1,632 @@
...
+// them: unknown `\p{...}` property names, a group name repeated within one
+// alternative, and v-mode's rule that a class `-` outside a range be escaped.
+func RegexCapturingGroups(pattern string, flags RegexFlags) (groups []RegexCapturingGroup, ok bool) {
+	if flags.Unicode && flags.UnicodeSets {
+		return nil, false
+	}
+	classesValid := true
+	if !IterateRegexCharacterClasses(pattern, flags, func(start, end int) {
+		elements, _, parsed := ParseRegexCharacterClassWithEnd(pattern, start, end, flags)
+		if !parsed {
```
elecmonkey, line 49: [P2] Treat failed character-class parsing as invalid

For `new RegExp('(a)' + '[a-\\d]', 'u')`, `ParseRegexCharacterClassWithEnd` fails here, but the callback returns without clearing `classesValid`. ESLint 10.8.1 therefore emits no rule diagnostic, while Rslint reports `(a)`. Failed class parsing—and strict `u`/`v` class escapes such as `[\\a]`—must reject the entire pattern.

internal/utils/regex_capture_groups.go
```diff
@@ -0,0 +1,632 @@
...
+		if len(rest) >= 2 && isHex(rest[0]) && isHex(rest[1]) {
+			return 4, true
+		}
+	case 'u':
+		if len(rest) > 0 && rest[0] == '{' {
+			n := 1
+			for n < len(rest) && isHex(rest[n]) {
+				n++
+			}
+			if n > 1 && n < len(rest) && rest[n] == '}' {
```
elecmonkey, line 613: [P2] Reject out-of-range Unicode code point escapes

This branch accepts any non-empty hexadecimal `\u{...}` escape without checking that its value is at most `0x10FFFF`. For `new RegExp('(a)' + '\u{110000}', 'u')`, ESLint 10.8.1 rejects the pattern and emits no rule diagnostic, while Rslint reports `(a)`.

internal/utils/regex_capture_groups.go
```diff
@@ -0,0 +1,632 @@
...
+	}
+
+	pos := 0
+	if !scanRegexAlternatives(pattern, flags, &pos, &groups, false) {
+		return nil, false
+	}
+	if pos != len(pattern) {
+		return nil, false
+	}
+	if flags.UV() && !unicodeBackrefsResolve(pattern, flags, groups) {
```
elecmonkey: [P2] Validate named references outside Unicode mode

Named backreferences must resolve whenever the pattern contains named captures, even without `u` or `v`. With `new RegExp('(?<x>a)' + '\k<y>(b)')`, regexpp rejects the unresolved `y` reference and ESLint emits nothing, but this `flags.UV()` gate skips validation and Rslint reports `(b)`.

internal/utils/regex_capture_groups.go
```diff
@@ -0,0 +1,632 @@
...
+		case c >= utf8.RuneSelf:
+			_, w := utf8.DecodeRuneInString(name[i:])
+			if w == 0 {
+				return false
+			}
+			i += w
+			continue
+		case c == '\\':
+			// A `\u` escape spells the code point out; its value isn't checked.
+			if i+1 >= len(name) || name[i+1] != 'u' {
```
elecmonkey: [P2] Support braced Unicode escapes in capture names

A capture name may contain a braced Unicode escape under Unicode mode. ESLint 10.8.1 reports `(b)` for `new RegExp('(?<\\u{61}>a)' + '(b)', 'u')`, but this scanner advances only past `\u`, rejects the following `{`, and suppresses the diagnostic entirely.

internal/rules/prefer_named_capture_group/prefer_named_capture_group.go
```diff
@@ -0,0 +1,224 @@
...
+// the final value of a comma expression. Each name must resolve to the real
+// global — not a local declaration or an `off`-disabled global.
+func isGlobalRegExpCallee(ctx rule.RuleContext, callee *ast.Node, eval *utils.StaticStringEvaluator) bool {
+	if callee == nil {
+		return false
+	}
+
+	switch callee.Kind {
+	case ast.KindIdentifier:
+		if callee.AsIdentifier().Text != "RegExp" || utils.IsShadowed(callee, "RegExp") {
```
elecmonkey: [P2] Stop tracking a reassigned RegExp global

ESLint's `ReferenceTracker` suppresses all references when the global binding has any write, but this check only detects declarations. Consequently `RegExp = custom; RegExp('(a)' + '')` produces no ESLint diagnostic while Rslint reports `(a)` as though the call still targeted the built-in constructor.

internal/rules/prefer_named_capture_group/prefer_named_capture_group.go
```diff
@@ -0,0 +1,224 @@
...
+			return false
+		}
+		value, ok := eval.EvalToString(utils.SkipAssertionsAndParens(access.ArgumentExpression))
+		if !ok || value != "RegExp" {
+			return false
+		}
+		return isKnownGlobalObject(ctx, access.Expression)
+	case ast.KindBinaryExpression:
+		binary := callee.AsBinaryExpression()
+		if binary == nil || binary.OperatorToken == nil || binary.OperatorToken.Kind != ast.KindCommaToken {
```
elecmonkey: [P2] Trace all ReferenceTracker pass-through expressions

The callee matcher only follows comma expressions, while `ReferenceTracker` also propagates references through conditional and logical expressions. ESLint reports `prefer-named-capture-group` for `(true ? RegExp : other)('(a)' + '')` and `(RegExp || other)('(a)' + '')`, but Rslint emits no diagnostic; wrapped global objects such as `(0, globalThis).RegExp(...)` are missed for the same reason.

internal/utils/regex_capture_groups.go
```diff
@@ -0,0 +1,654 @@
...
+// left out, so those patterns come back ok even though the host engine rejects
+// them: unknown `\p{...}` property names, a group name repeated within one
+// alternative, and v-mode's rule that a class `-` outside a range be escaped.
+func RegexCapturingGroups(pattern string, flags RegexFlags) (groups []RegexCapturingGroup, ok bool) {
+	if flags.Unicode && flags.UnicodeSets {
+		return nil, false
+	}
+	classesValid := true
+	if !IterateRegexCharacterClasses(pattern, flags, func(start, end int) {
+		elements, _, parsed := ParseRegexCharacterClassWithEnd(pattern, start, end, flags)
```
elecmonkey, line 48: [P2] Reject invalid escapes inside Unicode character classes

For `new RegExp('(a)' + '[\\a]', 'u')`, ESLint 10.8.1 rejects the pattern and reports nothing, while Rslint reports `(a)`. `ParseRegexCharacterClassWithEnd` accepts this strict-mode identity escape, so `classesValid` remains true; incomplete `\x`, `\01`, and out-of-range `\u{110000}` escapes inside `u`/`v` classes produce the same false positive.

internal/rules/prefer_named_capture_group/prefer_named_capture_group.go
```diff
@@ -0,0 +1,251 @@
...
+	if callee == nil {
+		return false
+	}
+
+	switch callee.Kind {
+	case ast.KindIdentifier:
+		if callee.AsIdentifier().Text != "RegExp" || utils.IsShadowed(callee, "RegExp") {
+			return false
+		}
+		if symbol := ctx.Refs.Resolve(callee); symbol != nil {
```
elecmonkey: [P2] Detect writes to RegExp in source-only runs

In a source-only lint run, `ctx.Refs.Resolve(callee)` cannot resolve the ambient `RegExp` symbol, so this mutation scan is skipped. As a result, `RegExp = custom; RegExp('(a)' + '')` produces no diagnostic in ESLint 10.8.1 but one in Rslint; write detection here must not depend on TypeChecker fallback.

internal/rules/prefer_named_capture_group/prefer_named_capture_group.go
```diff
@@ -0,0 +1,251 @@
...
+			return isKnownGlobalObject(ctx, binary.Right)
+		}
+	}
+	if node.Kind != ast.KindIdentifier {
+		return false
+	}
+	name := node.AsIdentifier().Text
+	switch name {
+	case "globalThis", "window", "self", "global":
+		return !utils.IsShadowed(node, name) && ctx.Globals.Access(name).IsDeclared()
```
elecmonkey: [P2] Stop tracking reassigned global-object bindings

For `globalThis = custom; globalThis.RegExp('(a)' + '')`, ESLint 10.8.1 suppresses the reference because the global-object binding has a write, while Rslint still reports `(a)`. This check handles local declarations and disabled globals but does not detect writes to `globalThis` (or the other recognized global-object names).

internal/rules/prefer_named_capture_group/prefer_named_capture_group.go
```diff
@@ -0,0 +1,251 @@
...
+
+	return false
+}
+
+func isKnownGlobalObject(ctx rule.RuleContext, node *ast.Node) bool {
+	node = utils.SkipAssertionsAndParens(node)
+	if node == nil {
+		return false
+	}
+	if node.Kind == ast.KindBinaryExpression {
```
elecmonkey, line 234: [P2] Trace pass-through expressions around global objects

ESLint's `ReferenceTracker` reports `(globalThis || other).RegExp('(a)' + '')` and `(true ? globalThis : other).RegExp('(a)' + '')`, but Rslint reports neither because `isKnownGlobalObject` unwraps only comma expressions. Logical and conditional pass-through forms need the same traversal used for direct `RegExp` callees.

internal/rules/prefer_named_capture_group/prefer_named_capture_group.go
```diff
@@ -0,0 +1,251 @@
...
+		return isKnownGlobalObject(ctx, access.Expression)
+	case ast.KindBinaryExpression:
+		binary := callee.AsBinaryExpression()
+		if binary == nil || binary.OperatorToken == nil {
+			return false
+		}
+		switch binary.OperatorToken.Kind {
+		case ast.KindCommaToken:
+			return isGlobalRegExpCallee(ctx, utils.SkipAssertionsAndParens(binary.Right), eval)
+		case ast.KindBarBarToken, ast.KindAmpersandAmpersandToken, ast.KindQuestionQuestionToken:
```
elecmonkey, line 214: [P2] Preserve one report per tracked global reference

For `(RegExp || RegExp)('(a)' + '')`, ESLint 10.8.1 emits two diagnostics—one for each tracked global reference—while Rslint emits one because this boolean `||` collapses both matching branches. The traversal must preserve the number of matching roots rather than only whether any root matched.

internal/utils/regex_capture_groups.go
swwind: Fixed. Strict `u`/`v` character-class scanning now rejects identity escapes, malformed `\x`, legacy octal and out-of-range `\u{...}` escapes, class-invalid `\B`/`\k`/decimal escapes, and incomplete or mixed v-mode set operators. Added regression coverage for these forms and valid v-mode counterparts.

internal/rules/prefer_named_capture_group/prefer_named_capture_group.go
swwind: Fixed. Global-root writes are now collected from authored references, so source-only runs suppress `RegExp` after reassignment without depending on TypeChecker resolution. Regression coverage includes assignment, update, destructuring, and loop write forms.

internal/rules/prefer_named_capture_group/prefer_named_capture_group.go
swwind: Fixed. The same write tracking now covers recognized global-object roots, including `globalThis`, so a reassigned global object no longer resolves `RegExp`.

internal/rules/prefer_named_capture_group/prefer_named_capture_group.go
swwind: Fixed. Global-object tracing now follows logical and conditional pass-through expressions. Added regression cases for both forms, including repeated roots.

internal/rules/prefer_named_capture_group/prefer_named_capture_group.go
swwind: Fixed. Tracing now preserves the number of matching global roots; `(RegExp || RegExp)` correctly emits two diagnostics, as do the other ReferenceTracker pass-through operators.

elecmonkey (dismissed):

elecmonkey:
cspell?

elecmonkey (approved):

PR #1831, opened 2026-08-21
feat: port rule require-unicode-regexp

swwind:
## Summary

Port the `require-unicode-regexp` rule from ESLint to rslint.

Enforces the use of a `u` or `v` flag on regular expressions (regex literals and `RegExp()`/`new RegExp()` calls), with an optional `requireFlag` option to pin one specifically, and autofix suggestions.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/require-unicode-regexp
- Source code: https://github.com/eslint/eslint/blob/v10.8.1/lib/rules/require-unicode-regexp.js

## Checklist


internal/rules/require_unicode_regexp/require_unicode_regexp.go
```diff
@@ -0,0 +1,260 @@
...
+	Name:   "require-unicode-regexp",
+	Schema: rule.NewSchema(schemaJSON),
+	Run: func(ctx rule.RuleContext, options []any) rule.RuleListeners {
+		requireFlag := parseOptions(options)
+		evaluator := utils.NewStaticStringEvaluatorWithSourceFile(ctx.TypeChecker, ctx.SourceFile)
+
+		checkCall := func(node *ast.Node, callee *ast.Node, argsList *ast.NodeList) {
+			if !utils.IsBuiltinGlobalCallee(callee, "RegExp", func(name string) bool {
+				return ctx.Globals.Access(name).IsDeclared()
+			}) {
```
elecmonkey: [P1] Track aliases and modified globals like ReferenceTracker

When the built-in is propagated, such as `const R = RegExp; R("foo")`, ESLint's `ReferenceTracker` follows the alias and reports, while this direct callee-shape check returns early and Rslint emits nothing. It also does the inverse after `RegExp = custom; RegExp("foo")`: ESLint ignores the modified global, but Rslint reports it. Preserve ReferenceTracker-style assignment, destructuring, pass-through, and modified-root tracking rather than checking only the immediate callee.

internal/rules/require_unicode_regexp/require_unicode_regexp.go
```diff
@@ -0,0 +1,260 @@
...
+	return rule.RuleFixReplaceRange(core.NewTextRange(tok.End, tok.End), insertText), true
+}
+
+// https://eslint.org/docs/latest/rules/require-unicode-regexp
+var RequireUnicodeRegexpRule = rule.Rule{
+	Name:   "require-unicode-regexp",
+	Schema: rule.NewSchema(schemaJSON),
+	Run: func(ctx rule.RuleContext, options []any) rule.RuleListeners {
+		requireFlag := parseOptions(options)
+		evaluator := utils.NewStaticStringEvaluatorWithSourceFile(ctx.TypeChecker, ctx.SourceFile)
```
elecmonkey: [P1] Resolve constant flags in source-only programs

When linting a supported source-only Program where `ctx.TypeChecker` is nil, this evaluator cannot resolve local identifier initializers and is not given `ctx.Refs`; consequently `const flags = "g"; RegExp("foo", flags)` is ignored by Rslint while ESLint 10.8.1 reports `requireUFlag`. This non-type-aware rule needs a binder/RefStore-backed path for folding local constants when no checker is available.

internal/rules/require_unicode_regexp/require_unicode_regexp.go
```diff
@@ -0,0 +1,260 @@
...
+	if requireFlag == "v" {
+		if ecmaVersion <= 2023 {
+			return false
+		}
+		return utils.IsValidRegexPattern(pattern, utils.RegexFlags{UnicodeSets: true})
+	}
+	if ecmaVersion <= 5 {
+		return false
+	}
+	return utils.IsValidRegexPattern(pattern, utils.RegexFlags{Unicode: true})
```
elecmonkey: [P2] Validate against the configured ECMAScript edition

For `ecmaVersion: 2015`, Rslint offers `addUFlag` for `new RegExp("(?<a>x)")`, whereas ESLint reports without a suggestion because named groups are unavailable in that edition; likewise `(?i:a)` only becomes valid in ES2025, but Rslint suggests it under ES2024. These calls only gate availability of `u`/`v` and otherwise validate against the latest grammar, so the configured edition must participate in pattern validation.

internal/utils/regex_validate.go
```diff
@@ -0,0 +1,341 @@
...
+		c := pattern[i]
+		// A `-` here has no left-hand character to open a range with.
+		if c == '-' {
+			return true
+		}
+		if strings.IndexByte(reservedClassSetPunctuators, c) >= 0 && i+1 < last && pattern[i+1] == c {
+			// `&&` is the intersection operator, legal between two operands —
+			// and a range is not an operand. Every other doubled punctuator is
+			// reserved.
+			if c != '&' || elements == 0 || lastWasRange || i+2 >= last || pattern[i+2] == '&' {
```
elecmonkey: [P2] Reject invalid v-class intersection operands

With `{ requireFlag: "v" }`, `/[a&&bc]/` is valid without `v` but becomes a syntax error with it because a multi-character union cannot be one intersection operand. ESLint therefore withholds the suggestion, while Rslint suggests `/[a&&bc]/v`; this condition only checks that bytes exist around `&&` and permits multiple elements on the right, so the full ClassIntersection operand constraints must be enforced.

internal/utils/regex_validate.go
```diff
@@ -0,0 +1,341 @@
...
+			return false
+		}
+	}
+
+	compileFlags := ""
+	if flags.UV() {
+		compileFlags = "u"
+	}
+	if _, err := esregexp.Compile(pattern, compileFlags); err != nil {
+		return false
```
elecmonkey, line 54: [P2] Accept v-only escaped class punctuators

With `{ requireFlag: "v" }`, `/[\&]/` can safely become `/[\&]/v` because the `v` grammar allows escaping reserved class punctuators, although the `u` grammar rejects `\&`. ESLint offers `addVFlag`, but Rslint offers no suggestion because every `v` pattern is first compiled as `u`; avoid applying the `u` character-class escape grammar when validating `v` patterns.

internal/utils/regex_validate.go
```diff
@@ -0,0 +1,341 @@
...
+		}
+		// `(?<name>` declares a group; `(?<=` and `(?<!` are lookbehinds.
+		if strings.HasPrefix(pattern[i:], "(?<") && i+3 < len(pattern) &&
+			pattern[i+3] != '=' && pattern[i+3] != '!' {
+			name, next, ok := readAngleName(pattern, i+2)
+			if !ok {
+				return false
+			}
+			names = append(names, name)
+			i = next
```
elecmonkey, line 443: [P2] Ignore group-like text inside character classes

For `/[(?<a>x)]\k<a>/`, `(?<a>` is ordinary character-class content, so the external `\k<a>` remains unresolved and adding `u` produces invalid syntax. ESLint withholds the suggestion, but Rslint suggests adding `u` because this scan records every textual `(?<name>` without tracking whether it is inside a class; skip character classes and escapes while collecting named groups.

elecmonkey (commented):
The port diverges from ESLint's global-reference and source-only constant evaluation behavior, causing both missed and spurious diagnostics. Its custom regex validator also emits syntax-breaking suggestions and omits valid suggestions for several u/v and ecmaVersion-dependent patterns.

internal/rules/require_unicode_regexp/require_unicode_regexp.go
swwind: Fixed. The rule now uses a root-driven ReferenceTracker-style call tracker. It follows RegExp through local aliases, assignments, destructuring, and value-preserving expressions, while dropping the built-in root when RegExp itself is modified. Focused regressions cover const aliases, assignment aliases, globalThis destructuring, comma pass-through, and a modified global root.

internal/rules/require_unicode_regexp/require_unicode_regexp.go
swwind: Fixed. The static string evaluator is now constructed with ctx.Refs, so binder-backed identifier resolution and mutation checks remain available when ctx.TypeChecker is nil. A source-only Program regression verifies that `const flags = "g"; RegExp("foo", flags)` reports requireUFlag without type services.

internal/rules/require_unicode_regexp/require_unicode_regexp.go
swwind: Fixed. Pattern validation now considers the configured ECMAScript edition. Named captures, named backreferences, lookbehind, and Unicode property escapes require ES2018, while inline modifier groups require ES2025. Regressions cover named groups under ES2015 and inline modifiers under ES2024.

internal/utils/regex_validate.go
swwind: Fixed. The v-class validator now requires exactly one complete class-set operand on each side of every `&&`, so `[a&&bc]` is rejected for a v-flag suggestion while `[a&&b]` remains accepted.

internal/utils/regex_validate.go
swwind: Fixed. Escaped reserved v-class punctuators are normalized only for the intermediate u-mode compiler check, while the original pattern is still checked by the v-specific validator. `/[\&]/` now receives the addVFlag suggestion.

internal/utils/regex_validate.go
swwind: Fixed. Named-group and backreference collection now skips complete character classes, so group-like text inside a class cannot satisfy an external named backreference. The focused regression for `/[(?<a>x)]\k<a>/` reports without an unsafe suggestion.

internal/rules/require_unicode_regexp/regexp_call_tracker.go
```diff
@@ -0,0 +1,335 @@
...
+			if name, ok := tracker.staticPropertyName(propertyName); ok && name == "RegExp" {
+				tracker.trackAssignmentTarget(element.Name(), regexpTraceConstructor)
+			}
+		}
+	case ast.KindObjectLiteralExpression:
+		if value != regexpTraceGlobalObject {
+			return
+		}
+		for _, propertyNode := range node.AsObjectLiteralExpression().Properties.Nodes {
+			if propertyNode.Kind != ast.KindPropertyAssignment {
```
elecmonkey: [P2] Handle shorthand destructuring assignments

The `ObjectLiteralExpression` branch skips `ShorthandPropertyAssignment`, so `let RegExp; ({RegExp} = globalThis); RegExp('x')` is not tracked. ESLint 10.8.1 reports `requireUFlag` with an `addUFlag` suggestion, while Rslint emits no diagnostic. Handle shorthand assignment properties in addition to `PropertyAssignment` when propagating the global object.

internal/utils/regex_validate.go
```diff
@@ -0,0 +1,460 @@
...
+				}
+				refs = append(refs, name)
+				i = next
+				continue
+			}
+			i += 2
+			continue
+		}
+		// `(?<name>` declares a group; `(?<=` and `(?<!` are lookbehinds.
+		if strings.HasPrefix(pattern[i:], "(?<") && i+3 < len(pattern) &&
```
elecmonkey, line 432: [P2] Validate named capture names and uniqueness

This scan accepts any text between `(?<` and `>` as a capture name and never rejects duplicate names. For `new RegExp("(?<a>x)(?<a>y)")` or `new RegExp("(?<1>x)")`, ESLint 10.8.1 reports without a suggestion, while Rslint offers `addUFlag`. Because this validator is also used by `no-regex-spaces`, that existing rule now reports and fixes invalid patterns that ESLint skips, such as `new RegExp("(?<a>  )(?<a>x)", "u")`.

internal/rules/require_unicode_regexp/regexp_call_tracker.go
elecmonkey: [P2] Track shorthand assignment properties

ESLint's `ReferenceTracker` follows shorthand object assignment patterns, so `let RegExp; ({RegExp} = globalThis); RegExp('x')` reports `requireUFlag` with an `addUFlag` suggestion. This branch currently skips every non-`PropertyAssignment`, so Rslint never propagates `globalThis.RegExp` into the shorthand target and emits no diagnostic for that call.

internal/utils/regex_validate.go
elecmonkey, line 432: [P2] Reject invalid named capture groups

The validator records any text between `(?<` and `>` as a capture name and never rejects duplicates, so patterns like `new RegExp("(?<a>x)(?<a>y)")` and `new RegExp("(?<1>x)")` are treated as valid with `u`. ESLint 10.8.1 reports these without suggestions, but Rslint offers `addUFlag`; because this helper is shared, `no-regex-spaces` also now reports invalid `u` patterns that ESLint skips, such as `new RegExp("(?<a>  )(?<a>x)", "u")`.

elecmonkey (approved):
lgtm

PR #1832, opened 2026-08-21
perf(rstest): optimize valid-expect-in-promise rule

elecmonkey:
## Summary

- Add a prescan that skips shared Rstest call analysis, callback collection, function traversal, and control-flow analysis for files without a promise-chain call. The source file's identifier table answers nearly every form in O(1): the parser interns a property name as an identifier, and it also interns the cooked text of a string, no-substitution template or numeric element-access key, so `promise.then(...)`, `promise["then"](...)`, `` promise[`finally`](...) `` and `promise["\x74hen"](...)` all put the member name into that table. The table excludes the `catch` of a `try`/`catch`, which stays a keyword token.
- The one form the table cannot answer is a key behind parentheses, `promise[("then")](...)`: `IsPromiseChainCall` reaches through them with `ast.SkipParentheses`, while the parser interns only a key that is a literal itself. A narrow scan covers it — such a call opens a bracket on a parenthesis, past whitespace and comments, and spells the member name unless the key is escaped. That scan runs before the substring searches, because it answers almost every file in a single `IndexByte` pass while each substring search costs a pass over the whole file.
- Use the existing promise-chain predicate for the exact AST check, preserving dot access, computed string and template access, escaped property names, parentheses, and method arity behavior.
- Add focused tests covering accepted promise-chain forms, the parenthesized key, and rejected lookalikes.

## Performance

Measured with `--timing all` while running the same 15 configured Rstest rules over all `*.test.*`, `*.spec.*`, and configured in-source test files in six Rstack repositories. Each result is the median `rstest/valid-expect-in-promise` rule time from 27 timing runs, collected in three rounds alternating with the same `7b4c4f95` baseline. The benchmark covers 1,978 files in total.

The complete JSONL diagnostics were identical before and after the optimization across all six repositories, covering 528 diagnostics.

| Rspack | 147 | 0.4ms | 0.1ms | 4.00× |
| Rsbuild | 629 | 0.8ms | 0.3ms | 2.67× |
| Rslib | 113 | 6.4ms | 0.3ms | 21.33× |
| Rspress | 134 | 1.8ms | 0.2ms | 9.00× |
| Rsdoctor | 89 | 0.7ms | 0.1ms | 7.00× |
| Rstest | 866 | 23.7ms | 1.5ms | 15.80× |
| **Combined** | **1,978** | **33.8ms** | **2.5ms** | **13.52×** |

Gating on the identifier table rather than the raw source text is what keeps the small repositories on the winning side: across the six repositories only 16 of the test files spell `then`, `catch` or `finally` as an identifier, and 15 of those actually contain a promise chain, so almost every file is answered by three map lookups.

Where a file is not answered there, the order of the remaining checks matters as much as the gate itself, because a rule this cheap can lose to its own prescan. Timed over the same parsed test files, the four substring searches cost four passes over the file and come close to the walk they guard — 0.76ms against 0.91ms on Rsbuild's 629 files, 1.52ms against 1.71ms on Rstest's 860. The single-pass bracket scan costs 0.05ms and 0.18ms, so it runs first.

### DIFF
--- a/internal/plugins/rstest/rules/valid_expect_in_promise/valid_expect_in_promise.go
+++ b/internal/plugins/rstest/rules/valid_expect_in_promise/valid_expect_in_promise.go
@@ -5,6 +5,7 @@ import (
 	"strings"
 
 	"github.com/microsoft/typescript-go/shim/ast"
+	"github.com/microsoft/typescript-go/shim/scanner"
 	rstestUtils "github.com/web-infra-dev/rslint/internal/plugins/rstest/utils"
 	"github.com/web-infra-dev/rslint/internal/rule"
 	internalUtils "github.com/web-infra-dev/rslint/internal/utils"
@@ -66,7 +67,7 @@ func sourceMayContainPromiseChain(sourceFile *ast.SourceFile) bool {
 // only costs a walk the caller would otherwise skip, and never hides one.
 func bracketOpensOnParenthesis(text string) bool {
 	for index := strings.IndexByte(text, '['); index >= 0; {
-		if next := skipTrivia(text, index+1); next < len(text) && text[next] == '(' {
+		if next := scanner.SkipTrivia(text, index+1); next < len(text) && text[next] == '(' {
 			return true
 		}
 		offset := strings.IndexByte(text[index+1:], '[')
@@ -76,40 +77,6 @@ func bracketOpensOnParenthesis(text string) bool {
 		index += offset + 1
 	}
 	return false
-}
-
-// skipTrivia returns the first index at or after index that starts neither
-// whitespace nor a comment.
-func skipTrivia(text string, index int) int {
-	for index < len(text) {
-		switch text[index] {
-		case ' ', '\t', '\r', '\n', '\v', '\f':
-			index++
-		case '/':
-			if index+1 >= len(text) {
-				return index
-			}
-			switch text[index+1] {
-			case '/':
-				end := strings.IndexByte(text[index:], '\n')
-				if end < 0 {
-					return len(text)
-				}
-				index += end + 1
-			case '*':
-				end := strings.Index(text[index+2:], "*/")
-				if end < 0 {
-					return len(text)
-				}
-				index += end + 4
-			default:
-				return index
-			}
-		default:
-			return index
-		}
-	}
-	return index
 }
 
 var promiseChainMemberNames = [...]string{"then", "catch", "finally"}
--- a/internal/plugins/rstest/rules/valid_expect_in_promise/valid_expect_in_promise.go
+++ b/internal/plugins/rstest/rules/valid_expect_in_promise/valid_expect_in_promise.go
@@ -17,35 +17,38 @@ import (
 // `finally` reached either as a property access or as a bracket access keyed by
 // a string or no-substitution template literal.
 //
-// The two forms are gated differently. A property access spells the member as
-// an identifier — a keyword-named property is interned like any other — so the
-// source file's identifier table answers that form exactly and in O(1). It also
-// excludes the `catch` of a `try`/`catch`, which stays a keyword token and never
-// reaches the table; gating on the raw source text instead let every file with a
-// `try`/`catch`, and every file with a backslash, through to the AST walk this
-// function exists to avoid.
+// The source file's identifier table answers nearly every form exactly and in
+// O(1). A property access spells the member as an identifier, and a keyword-named
+// property is interned like any other; the parser also interns the cooked text of
+// a string, no-substitution template or numeric element-access key, so
+// promise["then"](), promise[`finally`]() and the escaped promise["\x74hen"]()
+// put the member name in the same table. The table excludes the `catch` of a
+// `try`/`catch`, which stays a keyword token and never reaches it; gating on the
+// raw source text instead let every file with a `try`/`catch`, and every file
+// with a backslash, through to the AST walk this function exists to avoid.
 //
-// A bracket access keys the member off a literal that never reaches the
-// identifier table, so it keeps a text scan. That scan stays narrow by first
-// requiring a quote or backtick directly after `[`. The backslash check remains
-// for escaped keys such as promise["\x74hen"](), whose source text does not
-// spell the member name.
+// The one form the table cannot answer is a key behind parentheses, as in
+// promise[("then")](): the parser interns a key that is a literal itself, while
+// IsPromiseChainCall reaches through parentheses with ast.SkipParentheses. Such
+// a call opens its bracket on a parenthesis, and its raw text spells the member
+// name unless the key is escaped, so requiring both keeps the fallback sound
+// without paying for the brackets a test file spells everywhere else.
 func sourceMayContainPromiseChain(sourceFile *ast.SourceFile) bool {
 	if sourceFile == nil {
 		return true
 	}
 	// Reading a nil identifier table is well defined; a file with no identifiers
-	// has no property-access chain either way.
+	// has no promise chain either way.
 	for _, name := range promiseChainMemberNames {
 		if _, ok := sourceFile.Identifiers[name]; ok {
 			return nodeContainsPromiseChain(sourceFile.AsNode())
 		}
 	}
 
+	// The bracket scan runs first: it answers almost every file with a single
+	// IndexByte pass, while the four substring searches below cost a pass each.
 	text := sourceFile.Text()
-	if !strings.Contains(text, `["`) &&
-		!strings.Contains(text, `['`) &&
-		!strings.Contains(text, "[`") {
+	if !bracketOpensOnParenthesis(text) {
 		return false
 	}
 	if !strings.Contains(text, "then") &&
@@ -55,6 +58,58 @@ func sourceMayContainPromiseChain(sourceFile *ast.SourceFile) bool {
 		return false
 	}
 	return nodeContainsPromiseChain(sourceFile.AsNode())
+}
+
+// bracketOpensOnParenthesis reports whether any `[` in the text is followed,
+// past whitespace and comments, by `(`. It reads the raw text rather than the
+// AST, so a bracket inside a string or a regular expression counts too; that
+// only costs a walk the caller would otherwise skip, and never hides one.
+func bracketOpensOnParenthesis(text string) bool {
+	for index := strings.IndexByte(text, '['); index >= 0; {
+		if next := skipTrivia(text, index+1); next < len(text) && text[next] == '(' {
+			return true
+		}
+		offset := strings.IndexByte(text[index+1:], '[')
+		if offset < 0 {
+			return false
+		}
+		index += offset + 1
+	}
+	return false
+}
+
+// skipTrivia returns the first index at or after index that starts neither
+// whitespace nor a comment.
+func skipTrivia(text string, index int) int {
+	for index < len(text) {
+		switch text[index] {
+		case ' ', '\t', '\r', '\n', '\v', '\f':
+			index++
+		case '/':
+			if index+1 >= len(text) {
+				return index
+			}
+			switch text[index+1] {
+			case '/':
+				end := strings.IndexByte(text[index:], '\n')
+				if end < 0 {
+					return len(text)
+				}
+				index += end + 1
+			case '*':
+				end := strings.Index(text[index+2:], "*/")
+				if end < 0 {
+					return len(text)
+				}
+				index += end + 4
+			default:
+				return index
+			}
+		default:
+			return index
+		}
+	}
+	return index
 }
 
 var promiseChainMemberNames = [...]string{"then", "catch", "finally"}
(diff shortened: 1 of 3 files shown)


internal/plugins/rstest/rules/valid_expect_in_promise/valid_expect_in_promise.go
```diff
@@ -11,10 +12,46 @@ import (
...
+
+func nodeContainsPromiseChain(node *ast.Node) bool {
+	if node == nil {
+		return false
+	}
+	if testFramework.IsPromiseChainCall(node) {
+		return true
+	}
+	found := false
+	node.ForEachChild(func(child *ast.Node) bool {
```
swwind: This closure captures `found`, so it escapes and costs one heap allocation per visited node, on every file that passes the text gate.

`Node.ForEachChild` already returns whether any visit returned true, so the body collapses to an allocation-free, non-capturing recursion:

```go
func nodeContainsPromiseChain(node *ast.Node) bool {
	if node == nil {
		return false
	}
	if testFramework.IsPromiseChainCall(node) {
		return true
	}
	return node.ForEachChild(nodeContainsPromiseChain)
}
```

Given how often the text gate lets a file through (see the other comment), this is plausibly part of the Rsbuild 0.61× row.

internal/plugins/rstest/rules/valid_expect_in_promise/valid_expect_in_promise.go
```diff
@@ -11,10 +12,46 @@ import (
 	shared "github.com/web-infra-dev/rslint/internal/utils/test_framework/rules/valid_expect_in_promise"
 )
 
+// sourceMayContainPromiseChain reports whether the file contains a call accepted
+// by testFramework.IsPromiseChainCall. The text check avoids an AST walk for
+// the common negative case. A backslash keeps escaped static names
+// such as promise["\x74hen"]() in the exact AST scan.
+func sourceMayContainPromiseChain(sourceFile *ast.SourceFile) bool {
+	if sourceFile == nil || sourceFile.Identifiers == nil {
```
swwind: `Identifiers` is never read after this nil check, so the guard is dead — the parser always assigns it — and the field it probes is the one that would make the gate work.

The text gate is weak in practice: `strings.Contains(text, "\\")` is true for any file containing a regex, an escaped quote, or a Windows path, and `"catch"` is true for any `try`/`catch`. So the cheap gate rarely prevents the AST walk it exists to avoid.

`sourceFile.Identifiers` is a map of every identifier text in the file, so `Identifiers["then"] || Identifiers["catch"] || Identifiers["finally"]` is an O(1) exact test for the dominant `promise.then(...)` property-access form, leaving the substring and backslash scan needed only for the bracket-access forms.

swwind (commented):
Two notes on the fast path, both in `valid_expect_in_promise.go`. The core premise holds: reports only ever originate from a `chainGroup`, and `groupFor` is reached only from the `IsPromiseChainCall` branch of the shared analyzer's walk, so a file with no promise-chain call has no diagnostic to lose. The zero-value `shared.Runtime{}` is safe as well — `NewRule` iterates a nil map, and both `isAssertionCall` and the async-sink helper nil-check their callbacks.

internal/plugins/rstest/rules/valid_expect_in_promise/valid_expect_in_promise.go
elecmonkey: Fixed in a2e72472. `ForEachChild` returns whether any visit returned true, so the body is now `return node.ForEachChild(nodeContainsPromiseChain)` and the walk no longer allocates per visited node.

internal/plugins/rstest/rules/valid_expect_in_promise/valid_expect_in_promise.go
elecmonkey: Fixed in a2e72472, and the gate was weaker than the comment suggested. Across the six Rstack repositories, 450 of 1,972 test files passed the text gate while only 17 contain a promise chain — 433 files paid a full AST walk for nothing, which is what produced the Rsbuild row.

The property-access form now gates on `sourceFile.Identifiers`, which answers it exactly and in O(1) and excludes the `catch` of a `try`/`catch`. Bracket access keys the member off a string or no-substitution template literal that never reaches the table, so it keeps a text scan, now narrowed to files spelling a quote or backtick directly after `[`; the backslash check stays for escaped keys such as `promise["\x74hen"]()`.

One correction on the nil guard: `parser.go` assigns `result.Identifiers = p.identifiers` unconditionally, but `p.identifiers` is created lazily, so a file with no identifiers does leave it nil. Reading a nil map is well defined, so the guard was removable regardless — it is gone now.

Rsbuild goes from 0.61× to 1.22×, and the combined figure from 5.20× to 6.19×.

internal/plugins/rstest/rules/valid_expect_in_promise/valid_expect_in_promise.go
```diff
@@ -11,10 +12,72 @@ import (
...
+// excludes the `catch` of a `try`/`catch`, which stays a keyword token and never
+// reaches the table; gating on the raw source text instead let every file with a
+// `try`/`catch`, and every file with a backslash, through to the AST walk this
+// function exists to avoid.
+//
+// A bracket access keys the member off a literal that never reaches the
+// identifier table, so it keeps a text scan. That scan stays narrow by first
+// requiring a quote or backtick directly after `[`. The backslash check remains
+// for escaped keys such as promise["\x74hen"](), whose source text does not
+// spell the member name.
```
swwind: "A bracket access keys the member off a literal that never reaches the identifier table" is the opposite of what the parser does — the key is interned, which is why the block below is unreachable for every literal key. The escaped-key sentence has the same problem: the interned text is cooked, so `promise["\x74hen"]()` sets `Identifiers["then"]` and the loop above returns before the backslash check is ever reached. This paragraph needs to follow whatever the block below becomes.

internal/plugins/rstest/rules/valid_expect_in_promise/valid_expect_in_promise.go
```diff
@@ -11,10 +12,72 @@ import (
...
+		!strings.Contains(text, "[`") {
+		return false
+	}
+	if !strings.Contains(text, "then") &&
+		!strings.Contains(text, "catch") &&
+		!strings.Contains(text, "finally") &&
+		!strings.Contains(text, `\`) {
+		return false
+	}
+	return nodeContainsPromiseChain(sourceFile.AsNode())
```
swwind, line 61: `parseElementAccessExpressionRest` interns a string, no-substitution template, or numeric element-access key into `p.identifiers`, and it interns the cooked text, so `promise["then"]`, `` promise[`finally`] `` and `promise["\x74hen"]` all put the member name into `sourceFile.Identifiers` and are already answered by the loop above. Trivia does not matter either — `promise[/*c*/"then"]()` is interned the same way, so the "quote directly after `[`" shape buys nothing. Deleting this whole block leaves both the new prescan test and the rule suite green.

What it does not cover is the one form the loop cannot answer, a key behind parentheses:

```js
test("case", () => { promise[("then")](value => expect(value).toBe(1)); });
```

That reports on `main` and on upstream `jest/valid-expect-in-promise`, and it reports here once the prescan is bypassed, but the file spells no `["`, so the prescan answers false and the diagnostic is lost. Adding an unrelated `other["x"]` anywhere in the same file brings it back, which is the sign that the scan is not testing what it needs to test.

So the fallback's only remaining job is the parenthesized key:

```go
	// Every literal key that reaches the callee directly is interned with the
	// identifiers above, so the only chain left is one whose key hides behind
	// the parentheses IsPromiseChainCall skips.
	if !strings.Contains(sourceFile.Text(), "[(") {
		return false
	}
	return nodeContainsPromiseChain(sourceFile.AsNode())
```

`promise[ ("then") ]` still slips past that. If you would rather not carry a second heuristic, dropping the fallback entirely is defensible, but then the divergence from upstream should be pinned by a test rather than left implicit. Either way the prescan test wants a `promise[("then")](value => expect(value).toBe(1))` case, and the rule test an invalid case for it.

swwind (commented):
The identifier gate is the right instrument, and the Rsbuild row shows it. One correctness problem in the fallback beneath it: `parseElementAccessExpressionRest` interns a string, no-substitution template, or numeric element-access key into the identifier table, and it interns the cooked text, so the table already answers the bracket form as well — escaped keys included. The only form left for the text scan is a key hidden behind parentheses, which `promiseChainMethodName` reaches through `SkipParentheses` but the scan is not shaped to find, so the diagnostic is dropped. 0f8ba957 did not have this gap; its `strings.Contains(text, "then")` covered the parenthesized key.

| Code | ESLint | This PR |
| --- | --- | --- |
| `promise.then(value => expect(value).toBe(1))` | error | error |
| `promise["then"](value => expect(value).toBe(1))` | error | error |
| `promise["\x74hen"](value => expect(value).toBe(1))` | error | error |
| `promise[("then")](value => expect(value).toBe(1))` | error | no error |
| the same call, in a file that also contains `other["x"]` | error | error |

internal/plugins/rstest/rules/valid_expect_in_promise/valid_expect_in_promise.go
elecmonkey: Fixed in 6757fe9e. Both halves of the comment are right: `parseElementAccessExpressionRest` interns the key, and it interns the cooked text, so `promise["then"]()`, `` promise[`finally`]() `` and `promise["\x74hen"]()` are all answered by the identifier loop above, and the block below was unreachable for them.

The paragraph now says that, and describes the fallback as covering only the key behind parentheses, which is what the block became.

internal/plugins/rstest/rules/valid_expect_in_promise/valid_expect_in_promise.go
elecmonkey: Fixed in 6757fe9e. The quote-adjacency scan is gone; the fallback now covers only the key behind parentheses, and the prescan test gained `promise[("then")](value => expect(value).toBe(1))` while the rule test gained the matching invalid case.

I did not take the `strings.Contains(text, "[(")` form, because `promise[ ("then") ]` still slips past it. `bracketOpensOnParenthesis` looks for a `[` followed, past whitespace and comments, by `(`, so `promise[ ("then") ]` and `promise[/*c*/("then")]` are covered too. It reads raw text rather than the AST, so a bracket inside a string or a regular expression counts as well — that only costs a walk, and never hides one.

What the fix surfaced is that the order of the text checks matters as much as the gate. The four substring searches cost four passes over the file and come close to the walk they guard: timed over the same parsed test files, 0.76ms against 0.91ms for the unguarded walk on Rsbuild's 629 files, and 1.52ms against 1.71ms on Rstest's 860. The bracket scan is a single `IndexByte` pass at 0.05ms and 0.18ms, so it now runs first, ahead of the substring searches. The revision this PR started from paid the substring passes first, which is part of why Rsbuild only reached 1.22×.

Tests: the prescan test also covers `promise[ /* c */ ("finally") ](...)`, `promise[("\x74hen")](...)`, and a non-chain `rows[(index + 1)]`; `bracketOpensOnParenthesis` has its own test, including unterminated comments.

Re-ran the benchmark against the same `7b4c4f95` baseline, three rounds alternating, 27 timing samples per repository: combined 33.8ms → 2.5ms (13.52×), every repository positive, Rsbuild 2.67×. Diagnostics stay identical, 528 across the six repositories.

internal/plugins/rstest/rules/valid_expect_in_promise/valid_expect_in_promise.go
```diff
@@ -11,10 +12,127 @@ import (
...
+		index += offset + 1
+	}
+	return false
+}
+
+// skipTrivia returns the first index at or after index that starts neither
+// whitespace nor a comment.
+func skipTrivia(text string, index int) int {
+	for index < len(text) {
+		switch text[index] {
```
swwind: `skipTrivia` recognizes only ASCII whitespace, so the newly added `promise[\uFEFF("then")]` case never reaches the `(` and the prescan disables the rule, dropping its expected diagnostic (the focused package test currently fails). Other valid ECMAScript whitespace and CR/Unicode line terminators after `//` have the same false-negative; use the parser scanner or the repository’s ECMAScript trivia helper instead.

elecmonkey: Fixed in c25ebb3c. `bracketOpensOnParenthesis` now delegates trivia skipping to ts-go's `scanner.SkipTrivia`, so it shares the parser's handling of ECMAScript whitespace and Unicode line terminators instead of maintaining ASCII-only logic.

Added focused coverage for U+FEFF and line comments terminated by U+2028/U+2029, plus a full rule regression for a parenthesized key after U+FEFF.

Tests:
- `go test ./internal/plugins/rstest/rules/valid_expect_in_promise -count=1`
- `go test ./internal/plugins/rstest/... -count=1`

swwind (approved):
LGTM

PR #1835, opened 2026-08-21
feat(rstest): add no-hooks rule

elecmonkey:
## Summary

Report Rstest's four setup and teardown hooks — `beforeAll`, `beforeEach`, `afterEach`, `afterAll` — so a codebase can keep test state inside the test that owns it. The `allow` option exempts hooks by their resolved semantic name, so an aliased import such as `import { beforeEach as teardown }` is exempted by `{ "allow": ["beforeEach"] }` rather than by its local spelling.

Jest and Rstest have the same diagnostic contract here — same reported set, message, option, range, and absence of a fix — so the rule body moves to `internal/utils/test_framework/rules/no_hooks/` and both plugins register a wrapper that injects only the framework's parser. Hook recognition goes through the shared `ParseFnCall`, so globals, named and aliased imports, namespace and whole-module `require`, `import.meta.rstest`, same-file aliases, and `@rstest/playwright` hooks all resolve without the rule re-deriving provenance. Registered in the rstest plugin but left out of the `recommended` preset, matching upstream.

## Credits

Port from [`eslint-plugin-jest`](https://github.com/jest-community/eslint-plugin-jest)'s `no-hooks`, cross-checked against [`eslint-plugin-vitest`](https://github.com/vitest-dev/eslint-plugin-vitest)'s `no-hooks`.

- **`aroundAll` and `aroundEach` are not hooks.** `eslint-plugin-vitest` recognises both in its rule body, schema, and tests. Rstest's `RunnerAPI`, its globals declaration, and its runtime expose neither, so treating them as hooks would report calls to whatever function the user actually defined under those names.

- **`onTestFinished` and `onTestFailed` are not hooks either.** Rstest declares them separately from the four lifecycle hooks: they register a callback from inside a running test body rather than describing a setup or teardown stage, so they are outside this rule's subject even though they take a callback.

- **`@rstest/playwright` member hooks are reported.** `PlaywrightTest` declares `test.beforeAll`, `test.beforeEach`, `test.afterEach` and `test.afterAll`, and `test.extend(...)` returns another `PlaywrightTest`, so `test.beforeEach()` and `test.extend({}).afterAll()` are genuine hook registrations. Upstream has no such member form to consider. The boundary is exact rather than "any member named like a hook": `@rstest/core`'s `test.beforeEach`, and Playwright's `test.skip.beforeEach` and `test.beforeEach.only`, do not resolve as hooks and are left alone.

- **The `allow` schema constrains the four hook names.** Upstream's schema uses draft-6 `contains` with a value that is not itself a schema, so it constrains nothing; this port keeps the `items.enum` form the rslint Jest rule already ships, which rejects a misspelled hook name at configuration time. The accepted set is unchanged.

## Related Links

Related #935

## Checklist


internal/plugins/rstest/rules/no_hooks/no_hooks.go
```diff
@@ -0,0 +1,25 @@
...
+	"github.com/microsoft/typescript-go/shim/ast"
+	rstestUtils "github.com/web-infra-dev/rslint/internal/plugins/rstest/utils"
+	"github.com/web-infra-dev/rslint/internal/rule"
+	shared "github.com/web-infra-dev/rslint/internal/utils/test_framework/rules/no_hooks"
+)
+
+var NoHooksRule = shared.NewRule(shared.Config{
+	Name:             "rstest/no-hooks",
+	RequiresTypeInfo: true,
+	Prepare: func(ctx rule.RuleContext) shared.Runtime {
```
swwind, line 13: `rstest/no-conditional-expect` gates its analysis behind a `SourceFile.Identifiers` lookup and returns `Runtime{Skip: true}` when the file cannot contain a match. This adapter has no such fast path, so it builds the full rstest call-analysis index for every linted file. `GetRstestCallAnalysis` is `CachedByFile`, so the cost only shows up when no other rstest rule is enabled — but the same shape applies: if none of `beforeAll` / `beforeEach` / `afterAll` / `afterEach` appear in `ctx.SourceFile.Identifiers`, no hook call can exist in the file.

internal/plugins/rstest/rules/no_hooks/no_hooks_extras_test.go
```diff
@@ -0,0 +1,320 @@
...
+	"github.com/web-infra-dev/rslint/internal/utils"
+)
+
+func TestNoHooksExtras(t *testing.T) {
+	rule_tester.RunRuleTester(
+		fixtures.GetRootDir(),
+		"tsconfig.json",
+		t,
+		&no_hooks.NoHooksRule,
+		[]rule_tester.ValidTestCase{
```
swwind, line 28: The valid list is thorough on receiver and chain shapes, but a hook nested inside a `describe`/`test` callback — the shape almost every real suite uses — is not asserted in either test file, and neither is a multi-diagnostic file (`beforeEach(() => { afterEach(() => {}) })` reports twice, outer first).

These all behave correctly today with nothing pinning them:

```go
{Code: `new A().beforeEach()`},
{Code: `arr[0].beforeEach()`},
{Code: `this.beforeEach()`},
{Code: `(<any>x).beforeEach()`},
{Code: `beforeEach.skip(() => {})`},
{Code: `beforeEach.each([1])(() => {})`},
{Code: `new beforeEach()`},
{Code: "beforeEach`x`"},
{Code: `(0, beforeEach)(() => {})`},
{Code: `globalThis.beforeEach(() => {})`},
```

The four member-receiver cases are the ones I would prioritise, since `jest/no-hooks` reports all four. On the invalid side, `beforeEach()` with no arguments, `beforeEach?.(() => {})` and `beforeEach(...fns)` are all reported and all unasserted.

swwind (commented):
Ran an edge-case sweep over the new rule (~70 shapes) and diffed the behaviour against upstream `eslint-plugin-jest` 29.16.1 wherever the two are comparable. No correctness bugs in the port: `rstest/no-hooks` gets every realistic shape right, and the extraction leaves `jest/no-hooks` behaviourally unchanged.

Two behaviour notes, plus a coverage gap in the test files.

### Type-only imports are treated as value imports

| Code | ESLint | This PR |
| --- | --- | --- |
| `import type { beforeEach } from '@rstest/core'; beforeEach(() => {})` | ok | `Unexpected 'beforeEach' hook` |
| `import { type beforeEach } from '@rstest/core'; beforeEach(() => {})` | `Unexpected 'beforeEach' hook` | ok |

`resolveModuleImportSpecifier` in `internal/utils/test_framework/module_reference.go` never looks at `IsTypeOnly`, on either the import clause or the specifier, so a type-only binding resolves as a value import — and the two spellings land on opposite answers. This predates the PR (#1475) and is only reachable on code that already fails to compile, so it is fine to leave here; flagging it because this PR is what puts both plugins on one shared rule.

### The two parsers disagree on non-identifier receivers

| Code | ESLint | rstest/no-hooks | jest/no-hooks |
| --- | --- | --- | --- |
| `new A().beforeEach()` | ok | ok | `Unexpected 'beforeEach' hook` |
| `arr[0].beforeEach()` | ok | ok | `Unexpected 'beforeEach' hook` |
| `this.beforeEach()` | ok | ok | `Unexpected 'beforeEach' hook` |
| `(<any>x).beforeEach()` | ok | ok | `Unexpected 'beforeEach' hook` |

rstest is the correct column; the jest false positives come from `ParseJestFnCall` and are out of scope here. Worth a lock-in test on the rstest side so the difference stops being invisible — see the inline comment.

Everything else matched: schema, message text, report range, `allow` handling (all four names, duplicates, empty list), and every parser shape I tried — `describe.each` nesting, IIFEs, class static blocks, statement positions, `.tsx`, renamed `require` destructuring, and cross-file barrel re-exports (`export *`, named, renamed, import-then-export relay), which stay unreported exactly as upstream does.

internal/plugins/rstest/rules/no_hooks/no_hooks_extras_test.go
elecmonkey: Good catch — all three gaps were real. Pinned in 15fbae06.

All ten valid shapes you listed behave as you described and are now asserted, with the four member-receiver cases carrying a comment noting that `jest/no-hooks` reports them and `ParseRstestFnCall` does not, so the divergence is visible at the test site.

On the invalid side: `beforeEach()`, `beforeEach?.(() => {})` and `beforeEach(...fns)` are asserted with full ranges, along with a hook inside a `describe` callback and `beforeEach(() => { afterEach(() => {}) })` reporting twice, outermost first.

internal/plugins/rstest/rules/no_hooks/no_hooks.go
elecmonkey: I looked at this and decided against the fast path, because the two rules are not in the same position.

`expect` is only ever reachable as an identifier, so `no-conditional-expect`'s `SourceFile.Identifiers` guard is exact. Hook names are not: `elementAccessName` in `internal/utils/test_framework/member_chain.go` accepts a `StringLiteral` key, so `import * as rstest from '@rstest/core'; rstest["beforeEach"](() => {})` resolves and is reported today. A string literal is not in `Identifiers`, so the guard would silently drop it.

The shape is contrived, but it is currently correct behaviour, and the cost it would buy back only appears when no other rstest rule is enabled — `GetRstestCallAnalysis` is `CachedByFile` otherwise, as you note. I would rather not trade a real answer for that. 15fbae06 adds the element-access case as an invalid lock-in so the constraint is recorded where a future fast-path attempt would trip over it.

swwind (approved):
LGTM

PR #1836, opened 2026-08-21
feat(rstest): add prefer-hooks-in-order rule

elecmonkey:
## Summary

Report a lifecycle hook that appears after a hook from a later runtime stage, in the order Rstest runs them: `beforeAll`, `beforeEach`, `afterEach`, `afterAll`. The subject is each consecutive run of hook calls, not every hook in a file or block — any non-hook call encountered in the traversal ends the current run, while comments, blank lines, and statements containing no call do not. So `afterAll(); doSomething(); beforeAll();` is two independent runs and is left alone, and `afterAll(); afterEach();` is one run with an inversion and is reported.

Jest and Rstest share the whole contract, so the rule body moves to `internal/utils/test_framework/rules/prefer_hooks_in_order/` on top of the existing `test_framework.HooksOrder` / `HookOrderIndex` primitives, and each plugin's wrapper injects only its parser. Registered in the rstest plugin but left out of the `recommended` preset, matching upstream.

## Credits

Port from [`eslint-plugin-jest`](https://github.com/jest-community/eslint-plugin-jest)'s `prefer-hooks-in-order`, cross-checked against [`eslint-plugin-vitest`](https://github.com/vitest-dev/eslint-plugin-vitest)'s `prefer-hooks-in-order`.

Upstream tracks "inside a hook" as a single boolean and ignores every call made while it is set. That conflates two separate scopes — the run that surrounds a hook, and the run inside its callback — and the shared body scopes each run to its own nesting level instead. Both consequences only show up where a hook call is nested inside another hook call's subtree; without such nesting the two are equivalent.

- **A nested hook callback no longer ends the surrounding run.** Upstream clears the boolean as soon as any nested hook call exits, so the next outer hook is compared as if it started a fresh run and an inversion after that point goes unreported. Here the surrounding run survives the callback, so `afterAll(() => { beforeEach(() => {}); }); beforeAll(() => {});` is reported.

- **A nested hook is never compared against a hook from the enclosing run.** Upstream lets the enclosing run's index leak inward, which both reports correctly ordered nested runs and attributes reports to a hook at another nesting level. Here `afterAll(() => { beforeEach(() => {}); afterEach(() => {}); })` is accepted, and a genuinely inverted nested run is reported against the hook that precedes it at its own level.

- **`@rstest/playwright` member hooks take part in the ordering.** `test.beforeEach()` and `test.extend({}).afterAll()` resolve as hooks through the shared parser, so they are ordered against their neighbours; upstream has no member hook form. Conversely `@rstest/core`'s `test.beforeEach`, and `test.skip.beforeEach` / `test.beforeEach.only`, do not resolve as hooks, which makes them ordinary calls that end the current run — the same treatment any other non-hook call gets.

Every hook-inside-hook shape is covered on both the Jest and the Rstest side, asserting the full message so the hook a report is attributed to is pinned, not just its position.

## Related Links

Related #935

## Checklist

### DIFF
--- a/internal/plugins/rstest/rules/prefer_hooks_in_order/prefer_hooks_in_order.md
+++ b/internal/plugins/rstest/rules/prefer_hooks_in_order/prefer_hooks_in_order.md
@@ -1,0 +1,56 @@
+# prefer-hooks-in-order
+
+## Rule Details
+
+Prefer Rstest lifecycle hooks in the same order Rstest runs them:
+
+1. `beforeAll`
+2. `beforeEach`
+3. `afterEach`
+4. `afterAll`
+
+This rule checks each consecutive run of hook calls and reports a hook that
+appears after another hook from a later runtime stage. A normal non-hook call
+starts a new run. Comments, blank lines, and statements that do not contain a
+call do not.
+
+Examples of **incorrect** code for this rule:
+
+```ts
+afterAll(() => {});
+beforeAll(() => {});
+
+import { test } from "@rstest/playwright";
+
+test.afterEach(() => {});
+test.beforeEach(() => {});
+```
+
+Examples of **correct** code for this rule:
+
+```ts
+beforeAll(() => {});
+beforeEach(() => {});
+afterEach(() => {});
+afterAll(() => {});
+
+afterAll(() => {});
+doSomething();
+beforeAll(() => {});
+```
+
+The rule follows Rstest hooks imported from `@rstest/core`, globals,
+CommonJS/namespace forms, `import.meta.rstest`, same-file aliases, and
+`@rstest/playwright` member hooks such as `test.beforeEach()` and
+`test.extend({}).afterAll()`.
+
+## Differences from ESLint
+
+- Nested hook callbacks do not accidentally end the surrounding hook run. A
+  later outer hook is still compared against the outer hook sequence after an
+  inner hook callback exits.
+
+## Original Documentation
+
+- [eslint-plugin-jest: prefer-hooks-in-order](https://github.com/jest-community/eslint-plugin-jest/blob/v29.16.1/docs/rules/prefer-hooks-in-order.md)
+- [Source code](https://github.com/jest-community/eslint-plugin-jest/blob/v29.16.1/src/rules/prefer-hooks-in-order.ts)
--- a/internal/plugins/rstest/rules/prefer_hooks_in_order/prefer_hooks_in_order.md
+++ b/internal/plugins/rstest/rules/prefer_hooks_in_order/prefer_hooks_in_order.md
@@ -44,11 +44,41 @@ CommonJS/namespace forms, `import.meta.rstest`, same-file aliases, and
 `@rstest/playwright` member hooks such as `test.beforeEach()` and
 `test.extend({}).afterAll()`.
 
+Hooks written inside a hook callback form their own run at that nesting level.
+They are not compared against the hook that encloses them, and they do not end
+the run that surrounds it:
+
+```ts
+// reported: `beforeEach` comes after `afterAll` in the same run
+afterAll(() => {});
+beforeEach(() => {});
+
+// not reported: the inner run is ordered correctly on its own terms
+afterAll(() => {
+  beforeEach(() => {});
+  afterEach(() => {});
+});
+```
+
 ## Differences from ESLint
 
-- Nested hook callbacks do not accidentally end the surrounding hook run. A
-  later outer hook is still compared against the outer hook sequence after an
-  inner hook callback exits.
+The upstream rule tracks "inside a hook" as a single boolean and ignores every
+call made while it is set. That conflates two separate scopes, and this rule
+scopes each run to its own nesting level instead. Two consequences, both of
+which this rule reports differently from upstream on hook-inside-hook shapes:
+
+- A nested hook callback no longer ends the surrounding run. Upstream clears the
+  boolean as soon as any nested hook call exits, so a later outer hook is
+  compared as if it started a fresh run and an inversion after that point goes
+  unreported. Here the surrounding run survives the callback, so
+  `afterAll(() => { beforeEach(() => {}); }); beforeAll(() => {});` is reported.
+
+- A nested hook is never compared against a hook from the enclosing run.
+  Upstream lets the enclosing run's index leak in, which both reports correctly
+  ordered nested runs and attributes reports to a hook from another nesting
+  level. Here `afterAll(() => { beforeEach(() => {}); afterEach(() => {}); })`
+  is accepted, and a genuinely inverted nested run is reported against the hook
+  that precedes it at its own level.
 
 ## Original Documentation
 
--- a/internal/plugins/rstest/rules/prefer_hooks_in_order/prefer_hooks_in_order.md
+++ b/internal/plugins/rstest/rules/prefer_hooks_in_order/prefer_hooks_in_order.md
@@ -67,11 +67,12 @@ call made while it is set. That conflates two separate scopes, and this rule
 scopes each run to its own nesting level instead. Two consequences, both of
 which this rule reports differently from upstream on hook-inside-hook shapes:
 
-- A nested hook callback no longer ends the surrounding run. Upstream clears the
-  boolean as soon as any nested hook call exits, so a later outer hook is
-  compared as if it started a fresh run and an inversion after that point goes
-  unreported. Here the surrounding run survives the callback, so
-  `afterAll(() => { beforeEach(() => {}); }); beforeAll(() => {});` is reported.
+- A nested hook callback no longer ends the surrounding run. After a nested hook
+  exits, an ordinary call resets upstream's run, so a later outer hook starts
+  fresh and an inversion can go unreported. Here the surrounding run survives
+  the callback, so
+  `afterAll(() => { beforeEach(() => {}); doSomething(); }); beforeAll(() => {});`
+  is reported.
 
 - A nested hook is never compared against a hook from the enclosing run.
   Upstream lets the enclosing run's index leak in, which both reports correctly
--- a/internal/plugins/rstest/rules/prefer_hooks_in_order/prefer_hooks_in_order_extras_test.go
+++ b/internal/plugins/rstest/rules/prefer_hooks_in_order/prefer_hooks_in_order_extras_test.go
@@ -1,0 +1,317 @@
+package prefer_hooks_in_order_test
+
+import (
+	"testing"
+
+	"github.com/web-infra-dev/rslint/internal/plugins/rstest/fixtures"
+	"github.com/web-infra-dev/rslint/internal/plugins/rstest/rules/prefer_hooks_in_order"
+	"github.com/web-infra-dev/rslint/internal/rule_tester"
+)
+
+// TestPreferHooksInOrderExtras locks in branches and edge shapes that the
+// upstream Jest suite does not exercise for Rstest. Each case carries an inline
+// comment for the continuity rule, source shape, or hook-depth branch it
+// protects so future refactors cannot silently regress it.
+func TestPreferHooksInOrderExtras(t *testing.T) {
+	rule_tester.RunRuleTester(
+		fixtures.GetRootDir(),
+		"tsconfig.json",
+		t,
+		&prefer_hooks_in_order.PreferHooksInOrderRule,
+		[]rule_tester.ValidTestCase{
+			// ---- Dimension 4: Rstest sources resolve to the same hook names ----
+			{Code: `import { beforeAll, beforeEach, afterEach, afterAll } from "@rstest/core";
+beforeAll(() => {});
+beforeEach(() => {});
+afterEach(() => {});
+afterAll(() => {});`},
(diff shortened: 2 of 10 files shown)


internal/utils/test_framework/rules/prefer_hooks_in_order/prefer_hooks_in_order.go
```diff
@@ -0,0 +1,112 @@
...
+			hookCallDepth := 0
+			enteredHooks := map[*ast.Node]bool{}
+			parsedCalls := map[*ast.Node]*testFramework.ParsedCall{}
+
+			return rule.RuleListeners{
+				ast.KindCallExpression: func(node *ast.Node) {
+					parsed := runtime.Parse(node)
+					parsedCalls[node] = parsed
+
+					if hookCallDepth > 0 {
```
swwind: Depth tracking is where the divergence comes from. `inHook` is cleared by the *first* nested hook's exit, so ESLint resumes processing calls in the rest of the callback; the counter keeps every one of them invisible until the outer hook exits.

That produces both directions of error. A non-hook call after a nested hook no longer resets `previousHookIndex`, so the outer run survives where ESLint starts a new one, and hooks written after a nested hook in the same callback are never compared at all.

Restoring the boolean matches both `eslint-plugin-jest` and `@vitest/eslint-plugin` on all 25 shapes I tried.

internal/utils/test_framework/rules/prefer_hooks_in_order/prefer_hooks_in_order.go
```diff
@@ -0,0 +1,112 @@
...
+					if !testFramework.IsCallOfKind(parsed, testFramework.FnKindHook) {
+						previousHookIndex = -1
+						return
+					}
+
+					enteredHooks[node] = true
+					hookCallDepth++
+
+					currentHookIndex := testFramework.HookOrderIndex(parsed.Name)
+					if currentHookIndex < 0 {
```
swwind, line 75: Unreachable. `FnKindHook` is only assigned after `testFramework.IsHookName` passes — `internal/plugins/jest/utils/jest.go:157` and `internal/plugins/rstest/utils/parse_rstest_fn.go:561` — so `HookOrderIndex` cannot return a negative here.

internal/utils/test_framework/rules/prefer_hooks_in_order/prefer_hooks_in_order.go
```diff
@@ -0,0 +1,112 @@
...
+						return
+					}
+
+					previousHookIndex = currentHookIndex
+				},
+				rule.ListenerOnExit(ast.KindCallExpression): func(node *ast.Node) {
+					parsed := parsedCalls[node]
+					delete(parsedCalls, node)
+
+					if enteredHooks[node] && testFramework.IsCallOfKind(parsed, testFramework.FnKindHook) {
```
swwind: `enteredHooks` is redundant: both assignment paths cover every hook call, so `enteredHooks[node]` is already equivalent to the `IsCallOfKind(parsed, FnKindHook)` next to it. `parsedCalls` only carries the parse result from enter to exit; upstream just re-parses in the exit handler. Two per-node maps for what is a single scalar's worth of state.

internal/plugins/jest/rules/prefer_hooks_in_order/prefer_hooks_in_order_test.go
```diff
@@ -577,6 +577,18 @@ func TestPreferHooksInOrderRule(t *testing.T) {
 					{MessageId: "reorderHooks", Line: 38, Column: 13},
 				},
 			},
+			{
```
swwind, line 580: Real `eslint-plugin-jest` accepts this source. The nested `beforeEach` exit clears `inHook`, then `doSomething()` resets `previousHookIndex`, so the trailing `beforeAll` starts a fresh run.

This is the only case in the jest suite that requires the counter — with the boolean restored, every other case here still passes. Locking a shipped rule to non-upstream behavior belongs in its own change, not in an rstest rule port.

internal/plugins/rstest/rules/prefer_hooks_in_order/prefer_hooks_in_order_extras_test.go
```diff
@@ -0,0 +1,317 @@
...
+  beforeEach(() => {});
+  doSomething();
+});
+beforeAll(() => {});`,
+				Errors: []rule_tester.InvalidTestCaseError{
+					reorderHooksError("beforeAll", "afterAll", 5, 1),
+				},
+			},
+			// ---- Locks in shared depth fix: nested hook itself is ignored, not compared ----
+			{
```
swwind, line 297: ESLint emits two diagnostics for this source: `afterEach` at `3:3` and `beforeAll` at `5:1`. The first is dropped here, and the comment presents that as intended rather than as a divergence.

internal/plugins/rstest/rules/prefer_hooks_in_order/prefer_hooks_in_order.md
```diff
@@ -0,0 +1,56 @@
...
+doSomething();
+beforeAll(() => {});
+```
+
+The rule follows Rstest hooks imported from `@rstest/core`, globals,
+CommonJS/namespace forms, `import.meta.rstest`, same-file aliases, and
+`@rstest/playwright` member hooks such as `test.beforeEach()` and
+`test.extend({}).afterAll()`.
+
+## Differences from ESLint
```
swwind, line 63: This lists only the continuity half. The observable consequences are missing: extra reports where ESLint is silent, and missing reports for hooks written after a nested hook in the same callback.

`A normal non-hook call starts a new run` at the top of the file also no longer holds inside a hook callback.

swwind (commented):
The shared body rewrites eslint-plugin-jest's `inHook` boolean as a `hookCallDepth` counter. That is not behavior-preserving, and because `jest/prefer-hooks-in-order` now runs on the shared body, this PR silently changes a shipped rule while being scoped to adding an rstest rule.

I ran 25 hook-ordering shapes through `eslint-plugin-jest@29.16.1` and `@vitest/eslint-plugin` — the two agree on all 25 — and through this branch. 11 diverge. Every divergence requires a hook call nested inside another hook call's subtree; without such nesting `hookCallDepth` is exactly `inHook`.

False positives (ESLint silent, this PR reports):

| Code | ESLint | This PR |
| --- | --- | --- |
| `afterAll(() => { beforeEach(() => {}); doSomething(); }); beforeAll(() => {});` | no error | `5:1` reorderHooks |
| `afterAll(() => { describe('inner', () => { beforeAll(() => {}); }); }); beforeAll(() => {});` | no error | `6:1` reorderHooks |
| `afterAll(() => { function helper() { beforeAll(() => {}); } helper(); }); beforeAll(() => {});` | no error | `5:1` reorderHooks |
| `afterAll(() => { beforeEach(() => {}); it('inner', () => {}); }); beforeAll(() => {});` | no error | `5:1` reorderHooks |

False negatives (ESLint reports, this PR silent):

| Code | ESLint | This PR |
| --- | --- | --- |
| `afterAll(() => { beforeEach(() => {}); afterEach(() => {}); });` | `3:3` afterEach before afterAll | no error |
| `afterAll(() => { describe('x', () => { afterEach(() => {}); beforeEach(() => {}); }); });` | `4:5` beforeEach before afterAll | no error |
| `beforeAll(() => { afterAll(() => {}); beforeEach(() => {}); }); beforeAll(() => {});` | `5:1` beforeAll before beforeEach | no error |
| `afterAll(() => { beforeEach(() => { afterEach(() => {}); }); beforeAll(() => {}); });` | `5:3` beforeAll before afterAll | no error |

Same position, different `previousHook` in the message — line/column-only assertions do not catch these:

| Code | ESLint | This PR |
| --- | --- | --- |
| `afterEach(() => { beforeEach(() => {}); afterAll(() => {}); }); beforeAll(() => {});` | `5:1` before any `afterAll` | `5:1` before any `afterEach` |
| `afterAll(() => { describe('d', () => { beforeEach(() => {}); }); afterEach(() => {}); }); beforeAll(() => {});` | `5:1` before any `afterEach` | `5:1` before any `afterAll` |

Different position:

| Code | ESLint | This PR |
| --- | --- | --- |
| `afterAll(() => { beforeEach(() => {}); afterEach(() => {}); doSomething(); }); beforeAll(() => {});` | `3:3` afterEach before afterAll | `6:1` beforeAll before afterAll |

Nothing outside this PR's own new assertions asks for the rewrite. With the shared body restored to the `inHook` boolean, the whole upstream jest suite passes except `invalid-17` (the case added here), and the rstest upstream suite passes in full — only `extras/invalid-17` and `extras/invalid-18` fail, both added here. Under `inHook`, `extras/invalid-18` produces exactly the two diagnostics real ESLint produces for that source.

None of the shapes above are covered by either suite today — worth landing the divergent ones as cases once the semantics are settled.

Rest of the diff looks good: registration in `internal/plugins/rstest/all.go`, omission from the rstest recommended preset, the `rstest.config.mts` entry, and the JS mirror test expectations.

elecmonkey:
Thanks for running the 25 shapes — the mechanism you describe is accurate, and the two gaps it exposed are fixed in 7a91547. The depth counter is replaced by one `previousHookIndex` per hook frame: entering a hook pushes a frame for its callback, exiting pops it, so a run is only ever compared against hooks declared at its own nesting level.

On the buckets:

**The 4 "false positives".** I don't think these are false positives. All four are a top-level `afterAll(...)` followed by a top-level `beforeAll(...)`, which the rule reports whenever nothing is nested. Upstream is silent only because the nested hook's exit clears `inHook`, after which the following statement resets `previousHookIndex`. These stay reported.

**2 of the 4 false negatives are real, and are now fixed.** `afterAll(() => { describe('x', () => { afterEach(); beforeEach(); }); });` and `afterAll(() => { beforeEach(() => { afterEach(); }); beforeAll(); });` are genuine inversions written inside a callback, and ignoring everything below depth 0 meant they were never checked at all. They are now reported at the position upstream reports, attributed to the hook that actually precedes them at their own level (`afterEach` and `beforeEach`) rather than the enclosing `afterAll`.

**The other 2 are upstream reporting across nesting levels.** In `afterAll(() => { beforeEach(); afterEach(); })` the callback's `beforeEach`/`afterEach` pair is correctly ordered and is only reported because the enclosing `afterAll`'s index leaked inward; in `beforeAll(() => { afterAll(); beforeEach(); }); beforeAll();` a top-level `beforeAll` is compared against a `beforeEach` written inside another hook's callback. Frame scoping accepts the first and reports the second where the inversion actually is — `beforeEach` after `afterAll`, inside the callback.

**The `previousHook` and position rows** come from the same leak, so they now resolve to the hook at the reporting level: `beforeAll` before `afterEach` and `beforeAll` before `afterAll` respectively, and `6:1` instead of `3:3` for the last one.

All 11 shapes are covered as cases in `prefer_hooks_in_order_nesting_test.go` on both sides, asserting the full message rather than line/column so attribution is pinned; two also went into the JS mirror, which asserts `data`. The upstream Jest and Rstest suites are untouched and pass. The PR body now records both divergences from upstream instead of just the first.

swwind (commented):
nit: The example here is not actually a difference from upstream. `eslint-plugin-jest` also reports `afterAll(() => { beforeEach(() => {}); }); beforeAll(() => {});`, because no non-hook call resets its `previousHookIndex` after the nested hook exits.

Could we use the shape that demonstrates the real divergence instead?

```suggestion
A nested hook callback no longer ends the surrounding run. After a nested hook
exits, an ordinary call resets upstream's run, so a later outer hook starts
fresh and an inversion can go unreported. Here the surrounding run survives
the callback, so
`afterAll(() => { beforeEach(() => {}); doSomething(); }); beforeAll(() => {});`
is reported.
```

internal/plugins/rstest/rules/prefer_hooks_in_order/prefer_hooks_in_order.md
```diff
@@ -0,0 +1,86 @@
...
+The upstream rule tracks "inside a hook" as a single boolean and ignores every
+call made while it is set. That conflates two separate scopes, and this rule
+scopes each run to its own nesting level instead. Two consequences, both of
+which this rule reports differently from upstream on hook-inside-hook shapes:
+
+- A nested hook callback no longer ends the surrounding run. Upstream clears the
+  boolean as soon as any nested hook call exits, so a later outer hook is
+  compared as if it started a fresh run and an inversion after that point goes
+  unreported. Here the surrounding run survives the callback, so
+  `afterAll(() => { beforeEach(() => {}); }); beforeAll(() => {});` is reported.
```
swwind: nit: The example here is not actually a difference from upstream. `eslint-plugin-jest` also reports `afterAll(() => { beforeEach(() => {}); }); beforeAll(() => {});`, because no non-hook call resets its `previousHookIndex` after the nested hook exits.

Could we use the shape that demonstrates the real divergence instead?

```suggestion
- A nested hook callback no longer ends the surrounding run. After a nested hook
  exits, an ordinary call resets upstream's run, so a later outer hook starts
  fresh and an inversion can go unreported. Here the surrounding run survives
  the callback, so
  `afterAll(() => { beforeEach(() => {}); doSomething(); }); beforeAll(() => {});`
  is reported.
```

swwind (approved):
LGTM

PR #1837, opened 2026-08-21
feat(rstest): add no-alias-methods rule

elecmonkey:
## Summary

Report the eleven deprecated matcher aliases that `@vitest/expect` still ships — `toBeCalled`, `lastCalledWith`, `toThrowError` and the rest — and rewrite each to its canonical name. The alias table is read from the existing `rstestUtils.RSTEST_MATCHER_ALIASES` rather than copied, so it stays tied to the `@vitest/expect` version Rstest depends on.

Matcher recognition goes through the shared `ParseExpectCall`, so every Rstest expect source — global, import, alias, namespace, `import.meta.rstest`, and the test-context `expect` — is covered, and an `expect` imported from another framework is left alone. The fix rewrites the matcher accessor in place through the shared `test_framework.AccessorReplacement` and the deferred fix API. Registered in the rstest plugin but left out of the `recommended` preset, matching upstream.

A second commit fixes two chain-parsing bugs the rule surfaced. `findRstestExpectModifiersAndMatchers` failed the whole parse when a chain carried a computed identifier key anywhere in it, or any unrecognised member after the matcher. Both shapes are ordinary code: a member after the matcher call reads a property of the assertion's own result, and a computed key names a matcher only at runtime. Neither undoes the matcher that already parsed, so the chain is now truncated at that member instead. This was already producing false positives in `rstest/valid-expect` on `main` — `const r = expect(a).toHaveBeenCalled().message;` was reported as an unknown modifier and `expect(a).toReturn()[key]()` as a missing matcher call — and matching false negatives here. A member the grammar rejects *before* any matcher still fails the parse, unchanged.

## Credits

Port from [`eslint-plugin-vitest`](https://github.com/vitest-dev/eslint-plugin-vitest)'s `no-alias-methods`, cross-checked against [`eslint-plugin-jest`](https://github.com/jest-community/eslint-plugin-jest)'s `no-alias-methods`.

- **The fix preserves the accessor's original delimiter.** Upstream rewrites an element access to a single-quoted string, so `expect(spy)["toBeCalled"]()` comes back with different quotes and ``expect(spy)[`toBeCalled`]()`` loses its template literal. This port replaces only the name inside the existing accessor, leaving `'`, `"` and `` ` `` as the author wrote them; a plain property access is rewritten as a property access.

- **The message says "canonical name of".** Upstream's text is `Replace {{ alias }}() with its canonical name {{ canonical }}()`. This port keeps the wording and messageId (`replaceAlias`) the rslint Jest rule already ships, so both plugins report the same sentence.

- **Computed identifier keys are never reported.** Upstream accepts `expect(spy)[matcherName]()` as a supported accessor, matches the *variable's own name* against the alias table, and autofixes it by rewriting the variable reference. That answer does not follow from the code — the variable's value is what names the matcher — so `const toBeCalled = 'toHaveBeenCalled'; expect(spy)[toBeCalled]()` is reported and rewritten upstream even though it already calls the canonical matcher. This port stays silent on every computed key rather than report on a name that may not be the matcher's.

- **Template-literal keys are compared by their cooked value.** Upstream compares `quasis[0].value.raw`, so an escaped key such as ``expect(spy)[`toBeCall\u0065d`]()`` is not reported there; this port reports and fixes it. The cooked value is the property name the runtime actually looks up, so this is the more accurate answer, but it does turn upstream-valid code into an error. String-literal keys agree in both implementations, which also use the cooked value.

- **Only the first matcher in a chain is examined.** This matches upstream's parser, which stops at the first matcher even though a Chai chain can execute several. Stated explicitly because Rstest's parser does expose the full `Matchers` slice, and walking it would widen the reported set beyond the upstream contract.

## Related Links

Related #935

## Checklist


swwind (commented):
Checked against real `@vitest/eslint-plugin@1.6.27` with a 93-case edge-case corpus, running every case through both implementations and diffing diagnostics and one-pass fix output.

The core matcher and fix path matches upstream exactly: all eleven aliases, every expect source, optional chaining in all positions, modifier count and order rules, Chai language chains, `expect.soft` / `poll` / `element` / `any`, TS `!` / `as` / `satisfies` / type arguments, nesting and argument positions, astral-character columns, and spread arguments. 71 of the 93 cases are identical, and 4 of the remaining 22 are the documented quote-preservation divergence.

The rest fall into three groups:

| Code | ESLint | This PR |
| --- | --- | --- |
| `expect(a).toBeCalled().foo` | `1:11`, fixed | not reported |
| `expect(a).toReturn()['x']` | `1:11`, fixed | not reported |
| `expect(a).toBeCalled()[0]` | `1:11`, fixed | not reported |
| `const r = expect(a).toBeCalled().message;` | `1:21`, fixed | not reported |
| `expect(a).toReturn()[toBeCalled]()` | `1:11`, fixed | not reported |
| `expect(a).toBeCalledWith(1)[key]()` | `1:11`, fixed | not reported |
| `expect(a).toBeCalled()[key]` | `1:11`, fixed | not reported |
| `expect(spy).to.have[toBeCalled]()` | `1:21`, fixed | not reported |
| `expect[toBeCalled]()` | `1:8`, fixed | not reported |
| `expect(spy)[toBeCalled]().and.toReturn()` | `1:13`, fixed | not reported |
| `expect(spy)[toBeCalled]()` | `1:11`, fixed | `1:11`, not fixed |
| ``expect(a)[`toBeCall\u0065d`]()`` | not reported | `1:11`, fixed |

Details inline.

internal/plugins/rstest/rules/no_alias_methods/no_alias_methods.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+var NoAliasMethodsRule = rule.Rule{
+	Name:   "rstest/no-alias-methods",
+	Schema: rule.EmptyArraySchema,
+	Run: func(ctx rule.RuleContext, options []any) rule.RuleListeners {
+		analysis := rstestUtils.GetRstestCallAnalysis(ctx)
+		return rule.RuleListeners{
+			ast.KindCallExpression: func(node *ast.Node) {
+				parsed := analysis.ParseExpectCall(node)
```
swwind, line 32: Two parse-level shapes drop a matcher that upstream resolves, so a plain, non-computed alias goes unreported.

A trailing member access after the matcher call: `expect(a).toBeCalled().foo`, `expect(a).toReturn()['x']`, `expect(a).toBeCalled()[0]`, `const r = expect(a).toBeCalled().message;`. `findTopMostRstestExpectExpression` extends the expression through trailing members, so `foo` joins the chain and `findRstestExpectModifiersAndMatchers` returns `modifier-unknown` on the member after the matcher. Upstream's chain comes from the visited call's callee only, so a trailing member is never in it and the alias is reported and fixed. A trailing member that happens to be a Chai chainable (`.length`) or a modifier (`.not`) is reported by both, which is what makes this hard to spot.

A computed key anywhere in the chain: `expect(a).toReturn()[toBeCalled]()`, `expect(a).toBeCalledWith(1)[key]()`, `expect(a).toBeCalled()[key]`. `isComputedDynamicMemberName` makes the whole parse return `matcher-not-found`, including for an ordinary alias earlier in the chain that has nothing to do with the computed key. Upstream treats a computed identifier as a supported accessor, resolves the first matcher normally and reports it.

Both live in `parse_rstest_expect.go` rather than in this PR, so they aren't regressions it introduces — but this rule surfaces them as silent false negatives, and the differences section doesn't mention them.

internal/plugins/rstest/rules/no_alias_methods/no_alias_methods.go
```diff
@@ -0,0 +1,148 @@
...
+		return aliasHit{}, false
+	}
+	return aliasHit{
+		alias:     parsed.Matcher,
+		canonical: canonical,
+		node:      parsed.MatcherEntry.Node,
+	}, true
+}
+
+func strictComputedIdentifierFallback(
```
swwind: `strictComputedIdentifierFallback` is narrower than upstream, and the PR body's premise for it is not what upstream does. The body says `expect(spy)[matcherName]()` leaves upstream seeing "nothing at all"; upstream's `getNodeChain` accepts a computed identifier as a supported accessor and `getAccessorValue` returns its name, so upstream both reports and autofixes it (the fix rewrites the variable reference, which is arguably wrong — but it is not silence).

Three shapes this fallback rejects are upstream errors:

- `expect(spy).to.have[toBeCalled]()` → `1:21` (rejected because `to` / `have` aren't in `RSTEST_EXPECT_MODIFIER_NAMES`)
- `expect(spy)[toBeCalled]().and.toReturn()` → `1:13` (rejected by `last.Call != listenerNode`)
- `expect[toBeCalled]()` → `1:8` (rejected by the `parsed.Head == nil` guard)

Either widen the fallback to cover them, or keep the narrow behaviour and restate the rationale accurately: the port is deliberately stricter than upstream here, rather than filling a gap upstream leaves empty.

internal/plugins/rstest/rules/no_alias_methods/no_alias_methods_extras_test.go
```diff
@@ -0,0 +1,451 @@
...
+			{Code: `expect(spy).ok.and.toBeCalled()`},
+			{Code: `expect(spy).a("function").and.toBeCalled()`},
+
+			// ---- C. strict computed-identifier fallback fail-closed boundaries ----
+			{Code: `expect(spy).to.have[toBeCalled]()`},
+			{Code: `expect(spy).equal(other)[toBeCalled]()`},
+			{Code: `expect(spy).ok[toBeCalled]()`},
+			{Code: `expect(spy).unknown[toBeCalled]()`},
+			{Code: `expect(spy)[toBeCalled]().and.toReturn()`},
+			{Code: `expect[toBeCalled]()`},
```
swwind, line 64: Three of these are upstream errors rather than valid code: `expect(spy).to.have[toBeCalled]()` (`1:21`), `expect(spy)[toBeCalled]().and.toReturn()` (`1:13`) and `expect[toBeCalled]()` (`1:8`) are all reported and fixed by `@vitest/eslint-plugin@1.6.27`. Locking them in as valid is fine as a deliberate stricter-than-upstream choice, but "fail-closed boundaries" reads as if upstream ignores them too.

internal/plugins/rstest/rules/no_alias_methods/no_alias_methods.md
```diff
@@ -0,0 +1,48 @@
...
+This rule has no options.
+
+## Autofix
+
+This rule autofixes supported accessor forms and preserves the original accessor delimiter:
+
+- `expect(spy).toBeCalled()` becomes `expect(spy).toHaveBeenCalled()`
+- `expect(spy)['toBeCalled']()` becomes `expect(spy)['toHaveBeenCalled']()`
+- `expect(spy)["toBeCalled"]()` becomes `expect(spy)["toHaveBeenCalled"]()`
+- ``expect(spy)[`toBeCalled`]()`` becomes ``expect(spy)[`toHaveBeenCalled`]()```
```
swwind: The inline code span ends with three backticks instead of two, so it never closes and the line renders wrong.

internal/plugins/rstest/rules/no_alias_methods/no_alias_methods.md
```diff
@@ -0,0 +1,48 @@
...
+
+Terminal computed identifier keys such as `expect(spy)[matcherName]()` are reported only with no prefix or a valid `expect` modifier prefix. Chai-prefixed dynamic keys are ignored, and reported dynamic keys are never fixed.
```
swwind: This example never triggers. The fallback looks up `RSTEST_MATCHER_ALIASES[last.Name]`, i.e. the identifier's own text, so `expect(spy)[matcherName]()` is reported only when the variable is literally named `toBeCalled` or another alias. As written, a reader will expect a diagnostic on their own `expect(spy)[matcherName]()` and never get one.

internal/plugins/rstest/rules/no_alias_methods/no_alias_methods.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Differences from eslint-plugin-vitest
+
+Rslint uses the same matcher mapping as `eslint-plugin-vitest@v1.6.27`, but two user-visible details are different:
```
swwind: "two user-visible details" undercounts. This section should also carry:

- computed identifier keys: upstream reports and fixes them, this rule reports without a fix, and stays silent on the Chai-prefixed, static and trailing-chain forms;
- template-literal keys with escapes: upstream compares `quasis[0].value.raw`, so ``expect(a)[`toBeCall\u0065d`]()`` is not reported there, while this rule compares the cooked text and reports and fixes it. The cooked value is the property name the runtime actually uses, so this rule is the more correct of the two — but it turns upstream-valid code into an error, which is what this section is for. (String-literal keys agree: both use the cooked value.)

internal/plugins/rstest/rules/no_alias_methods/no_alias_methods.go
elecmonkey: Both confirmed, and worse than false negatives: the same parse failures are active false positives in `rstest/valid-expect` on `main`. `const r = expect(a).toHaveBeenCalled().message;` reports "Expect has an unknown modifier", and `expect(a).toReturn()[key]()` reports "Expect must have a corresponding matcher call". Fixed in ad53d2e7 rather than deferred.

`findRstestExpectModifiersAndMatchers` now classifies a computed identifier key as an unrecognised member instead of failing the parse, and ends the chain at the first unrecognised member once a matcher has been found. A member the grammar rejects before any matcher still fails the parse, so `expect(x).to.be("string").that.does.not.contain("world")` and `expect(x).to.have.foo.equal(y)` are unchanged. The one existing case that moved is `expect(x).to.be.a("string").that.foo.contain("x")`, which now parses with matcher `a`; it has a dedicated test with the reasoning. The six shapes from your table are locked in as invalid with fix output, and the `valid-expect` false positives as valid.

internal/plugins/rstest/rules/no_alias_methods/no_alias_methods.go
elecmonkey: You're right that the premise was wrong, and I took the third option: the fallback is gone.

It looked up `RSTEST_MATCHER_ALIASES[last.Name]` — the variable's own name, not its value. So `const toBeCalled = "toHaveBeenCalled"; expect(spy)[toBeCalled]()` was reported as an alias use even though it already calls the canonical matcher. Upstream has the same defect and autofixes it, which rewrites working code; widening to match would have widened that. The rule now stays silent on every computed key, and the PR body says so accurately.

internal/plugins/rstest/rules/no_alias_methods/no_alias_methods_extras_test.go
elecmonkey: Reworded. The section now says these are deliberately stricter than upstream and that upstream reports and autofixes them, and the four former fallback hits moved into it with the same note.

internal/plugins/rstest/rules/no_alias_methods/no_alias_methods.md
elecmonkey: Fixed.

internal/plugins/rstest/rules/no_alias_methods/no_alias_methods.md
elecmonkey: Removed along with the fallback.

internal/plugins/rstest/rules/no_alias_methods/no_alias_methods.md
elecmonkey: Both added, and the count is now four. The computed-key entry explains why upstream's answer is unsound rather than just noting the difference; the template-literal entry states that the cooked value is what the runtime looks up, so this rule is the more accurate one, and that it turns upstream-valid code into an error. An escaped-key case is locked in as invalid with its fix output.

internal/plugins/rstest/rules/no_alias_methods/no_alias_methods_upstream_test.go
```diff
@@ -0,0 +1,73 @@
+// TestNoAliasMethodsUpstream migrates the complete
+// eslint-plugin-vitest@v1.6.27 no-alias-methods suite. Rstest-only edge cases,
+// real-user shapes, computed-identifier fallback, and edit-demand coverage
```
swwind: Nit: this header still says “computed-identifier fallback”, but the fallback was removed. Could this instead describe the computed-key policy (or omit that phrase)?

internal/plugins/rstest/rules/no_alias_methods/no_alias_methods_extras_test.go
```diff
@@ -0,0 +1,487 @@
+// TestNoAliasMethodsExtras locks in the Rstest-only augmentation required by
+// the port spec: first-matcher contract, strict computed-identifier fallback,
```
swwind: Nit: this header still refers to the removed “strict computed-identifier fallback”. Please update it to describe the current “computed keys are never reported” policy.

internal/plugins/rstest/rules/no_alias_methods/no_alias_methods_upstream_test.go
elecmonkey: Updated the header to describe the current computed-key non-reporting policy in 07552ad5.

internal/plugins/rstest/rules/no_alias_methods/no_alias_methods_extras_test.go
elecmonkey: Updated the header to describe the current computed-key non-reporting policy in 07552ad5.

internal/plugins/rstest/rules/no_alias_methods/no_alias_methods.md
```diff
@@ -0,0 +1,50 @@
...
+- computed identifier keys are never reported. Upstream treats `expect(spy)[matcherName]()` as a supported accessor, matches the *variable's own name* against the alias table, and autofixes it by rewriting the variable reference. That answer does not follow from the code: the variable's value is what names the matcher, so `const toBeCalled = 'toHaveBeenCalled'` is reported and rewritten even though it already calls the canonical matcher. This rule stays silent instead;
+- template-literal keys are compared by their cooked value, so ``expect(spy)[`toBeCall\u0065d`]()`` is reported and fixed. Upstream compares `quasis[0].value.raw` and reports nothing. The cooked value is the property name the runtime actually looks up, so this is the more accurate answer, but it does turn upstream-valid code into an error. String-literal keys agree in both implementations, which also use the cooked value.
+
+## Original Documentation
+
+- [eslint-plugin-vitest: no-alias-methods](https://github.com/vitest-dev/eslint-plugin-vitest/blob/v1.6.27/docs/rules/no-alias-methods.md)
```
swwind: Would you like to remove this? As this is not a port of `vitest` rule. You can cite it but not in the "Original Documentation" section. We may not follow upstream continuously.

elecmonkey: done

swwind (approved):
LGTM

PR #1838, opened 2026-08-21
feat(rstest): add max-expects rule

elecmonkey:
## Summary

Report each assertion past a configurable maximum (default 5) within one test body, so a test that has grown into a checklist becomes visible. Counting goes through the shared `ParseExpectCall`, so `expect(value)`, `expect.soft(value)`, `expect.poll(fn)` and `expect.element(locator)` all count, while static helpers such as `expect.assertions(2)`, `expect.hasAssertions()` and asymmetric matchers do not, and neither do broken chains such as a bare `expect(value)` — those belong to `rstest/valid-expect`. A Chai chain counts once per assertion factory, so `expect(value).to.be.a("string").and.not.be.empty` is one assertion.

Test boundaries come from `analysis.Callbacks().Functions`, which resolves inline callbacks, all three overloads, hoisted and variable-held functions, `.each`, `.for` and `test.extend`. A lifecycle hook body carries its own count as well: `beforeAll`, `beforeEach`, `afterEach` and `afterAll` are test code and are limited the same way a test is. Registered in the rstest plugin but left out of the `recommended` preset, matching upstream.

## Credits

Port from [`eslint-plugin-jest`](https://github.com/jest-community/eslint-plugin-jest)'s `max-expects`, cross-checked against [`eslint-plugin-vitest`](https://github.com/vitest-dev/eslint-plugin-vitest)'s `max-expects`.

- **Counts live on a stack of frames, not in one global integer.** Upstream keeps a single counter and resets it to 0 on both entry and exit of any function that is not the registration's direct callback, which discards the enclosing test's tally — the assertions that follow are counted from zero and the ones over the limit go unreported. This port pushes a frame for each counting boundary and pops back to the enclosing count, so the inner function gets its own tally without erasing the test's. The divergence is not limited to a named helper: any function-valued expression between two assertions — a nested registration, an object method, a `new` with a callback — behaves this way.

- **An unresolved callback is recognised by position, anywhere inside the registration's callback argument.** A callback passed by name has no syntactic ancestor tying it to its registration, so its assertions are attributed through the resolved callback set. When the parser cannot resolve the callback, the rule falls back to a registration frame that any function written inside the callback argument activates, whether it is a direct argument of the wrapping call (`test('x', wrap(() => …))`) or nested further in (`new Wrapper(() => …)`, `wrap([() => …])`, `wrap({ cb: () => … })`, `cond ? () => … : null`). A function held in the options object of the `(name, options, callback)` overload is not a callback and does not activate it. The fallback uses `analysis.ParseTestCall` only; nothing is added to the shared analysis.

- **A callback written outside the registration and reached through a member expression is not counted.** `const suite = { run: () => … }; test('a', suite.run)`, and the same shape with a class property, stay silent: the function is written outside the registration and nothing at the registration ties the two together. Upstream counts them because it counts assertions everywhere rather than only in test bodies — which is also why assertions at the top level of a file, or directly in a `describe` body, are ignored here.

- **Ordinary function declarations do not reset the count.** Upstream listens on function expressions and arrow functions; the rslint Jest port additionally covers methods, accessors and constructors because tsgo represents them differently from ESTree. A plain `function` declaration is a boundary in neither, so it inherits the current frame here as well.

## Related Links

Related #935

## Checklist

### DIFF
--- a/internal/plugins/rstest/rules/max_expects/max_expects.go
+++ b/internal/plugins/rstest/rules/max_expects/max_expects.go
@@ -183,6 +183,29 @@ func containsFallbackFunction(node *ast.Node) bool {
 	return found
 }
 
+// fallbackCandidateRoot is the argument subtree a callback may be found in. An
+// object literal is never one: it is the options argument of an overload, and
+// a function inside it is a property, not the registered callback.
+func fallbackCandidateRoot(node *ast.Node) *ast.Node {
+	node = internalUtils.SkipAssertionsAndParens(node)
+	if node == nil || isObjectLiteralLike(node) {
+		return nil
+	}
+	return node
+}
+
+// hookFallbackCandidateRoots answers the same question for a lifecycle hook,
+// whose callback is the first argument rather than the second.
+func hookFallbackCandidateRoots(call *ast.CallExpression) []*ast.Node {
+	if call == nil || call.Arguments == nil || len(call.Arguments.Nodes) == 0 {
+		return nil
+	}
+	if root := fallbackCandidateRoot(call.Arguments.Nodes[0]); root != nil {
+		return []*ast.Node{root}
+	}
+	return nil
+}
+
 func registrationFallbackCandidateRoots(call *ast.CallExpression) []*ast.Node {
 	if call == nil || call.Arguments == nil {
 		return nil
@@ -193,13 +216,7 @@ func registrationFallbackCandidateRoots(call *ast.CallExpression) []*ast.Node {
 		return nil
 	}
 
-	candidateRoot := func(node *ast.Node) *ast.Node {
-		node = internalUtils.SkipAssertionsAndParens(node)
-		if node == nil || isObjectLiteralLike(node) {
-			return nil
-		}
-		return node
-	}
+	candidateRoot := fallbackCandidateRoot
 
 	if len(args) == 2 {
 		if root := candidateRoot(args[1]); root != nil {
@@ -266,7 +283,14 @@ var MaxExpectsRule = rule.Rule{
 
 		enterFunction := func(node *ast.Node) {
 			kind := functionPushForNode(node, callbacks)
-			if kind == functionPushNone &&
+			// A callback the parser could not resolve is recognised by position:
+			// any function inside the registration's callback argument activates
+			// the registration frame. That has to be checked on the detached
+			// branch as well, because only a function whose immediate parent is
+			// the wrapping call reaches functionPushNone — nesting it in a `new`,
+			// an array, an object property, or a conditional makes it detached,
+			// and pushing an inactive frame there would leave the body uncounted.
+			if (kind == functionPushNone || kind == functionPushDetached) &&
 				node != nil &&
 				node.Body() != nil &&
 				canActivateRegistrationFallback(stack.top(), node) {
@@ -321,6 +345,21 @@ var MaxExpectsRule = rule.Rule{
 						false,
 						node,
 						registrationFallbackCandidateRoots(node.AsCallExpression()),
+					)
+					pushedRegistrations[node] = true
+					return
+				}
+
+				// A hook body is test code and gets its own tally, the same as a
+				// test body. Hooks have no resolved callback set, so the frame is
+				// activated positionally, through the same candidate roots.
+				if parsed := analysis.ParseFnCall(node); parsed != nil &&
+					parsed.Kind == rstestUtils.RstestFnTypeHook {
+					stack.push(
+						frameRegistrationFallback,
+						false,
+						node,
+						hookFallbackCandidateRoots(node.AsCallExpression()),
 					)
 					pushedRegistrations[node] = true
 					return
--- a/internal/plugins/rstest/rules/max_expects/max_expects.go
+++ b/internal/plugins/rstest/rules/max_expects/max_expects.go
@@ -1,0 +1,348 @@
+package max_expects
+
+import (
+	_ "embed"
+	"fmt"
+	"strconv"
+
+	"github.com/microsoft/typescript-go/shim/ast"
+	rstestUtils "github.com/web-infra-dev/rslint/internal/plugins/rstest/utils"
+	"github.com/web-infra-dev/rslint/internal/rule"
+	internalUtils "github.com/web-infra-dev/rslint/internal/utils"
+)
+
+//go:embed max_expects.schema.json
+var schemaJSON []byte
+
+const defaultMax = 5
+
+type options struct {
+	Max int
+}
+
+type frameKind uint8
+
+const (
+	frameInactive frameKind = iota
+	frameTestCallback
+	frameDetachedFunction
+	frameRegistrationFallback
+)
+
+type countFrame struct {
+	kind                   frameKind
+	count                  int
+	active                 bool
+	registration           *ast.Node
+	callbackCandidateRoots []*ast.Node
+}
+
+type frameStack struct {
+	frames []countFrame
+}
+
+type functionPushKind uint8
+
+const (
+	functionPushNone functionPushKind = iota
+	functionPushTestCallback
+	functionPushDetached
+	functionActivateRegistrationFallback
+)
+
+func parseOptions(rawOptions []any) options {
+	opts := options{Max: defaultMax}
+	if len(rawOptions) == 0 {
+		return opts
+	}
+
+	optsMap, _ := rawOptions[0].(map[string]any)
+	if maxVal, ok := internalUtils.CoerceIntegral(optsMap["max"]); ok && maxVal >= 1 {
+		opts.Max = maxVal
+	}
+
+	return opts
+}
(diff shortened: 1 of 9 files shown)


internal/plugins/rstest/rules/max_expects/max_expects.go
```diff
@@ -0,0 +1,348 @@
...
+				node.Body() != nil &&
+				canActivateRegistrationFallback(stack.top(), node) {
+				kind = functionActivateRegistrationFallback
+			}
+
+			switch kind {
+			case functionPushTestCallback:
+				stack.push(frameTestCallback, true, nil, nil)
+				pushedFunctions[node] = functionPushTestCallback
+			case functionPushDetached:
```
swwind, line 304: `canActivateRegistrationFallback` is only consulted on the `functionPushNone` branch, and `functionPushForNode` returns that only when the parent is a `CallExpression`. Any other way of nesting the callback inside the registration argument falls into `functionPushDetached`, which inherits `active: false` from the still-inactive registration frame, so the whole body goes uncounted:

```ts
test('a', new Wrapper(() => { expect(1).toBe(1); expect(2).toBe(2); }));
test('a', wrap([() => { expect(1).toBe(1); expect(2).toBe(2); }]));
test('a', wrap({ cb: () => { expect(1).toBe(1); expect(2).toBe(2); } }));
test('a', cond ? () => { expect(1).toBe(1); expect(2).toBe(2); } : null);
```

All four report at `max: 1` upstream. The `fakeAsync` cases only cover the direct-argument shape, so the hole is not visible from the suite. Consulting `canActivateRegistrationFallback` on the detached branch too — activating the registration frame instead of pushing an inactive detached frame — would close it.

internal/plugins/rstest/rules/max_expects/max_expects.go
```diff
@@ -0,0 +1,348 @@
...
+}
+
+func (stack *frameStack) pop() {
+	if len(stack.frames) <= 1 {
+		return
+	}
+	stack.frames = stack.frames[:len(stack.frames)-1]
+}
+
+func (stack *frameStack) increment() int {
```
swwind, line 108: The root frame is never active, so an assertion that is not lexically inside a resolved test callback is never counted. Beyond top-level statements this also silences hook bodies and callbacks the resolver cannot attribute:

```ts
beforeEach(() => { expect(1).toBe(1); expect(2).toBe(2); });
describe('d', () => { beforeEach(() => { expect(1).toBe(1); expect(2).toBe(2); }); });
const s = { cb: () => { expect(1).toBe(1); expect(2).toBe(2); } };
test('a', s.cb);
class S { run = () => { expect(1).toBe(1); expect(2).toBe(2); }; }
test('a', new S().run);
```

Upstream reports all of these. The `.md` says assertions outside test callbacks are ignored, so the top-level cases read as deliberate, but `beforeEach`/`afterEach` bodies are test code by any reading and are the case most likely to surprise a migrating user — worth an explicit decision rather than falling out of the frame model.

internal/plugins/rstest/rules/max_expects/max_expects_extras_test.go
```diff
@@ -0,0 +1,580 @@
...
+				Options: max1Option,
+			},
+			// ---- Dimension 4: negative sources ----
+			{Code: `import { expect } from 'vitest'; expect(1).toBe(1);`, Options: max1Option},
+			{Code: `import { expect } from '@jest/globals'; expect(1).toBe(1);`, Options: max1Option},
+			{Code: `import { test, expect } from '@playwright/test'; test('case', () => { expect(1).toBe(1); });`, Options: max1Option},
+			{Code: `const expect = () => ({ toBe() {} }); test('case', () => { expect(1).toBe(1); });`, Options: max1Option},
+		},
+		[]rule_tester.InvalidTestCase{
+			// ---- Locks in upstream create() arm: detached helper restores outer count ----
```
swwind: This case is labelled as locking in the upstream `create()` arm, but it is a deliberate divergence: upstream resets the counter to 0 on both entry and exit of a non-test function, while this frame restores the enclosing count. The divergence is also not limited to a named helper — it fires for any function-valued expression sitting between two assertions:

```ts
test('a', () => { expect(1).toBe(1); test('b', () => {}); expect(2).toBe(2); });
test('a', () => { expect(1).toBe(1); const o = { m() {} }; expect(2).toBe(2); });
test('a', () => { expect(1).toBe(1); new Foo(() => {}); expect(2).toBe(2); });
```

Upstream is silent on all three; this rule reports the second assertion. The per-frame counting is the more accurate answer, so the comment is probably what should change rather than the behaviour.

internal/plugins/rstest/rules/max_expects/max_expects.md
```diff
@@ -0,0 +1,78 @@
...
+## Rstest specifics
+
+- `expect.assertions(2)`, `expect.hasAssertions()`, `expect.any(String)`, and
+  other static `expect` helpers do not count as assertions.
+- Broken or incomplete chains such as `expect(value)` and `expect(value).toBe`
+  do not count; those are covered by `rstest/valid-expect`.
+- Chai property and chained assertions still count once per assertion factory:
+  `expect(value).to.be.ok` and
+  `expect(value).to.be.a("string").and.not.be.empty` each count as one.
+- Assertions outside test callbacks are ignored. Detached helpers inside a test
```
swwind: If the wrapper behaviour above stays as-is, this section should also say that a wrapped callback is only recognised when it is a direct argument of the wrapping call. With `fakeAsync(() => ...)` supported, a reader would reasonably expect `new Wrapper(() => ...)` or `wrap({ cb: () => ... })` to be linted too.

swwind (commented):
Compared against real ESLint (`eslint-plugin-jest` v29 `max-expects`, `@typescript-eslint/parser`) across ~50 shapes at `{ "max": 1 }`. Option parsing, schema, message data, chai chain counting, `expect.assertions`/asymmetric-matcher exclusion, `test.each`/`test.for`, `ctx.expect`, optional calls, IIFEs, `as`-wrapped callbacks, generator and `async` callbacks, and the `recommended` preset omission all behave correctly. Three behavioural gaps are worth a look.

| Code (`max: 1`) | ESLint | This PR |
| --- | --- | --- |
| `test('a', wrap(() => { 2 expects }))` | reports | reports |
| `test('a', new Wrapper(() => { 2 expects }))` | reports | silent |
| `test('a', wrap([() => { 2 expects }]))` | reports | silent |
| `test('a', cond ? () => { 2 expects } : null)` | reports | silent |
| `test('a', wrap({ cb: () => { 2 expects } }))` | reports | silent |
| `test('a', wrap(withDb({ run: () => { 2 expects } })))` | reports | silent |
| `class S { run = () => { 2 expects } } test('a', new S().run)` | reports | silent |
| `beforeEach(() => { 2 expects })` | reports | silent |
| `test('a', () => { expect(1); test('b', () => {}); expect(2); })` | silent | reports |
| `test('a', () => { expect(1); const o = { m() {} }; expect(2); })` | silent | reports |
| `test('a', () => { expect(1); new Foo(() => {}); expect(2); })` | silent | reports |

`2 expects` stands for a body containing two `expect(...).toBe(...)` calls.

Details inline.

elecmonkey:
Thanks — all three are addressed in cdf6bf7.

**The wrapper shapes.** Fixed as you suggested: the positional recognition is now consulted on the detached branch as well, so any function written inside the registration's callback argument activates the registration frame rather than pushing an inactive frame of its own. `new Wrapper(() => …)`, `wrap([() => …])`, `wrap({ cb: () => … })`, `cond ? () => … : null` and `wrap(withDb({ run: () => … }))` all report now. A function held in the options object of the `(name, options, callback)` overload still does not activate it, since that argument is not a callback.

**Hook bodies.** Made explicit rather than left to fall out of the frame model: a hook registration now opens a frame of its own, activated through the same candidate roots with the callback in first position. `beforeEach`, `afterAll`, a hook nested in a `describe`, and `beforeEach(wrap(() => …))` all count, and a hook body and the test bodies around it are separate counts.

Three shapes from your list stay silent, and are now recorded as deliberate in the rule doc and the PR body: assertions at the top level of a file, assertions directly in a `describe` body, and a callback reached through a member expression — `const suite = { run: () => … }; test('a', suite.run)` and the class-property form. In those the function is written outside the registration and nothing at the registration ties the two together, so attributing it would be a guess. Upstream reports them because it counts assertions everywhere rather than only in test bodies.

**The `create()` arm comment.** Agreed on both halves — the behaviour is the more accurate answer, so the comment changed instead. It now says the restored count is a deliberate divergence, and your three shapes (a nested registration, an object method, a `new` with a callback between two assertions) are cases.

Everything above is covered in `max_expects_boundaries_test.go` on the Go side, with the hook and nested-wrapper shapes mirrored into the JS test, and the rule doc now spells out which bodies carry a count and which do not.

swwind (approved):
LGTM

PR #1839, opened 2026-08-21
feat(rstest): add no-conditional-in-test rule

elecmonkey:
## Summary

Report conditional control flow inside a test body — `if`, `switch`, ternaries, and `&&` / `||` / `??` — because a branch means one test is covering several execution paths. Optional chaining is reported only under `{ "allowOptionalChaining": false }`, which defaults to `true`. Conditions in a `describe` callback, in a hook, or in a helper declared outside the test are left alone.

Test boundaries come from the test registration call itself, entered and exited with a depth counter, so every argument of the call is inside the test, a function declared outside it is not even when the call names it as its callback, and a nested registration exiting cannot clear the outer test's state. This PR also fixes five call-analysis unwrap points that stopped at parentheses and did not see through TypeScript's `as`, `satisfies`, `!` and `<T>`: without it `test('x', (() => {}) as Fn)` has no recognised callback, which produces a false positive in `no-standalone-expect`. Registered in the rstest plugin but left out of the `recommended` preset, matching upstream.

The traversal itself lives in `internal/utils/test_framework/rules/no_conditional_in_test/`, and both `jest/no-conditional-in-test` and `rstest/no-conditional-in-test` are adapters over it that only say which `CallExpression` registers a test case. The jest rule therefore gains the two behaviors below as well: registration depth is counted rather than flagged, and the reported range of an optional chain covers a trailing `!` (`a?.b!` was reported one character short).

## Credits

Port from [`eslint-plugin-jest`](https://github.com/jest-community/eslint-plugin-jest)'s `no-conditional-in-test`.

- **The Jest contract is used rather than `eslint-plugin-vitest`'s.** The Vitest rule of the same name reports only an `if` statement sitting directly in a callback's block: no `switch`, no ternary, no logical expression, no `allowOptionalChaining`, nothing inside a nested function, and no support for a callback passed by name. That is a narrower implementation, not a framework difference, so this port takes the complete Jest node set and option contract and uses the Vitest cases as a check against accidentally narrowing.

- **An inner registration exiting does not clear the outer test's scope.** Upstream keeps a single flag, so `test('a', () => { test('b', () => {}); if (x) {} })` stops being treated as inside a test once the inner registration exits and the trailing `if` goes unreported. This port counts registration depth instead, so the outer test's scope survives the inner one.

- **Optional chains are matched on tsgo's representation.** tsgo has no `ChainExpression` wrapper — the optional flag sits on individual property, element, call and non-null links — so the rule reports each outermost chain exactly once instead of once per link, and treats a chain inside an argument or a computed key as its own outermost chain. Logical expressions are likewise `BinaryExpression` nodes filtered by operator, so arithmetic, comparison, assignment and comma expressions are never reported.

- **Five shared unwrap points now see through TypeScript assertions.** The callback argument, the named callback's initializer, the file-level function-name index, and the destructured `require('@rstest/core')` initializer all switch from `ast.SkipParentheses` to `SkipAssertionsAndParens`, and the module-top-level check that the file-level index feeds walks the same assertion wrappers. Each of them uses the unwrapped node only to decide "is this a function / is this the rstest module", never as a diagnostic range or fix anchor, so the change is one-directional: rules that report inside a callback gain real reports, and `no-standalone-expect` loses false ones. The remaining nine `SkipParentheses` call sites are deliberately untouched — the six that decide whether a call counts as a registration or an assertion would change the reported set of fifteen merged rules and need their own evaluation, and the three that read a computed member name are retained as fix anchors, where narrowing the range could produce a fix that breaks syntax.

## Related Links

Related #935

## Checklist


internal/plugins/rstest/utils/test_callback_candidate_test.go
```diff
@@ -145,75 +145,88 @@ func TestRstestCallAnalysisCallbacksSkipFilesWithoutTests(t *testing.T) {
...
 
-func TestRstestCallAnalysisCallbacksRemainTestOnly(t *testing.T) {
-	sourceFile := parser.ParseSourceFile(
-		ast.SourceFileParseOptions{
-			FileName: "/describe-only.test.ts",
-			Path:     "/describe-only.test.ts",
-		},
-		`describe.concurrent("suite", () => {});`,
-		core.ScriptKindTS,
 	)
```
swwind: This file is line-sorted from here to the end of the diff: three stray `)`, three copies of `ast.SourceFileParseOptions{`, three `analysis :=` declarations, an unterminated string literal. `gofmt -e` reports errors at 148:2, 212:2 and 245:2, so `internal/plugins/rstest/utils` has no compilable tests — `TestRstestCallAnalysisCallbacksRemainTestOnly` and `TestRstestConcurrentContextBuildsOwnershipLazily` are destroyed, and `TestRstestCallAnalysisCallbacksUnwrapTypeScriptAssertions`, the only coverage for the `SkipAssertionsAndParens` change, never runs.

internal/plugins/rstest/rules/no_conditional_in_test/rule.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+}
+
+var NoConditionalInTestRule = rule.Rule{
+	Name:   "rstest/no-conditional-in-test",
+	Schema: rule.NewSchema(schemaJSON),
+	Run: func(ctx rule.RuleContext, rawOptions []any) rule.RuleListeners {
+		opts := parseOptions(rawOptions)
+		analysis := rstestUtils.GetRstestCallAnalysis(ctx)
+		testCallbackFunctions := analysis.Callbacks().Functions
```
swwind: Using the callback-function set as the scope is what produces both divergence groups in the summary. Upstream flips `inTestCase` on entering and exiting the test `CallExpression`, which means the call's other arguments are in scope and a function that merely happens to be named as the callback is not; `analysis.Callbacks().Functions` is the opposite on both counts. That set exists to bind TestContext parameters, which is a different question from "is this node inside the test call".

internal/plugins/rstest/rules/no_conditional_in_test/no_conditional_in_test_extras_test.go
```diff
@@ -0,0 +1,645 @@
...
+});
+
+beforeEach(() => {
+  if (flag) {
+    prepare();
+  }
+});
+    `},
+
+		// ---- Dimension 4: unresolved wrapper is fail-closed ----
```
swwind: ESLint reports the `if` here at 1:27-1:47 — the arrow is still inside the test `CallExpression` — and `jest/no-conditional-in-test` in this repo reports it too. This belongs in the invalid list.

swwind (commented):
Compared this port against `jest/no-conditional-in-test` from eslint-plugin-jest v29.16.1 (the upstream it names) and against the existing `jest/no-conditional-in-test` in this repo, over 75 cases. This port diverges from ESLint on 19 of them; the in-repo `jest` port diverges on 1.

One blocking item first: `internal/plugins/rstest/utils/test_callback_candidate_test.go` is line-sorted from line 148 onward and no longer parses, so `internal/plugins/rstest/utils` has no compilable tests. Details inline.

### Scope is the callback function, upstream's is the whole test call

Upstream sets `inTestCase` when it enters a test `CallExpression` and clears it on exit, so every argument of the call is in scope. This port starts counting only inside the callback function, so conditionals in the other arguments are missed:

| Code | ESLint | This PR |
| --- | --- | --- |
| `test(cond ? 'a' : 'b', () => {});` | 1:6-1:22 | clean |
| `it(name ?? 'x', () => {});` | 1:4-1:15 | clean |
| ``test(`${a \|\| b}`, () => {});`` | 1:9-1:15 | clean |
| `test('a', () => {}, timeout \|\| 5);` | 1:21-1:33 | clean |
| `it('a', () => {}, opts ? 1 : 2);` | 1:19-1:31 | clean |
| `test.each([a \|\| b])('x', () => {});` | 1:12-1:18 | clean |
| `test.skip.each([a \|\| b])('x', () => {});` | 1:17-1:23 | clean |
| ``test.each`${a \|\| b}`('x', () => {});`` | 1:13-1:19 | clean |
| `test(a?.b, () => {});` with `allowOptionalChaining: false` | 1:6-1:10 | clean |
| `test('case', wrap(() => { if (flag) { act(); } }));` | 1:27-1:47 | clean |

### Callbacks passed by name are resolved file-wide

`analysis.Callbacks().Functions` attributes a module-level function to the test that names it. Upstream has no such notion — the function body is not inside the test call, so it is never reported:

| Code | ESLint | This PR |
| --- | --- | --- |
| `test('a', cb); function cb() { if (x) {} }` | clean | 1:32-1:41 |
| `const cb = () => { if (x) {} }; test('a', cb);` | clean | 1:20-1:29 |
| `test('a', cb); const cb = function () { if (x) {} };` | clean | 1:41-1:50 |
| `test.each([1])('a', cb); function cb() { if (x) {} }` | clean | 1:42-1:51 |
| `const cb = () => { if (x) {} }; test('a', cb); useIt(cb);` | clean | 1:20-1:29 |
| `test('a', cb); let cb = () => { if (x) {} }; cb = other;` | clean | 1:33-1:42 |
| `test('a', cb); function cb() { a?.b; }` with `allowOptionalChaining: false` | clean | 1:32-1:36 |

In the `let cb` case the alias is reassigned to `other` on the next statement and the body is still attributed to the test.

### Nested test calls

`test('a', () => { test('b', () => {}); if (x) {} });` — ESLint is clean (it clears `inTestCase` when the inner call exits), this PR reports 1:40-1:49. Defensible as a fix, but it contradicts the 1:1 upstream claim, so it should be a deliberate and documented divergence rather than a side effect.

### Suggestion

Track the test `CallExpression` itself — `analysis.ParseTestCall` on enter/exit, keeping rstest's own import scoping — instead of the callback-function set. That is what `jest/no-conditional-in-test` in this repo already does, and it matched ESLint on 74 of the 75 cases.

That also makes it worth sharing one implementation under `internal/utils/test_framework/rules/`, the way the preceding no-hooks port does, and carrying this PR's `NonNullExpression` range handling into it: for `a?.b!` ESLint reports 1:19-1:24 and this PR matches, while `jest/no-conditional-in-test` stops at 1:19-1:23.

The extras suite and the rule doc currently lock the above behaviors in as intended, so they need to change with the fix.

internal/plugins/rstest/rules/no_conditional_in_test/no_conditional_in_test_extras_test.go
```diff
@@ -0,0 +1,645 @@
...
+      return 1;
+    }
+    return 2;
+  };
+});
+      `,
+			Errors: []rule_tester.InvalidTestCaseError{{MessageId: "conditionalInTest", Line: 4, Column: 5}},
+		},
+
+		// ---- Dimension 4: named callback ownership via analysis.Callbacks().Functions ----
```
swwind: Both cases in this block are clean under ESLint and under `jest/no-conditional-in-test`: the function body is outside the test call, so upstream never enters it. Locking them in as invalid fixes the false positive in place.

internal/plugins/rstest/rules/no_conditional_in_test/no_conditional_in_test.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Limitations
+
+The rule checks conditionals inside test callbacks reached through supported
+Rstest forms, including global `test` / `it`, imports from `@rstest/core` and
+`@rstest/playwright`, namespace and CommonJS access, `import.meta.rstest`,
+parameterized `.each` / `.for`, and same-file callback aliases such as
+`test('case', callback)` when `callback` is defined in the same file.
+
+Wrapper callbacks are not followed. For example, conditionals inside
```
swwind: This paragraph and the `same-file callback aliases` sentence above document the two divergences from upstream as intended behavior. Upstream reports `test('case', wrap(() => { if (x) {} }))` and does not report a named callback declared outside the call.

internal/plugins/rstest/utils/test_callback_candidate_test.go
elecmonkey: Fixed in 9c8f596d0. The tail of this file had been reordered into something that does not parse. `TestRstestCallAnalysisCallbacksRemainTestOnly` and `TestRstestConcurrentContextBuildsOwnershipLazily` are restored as they were, and `TestRstestCallAnalysisCallbacksUnwrapTypeScriptAssertions` is rewritten. `gofmt -e` is clean and the package's tests compile and pass.

That test then found a fifth unwrap point: `isModuleTopLevelFunction`, which the file-level function index feeds, walked up through `ParenthesizedExpression` but stopped at `AsExpression`, so `const cb = (() => {}) as () => void; test('x', cb)` never resolved. It now walks the same assertion wrappers as the other four.

internal/plugins/rstest/rules/no_conditional_in_test/rule.go
elecmonkey: You are right, and the rule now tracks the registration `CallExpression` on enter and exit via `analysis.ParseTestCall`, as `jest/no-conditional-in-test` does. Both divergence groups are gone: all ten missed cases in the first table report, and all seven false positives in the second are clean, including the reassigned `let cb`. The upstream suite is untouched and still passes.

The one thing kept from the old shape is a depth counter rather than a flag, so nested registrations stay as the improvement you called out — now deliberate rather than incidental, documented in the PR body and the rule doc, with its own entry in the extras suite.

internal/plugins/rstest/rules/no_conditional_in_test/no_conditional_in_test_extras_test.go
elecmonkey: Moved to the invalid list in 9c8f596d0. Every argument of the registration call is in scope now, so this reports at 3:3.

internal/plugins/rstest/rules/no_conditional_in_test/no_conditional_in_test_extras_test.go
elecmonkey: Both moved to the valid list in 9c8f596d0, along with the two named-callback cases in the `(name, cb, timeout)` and `(name, options, cb)` overload blocks. A function declared outside the registration call is no longer attributed to it.

internal/plugins/rstest/rules/no_conditional_in_test/no_conditional_in_test.md
elecmonkey: Rewritten in 9c8f596d0. The Limitations section now says the scope is the registration call and that everything inside it — title, options, timeout, `.each` data, and a callback passed through a wrapper — is checked, that a callback declared outside the call is not, and that an inner registration exiting does not clear the outer test's scope.

elecmonkey:
Thanks for the 75-case comparison — all 19 divergences are fixed in 9c8f596d0, with replies on the individual threads.

The shared implementation is in as well (a95c34f5a): the traversal moved to `internal/utils/test_framework/rules/no_conditional_in_test/`, and both plugin rules are now adapters that only say which `CallExpression` registers a test case. `jest/no-conditional-in-test` picks up the depth counter and the `a?.b!` range along with it, so its nested-registration case now expects both reports and there is a case locking in the range.

internal/plugins/rstest/rules/no_conditional_in_test/rule.go
elecmonkey: Done in a95c34f5a — the traversal is now `internal/utils/test_framework/rules/no_conditional_in_test/`, following the shape `no_conditional_expect` already uses: a `Runtime` whose only member is `IsTestCall`, so each plugin keeps its own registration parsing and import scoping. The file is `rule.go` rather than `no_conditional_in_test.go`, since Go would treat the latter as a test file.

Carrying the `NonNullExpression` range handling over changes two things in `jest/no-conditional-in-test`: `a?.b!` now reports 1:19-1:24, with a case added for it, and the nested-registration case expects both reports instead of one. Its doc gained a Limitations note for the latter.

swwind (approved):
LGTM

fansenze (approved):

PR #1840, opened 2026-08-21
feat(rstest): add prefer-todo rule

elecmonkey:
## Summary

Report a test registration that is in the file but has no implementation — either the callback is missing entirely, or it is an inline empty function body — and rewrite it to `.todo`, so the runner reports it as a todo instead of an empty pass. Rstest accepts `(title, fn, timeout)` and `(title, options, fn?)`, and the rule recognises the options overload, keeping the options object when it rewrites the call.

Registrations resolve through the shared `ParseFnCall`. Reporting and fixing are two separate decisions: a registration the parser has proved to be an Rstest test is reported even when no rewrite delivers a todo, while a call site the parser cannot prove still reaches the test API is not reported at all. Registered in the rstest plugin but left out of the `recommended` preset, matching upstream.

## Credits

Port from [`eslint-plugin-jest`](https://github.com/jest-community/eslint-plugin-jest)'s `prefer-todo`.

- **The Rstest options overload is recognised and preserved.** Jest has no `(title, options, fn?)` form. `test('retry later', { timeout: 1000 }, () => {})` is reported and rewritten to `test.todo('retry later', { timeout: 1000 })`, keeping the options object. Recognition requires the second argument to be written as an object literal: `test('case', options)` and `test('case', options, () => {})` are left alone, because the identifier may itself be the callback.

- **A whole-module CommonJS namespace is not reported.** `const core = require('@rstest/core'); core.test('x')` looks like a resolved Rstest registration, but `const` freezes only the binding — `core.test = custom` reassigns the property, and the call then runs the user's function. Reporting it would be a false positive with a fix that changes what the code calls, so the rule requires the call site to be provably a bare test API and fails closed here. Jest has no equivalent gate.

- **A repeated `.skip` chain is reported without a fix.** `test.skip.skip('x', () => {})` would, with only the first accessor replaced, become `test.todo.skip(...)`, and Rstest's modifier getters let the later write win — the test stays skipped, which is not what the diagnostic promised. Upstream does not parse repeated modifiers at all, so there is no behaviour to match; the diagnostic still fires, because the registration is a real empty test, but it carries no edit.

- **An alias that hides `.skip` is reported without a fix; one that hides anything else is not reported.** A same-file `const` alias of `test` is followed and fixed. `const skipped = test.skip; skipped('x')` resolves to a skipped test with no body, so it is reported, but the `.skip` lives in the alias initializer and there is no accessor at the call site to rewrite. An alias hiding a different modifier or a factory — `.fails`, `.each(...)`, `.for(...)`, `.extend(...)` — is left alone entirely, since the rewrite would have to reason about what the hidden member does to the run mode. Computed access such as `test[skip]('case', () => {})` is left alone for the same reason: the member is not statically known, so it may name a run mode this rule does not report on at all.

- **Interpolated template titles are reported and fixed.** ``test(`case ${name}`)`` still registers an unimplemented Rstest test: the dynamic title does not make a missing or inline-empty callback ambiguous. The rule therefore reports it and rewrites the callee to `.todo`. Upstream accepts only string literals and no-substitution template literals, so it does not report this form.

- **Nested test registrations are reported and fixed.** In `wrap(test('case'))`, the inner `test(...)` call still executes and registers an Rstest test. The shared Rstest parser recognises that registration, so the rule reports it and rewrites the inner call to `test.todo(...)`. Upstream rejects test calls nested inside another call or member chain.

- **Optional registrations are reported without a fix.** `test?.('case')` is reported when the callback is missing or inline-empty, but no edit is attached: inserting `.todo` would move the optional boundary and may change runtime behavior. Upstream reports and autofixes this form.

- **Unsupported extra arguments do not hide an empty callback.** A call such as `test('case', () => {}, 1000, extra)` is outside Rstest's supported overloads, but the callback is still provably inline-empty, so the rule reports it. It carries no fix because dropping or preserving the unsupported trailing arguments would guess at runtime behavior; upstream reports and autofixes this form.

- **Parentheses around an inline empty callback are removed together with the callback.** The fix range covers the whole argument, so `test('case', (() => {}))` becomes `test.todo('case')`, in the plain callback position, the options overload, and alongside a parenthesized timeout. Upstream's fix range ends at the inner function and can leave an unmatched closing parenthesis; this rule keeps the fixed call parseable.

## Related Links

Related #935

## Checklist


internal/plugins/rstest/rules/prefer_todo/prefer_todo.go
```diff
@@ -0,0 +1,756 @@
...
+	return sameFileDecls > 0
+}
+
+func isSafeBareAliasDeclaration(
+	ctx rule.RuleContext,
+	declaration *ast.Node,
+	visited map[*ast.Symbol]bool,
+	depth int,
+) bool {
+	if declaration == nil || declaration.Kind != ast.KindVariableDeclaration {
```
swwind, line 549: `isSafeBareAliasDeclaration` only accepts `ast.KindVariableDeclaration`, so a destructured binding (a `BindingElement`) never qualifies and `isSafeBareIdentifierReference` falls through to the `sameFileDecls > 0` check and fails.

Effect: every `import.meta.rstest` destructure is skipped, including forms that hide no modifier at all.

```js
const { test } = import.meta.rstest; test("case");             // no report
const { test } = import.meta.rstest; test("case", () => {});   // no report
const { it }   = import.meta.rstest; it("case");               // no report
const { test: t } = require("@rstest/core"); t("case");        // t.todo("case")
```

The `require` form works only because it takes the `resolveModuleRequireBinding` path instead. `rstest/no-disabled-tests` reports `Test is missing function argument` on the first line, so the parser resolves the binding; `test.todo(...)` is a safe fix for all of these, and `prefer_todo.md:34` lists `import.meta.rstest` as a supported source.

internal/plugins/rstest/rules/prefer_todo/prefer_todo.go
```diff
@@ -0,0 +1,756 @@
...
+				if len(args) == 0 {
+					return
+				}
+
+				title := ast.SkipParentheses(args[0])
+				if !isStringNode(title) {
+					return
+				}
+
+				callSite, ok := classifyCallSite(ctx, call, parsed)
```
swwind, line 86: A `classifyCallSite` miss skips the report, not just the fix:

```js
const core = require("@rstest/core"); core.test("case");   // no report
const skipped = test.skip; skipped("case");                // no report
```

`rstest/no-disabled-tests` reports `skippedTest` *and* `Test is missing function argument` on the second one, so the plugin already tells the user that test has no body — only this rule stays quiet. The doc comment on lines 41-45 says the rule refuses *fixes* for these shapes, which is not what the code does.

`ReportNodeWithDeferredFixes` accepts a builder that returns `nil`, so the gate can decide the fix and let the report fire unconditionally.

internal/plugins/rstest/rules/prefer_todo/prefer_todo.go
```diff
@@ -0,0 +1,756 @@
...
+	case ast.KindIdentifier:
+		return isSafeBareIdentifierReference(ctx, initializer, visited, depth)
+	case ast.KindPropertyAccessExpression, ast.KindElementAccessExpression:
+		return isSafeBarePropertyReference(ctx, initializer, visited, depth)
+	default:
+		return false
+	}
+}
+
+func isImportMetaRstestAliasSymbol(symbol *ast.Symbol) bool {
```
swwind: `isImportMetaRstestAliasSymbol` accepts any variable declaration initialized to `import.meta.rstest`, skipping both guards that `isSafeRstestNamespaceAliasDeclaration` (line 291) applies to every other alias form: the `const` requirement and the same-source-file check. `containsImportMeta` (line 471) is loose in the same direction — it accepts anything with `import.meta` at the root of a property/element/call chain, rather than `import.meta` itself.

Both relaxations are unreachable, because the parser rejects these earlier:

```js
let api = import.meta.rstest; api = other; api.test("case");  // no report
import.meta.env.rstest.test("case");                          // no report
import.meta.glob("x").rstest.test("case");                    // no report
```

So they contradict the const/same-file rule stated at line 291 without buying any coverage. Matching `import.meta` exactly and routing through the line 291 gate removes the contradiction.

internal/plugins/rstest/rules/prefer_todo/prefer_todo.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Limitations
+
+The rule follows Rstest globals, named imports, ES module namespace imports,
+`import.meta.rstest` and same-file `const` aliases only when the alias chain is
+provably a bare test API reference. It intentionally fails closed for aliases
+that hide modifiers or factories such as `.skip`, `.fails`, `.each(...)`,
+`.for(...)` or `.extend(...)`.
+
+Whole-module CommonJS namespace objects, such as
```
swwind: "left unchanged" reads as "reported but not autofixed"; these forms produce no diagnostic at all. The repeated-`.skip` sentence has the same problem — the stated reason only explains the missing fix. However the report/fix split gets resolved, this section should describe what the user actually sees.

swwind (commented):
Reviewed the rule against `eslint-plugin-jest@29.16.1` and against how the shared rstest parser resolves the same call shapes. The autofixes hold up; every finding below is about a call that is recognized as an unimplemented test but produces no diagnostic.

| --- | --- | --- |
| `const { test } = import.meta.rstest; test("case");` | n/a | no report |
| `const { test: t } = require("@rstest/core"); t("case");` | n/a | `t.todo("case")` |
| `const core = require("@rstest/core"); core.test("case");` | no report | no report |
| `const skipped = test.skip; skipped("case");` | no report | no report |

Row 1 is the one I would call a bug: `rstest/no-disabled-tests` reports `Test is missing function argument` on that exact line, so the parser does resolve the binding — only this rule's own alias resolver does not. Rows 3 and 4 match upstream, but `rstest/no-disabled-tests` recognizes both shapes too, so the silence is an inconsistency inside the plugin rather than a divergence from jest.

One guardrail while reworking the alias handling: don't anchor the `.todo` insertion at `Head.Local.Node` the way `jest/prefer-todo` does. In rstest that node is the chain root, so `core.test("case")` would fix to `core.todo.test("case")`. Inserting after the whole callee is correct as written.

Details inline.

internal/plugins/rstest/rules/prefer_todo/prefer_todo.go
elecmonkey: Confirmed for all three, and fixed. `isSafeBareAliasDeclaration` now takes a `BindingElement` as well, so `const { test } = import.meta.rstest` and its rename and `it` forms report and fix.

The binding has to read the namespace's own `test` or `it`: a rename is matched on the original property name, and a default value, a rest element, or a nested pattern is refused. Walking up through nested patterns would have accepted `const { runner: { test } } = ns`, where the binding names a property of something else entirely.

One neighbouring shape turned out not to resolve at all: `import * as rstest from "@rstest/core"; const { test } = rstest; test("case")` is invisible to `ParseFnCall`, so no rstest rule sees it as a registration. That is a parser gap rather than a gate in this rule, so it is locked in as valid with a note rather than worked around here.

internal/plugins/rstest/rules/prefer_todo/prefer_todo.go
elecmonkey: Right — the doc comment described the intended design and the code did something stricter. `classifyCallSite` now answers "report?" and "is there a fix?" separately, and `buildFixes` returns nil for the second.

`const skipped = test.skip; skipped("case", () => {})` and both repeated-`.skip` forms now report with no edit attached.

Two shapes stay unreported on purpose, and I have separated them from the rest in the wording. `core.test("case")` is one: the mutability argument says the call may not be an Rstest test at all, so a diagnostic there would be a false positive, not merely an unfixable one. `test[skip]("case")` is the other: the member is not statically known, so it may name `.only` or `.each`, run modes this rule does not report on in their static form either — reporting would mean guessing at the run mode.

The doc comment, the rule doc and the PR body now say "not reported" for those two and "reported without a fix" for the rest.

internal/plugins/rstest/rules/prefer_todo/prefer_todo.go
elecmonkey: Both removed. `isImportMetaRstestAliasSymbol` is gone, so an `import.meta.rstest` alias goes through the same const and same-file gate as every other alias form; `isSafeRstestNamespaceAliasDeclaration` already had an `isImportMetaRstest` branch, so `const api = import.meta.rstest; api.test("x")` is unchanged. `containsImportMeta` is replaced by `isImportMeta`, which matches `import.meta` exactly. Your three shapes are locked in as valid.

internal/plugins/rstest/rules/prefer_todo/prefer_todo.md
elecmonkey: Rewritten. Limitations is now split into "Reported without a fix" and "Not reported at all", with each shape under the heading that matches what the user sees, and each reason stated for that outcome rather than for the missing fix. The PR body's differences section follows the same split.

internal/plugins/rstest/rules/prefer_todo/prefer_todo.go
```diff
@@ -0,0 +1,809 @@
...
+func isObjectLiteral(node *ast.Node) bool {
+	return node != nil && node.Kind == ast.KindObjectLiteralExpression
+}
+
+func isStringNode(node *ast.Node) bool {
+	if node == nil {
+		return false
+	}
+	switch node.Kind {
+	case ast.KindStringLiteral, ast.KindNoSubstitutionTemplateLiteral, ast.KindTemplateExpression:
```
swwind, line 841: `isStringNode` accepts `TemplateExpression`, so `test(`case ${name}`)` is reported and autofixed. Upstream `eslint-plugin-jest@29.16.1` accepts that case: its string-node helper only accepts template literals with no expressions. This is an undocumented false positive; restrict this branch to no-substitution templates.

internal/plugins/rstest/rules/prefer_todo/prefer_todo.go
```diff
@@ -0,0 +1,809 @@
...
+				}
+				if parsed.Todo || parsed.IsParameterized() {
+					return
+				}
+
+				call := node.AsCallExpression()
+				if call == nil || call.Arguments == nil {
+					return
+				}
+				if call.QuestionDotToken != nil ||
```
swwind: Optional registrations are discarded before the report/fix split. `test?.("case")` produces `unimplementedTest` in upstream (with `test.todo?.("case")`), while this rule emits nothing. If an optional call has no safe Rstest rewrite, it should at least report without a fix, consistent with the rule’s documented report-without-fix behavior.

internal/plugins/rstest/rules/prefer_todo/prefer_todo.go
```diff
@@ -0,0 +1,809 @@
...
+func classifyCallback(args []*ast.Node) callbackClassification {
+	if len(args) == 1 {
+		return callbackClassification{
+			kind:          callbackMissing,
+			callbackIndex: -1,
+			optionsIndex:  -1,
+			timeoutIndex:  -1,
+		}
+	}
+	if len(args) == 0 || len(args) > 3 {
```
swwind: `classifyCallback` treats every call with more than three arguments as implemented/unknown. Upstream reports `test("case", () => {}, 1, 2)` as `emptyTest`; this implementation is silent. The runner still receives an empty callback, so this is an undocumented false negative. Either preserve the extra arguments in a safe rewrite or report without a fix.

internal/plugins/rstest/rules/prefer_todo/prefer_todo.go
```diff
@@ -0,0 +1,809 @@
...
+// Rstest test at run time, such as a whole-module CommonJS namespace whose
+// properties are writable, is not reported at all.
+var PreferTodoRule = rule.Rule{
+	Name:   "rstest/prefer-todo",
+	Schema: rule.EmptyArraySchema,
+	Run: func(ctx rule.RuleContext, options []any) rule.RuleListeners {
+		analysis := rstestUtils.GetRstestCallAnalysis(ctx)
+		return rule.RuleListeners{
+			ast.KindCallExpression: func(node *ast.Node) {
+				parsed := analysis.ParseFnCall(node)
```
swwind, line 61: `wrap(test("case"))` is reported here, but the pinned upstream rule emits no diagnostic. The differential matrix found this is because `ParseFnCall` accepts the nested call while upstream’s parser does not. If this broader behavior is intended, document it as an Rstest difference; otherwise the rule needs to reject nested registrations for parity.

internal/plugins/rstest/rules/prefer_todo/prefer_todo.go
elecmonkey: Kept this behavior intentionally for Rstest. An interpolated title does not make a missing or inline-empty callback ambiguous, so both forms continue to report and autofix. I documented this as Rstest-specific behavior and added coverage for both callback cases.

internal/plugins/rstest/rules/prefer_todo/prefer_todo.go
elecmonkey: Fixed in 9eac203b. Optional registrations now report when the callback is missing or inline-empty, but carry no autofix because inserting `.todo` would move the optional boundary and may change runtime behavior. Tests cover both forms, including an optional `.skip` chain.

internal/plugins/rstest/rules/prefer_todo/prefer_todo.go
elecmonkey: Fixed in 9eac203b. Calls with unsupported extra arguments now still classify the supported callback positions. When the callback is provably inline-empty, the rule reports `emptyTest` without an autofix rather than guessing how the unsupported trailing arguments should be preserved.

internal/plugins/rstest/rules/prefer_todo/prefer_todo.go
elecmonkey: Kept this broader behavior intentionally for Rstest. The inner `test(...)` still executes and registers a real Rstest test even when its result is passed to another call, so it continues to report and autofix. I documented the divergence from the upstream Jest parser and added coverage for missing and inline-empty callbacks.

internal/plugins/rstest/rules/prefer_todo/prefer_todo.go
```diff
@@ -0,0 +1,823 @@
...
+	var third *ast.Node
+	if len(args) >= 3 {
+		third = skipParentheses(args[2])
+	}
+
+	secondInlineFn, secondEmpty := inlineFunctionState(second)
+	thirdInlineFn, thirdEmpty := inlineFunctionState(third)
+
+	if secondInlineFn {
+		if thirdInlineFn {
```
swwind: An unsupported third argument can still hide a proven-empty callback here. `test("case", () => {}, () => {})` emits `emptyTest` under `eslint-plugin-jest@29.16.1` and fixes to `test.todo("case", () => {})`, while this implementation returns `callbackImplementedOrUnknown` and reports nothing.

This also conflicts with the documented policy that unsupported extra arguments do not hide an empty callback. Since the second argument is certainly the callback, this should report `emptyTest`; because the third argument is not a valid timeout, reporting without a fix would be the conservative behavior already used for other unsupported argument lists.

elecmonkey: Fixed in d419d7929. A provably empty second-argument callback now reports `emptyTest` even when the unsupported third argument is another function. Since that third argument is not a valid Rstest timeout, the diagnostic is report-only and carries no autofix. Added the exact regression case; the targeted Go tests and JS integration tests pass.

internal/plugins/rstest/rules/prefer_todo/prefer_todo.go
```diff
@@ -0,0 +1,829 @@
...
+				return callbackClassification{
+					kind:          callbackInlineEmpty,
+					callbackIndex: 1,
+					optionsIndex:  -1,
+					timeoutIndex:  -1,
+					reportOnly:    true,
+				}
+			}
+			timeoutIndex := -1
+			if len(args) == 3 {
```
swwind, line 732: Any non-function third argument is treated as a valid timeout here and is deleted by the autofix. For example, each of these reports `emptyTest` and fixes to `test.todo("case")`:

```js
test("case", () => {}, "not a timeout");
test("case", () => {}, {});
test("case", () => {}, null);
```

These third arguments are outside Rstest's `(title, fn, timeout)` overload, just like the function-valued third argument handled by the `thirdInlineFn` branch above. Deleting the user's expression is therefore not a safe rewrite and conflicts with the documented policy that unsupported argument lists are report-only. Please classify syntactically unsupported third arguments as `reportOnly` and add regression coverage.

internal/plugins/rstest/rules/prefer_todo/prefer_todo.go
```diff
@@ -0,0 +1,829 @@
...
+			repeatedSkip = true
+			continue
+		}
+		skipEntry = &parsed.MemberEntries[i]
+	}
+	if skipEntry == nil {
+		return callSiteClassification{}, false
+	}
+	receiver, _ := testFramework.AccessorReceiverAndParent(skipEntry)
+	if !isSafeBareCallSiteReference(ctx, receiver) {
```
swwind, line 160: This receiver check drops the diagnostic when `.skip` is present both inside an alias and at the call site:

```js
const skipped = test.skip;
skipped.skip("case", () => {});
```

The shared parser resolves this as an Rstest test with `parsed.Skipped == true` and a call-site `Members == ["skip"]`, but `isSafeBareCallSiteReference` rejects the receiver because its initializer hides `.skip`, so this branch returns false and the rule emits nothing. The documentation says both a hidden `.skip` alias and a repeated `.skip` chain are reported without a fix; this case is both. Please preserve/prove the hidden `.skip` provenance and report this call without an autofix, while continuing to reject aliases that hide modifiers such as `.fails`.

internal/plugins/rstest/rules/prefer_todo/prefer_todo.go
elecmonkey: Fixed in 24e7ab3eb. A syntactically definite non-number third argument now keeps the empty-test diagnostic but suppresses the autofix, so string, object, null, and assertion-wrapped variants are not deleted. Numeric and potentially numeric expressions retain the existing fix. Added Go and JS regression coverage.

internal/plugins/rstest/rules/prefer_todo/prefer_todo.go
elecmonkey: Fixed in 24e7ab3eb. The rule now proves aliases whose initializer chain contains only `.skip` members, so a hidden `.skip` followed by a call-site `.skip` reports without a fix. Aliases containing `.fails`, `.only`, or factories remain rejected; regression coverage includes the reported case and a hidden `.fails` guard.

internal/plugins/rstest/rules/prefer_todo/prefer_todo.go
elecmonkey: Correction: the final commit is c10b86772 (the previous commit was rewritten only to remove an automatically injected attribution trailer).

internal/plugins/rstest/rules/prefer_todo/prefer_todo.go
elecmonkey: Correction: the final commit is c10b86772 (the previous commit was rewritten only to remove an automatically injected attribution trailer).

internal/plugins/rstest/rules/prefer_todo/prefer_todo.go
```diff
@@ -0,0 +1,970 @@
...
+) bool {
+	if identifier == nil || identifier.Kind != ast.KindIdentifier || depth > 16 {
+		return false
+	}
+	if ctx.Refs == nil {
+		text := identifier.AsIdentifier().Text
+		return text == "test" || text == "it"
+	}
+
+	symbol := ctx.Refs.Resolve(identifier)
```
swwind, line 461: The unresolved-reference fallback treats any bare `test`/`it` spelling as an Rstest global, and the later check considers only value declarations. This disagrees with the pinned upstream rule’s TypeScript scope model, where type-space bindings also shadow the Jest global.

For example, `eslint-plugin-jest@29.16.1` reports nothing for all of these:

```ts
type test = string;
test("case");

interface test {}
test("case");

namespace test {}
test("case");

function f<test>() {
  test("case");
}
```

This implementation emits `unimplementedTest` and offers a `.todo` fix for each call. Nested type aliases, interfaces, and namespaces fail in the same way, while ordinary value bindings and `import type` bindings are already handled correctly.

Please account for the lexical/type-scope bindings recognized by typescript-eslint before falling back to the Rstest global, and add focused regressions for the top-level and nested forms.

elecmonkey: Fixed in b10c8cbf9. Before falling back to the injected Rstest global, a bare `test`/`it` is now resolved in every declaration space — value, type, namespace, and alias — so a type alias, interface, namespace, or type parameter of that name shadows it the same way typescript-eslint's scope manager does. Both fallback paths are covered: the unresolved-reference one and the global-mode one that previously consulted only value declarations. Regressions cover the top-level and nested forms for `test` and `it`.

internal/plugins/rstest/rules/prefer_todo/prefer_todo_extras_test.go
```diff
@@ -0,0 +1,482 @@
...
+				Output: []string{`test
+  . /* keep */ todo("case");`},
+				Errors: []rule_tester.InvalidTestCaseError{{MessageId: "emptyTest", Line: 1, Column: 1}},
+			},
+			// ---- parenthesized callee remains parseable.
+			{
+				Code:   `(test)("case");`,
+				Output: []string{`(test).todo("case");`},
+				Errors: []rule_tester.InvalidTestCaseError{{MessageId: "unimplementedTest", Line: 1, Column: 1}},
+			},
```
swwind: Let's lock in the safe wrapper-range behavior for every supported callback position. The pinned upstream rule leaves unmatched closing parentheses for the first two callback cases, while this implementation correctly removes the complete argument.

```suggestion
				// ---- parenthesized callees and removable arguments remain parseable.
				{
					Code:   `(test)("case");`,
					Output: []string{`(test).todo("case");`},
					Errors: []rule_tester.InvalidTestCaseError{{MessageId: "unimplementedTest", Line: 1, Column: 1}},
				},
				{
					Code:   `test("case", (() => {}));`,
					Output: []string{`test.todo("case");`},
					Errors: []rule_tester.InvalidTestCaseError{{MessageId: "emptyTest", Line: 1, Column: 1}},
				},
				{
					Code:   `test("case", { retry: 1 }, (() => {}));`,
					Output: []string{`test.todo("case", { retry: 1 });`},
					Errors: []rule_tester.InvalidTestCaseError{{MessageId: "emptyTest", Line: 1, Column: 1}},
				},
				{
					Code:   `test("case", (() => {}), (1000));`,
					Output: []string{`test.todo("case");`},
					Errors: []rule_tester.InvalidTestCaseError{{MessageId: "emptyTest", Line: 1, Column: 1}},
				},
```

internal/plugins/rstest/rules/prefer_todo/prefer_todo.md
```diff
@@ -0,0 +1,95 @@
...
+### Rstest-specific behavior
+
+The rule intentionally recognizes two registration shapes that the upstream
+Jest parser leaves alone:
+
+- interpolated template titles such as ``test(`case ${name}`)`` are reported,
+  because the dynamic title does not make a missing or empty callback ambiguous;
+- nested registrations such as `wrap(test('case'))` are reported, because the
+  inner `test(...)` call still executes and registers an Rstest test. Rewriting
+  that inner call to `test.todo(...)` preserves its registration role.
```
swwind: This safe fixer difference should be documented alongside the other intentional divergences.

```suggestion
The rule intentionally differs from upstream in these cases:

- interpolated template titles such as ``test(`case ${name}`)`` are reported,
  because the dynamic title does not make a missing or empty callback ambiguous;
- nested registrations such as `wrap(test('case'))` are reported, because the
  inner `test(...)` call still executes and registers an Rstest test. Rewriting
  that inner call to `test.todo(...)` preserves its registration role;
- parentheses around an inline empty callback are removed together with the
  callback. Upstream's fix range ends at the inner function and can leave
  unmatched closing parentheses, while this rule keeps the fixed call parseable.
```

internal/plugins/rstest/rules/prefer_todo/prefer_todo_extras_test.go
elecmonkey: Applied in 0589f6b56. All four cases were already the rule's behavior, so this locks it in: the fix range covers the whole argument including its wrapping parentheses, in the plain callback position, the options overload, and alongside a parenthesized timeout. The parenthesized-callee case moved under the widened comment.

internal/plugins/rstest/rules/prefer_todo/prefer_todo.md
elecmonkey: Applied in 0589f6b56. The section now reads as a list of intentional divergences rather than of recognized registration shapes, and the fixer difference is documented as the third item.

swwind (approved):
LGTM

internal/plugins/rstest/rules/prefer_todo/prefer_todo.md
```diff
@@ -0,0 +1,97 @@
...
+- nested registrations such as `wrap(test('case'))` are reported, because the
+  inner `test(...)` call still executes and registers an Rstest test. Rewriting
+  that inner call to `test.todo(...)` preserves its registration role;
+- parentheses around an inline empty callback are removed together with the
+  callback. Upstream's fix range ends at the inner function and can leave
+  unmatched closing parentheses, while this rule keeps the fixed call parseable.
+
+## Original Documentation
+
+- [eslint-plugin-jest: prefer-todo](https://github.com/jest-community/eslint-plugin-jest/blob/v29.16.1/docs/rules/prefer-todo.md)
```
swwind, line 96: do this need to be cited here?

elecmonkey: i'll rewrite rstest/'s doc in another PR. resolve it here.

swwind (approved):

PR #1841, opened 2026-08-21
feat(rstest): add prefer-each rule

elecmonkey:
## Summary

Report Rstest registrations wrapped in a hand-written `for`, `for...in` or `for...of` loop and point at `.each`, which keeps the cases grouped under one declaration and gives reporters a real title per case. A loop that registers a single test is reported against that test's own API name; a loop that registers several, or any `describe`, or only hooks, is reported against `describe.each`, since the suite-level form is what can carry more than one registration. A loop is judged by what it registers: a loop whose body holds no registration is business logic and is never reported, wherever it sits, while a loop that does register is reported even inside a test callback.

Registrations resolve through the shared `ParseFnCall`, so globals, named and aliased imports, namespace and whole-module `require`, `import.meta.rstest`, same-file aliases and `@rstest/playwright` are all followed. Registered in the rstest plugin but left out of the `recommended` preset, matching upstream.

## Credits

Port from [`eslint-plugin-jest`](https://github.com/jest-community/eslint-plugin-jest)'s `prefer-each`, cross-checked against [`eslint-plugin-vitest`](https://github.com/vitest-dev/eslint-plugin-vitest)'s `prefer-each`.

- **The message names the API the code actually uses.** A loop registering a single `it` is reported as `it.each`, a single `test` as `test.each`. Upstream hardcodes `it.each` for the single-test branch, so a file written with `test` is told to use an API it never mentions. The name comes from the resolved registration, so an alias reports the API it resolves to. The `describe.each` branch is unchanged.

- **`.for` is never recommended.** Rstest also offers `.for`, whose callback receives `(row, context)` rather than the spread row. That is not a mechanical replacement for a hand-written loop, so the rule keeps a single fixed recommendation instead of choosing between two.

- **Each loop is judged from its own registrations.** Upstream keeps one flat list of registrations for the whole file, gated on an `inTestCaseCall` boolean that every `test(...)` sets and every `test(...)` exit clears; both halves leak across scopes. This port records a registration against the innermost loop that lexically contains it and reports a loop from that frame alone. Three shapes come out differently: a loop inside a test callback is named for what it itself registers (`test.each`) rather than for the enclosing `test(...)` that upstream still holds in its list (`describe.each`); a loop inside a function in a non-callback argument position is reported instead of being swallowed by the still-open `test(...)`; and an outer loop keeps its own registration instead of having it discarded when an inner loop is entered.

- **Loop kinds stay at upstream's scope.** `while` / `do...while` loops and array callbacks such as `cases.forEach(row => test(…))` also register repeatedly, but they are outside upstream's subject and stay outside this port's; widening the reported set is a separate decision from porting the rule.

## Related Links

Related #935

## Checklist


internal/plugins/rstest/rules/prefer_each/prefer_each.go
```diff
@@ -0,0 +1,108 @@
...
+	return "describe"
+}
+
+var PreferEachRule = rule.Rule{
+	Name:   "rstest/prefer-each",
+	Schema: rule.EmptyArraySchema,
+	Run: func(ctx rule.RuleContext, options []any) rule.RuleListeners {
+		analysis := rstestUtils.GetRstestCallAnalysis(ctx)
+		callbacks := analysis.Callbacks().Functions
+		// NOTE: Unlike eslint-plugin-jest's single boolean, rstest tracks the
```
swwind: The rationale here does not hold: ESLint does not report the shape it describes. Upstream's `enterForLoop` carries the same guard this `enterLoop` does, so for `test('outer', () => { test('inner', () => {}); for (const row of rows) { consume(row); } })` the loop is entered with `inTestCaseCall` already false, `jestFnCalls` is cleared, and the empty-bodied loop exits with nothing to report. There is no false positive to fix.

What the counter does change is the opposite direction: any loop that registers tests inside a test callback is now silently unreported. It does not need a preceding nested `test` — `test('outer', () => { for (const row of rows) { test(row.name, () => {}); } })` is `describe.each` upstream and nothing here, and the same holds for `async`, `function` and `(done)` callbacks, for `test.each(...)` callbacks, and for a `describe` nested inside a test callback. Because `Callbacks().Functions` also attributes callbacks passed by name, `const cb = () => { for (const row of rows) { test(row.name, () => {}); } }; test('a', cb);` is suppressed too, even though the loop is lexically at file top level.

The divergence is not one-directional either: a function in a non-callback argument position sits at depth 0, so `test('a', () => {}, (function () { for (const row of rows) { beforeEach(() => {}); } return 5; })());` reports `describe.each` here and nothing upstream.

`internal/utils/test_framework/rules/no_conditional_expect` makes the same bool-to-depth swap deliberately, but there the guard runs the other way and the counter widens coverage. Here it narrows it.

None of these shapes are covered by the tests; the one nested-callback case in `prefer_each_extras_test.go` is the one ESLint also does not report.

internal/plugins/rstest/rules/prefer_each/prefer_each.go
```diff
@@ -0,0 +1,108 @@
...
+		},
+	}
+}
+
+type pendingRegistration struct {
+	kind rstestUtils.RstestFnType
+	name string
+}
+
+func recommendFn(pending []pendingRegistration) string {
```
swwind, line 55: The `CallExpression` listener returns early at `testCallbackDepth > 0`, so registrations inside a test callback never reach `pending`, while upstream pushes them unconditionally. That changes the recommendation: `for (const row of rows) { test(row.name, () => { beforeEach(() => {}); }); }` reports `test.each` here and `describe.each` upstream, and the same holds with a nested `test` in place of the hook. The node and range are right, only the message differs — but it differs against the contract stated in this rule's own docs ("Any loop containing `describe`, hooks, or multiple registrations suggests `describe.each`").

internal/plugins/rstest/rules/prefer_each/prefer_each.md
```diff
@@ -0,0 +1,98 @@
...
+the original API family:
+
+- `it(...)` suggests `it.each`
+- `test(...)` suggests `test.each`
+- Playwright test registrations also suggest `test.each`
+
+Any loop containing `describe`, hooks, or multiple registrations suggests
+`describe.each`.
+
+## Differences from ESLint
```
swwind, line 79: Same issue as the NOTE in the source: ESLint reports nothing on this example, so this section documents a difference that does not exist. The differences that do exist — loops registering tests inside a test callback are never reported, and the `describe.each`/`test.each` selection above — are undocumented and untested.

swwind (commented):
Reviewed against `eslint-plugin-jest@29.16.1` `prefer-each`, with 46 snippets run through both real ESLint and this rule. 40 match, including the deliberate `test`/`it` naming adaptation. The state machine itself is faithful — loop enter/exit clearing, nested loops, loop-header registrations, hook and `describe` callbacks, `.each` chains and templates, aliased imports, shadowed bindings, and `while`/`do...while` (correctly ignored) all behave identically.

The remaining six all come from one change: replacing upstream's `inTestCaseCall` boolean with a lexical `testCallbackDepth` counter.

| Code | ESLint | This PR |
| --- | --- | --- |
| `test('o', () => { for (const r of rows) { test(r.name, () => {}); } })` | `describe.each` | no report |
| same, with `async () => {}`, `function () {}` or `(done) => {}` | `describe.each` | no report |
| `const cb = () => { for (const r of rows) { test(r.name, () => {}); } }; test('a', cb);` | `it.each` | no report |
| `test.each(rows)('%s', () => { for (const r of rows) { test(r, () => {}); } })` | `describe.each` | no report |
| `for (const r of rows) { test(r.name, () => { beforeEach(() => {}); }); }` | `describe.each` | `test.each` |
| `test('a', () => {}, (function () { for (const r of rows) { beforeEach(() => {}); } return 5; })());` | no report | `describe.each` |

Details inline.

elecmonkey:
Thanks — all six rows check out, and the NOTE's stated rationale was wrong: upstream's `enterForLoop` clears `jestFnCalls` in that shape, so there was no false positive to fix in the first place.

Rather than restoring `inTestCaseCall`, the rule now gives each loop its own frame: a registration is recorded against the innermost loop that lexically contains it, and a loop is reported from that frame alone. The callback-depth counter is gone, and with it the dependency on `Callbacks().Functions`. A loop that registers nothing has an empty frame and is never reported, so the business-loop-inside-a-test case still stays silent — for the same reason ESLint has, not a special case.

Where the six rows land now:

| Code | ESLint | This PR |
| --- | --- | --- |
| `test('o', () => { for (const r of rows) { test(r.name, () => {}); } })` | `describe.each` | `test.each` |
| same, with `async () => {}`, `function () {}` or `(done) => {}` | `describe.each` | `test.each` |
| `const cb = () => { for (const r of rows) { test(r.name, () => {}); } }; test('a', cb);` | `it.each` | `test.each` |
| `test.each(rows)('%s', () => { for (const r of rows) { test(r, () => {}); } })` | `describe.each` | `test.each` |
| `for (const r of rows) { test(r.name, () => { beforeEach(() => {}); }); }` | `describe.each` | `describe.each` |
| `test('a', () => {}, (function () { for (const r of rows) { beforeEach(() => {}); } return 5; })());` | no report | `describe.each` |

Row 5 now agrees with ESLint and with the contract in the rule's own docs. Rows 1, 2 and 4 report where they previously did not; the name still differs because upstream's flat list holds the enclosing `test(...)` when the inner loop exits, so its `describe.each` counts a registration the loop does not make. Row 3 reports at the same node as ESLint and differs only through the `test`/`it` naming adaptation already noted in the PR body. Row 6 stays divergent in the other direction and I kept it: the loop does register a hook per iteration, and ESLint is silent there only because the enclosing `test(...)` is still open when the loop exits.

One shape outside the table changed from the same cleanup. In

```js
for (const suite of suites) {
  test(suite.name, () => {});
  for (const item of suite.items) { consume(item); }
}
```

ESLint clears the shared list when the inner loop is entered and reports nothing; the outer loop now reports `test.each`. That case was previously locked in as valid and has moved to `invalid`, together with the nested-callback case you flagged.

Coverage: a new `prefer_each_scoping_test.go` (3 valid, 11 invalid) asserts the full message and range for every shape above, and two of them are mirrored in `packages/rslint-test-tools/tests/rstest/rules/prefer-each.test.ts`. The rule doc's `Differences from ESLint` section now describes the three divergences that exist instead of the one that did not.

internal/plugins/rstest/rules/prefer_each/prefer_each.go
```diff
@@ -0,0 +1,100 @@
...
+		//
+		// This rule gives each loop its own frame instead. A registration is
+		// recorded against the innermost loop that lexically contains it, and a
+		// loop is reported from its own frame alone, so what the report says is
+		// exactly what the loop registers. A loop that only runs business logic
+		// has an empty frame and is never reported, whether or not it sits inside
+		// a test callback.
+		frames := make([][]pendingRegistration, 0, 4)
+
+		enterLoop := func(node *ast.Node) {
```
swwind, line 86: `ForInStatement` and `ForOfStatement` frames begin before their iterable expression is traversed, so a registration used to produce the iterable is attributed to the loop body. Upstream leaves both of these silent:

```js
for (const row of getRows(it('one', () => {}))) {}
for (const row in getRows(it('one', () => {}))) {}
```

This rule reports `prefer using `it.each` rather than a manual loop` for each. The `it(...)` call executes once while evaluating the iterable; it is not a repeated registration and cannot be replaced with `.each`.

Please open the loop frame only after the iterable is visited, or otherwise exclude registrations in the `ForIn`/`ForOf` expression. The frame entered at `prefer_each.go:56` currently captures every nested call.

elecmonkey: Fixed in d3dcc2cb. `for...in/of` iterable registrations are now excluded from that loop's frame. When nested, such a registration falls through to an enclosing loop because the iterable is evaluated once per outer iteration.

Added the two reported cases to both Go and JS tests, plus coverage proving that a body registration still determines the recommendation and that a nested iterable registration belongs to the outer loop.

Verified with `eslint-plugin-jest@29.16.1`, the full Rstest Go test suite, and the targeted JS integration test. Thanks for catching the traversal-order difference.

internal/plugins/rstest/rules/prefer_each/prefer_each.go
```diff
@@ -0,0 +1,129 @@
...
+				}
+				switch parsed.Kind {
+				case testFramework.FnKindTest,
+					testFramework.FnKindDescribe,
+					testFramework.FnKindHook:
+					for i := len(frames) - 1; i >= 0; i-- {
+						// The right-hand expression of for-in/of runs once before that
+						// loop starts. It belongs to an enclosing loop, if any, rather
+						// than to the loop whose rows it produces.
+						if isInsideForInOrOfExpression(node, frames[i].node) {
```
swwind: Classic `for` control clauses need the same exclusion applied to `for...in/of` iterable expressions. With `eslint-plugin-jest@29.16.1`, none of these report:

```js
for (let i = register(it('once', () => {})); i < 2; i++) {}
for (let i = 0; check(it('condition', () => {}), i < 2); i++) {}
for (let i = 0; i < 2; step(it('update', () => {}), i++)) {}
```

Rslint reports `it.each` for all three. In particular, the initializer executes only once, so that diagnostic is a false positive and cannot be replaced by `.each`.

The loop frame opens before all `ForStatement` children are visited, while `isInsideForInOrOfExpression` always returns false for this loop kind, so registrations in the initializer, condition, and incrementor are attributed to the loop. Please restrict classic-for attribution to its body, or explicitly exclude all three control clauses, and add regression coverage.

elecmonkey: Fixed in 83e5759e9. Registrations are now attributed only when they are inside the loop body, so classic `for` initializer, condition, and update expressions are excluded, as are `for...in/of` iterable expressions. When nested inside another loop body, such registrations still fall through to the enclosing loop frame.

Added regression coverage for all three reported classic `for` control clauses in both Go and JS tests, plus a nested initializer case.

Verified with:
- `go test ./internal/plugins/rstest/rules/prefer_each`
- `go test ./internal/plugins/rstest/...`
- `pnpm run build`
- targeted Rstest `prefer-each` integration test

internal/plugins/rstest/rules/prefer_each/prefer_each.go
```diff
@@ -0,0 +1,138 @@
...
+		// NOTE: eslint-plugin-jest keeps one flat list of registrations for the
+		// whole file and decides what to do with it from an `inTestCaseCall`
+		// boolean that is set by every `test(...)` call and cleared by every
+		// `test(...)` exit. Both halves of that leak across scopes: registrations
+		// made before a loop can survive into the loop's report, a nested test
+		// clears the flag while the outer test is still open, and a loop that
+		// runs while the flag happens to be set is skipped entirely.
+		//
+		// This rule gives each loop its own frame instead. A registration is
+		// recorded against the innermost loop that lexically contains it, and a
```
swwind: The “innermost loop that lexically contains it” description is not accurate for registrations in a nested loop header. For example:

```js
for (const suite of suites) {
  for (const row of getRows(test(suite.name, () => {}))) {}
}
```

This PR attributes the registration to the outer loop and reports `test.each`, while `eslint-plugin-jest@29.16.1` reports nothing. The same applies to registrations in a nested classic `for` initializer, condition, or update expression. This behavior is intentional and covered by the new scoping tests, but it is a fourth observable upstream difference beyond the three currently listed in the documentation. Please reword this comment to say “the innermost loop whose body contains it” and document the nested-loop-header difference in `prefer_each.md`.

elecmonkey: Fixed in e13c3bbaf. The frame comment now reads "the innermost loop whose body contains it", and `prefer_each.md` documents the nested-loop-header case as a fourth difference from upstream, with the `for...of` example you gave and a note that a nested classic `for` initializer, condition, or update expression behaves the same way. No behavior change; the existing scoping tests already cover both shapes.

swwind (approved):
LGTM

PR #1868, opened 2026-08-24
feat: port rule sort-imports

swwind:
## Summary

Port the `sort-imports` rule from ESLint to rslint.

Enforce consistent ordering of import declarations and named import members, including ESLint-compatible options and autofixes.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/sort-imports
- Source code: https://github.com/eslint/eslint/blob/v10.9.0/lib/rules/sort-imports.js

## Checklist


internal/rules/sort_imports/sort_imports.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+	return members, namedNode
+}
+
+func buildMemberFix(ctx rule.RuleContext, members []importMember, namedNode *ast.Node, ignoreCase bool) []rule.RuleFix {
+	if len(members) < 2 || namedNode == nil {
+		return nil
+	}
+	namedRange := utils.TrimNodeTextRange(ctx.SourceFile, namedNode)
+	if utils.HasCommentInSpan(ctx.Comments.All(), namedRange.Pos(), namedRange.End()) {
```
elecmonkey: [P2] Preserve fixability for comments inside a specifier

This rejects every comment anywhere inside the named-import braces, which is broader than ESLint’s per-specifier `getCommentsBefore`/`getCommentsAfter` check. For `import { b /* keep */ as c, a } from "x";`, ESLint 10.9.0 reports the diagnostic with a fix to `import { a, b /* keep */ as c } from "x";`, while Rslint reports the same diagnostic without any fix; the same mismatch occurs for comments after `as`, string-named imports, and type-only specifiers. Narrow the guard to comments attached before or after each specifier instead of suppressing fixes for comments contained within a specifier.

swwind: Fixed in 414f6e42. `buildMemberFix` now checks only the gaps before, between, and after import specifiers, so comments inside a specifier move with its complete source text. I added focused regression coverage for comments before `as`, after `as`, string-named imports, and type-only specifiers; the existing boundary-comment cases continue to suppress fixes. `go test ./internal/rules/sort_imports` passes.

elecmonkey (approved):
excellent

PR #1870, opened 2026-08-24
feat: port rule unicorn/no-this-assignment

rockm3:
## Summary

Port the `unicorn/no-this-assignment` rule from ESLint to rslint.

Disallow assigning `this` directly to an identifier, including variable declarations and assignment expressions.

## Related Links

- ESLint rule: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v73.0.0/docs/rules/no-this-assignment.md
- Source code: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v73.0.0/rules/no-this-assignment.js

## Checklist


internal/plugins/unicorn/rules/no_this_assignment/no_this_assignment.go
```diff
@@ -0,0 +1,55 @@
...
+			ast.KindVariableDeclaration: func(node *ast.Node) {
+				declaration := node.AsVariableDeclaration()
+				checkAssignment(node, declaration.Name(), declaration.Initializer)
+			},
+			ast.KindBinaryExpression: func(node *ast.Node) {
+				binary := node.AsBinaryExpression()
+				if binary.OperatorToken == nil || !ast.IsAssignmentOperator(binary.OperatorToken.Kind) {
+					return
+				}
+
```
elecmonkey, line 54: When a default appears inside an assignment pattern, such as `[foo = this] = array`, `({foo: bar = this} = object)`, or `for ([foo = this] of array) {}`, ts-go exposes the default as a `KindBinaryExpression` with `EqualsToken`, so this listener reports it. ESTree exposes the same node as `AssignmentPattern`, while upstream only listens to `AssignmentExpression`, so eslint-plugin-unicorn@73.0.0 reports nothing. Exclude binary expressions that are defaults within destructuring-assignment targets before calling `checkAssignment`.

rockm3: Fixed in f377982. The binary-expression listener now skips destructuring defaults via `utils.IsDefaultValueInDestructuringAssignment`, matching the upstream ESTree `AssignmentPattern` behavior. I also added regression cases for array, object, `for-of`, and nested destructuring defaults. Thanks for catching this!

internal/plugins/unicorn/rules/no_this_assignment/no_this_assignment.go
```diff
@@ -0,0 +1,59 @@
...
+			ast.KindVariableDeclaration: func(node *ast.Node) {
+				declaration := node.AsVariableDeclaration()
+				checkAssignment(node, declaration.Name(), declaration.Initializer)
+			},
+			ast.KindBinaryExpression: func(node *ast.Node) {
+				binary := node.AsBinaryExpression()
+				if binary.OperatorToken == nil || !ast.IsAssignmentOperator(binary.OperatorToken.Kind) {
+					return
+				}
+				if utils.IsDefaultValueInDestructuringAssignment(node) {
```
swwind, line 51: **[P2] Skip defaults in `for-in` destructuring targets**

`utils.IsDefaultValueInDestructuringAssignment` does not recognize a direct `ForInStatement` initializer as a destructuring-assignment pattern. As a result, `for ({key: self = this} in source) {}` reports `self = this`, while eslint-plugin-unicorn@73.0.0 emits no rule diagnostic because ESTree represents it as an `AssignmentPattern`. Array and nested forms have the same false positive. Please include direct `for-in` destructuring targets in this check and add regression cases alongside the existing `for-of` cases.

rockm3: Fixed in 1f237a5. The shared default-value helper now falls back to `IsInDestructuringAssignment`, which covers direct `for-in` targets while preserving the existing assignment and `for-of` handling. I added object, array, and nested `for-in` regression cases, and verified the shared utils package, the rule Go tests, Go lint, the rebuilt binary, and the JS parity tests. Thanks!

elecmonkey (dismissed):

internal/utils/write_reference.go
```diff
@@ -270,5 +270,6 @@ func IsDefaultValueInDestructuringAssignment(node *ast.Node) bool {
 }
 
 func isArrayOrObjectDestructuringAssignmentPattern(node *ast.Node) bool {
-	return node != nil && ast.IsArrayLiteralOrObjectLiteralDestructuringPattern(node)
+	return node != nil &&
+		(ast.IsArrayLiteralOrObjectLiteralDestructuringPattern(node) || IsInDestructuringAssignment(node))
```
swwind: **[P2] Stop destructuring walks at ordinary expression boundaries**

The `IsInDestructuringAssignment` fallback treats ordinary object and array expressions nested inside a member assignment target as part of the surrounding destructuring pattern. For example, with `({a: ({self: self = this}).x} = source);`, eslint-plugin-unicorn@73.0.0 reports `self = this` once, while rslint reports nothing because `IsDefaultValueInDestructuringAssignment` incorrectly suppresses the real assignment.

This also regresses the existing `block-scoped-var` rule. In:

```js
function f() {
  if (condition) {
    var self;
  }
  ({a: ({self: self = 1}).x} = source);
}
```

ESLint reports the out-of-scope reference once, while rslint reports the same diagnostic twice because `ordinaryReferenceMultiplicity` mistakes the assignment for an `AssignmentPattern`.

The mismatch reproduces through object and array expressions, calls, conditionals, sequences, assignment targets, and `for-in`/`for-of`/`for-await` headers. Please preserve structural expression boundaries when identifying destructuring defaults and add regression tests for both `no-this-assignment` and `block-scoped-var`.

rockm3: Fixed in bdf5be5. I replaced the broad IsInDestructuringAssignment fallback with a structural walk matching the ts-go destructuring-pattern helper, extended only at the root for direct for-in initializers. It now stops at ordinary expression boundaries while retaining nested array/object patterns. I added regression coverage for member, call, and conditional boundaries in no-this-assignment, plus the reference-multiplicity case in block-scoped-var. The shared utils package, all direct consumer packages, Go lint, rebuilt binary, and both JS rule suites pass. Thanks for the detailed cases!

swwind (approved):
LGTM

PR #1871, opened 2026-08-24
perf(rstest): optimize prefer-called-exactly-once-with rule

elecmonkey:
## Summary

- Add a file-level prescan that skips shared Rstest call analysis and block traversal unless the source contains both matcher roles required for a merge.
- Use the source file's identifier table for O(1) checks covering dot access, computed string/template access, and escaped property names for both Jest and Chai matcher spellings.
- Preserve parenthesized computed keys such as `expect(spy)[("calledOnce")]` with a narrow bracket scan and exact AST fallback.
- Add focused tests covering mixed matcher styles, computed and escaped keys, trivia, Unicode line separators, conservative unknown-source handling, and rejected lookalikes.

## Performance

Measured with `--timing all` while running `rstest/prefer-called-exactly-once-with` over all `*.test.*`, `*.spec.*`, and configured in-source test files in six Rstack repositories. Each result is the median rule time from 11 timing runs after two warmups, with baseline and candidate execution order alternated between rounds. Both variants were built from `1436d521`; the benchmark covers 1,978 files.

The complete JSONL diagnostics were identical before and after the optimization across all six repositories, including all three diagnostics produced by this rule.

| Rspack | 146 | 0.8ms | 0.1ms | 8.00× |
| Rsbuild | 629 | 6.9ms | 0.4ms | 17.25× |
| Rslib | 114 | 3.3ms | 0.1ms | 33.00× |
| Rspress | 134 | 3.6ms | 0.1ms | 36.00× |
| Rsdoctor | 89 | 2.7ms | 0.1ms | 27.00× |
| Rstest | 866 | 22.1ms | 1.0ms | 22.10× |
| **Combined** | **1,978** | **39.4ms** | **1.8ms** | **21.89×** |

A second benchmark with the complete 19-rule Rstest suite corroborated the result: summed target-rule timing across seven runs decreased from 148.9ms to 7.6ms, a 19.59× speedup. All 1,987 diagnostics from the complete suite were identical.

The identifier-table gate keeps the common negative path to four map lookups. Files containing a bracket that opens on a parenthesis take the narrow fallback so accepted computed matcher forms are not lost.

## Checklist

### DIFF
--- a/internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
+++ b/internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
@@ -46,16 +46,6 @@ const (
 	mergeMatcherRolePair = mergeMatcherRoleOnce | mergeMatcherRoleWith
 )
 
-var mergeMatcherNames = [...]struct {
-	name string
-	role mergeMatcherRoles
-}{
-	{name: "toHaveBeenCalledOnce", role: mergeMatcherRoleOnce},
-	{name: "calledOnce", role: mergeMatcherRoleOnce},
-	{name: "toHaveBeenCalledWith", role: mergeMatcherRoleWith},
-	{name: "calledWith", role: mergeMatcherRoleWith},
-}
-
 // sourceMayContainMergePair cheaply rejects files that cannot contain both
 // halves of a merge. The parser interns property names and the cooked text of
 // string/template element-access keys, so the identifier table covers normal,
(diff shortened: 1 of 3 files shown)


internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
```diff
@@ -37,6 +38,110 @@ var onceMatchers = map[string]bool{
...
 
+type mergeMatcherRoles uint8
+
+const (
+	mergeMatcherRoleOnce mergeMatcherRoles = 1 << iota
+	mergeMatcherRoleWith
+	mergeMatcherRolePair = mergeMatcherRoleOnce | mergeMatcherRoleWith
+)
+
+var mergeMatcherNames = [...]struct {
```
swwind: Could we derive these names from `onceMatchers` and `combinedMatchers` above instead of restating all four here? This pre-scan is correctness-critical: if a matcher is added to either canonical map but omitted here, the rule will silently return no listeners for files using it. Iterating the two existing maps would retain the same four identifier lookups while preventing the gate from drifting.

elecmonkey: Good point. Fixed in 57f82737d: the source pre-scan now iterates the keys of `onceMatchers` and `combinedMatchers`, and the parenthesized-element fallback derives roles from the same canonical maps. This keeps the same maximum of four identifier lookups without allowing the gate to drift when matchers are added. I also fixed the cspell failure in the escaped-key test fixtures. The focused Go test and the full `pnpm check-spell` check both pass. Thanks!

swwind (approved):
LGTM

PR #1872, opened 2026-08-24
feat: port rule sort-vars

swwind:
## Summary

Port the `sort-vars` rule from ESLint to rslint.

Require identifier variables within the same declaration block to be sorted alphabetically, with ESLint-compatible case-insensitive ordering and safe autofixes.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/sort-vars
- Source code: https://github.com/eslint/eslint/blob/v10.9.0/lib/rules/sort-vars.js

## Checklist


elecmonkey (approved):
The implementation matches ESLint 10.9.0 across the upstream suite and targeted differential cases covering declarations, destructuring, comments, TypeScript syntax, Unicode ordering, options, diagnostics, and autofixes. The affected Go tests also pass, and no actionable regression was identified.

PR #1874, opened 2026-08-24
perf(rstest): optimize max-expects rule

elecmonkey:
## Summary

- Replace per-node function and call listener bookkeeping maps with node ownership recorded directly on the existing count frames, so enter and exit pair through the traversal's own nesting.
- Each frame stores the node whose exit tears it down; a fallback frame activated by position also stores the function that activated it. An ordinary call, or a function that changes no state, does no bookkeeping at all.
- Keep frame activation, callback fallback, assertion counting, diagnostic order, messages, options, and schema unchanged.
- Add focused tests covering nested owner pairing, nodes that pushed nothing, the sentinel frame, and activation followed by reactivation.

## Correctness

The same 19 configured Rstest rules were run over all test, spec, and configured in-source test files in six Rstack repositories, against `origin/main@63cd63c0a` and against this branch, covering 1,978 files.

Complete JSONL diagnostics were identical before and after, on stdout and stderr, so files that fail to parse are covered too:

| Configuration | Diagnostics | Result |
| --- | ---: | --- |
| All 19 Rstest rules | 1,987 | byte-for-byte identical |
| `max-expects` alone | 809 | byte-for-byte identical |

This rule emits no fixes or suggestions, and both runs confirm none were produced.

## Performance

Repository-level `--timing` cannot resolve an effect of this size on the hardware available here. Across seven rounds, a single unchanged binary ranged from 94 ms to 149 ms on the largest repository, so round-to-round drift is an order of magnitude larger than the change being measured, and running the two variants in separate blocks reproduces whichever sign the drift happens to take.

The measurement that does resolve is an isolated benchmark that lints a generated corpus with only this rule enabled, over three shapes: many ordinary calls with few registrations, registrations whose callback the parser resolves precisely, and registrations behind an unresolved wrapper. The two variants alternate within one session, five rounds of three samples each with the execution order swapped every round, compared with `benchstat` (n=15):

| Shape | `sec/op` vs `main` | `allocs/op` vs `main` |
| --- | --- | --- |
| Ordinary calls | -2.70% (p=0.009) | -4.48% (p=0.000) |
| Direct callbacks | -2.15% (p=0.000) | -0.37% (p=0.000) |
| Wrapped callbacks | -1.72% (p=0.007) | -0.40% (p=0.000) |

Bytes allocated fall by 1.47% on the ordinary-call shape and about 0.3% on the other two. These figures cover a full lint run including shared Rstest call analysis, which dominates the measurement, so the reduction attributable to the rule's own listener work is larger than the totals suggest.

The temporary benchmark probe is not included in the production diff.

### DIFF
--- a/internal/plugins/rstest/rules/max_expects/max_expects.go
+++ b/internal/plugins/rstest/rules/max_expects/max_expects.go
@@ -51,6 +51,9 @@ const (
 )
 
 const (
+	// Listener events follow AST nesting, so ordinary files only use a few
+	// slots. Keep that path allocation-free while retaining a sound overflow
+	// path for generated or adversarially deep expressions.
 	inlineFunctionEventCapacity = 8
 	inlineCallEventCapacity     = 16
 )
--- a/internal/plugins/rstest/rules/max_expects/max_expects.go
+++ b/internal/plugins/rstest/rules/max_expects/max_expects.go
@@ -29,11 +29,20 @@ const (
 	frameRegistrationFallback
 )
 
+// owner is the node whose exit tears the frame down: the registration call for
+// a fallback frame, the function itself for a callback or detached frame. Node
+// identity is stable across the traversal and the traversal is strictly nested,
+// so matching the owner on exit is enough to pair enter with exit — no separate
+// record of what each enter did is needed.
+//
+// activationOwner is the function that positionally activated a fallback frame;
+// its exit deactivates the frame again without popping it.
 type countFrame struct {
 	kind                   frameKind
 	count                  int
 	active                 bool
-	registration           *ast.Node
+	owner                  *ast.Node
+	activationOwner        *ast.Node
 	callbackCandidateRoots []*ast.Node
 }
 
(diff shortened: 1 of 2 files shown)


swwind (commented):
Non-blocking optimization idea: I think both event stacks may be removable by recording node ownership directly on the existing `countFrame`.

The registration frame already stores its owning call node, so the call exit listener could pop only when `stack.top().registration == node`. Similarly, pushed function frames could store their owning function node, while fallback activation could store an `activationOwner`; the function exit listener would then either pop the matching owned frame or deactivate the matching fallback frame.

This would avoid `push(false)` / `pop()` for every ordinary call and avoid recording `functionPushNone` for functions that do not change state. It would also remove the inline capacities, overflow slices, and their deeply nested allocation path.

The traversal is strictly nested, and the previous implementation already relied on stable `*ast.Node` identity, so this seems compatible with the existing assumptions. It would still need the same differential matrix and benchmarks—the current stack implementation may already be faster in practice—but this ownership-based version could be worth comparing with `benchstat`, especially on files containing many ordinary calls and few registrations.

Would you like to try this approach and compare the benchmark?

elecmonkey:
Tried it, and the ownership version is now pushed as 25b90d613.

Both event stacks are gone. Each `countFrame` records the node whose exit tears it down — the registration call for a fallback frame, the function itself for a callback or detached frame — and a fallback frame activated by position additionally records the activating function in `activationOwner`. The call exit listener pops only when `stack.top().owner == node`; the function exit listener pops on the same comparison, otherwise deactivates when `activationOwner` matches, otherwise does nothing. That removed the inline capacities, the overflow slices and their deeply nested allocation path, about 90 lines in total.

Two things worth noting on the correctness side. The sentinel frame's owner is nil and an exiting node is never nil, so the previous underflow guards became structurally unnecessary rather than merely unused. And clearing `activationOwner` on deactivation preserves the existing behaviour where a later sibling function in the same candidate root can activate the frame again.

Differential against `origin/main@63cd63c0a` over 1,978 files in six repositories is byte-for-byte identical, on stdout and stderr, both with all 19 Rstest rules (1,987 diagnostics) and with `max-expects` alone (809 diagnostics).

On the benchmark, your instinct that the stack implementation might already be competitive was worth checking carefully, because the first comparison I ran said exactly that. It was an artifact: the two sample sets were collected about thirty minutes apart, and drift on this machine is larger than the effect. Alternating the variants within one session, five rounds of three samples with the order swapped each round, `benchstat` gives (n=15):

| Shape | sec/op vs main | allocs/op vs main |
| --- | --- | --- |
| Ordinary calls | -2.70% (p=0.009) | -4.48% (p=0.000) |
| Direct callbacks | -2.15% (p=0.000) | -0.37% (p=0.000) |
| Wrapped callbacks | -1.72% (p=0.007) | -0.40% (p=0.000) |

The same drift invalidates the repository-level timing table the description previously carried: a single unchanged binary ranges from 94 ms to 149 ms across rounds on the largest repository, which is an order of magnitude more than the effect. I have rewritten the description around the interleaved benchmark and dropped the numbers that do not reproduce.

So the honest summary is that this is a small, real improvement, largest on exactly the shape you predicted — many ordinary calls, few registrations — and the main win is that a whole layer of bookkeeping no longer exists. Thanks for pushing on it.

swwind (commented):
cspell. nothing from here now.

swwind (approved):

PR #1875, opened 2026-08-24
feat: port rule class-methods-use-this

swwind:
## Summary

Port the `class-methods-use-this` rule from ESLint to rslint.

Enforce that class instance methods use `this` or `super`, including class fields and TypeScript-specific override and implements options.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/class-methods-use-this
- Source code: https://github.com/eslint/eslint/blob/v10.9.0/lib/rules/class-methods-use-this.js

## Checklist


internal/rules/class_methods_use_this/class_methods_use_this.go
```diff
@@ -0,0 +1,424 @@
...
+	enterClassLikeMember := func(node *ast.Node) {
+		if name := ast.GetNameOfDeclaration(node); name != nil && name.Kind == ast.KindComputedPropertyName {
+			// Defer push to ComputedPropertyName:exit.
+			return
+		}
+		if node.Body() == nil {
+			pushAnonymous()
+			return
+		}
+		pushMember(node)
```
elecmonkey: Keep decorators outside the method frame

When a decorated method's decorator references `this` or `super`, this frame is pushed before ts-go traverses the decorators. For `class C { @dec(this) foo() {} }`, ESLint 10.9.0 still reports `foo` because decorators are visited before the method's `FunctionExpression`, whereas this implementation marks `foo` as using `this` and suppresses the diagnostic. The member context needs to begin only after its decorators have been traversed.

internal/rules/class_methods_use_this/class_methods_use_this.go
```diff
@@ -0,0 +1,424 @@
...
+					pushMember(parent)
+				}
+			}
+		},
+
+		// Static blocks have their own `this`; isolate them.
+		ast.KindClassStaticBlockDeclaration:                      func(*ast.Node) { pushAnonymous() },
+		rule.ListenerOnExit(ast.KindClassStaticBlockDeclaration): func(*ast.Node) { popContext() },
+
+		ast.KindThisKeyword:  markUsesThis,
```
elecmonkey, line 500: Count `this` references inside TypeScript type queries

For TypeScript inputs such as `class C { foo(x: typeof this) {} }`, `@typescript-eslint/parser` exposes the operand as a `ThisExpression`, so ESLint 10.9.0 considers the method to use `this` and emits no diagnostic. ts-go represents this as `KindThisType`, but the rule only listens for `KindThisKeyword`, causing Rslint to report `foo`; the same mismatch also occurs in return types and generic constraints containing `typeof this`.

internal/rules/class_methods_use_this/class_methods_use_this.go
swwind: Fixed in 96b5296f3. Class member and field frames are now deferred until the final decorator has been traversed, so `this` and `super` inside decorators remain in the enclosing context. The fix covers methods, getters, setters, nested decorated fields, and decorator-inclusive field diagnostic ranges, with focused regression tests for each boundary.

internal/rules/class_methods_use_this/class_methods_use_this.go
swwind: Fixed in 96b5296f3. The rule now recognizes both ts-go representations of `this` directly under a `TypeQuery`, while deliberately excluding a plain `this` type. Regression coverage includes parameter types, return types, generic constraints, nested type arguments, class-field arrow parameters, and the adjacent plain-`this` negative case.

internal/rules/class_methods_use_this/class_methods_use_this.go
```diff
@@ -0,0 +1,474 @@
...
+		exitFunction(node)
+	}
+
+	markUsesThis := func(*ast.Node) {
+		if stack != nil {
+			stack.usesThis = true
+		}
+	}
+
+	markTypeQueryThis := func(node *ast.Node) {
```
elecmonkey, line 386: This is fixable in Rslint rather than a TypeScript limitation. ts-go does preserve the `typeof this.value` shape; the current implementation just checks the direct parent of `this`, so it only handles `typeof this` and misses qualified type-query operands such as `typeof this.value` or `typeof this.value.deep`.

ESLint 10.9.0 with `@typescript-eslint/parser` treats `this` inside those qualified type queries as a `this` use, so `class C { method(value: typeof this.value) {} }` should not report. Rslint currently reports `missingThis` for that case. The rule can match upstream by walking from the `this` node through the property-access / qualified-name chain and checking whether the enclosing node is `KindTypeQuery` (similar to TypeScript's own `IsThisInTypeQuery` logic), instead of requiring `node.Parent.Kind == KindTypeQuery`.

elecmonkey:
Please also resolve conflict

internal/rules/class_methods_use_this/class_methods_use_this.go
swwind: Fixed in ede122b6a. The rule now follows `this` through the left/receiver side of ts-go qualified-name and property-access chains to the enclosing `TypeQuery`, so `typeof this.value` and `typeof this.value.deep` match ESLint. Focused regressions cover both qualified shapes, and the full `class_methods_use_this` package test passes.

internal/rules/class_methods_use_this/class_methods_use_this.go
```diff
@@ -0,0 +1,487 @@
...
+	// The computed-key deferral also applies to *bodyless* members:
+	// `abstract [this.foo](): void` must let `this` in the computed key flow
+	// to the enclosing scope, not be eaten by the bodyless anonymous frame.
+	// The matching ComputedPropertyName:exit branch handles bodyless and
+	// bodied identically — pushing anonymous when Body() == nil, member when
+	// non-nil — keeping the stack balanced against the unconditional pop on
+	// the member's exit listener.
+	pushClassLikeMember := func(node *ast.Node) {
+		if node.Body() == nil {
+			pushAnonymous()
```
elecmonkey: [P2] Do not isolate bodyless TypeScript signatures

For `class Outer { outer() { return class Inner { method(x: typeof this.x): void; }; } }`, ESLint 10.9.0 with `@typescript-eslint/parser` 8.65.0 emits no diagnostic, while this implementation reports `outer`. Upstream represents the signature as `TSEmptyBodyFunctionExpression`, which never enters the rule's `FunctionExpression` listener, so its `this` reference reaches the outer frame; pushing an anonymous frame here swallows that reference. The same mismatch affects abstract/static/get signatures and bodyless function declarations, so bodyless nodes should not push or pop a frame.

swwind: Fixed in 475ef848. Bodyless TypeScript methods, accessors, and function declarations now leave the enclosing frame active instead of pushing an anonymous frame, matching the absence of upstream function listeners. Focused regressions cover method, static method, abstract getter, and function declaration signatures containing `typeof this.x`.

internal/rules/class_methods_use_this/class_methods_use_this.go
```diff
@@ -0,0 +1,494 @@
...
+
+	// exitFunction pops the current stack frame and, if it represents a
+	// reportable class member that did not use `this`/`super`, emits the
+	// diagnostic. Mirrors upstream's `exitFunction`.
+	exitFunction := func(node *ast.Node) {
+		frame := popContext()
+		if frame == nil || frame.member == nil || frame.usesThis {
+			return
+		}
+		if opts.ignoreOverrideMethods && frame.member.ModifierFlags()&ast.ModifierFlagsOverride != 0 {
```
elecmonkey: [P2] Ignore JSDoc-synthesized override flags

When `ignoreOverrideMethods` is enabled for a source-only `.js` file, ts-go folds `/** @override */` into `ModifierFlags()`, although ESLint treats it only as a comment. For `class Example { /** @override */ method() {} }`, ESLint 10.9.0 reports `missingThis`, while Rslint emits nothing; the result also changes when type checking is enabled. Please use `utils.ESTreeModifierFlags(frame.member)` here so only source-authored `override` modifiers trigger this exemption.

internal/rules/class_methods_use_this/class_methods_use_this.go
```diff
@@ -0,0 +1,494 @@
...
+		}
+		for _, clause := range hc.Nodes {
+			if clause == nil {
+				continue
+			}
+			hcNode := clause.AsHeritageClause()
+			if hcNode == nil {
+				continue
+			}
+			if hcNode.Token == ast.KindImplementsKeyword && hcNode.Types != nil && len(hcNode.Types.Nodes) > 0 {
```
elecmonkey: [P2] Exclude JSDoc-synthesized implements clauses

For `/** @implements {Contract} */ class Example { method() {} }` with `ignoreClassesWithImplements: "all"`, ESLint 10.9.0 reports `missingThis`, while Rslint emits nothing because ts-go materializes the JSDoc tag as a heritage clause. `"public-fields"` is affected too. Please filter out `utils.IsJSDocSyntaxNode` clauses and heritage types so only source-authored `implements` clauses trigger this exemption.

internal/rules/class_methods_use_this/class_methods_use_this.go
swwind: Fixed in 38e8f6610, with additional coverage in b28516d8. `ignoreOverrideMethods` now reads `utils.ESTreeModifierFlags(frame.member)`, so JSDoc-synthesized `@override` metadata no longer suppresses `missingThis`; source-authored TypeScript `override` remains exempt. Regression coverage includes checked and unchecked JavaScript, methods, fields, and getters.

internal/rules/class_methods_use_this/class_methods_use_this.go
swwind: Fixed in 38e8f6610, with additional coverage in b28516d8. `classImplementsInterface` now excludes both JSDoc-synthesized heritage clauses and JSDoc heritage types, so `@implements` no longer activates either the `"all"` or `"public-fields"` exemption. Regression coverage includes checked and unchecked JavaScript and class expressions.

internal/rules/class_methods_use_this/class_methods_use_this.go
```diff
@@ -0,0 +1,511 @@
...
+	// parens, so we recover the same shape via `ast.WalkUpParenthesizedExpressions`.
+	classFieldOfFunctionLike := func(node *ast.Node) *ast.Node {
+		parent := node.Parent
+		if parent == nil {
+			return nil
+		}
+		// WalkUpParenthesizedExpressions advances past ParenthesizedExpression
+		// ancestors; passes through unchanged when `parent` is not itself a
+		// paren (covering the common, non-wrapped case).
+		parent = ast.WalkUpParenthesizedExpressions(parent)
```
elecmonkey: [P2] Strip JSDoc cast wrappers around class-field functions

For JavaScript such as `class A { field = /** @type {() => void} */ (() => {}); }`, ESLint 10.9.0 reports `missingThis`, while Rslint emits no `class-methods-use-this` diagnostic; the same mismatch occurs for function expressions and `@satisfies` casts. ts-go inserts an `AsExpression`/`SatisfiesExpression` between the function and `PropertyDeclaration`, but this helper skips only parentheses, so the function is not recognized as the field initializer. Resolve the ESTree-visible parent through these JSDoc-generated wrappers as well.

swwind: Fixed in 901bd734a. `classFieldOfFunctionLike` now uses `utils.ESTreeParent`, which skips source parentheses and JSDoc-generated `@type`/`@satisfies` cast wrappers while preserving authored TypeScript assertions. Focused regressions cover JSDoc-wrapped arrow functions, function expressions, and `@satisfies` casts; `go test ./internal/rules/class_methods_use_this/` passes.

elecmonkey (approved):
😯

PR #1882, opened 2026-08-25
feat: port rule new-cap

swwind:
## Summary

Port the `new-cap` rule from ESLint to rslint.

Require constructor names to begin with an uppercase letter and require uppercase function names to be called with `new`. Support all upstream options, ECMAScript-compatible exception patterns, static property names, optional chains, and TypeScript parser-recovery boundaries.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/new-cap
- Source code: https://github.com/eslint/eslint/blob/v10.9.1/lib/rules/new-cap.js

## Checklist


internal/rules/new_cap/new_cap.go
```diff
@@ -0,0 +1,264 @@
...
+	name string,
+	pattern *esregexp.RegExp,
+	skipProperties bool,
+) bool {
+	callee = ast.SkipParentheses(callee)
+	if callee == nil {
+		return false
+	}
+	sourceText := utils.TrimmedNodeText(ctx.SourceFile, callee)
+	if _, ok := allowed[name]; ok {
```
elecmonkey, line 189: [P2] Preserve inherited exception names from upstream

With the default options, ESLint v10.9.1 accepts `new constructor()`, `new toString()`, and the other lowercase names inherited from `Object.prototype`, because upstream stores exceptions in `{}` and tests them with `allowedMap[calleeName]`. This Go map has no prototype entries, so the same inputs reach the `lower` diagnostic path and produce false positives. Seed the relevant inherited names or otherwise reproduce upstream's lookup semantics.

internal/rules/new_cap/new_cap.go
```diff
@@ -0,0 +1,264 @@
...
+		}
+	}
+}
+
+func compilePattern(value any) *esregexp.RegExp {
+	pattern, _ := value.(string)
+	if pattern == "" {
+		return nil
+	}
+	compiled, err := esregexp.Compile(pattern, "u")
```
elecmonkey, line 63: [P2] Surface patterns rejected by the required Unicode mode

Patterns such as `\a`, `\1`, `(?=a)*`, or `]` pass the schema's flagless regex validation but fail `esregexp.Compile(pattern, "u")`; this branch silently converts that failure into a nil pattern. ESLint v10.9.1 constructs `new RegExp(pattern, "u")` and aborts rule initialization for these inputs, whereas Rslint continues and behaves as if no exception pattern were configured, potentially emitting unrelated diagnostics. Validate using `u` semantics or propagate the compilation failure instead of dropping the option.

internal/rules/new_cap/new_cap.go
swwind: Thanks. This behavior comes from upstream's exception map being a plain object, so inherited `Object.prototype` names are accidentally treated as exceptions. rslint intentionally uses a true string set and does not preserve that upstream bug. I've added a focused regression case for `new constructor()` and documented the divergence on the rule page.

internal/rules/new_cap/new_cap.go
swwind: Thanks. This is the known framework-owned mismatch between flagless schema validation and runtime `u` compilation (`REGEXP-02`). This PR already compiles with upstream's `u` flag, and it does not change the shared schema validator, so I'm leaving this unchanged.

internal/rules/new_cap/new_cap.go
```diff
@@ -0,0 +1,264 @@
...
+	if pattern != nil && pattern.TestOrTimeout(sourceText) {
+		return true
+	}
+	if name == "UTC" && ast.IsAccessExpression(callee) {
+		object := ast.SkipParentheses(utils.AccessExpressionObject(callee))
+		if object != nil && object.Kind == ast.KindIdentifier && object.AsIdentifier().Text == "Date" {
+			return true
+		}
+	}
+	return skipProperties && ast.IsAccessExpression(callee)
```
elecmonkey, line 202: [P2] Keep non-Date UTC calls reportable with properties:false

When `properties: false` and the callee is a non-`Date` member named `UTC`—for example `obj.UTC()` or `obj["UTC"]()`—ESLint 10.9.1 still emits `upper`, because its special `UTC` branch returns whether the receiver is exactly `Date`. This implementation falls through to the generic property exemption and emits nothing. Return the receiver check directly from the `UTC` branch so only `Date.UTC` is exempted.

internal/rules/new_cap/new_cap.go
```diff
@@ -0,0 +1,264 @@
...
+	capitalizationNonAlpha capitalization = iota
+	capitalizationLower
+	capitalizationUpper
+)
+
+// getCap mirrors upstream's `str.charAt(0)` classification. JavaScript reads
+// one UTF-16 code unit here, so an astral letter begins with a lone high
+// surrogate and is consequently non-alphabetic for this rule.
+func getCap(name string) capitalization {
+	units := ecmascript.StringCodeUnits(name)
```
elecmonkey: [P2] Read only the first UTF-16 code unit

Every handled call/new expression invokes `getCap`, but `StringCodeUnits(name)` converts and allocates for the entire callee name even though only index 0 is used. A benchmark with a 1.2 MB static property name measured about 2.37 ms and 2.4 MB allocated per call, while reading only the first code point/code unit took about 10 ns with no allocation. Decode only the leading rune and derive its first UTF-16 unit to keep this hot path constant-space.

internal/rules/new_cap/new_cap.go
swwind: Thanks. Fixed by returning the Date receiver check directly from the UTC special case, so `obj.UTC()` and `obj["UTC"]()` remain reportable with `properties: false`, while the corresponding `Date` calls stay exempt. I added focused regression cases for both dot and static element access.

internal/rules/new_cap/new_cap.go
swwind: Thanks. `getCap` now decodes only the leading rune and classifies an astral rune as a leading UTF-16 surrogate, avoiding allocation proportional to the complete callee name. I added an allocation regression using a 1 MiB name.

elecmonkey (approved):
Good!!!

PR #1885, opened 2026-08-25
feat(rstest): add prefer-expect-type-of rule

elecmonkey:
## Summary

Report `expect(typeof value).toBe(type)` and its `toEqual` form and rewrite them to `expect(value).toBeTypeOf(type)`, which states the intent directly and, when it fails, reports the value's actual type instead of comparing two strings. Only a matcher that is really called, with exactly one non-spread argument, is reported: a broken chain such as `expect(typeof value).toBe`, a spread argument, and a Chai property assertion such as `expect(typeof value).to.be.ok` are all left alone.

The assertion resolves through the shared `ParseExpectCall`, so globals, named and renamed imports, namespace and whole-module `require`, `import.meta.rstest`, `@rstest/playwright` and the `expect` a test callback receives through its context are all covered, and an `expect` from Vitest, Jest, Playwright or Chai is not. `expect.poll(fn)` and `expect.element(locator)` are excluded: the first takes a callback and the second a locator, so neither carries the value being type checked. The fix is deferred. Registered in the rstest plugin but left out of the `recommended` preset, matching upstream.

## Credits

Port from [`@vitest/eslint-plugin`](https://github.com/vitest-dev/eslint-plugin-vitest)'s `prefer-expect-type-of`. `eslint-plugin-jest` has no counterpart, so there is no second baseline to cross-check against.

- **The fix is two surgical edits rather than a whole-call rewrite.** Upstream replaces the entire assertion with `` `expect(${value})${modifiers}.toBeTypeOf(${type})` ``. This port removes the `typeof` operator from the assertion's argument and renames the matcher accessor, leaving everything else as written. Three shapes come out differently: `expect(actual, message)` keeps the second argument Rstest's `expect` accepts and upstream's rewrite drops; `expect.soft(...)` stays soft instead of becoming an assertion that aborts the test; and the expect root keeps its spelling, so `import.meta.rstest.expect(...)`, a renamed import and a test context's `expect` are not replaced by a bare `expect` that may not be bound in that scope. The accessor's own quoting is preserved for the same reason, so `expect(typeof x)['toBe']('string')` comes back as `['toBeTypeOf']`.

- **A computed identifier matcher key is not reported.** `expect(typeof value)[matcherName]('string')` chooses its matcher at runtime, so a key that happens to be spelled `toBe` names a variable rather than a matcher. Reporting it would flag whatever assertion that variable holds, and the fix would have to rewrite the variable's name; the rule stays silent instead. Upstream reads the key's own text and both reports and fixes it.

## Related Links

Related #935

## Checklist

### DIFF
--- a/internal/plugins/rstest/rules/prefer_expect_type_of/prefer_expect_type_of.md
+++ b/internal/plugins/rstest/rules/prefer_expect_type_of/prefer_expect_type_of.md
@@ -1,0 +1,90 @@
+# prefer-expect-type-of
+
+## Rule Details
+
+Prefer `expect(value).toBeTypeOf(type)` over `expect(typeof value).toBe(type)`.
+The dedicated matcher states the intent directly and, when it fails, reports
+the value's actual type instead of a plain string comparison.
+
+Examples of **incorrect** code:
+
+```ts
+expect(typeof value).toBe('string');
+expect(typeof value).toEqual('number');
+expect(typeof value).not.toBe('function');
+expect.soft(typeof value).toBe('object');
+```
+
+Examples of **correct** code:
+
+```ts
+expect(value).toBeTypeOf('string');
+expect(value).not.toBeTypeOf('function');
+expect.soft(value).toBeTypeOf('object');
+expect(typeof value === 'string').toBe(true);
+```
+
+The rule reports only a matcher that is actually called with exactly one
+argument. A broken chain such as `expect(typeof value).toBe`, a spread argument
+such as `expect(typeof value).toBe(...types)`, and a Chai property assertion
+such as `expect(typeof value).to.be.ok` are all left alone.
+
+## Fix
+
+The fix edits two places and leaves the rest of the call exactly as written: it
+removes the `typeof` operator from the assertion's argument, and renames the
+matcher accessor to `toBeTypeOf`. Everything else survives — the expect root as
+spelled at the call site, `expect.soft`, the optional second `message` argument
+of `expect(actual, message)`, the modifier chain, the matcher argument, the
+accessor's own quoting, and the surrounding line breaks and comments:
+
+```ts
+expect.soft(typeof value, 'should be a string')['toBe']('string');
+// becomes
+expect.soft(value, 'should be a string')['toBeTypeOf']('string');
+```
+
+A comment written between `typeof` and its operand sits inside the removed span
+and is removed with it. Every comment outside those two spans is preserved.
+
+## Rstest specifics
+
+The assertion is recognised through Rstest's shared expect parser, so every
+expect source is covered: globals, `@rstest/core` named imports and renamed
+imports, `require('@rstest/core')` destructuring, namespace imports, whole-module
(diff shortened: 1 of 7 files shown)


internal/plugins/rstest/rules/prefer_expect_type_of/prefer_expect_type_of.md
```diff
@@ -0,0 +1,90 @@
...
+The port also narrows one shape:
+
+- A computed identifier key, `expect(typeof value)[matcherName]('string')`, is
+  not reported. The matcher is chosen at runtime, so `toBe` there is the name of
+  a variable rather than of a matcher, and neither the report nor a rename of
+  that variable would be correct.
+
+## Original Documentation
+
+- [@vitest/eslint-plugin: prefer-expect-type-of](https://github.com/vitest-dev/eslint-plugin-vitest/blob/v1.6.27/docs/rules/prefer-expect-type-of.md)
```
swwind: No need to cite vitest plugin as upstream.

internal/plugins/rstest/rules/prefer_expect_type_of/prefer_expect_type_of.md
```diff
@@ -0,0 +1,90 @@
...
+`require`, `import.meta.rstest`, `@rstest/playwright`, and the `expect` a test
+callback receives through its context (`({ expect })` and `ctx.expect`). An
+`expect` imported from Vitest, Jest, Playwright or Chai, and a local variable
+that shadows `expect`, are not reported.
+
+`expect.poll(fn)` and `expect.element(locator)` are excluded: the first takes a
+callback and the second a locator, so neither carries the value being type
+checked.
+
+## Differences from ESLint
```
swwind: Do we really need this chapter to clarify the difference from vitest?

internal/plugins/rstest/rules/prefer_expect_type_of/prefer_expect_type_of.md
```diff
@@ -0,0 +1,90 @@
...
+  `expect` by upstream, which may not even be bound in that scope.
+
+The port also narrows one shape:
+
+- A computed identifier key, `expect(typeof value)[matcherName]('string')`, is
+  not reported. The matcher is chosen at runtime, so `toBe` there is the name of
+  a variable rather than of a matcher, and neither the report nor a rename of
+  that variable would be correct.
+
+## Original Documentation
```
swwind: Do we really need to reference vitest as upstream here?

elecmonkey: fixed

internal/plugins/rstest/rules/prefer_expect_type_of/prefer_expect_type_of.md
elecmonkey: fixed!!!

swwind (approved):
LGTM, thanks for you patience.

PR #1889, opened 2026-08-25
feat(rstest): add warn-todo rule

elecmonkey:
## Summary

Report `.todo` on Rstest `test`, `it` and `describe` registrations. A todo registration names work that has not been written yet and the runner reports it without running anything, so one left in a committed suite is a test that can never fail.

The registration resolves through the shared `ParseFnCall`, so globals, named and renamed imports, namespace and whole-module `require`, `import.meta.rstest`, same-file aliases and `@rstest/playwright` are all covered, and a `test.todo` from Vitest, Jest or `node:test` is not. The report anchors on the `todo` accessor written at the call site, in whatever spelling it was written — `.todo`, `['todo']`, `` [`todo`] ``. When `.todo` came in through an alias, the call site has no such accessor, so `const pending = test.todo; pending('x')` is reported on `pending`, the identifier that resolves to the todo registration. The rule only reports: there is no autofix and no suggestion, because the replacement for a todo is the test body. Registered in the rstest plugin but left out of the `recommended` preset, matching upstream.

## Credits

Port from [`@vitest/eslint-plugin`](https://github.com/vitest-dev/eslint-plugin-vitest)'s `warn-todo`. `eslint-plugin-jest` has no counterpart, so there is no second baseline to cross-check against.

- **`todo` written as a property of a test options object is not reported.** Upstream also reports `test('foo', { todo: true }, fn)`. Rstest's options object accepts only `timeout`, `retry`, `repeats` and `meta`, so `todo` there is an ordinary unknown property with no effect on the runner, and reporting it would be a false positive. Upstream's invalid `{ todo: true }` case is therefore valid here, alongside its already valid `{ todo: false }` case.
- **A computed `todo` accessor is reported.** Upstream matches the member only when it is written as an identifier, so `test['todo']('case')` and its template-literal spelling go unreported there. Both spellings register the same todo, so both are reported here.
- **A modifier chained after `.todo` keeps the registration.** Upstream validates the whole member chain against a fixed allowlist that has no entry for `todo.runIf`, `todo.skipIf` or `todo.todo`, so those calls are not recognized as registrations at all. Rstest's `test` and `describe` return a chainable API, so `test.todo.runIf(cond)('case')`, `test.todo.skipIf(cond)('case')` and `test.todo.todo('case')` each still register a todo and are reported.
- **Transparent TypeScript receiver wrappers are reported.** `test!.todo(...)`, `(test as typeof test).todo(...)`, `(test satisfies typeof test).todo(...)`, and angle-bracket assertions all invoke the same registration and are recognized.
- **Same-file aliases are resolved.** Plain const aliases and direct const object bindings, including renamed, shorthand, and static computed keys, resolve to their Rstest registration.
- **A registration nested inside another call is reported.** `consume(test.todo("case"))` still executes the inner todo registration, so the inner call is reported.

## Related Links

Related #935

## Checklist


internal/plugins/rstest/rules/warn_todo/warn_todo_upstream_test.go
```diff
@@ -0,0 +1,88 @@
...
+		Line:      line,
+		Column:    column,
+		EndLine:   line,
+		EndColumn: endColumn,
+	}
+}
+
+// Upstream: @vitest/eslint-plugin v1.6.27 tests/warn-todo.test.ts.
+//
+// Two upstream cases move because Rstest's TestOptions has no `todo` field
```
swwind: **[P2] Document and test the remaining upstream divergences**

This comment and the PR description currently document only the options-object divergence, but running the actual `@vitest/eslint-plugin@1.6.27` rule reveals two additional behavioral differences:

- Upstream only matches identifier members, so `test['todo']('case')` and ``test[`todo`]('case')`` are not reported, while this implementation reports them.
- Upstream’s call-chain allowlist rejects forms such as `test.todo.runIf(cond)('case')`, `test.todo.skipIf(cond)('case')`, and `test.todo.todo('case')`, while Rstest returns a chainable `TestAPI` and this implementation reports them.

These appear to be reasonable Rstest-specific extensions rather than implementation bugs, but they should be explicitly listed as intentional differences, with focused regression tests for the latter chain forms. Otherwise, “Two upstream cases move” may lead future maintainers to assume that the options object is the only upstream divergence.

elecmonkey: Both are real and both are intentional; I've listed them in the PR description and rewritten the comment so it names all three divergences instead of just the options object.

The computed accessor: upstream matches the member only when it is an identifier, so `test['todo']('case')` and its template-literal spelling go unreported there. Both spellings register the same todo, so both are reported here. Cases for all three spellings are already in `warn_todo_extras_test.go`.

The chain forms: upstream validates the whole member chain against a fixed allowlist, which has no entry for `todo.runIf`, `todo.skipIf` or `todo.todo`, so those calls are not recognized as registrations at all. Rstest's `test` and `describe` return a chainable API, so each of those forms still registers a todo. Added regression cases for `test.todo.runIf(cond)`, `test.todo.skipIf(cond)`, `describe.todo.runIf(cond)`, and `test.todo.todo` (two accessors register one todo, so one report, anchored on the first).

internal/plugins/rstest/rules/warn_todo/warn_todo_extras_test.go
```diff
@@ -0,0 +1,228 @@
...
+			{Code: `const test = createRunner(); test.todo("case");`},
+			{Code: `function run(test) { test.todo("case"); }`},
+			{Code: `{ let describe = makeSuite(); describe.todo("suite"); }`},
+			// A computed member the parser cannot read as `todo`.
+			{Code: `test[name]("case", () => {});`},
+			{Code: "test[`to${suffix}`](\"case\", () => {});"},
+			{Code: `test[0]("case", () => {});`},
+			// A declaration without an initializer must not be read as a require.
+			{Code: `declare const pending: any; pending("case");`},
+			// Dimension 4: the parser does not read a registration through a
```
swwind: **[P2] Do not treat transparent TypeScript wrappers as valid**

`test!.todo("case")`, `(test as typeof test).todo("case")`, and `(test satisfies typeof test).todo("case")` all invoke the same Rstest `test.todo` registration. Running these cases with Rstest 0.11.3 collects each one as a todo, but this rule reports none and these valid cases explicitly lock in that false negative.

Please unwrap the transparent TypeScript receiver wrappers during Rstest call resolution and move these cases to the invalid set. This rule promises to report `.todo` registrations, so silently accepting ordinary TypeScript assertions is not a safe “fail closed” behavior.

internal/plugins/rstest/rules/warn_todo/warn_todo.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+var WarnTodoRule = rule.Rule{
+	Name:   "rstest/warn-todo",
+	Schema: rule.EmptyArraySchema,
+	Run: func(ctx rule.RuleContext, options []any) rule.RuleListeners {
+		analysis := rstestUtils.GetRstestCallAnalysis(ctx)
+		return rule.RuleListeners{
+			ast.KindCallExpression: func(node *ast.Node) {
+				parsed := analysis.ParseFnCall(node)
```
swwind, line 23: **[P2] Resolve todo registrations introduced through destructuring aliases**

Same-file alias support does not cover destructuring the Rstest API:

```ts
const { todo: pending } = test;
pending("case");

const { todo } = test;
todo("another case");
```

Rstest 0.11.3 collects both calls as todo tests, while `ParseFnCall` returns no registration and this listener emits no diagnostic. The rule documentation says same-file aliases are recognized, so please extend alias resolution to binding elements and add both renamed and shorthand destructuring regression cases.

internal/plugins/rstest/rules/warn_todo/warn_todo_extras_test.go
elecmonkey: Implemented in 476bca965d. The Rstest call-chain parser and candidate-root resolver now unwrap non-null, as, satisfies, and angle-bracket type assertions. I moved the reported cases to the invalid set and added angle-bracket assertion coverage. The focused packages and the full Rstest plugin suite pass.

internal/plugins/rstest/rules/warn_todo/warn_todo.go
elecmonkey: Implemented in 476bca965d. Const object binding elements are now resolved as alias chain parts, covering both renamed and shorthand destructuring while keeping alias resolution limited to direct const object bindings. I added rule-level regressions and shared semantic-flag coverage. The focused packages and the full Rstest plugin suite pass.

internal/plugins/rstest/utils/parse_rstest_fn.go
```diff
@@ -441,6 +442,45 @@ func resolveRstestRoot(
...
+	variable := internalUtils.EnclosingVariableDeclarationOfBindingElement(declaration)
+	if !isConstRstestVariableDeclaration(variable) || declaration.Parent.Parent != variable {
+		return nil, nil, false
+	}
+
+	nameNode := binding.PropertyName
+	if nameNode == nil {
+		nameNode = binding.Name()
+	}
+	name, ok := internalUtils.GetStaticStringLiteralValue(nameNode)
```
swwind: **[P1] Normalize computed binding property names before resolving aliases**

Static computed destructuring keys currently either crash the rule or bypass the new alias resolution. For example:

```ts
const { ["test"]: check } = require("@rstest/core");
check.todo("case");
```

The shared candidate collector calls `Text()` on the `ComputedPropertyName` and panics; the same happens with `import.meta.rstest`. The same-file form:

```ts
const { ["todo"]: pending } = test;
pending("case");
```

does not panic, but emits no diagnostic because `GetStaticStringLiteralValue` receives the `ComputedPropertyName` wrapper instead of its expression. The no-substitution template spelling has the same false negative.

Please normalize static binding property names consistently in both the direct import/require candidate path and `constRstestAliasInitializer`—for example via the existing static-property-name helper—while continuing to reject dynamic keys. Add focused string and template computed-binding regressions for direct sources and same-file aliases.

internal/plugins/rstest/rules/warn_todo/warn_todo_upstream_test.go
```diff
@@ -0,0 +1,104 @@
...
+		Line:      line,
+		Column:    column,
+		EndLine:   line,
+		EndColumn: endColumn,
+	}
+}
+
+// Upstream: @vitest/eslint-plugin v1.6.27 tests/warn-todo.test.ts.
+//
+// Three divergences from upstream, all of them deliberate; the cases that show
```
swwind, line 26: **[P2] Keep the upstream divergence inventory complete**

Running `@vitest/eslint-plugin@1.6.27` reveals additional deliberate differences beyond the three listed here:

- TypeScript receiver wrappers such as `test!.todo(...)` and `(test as typeof test).todo(...)` are not recognized upstream but are reported here.
- Upstream does not resolve the plain or destructured same-file aliases supported by this implementation.
- Upstream also skips a real registration nested inside another call, such as `consume(test.todo("case"))`, because its parser rejects a Vitest call whose parent is another `CallExpression`; Rslint reports it.

These are reasonable Rstest extensions, but the current “Three divergences” statement is no longer accurate. Please list the complete intentional differences on the rule documentation page, update this comment, and add a focused nested-call regression alongside the existing wrapper and alias cases.

internal/plugins/rstest/utils/parse_rstest_fn.go
elecmonkey: Addressed in 990eaa1a47. Static computed binding keys are now normalized in the candidate collector, same-file alias resolver, `import.meta.rstest` binding resolver, and shared require-binding resolver; dynamic keys remain unresolved. Added string and no-substitution-template regressions for direct require bindings and same-file `todo` aliases. The focused Rstest rule, utility, and shared test-framework packages pass.

internal/plugins/rstest/rules/warn_todo/warn_todo_upstream_test.go
elecmonkey: Updated the PR description with the TypeScript-wrapper, same-file-alias, and nested-call behaviors, and added a focused `consume(test.todo("case"))` regression. I kept the rule documentation user-facing: it describes the supported Rstest forms without comparing them to another implementation.

elecmonkey:
Reverted the shared Rstest parsing changes. `warn-todo` now uses only the existing parser capabilities; the destructured-alias, TypeScript-wrapper, and static computed-destructuring edge cases are intentionally left unsupported to avoid changing behavior in existing Rstest rules such as `prefer-todo`.

Verified with:
`go test -count=1 ./internal/plugins/rstest/rules/warn_todo ./internal/plugins/rstest/rules/prefer_todo`

fansenze (approved):

PR #1890, opened 2026-08-25
feat(rstest): add prefer-called-times rule

elecmonkey:
## Summary

Report `expect(fn).toHaveBeenCalledOnce()` and point at `expect(fn).toHaveBeenCalledTimes(1)`, so that every call-count assertion in a file reads the same way and changing an expected count is an edit to the number rather than a switch to a different matcher. The rule is autofixable.

The matcher has to be called — `expect(fn).toHaveBeenCalledOnce` asserts nothing, and rewriting it would turn a no-op into a live assertion — and an assertion factory has to have run, so `expect.toHaveBeenCalledOnce()`, which never received a value to count calls on, is left alone. Chai's `calledOnce` is a property getter whose count-taking equivalent is `callCount(1)` rather than `toHaveBeenCalledTimes(1)`, so it is not covered either.

Assertions resolve through the shared `ParseExpectCall`, so globals, named and aliased imports, namespace and whole-module `require`, `import.meta.rstest`, `@rstest/playwright` and the `expect` a test callback receives through its context are all followed. Registered in the rstest plugin but left out of the `recommended` preset, matching upstream.

The fix edits two places and leaves the rest of the assertion exactly as written: it renames the matcher accessor, and inserts `1` directly after the matcher call's opening parenthesis. The expect root as spelled at the call site, `expect.soft`, the optional second `message` argument of `expect(actual, message)`, the modifier chain, the accessor's own quoting and the surrounding line breaks and comments all survive.

## Credits

Port from [`eslint-plugin-vitest`](https://github.com/vitest-dev/eslint-plugin-vitest)'s `prefer-called-times`.

- **`toBeCalledOnce` is not reported.** Upstream matches both `toBeCalledOnce` and `toHaveBeenCalledOnce`, and derives the replacement by substituting `Times` for `Once`. `@vitest/expect@4.1.10`, which Rstest depends on, has no `toBeCalledOnce` at all, so code calling it is broken rather than improvable and the rule stays out of it. That leaves one source name and one replacement, so the derivation becomes two constants.

- **A matcher that already has arguments is not reported.** `toHaveBeenCalledOnce` takes none, so anything written between the parentheses means the assertion is not the shape this rule rewrites — and upstream's fix would insert `1` in front of it and change what the assertion checks.

- **`expect.toHaveBeenCalledOnce()` is not reported.** No value was ever passed to `expect`, so the assertion cannot run whichever call-count matcher it names. Upstream reports it and renames the matcher, which only makes a broken assertion look correct.

- **The insertion point for `1` is scanned for.** Upstream computes it as one character past the end of the matcher name, which assumes the opening parenthesis sits immediately after it. That holds for `expect(fn).toHaveBeenCalledOnce()` alone; `expect(fn)['toHaveBeenCalledOnce']()`, a call split across lines, and a comment between the name and the parentheses each put other characters in between, and upstream's fix corrupts them. This port scans for the parenthesis instead, and inserts the `1` after it so a comment already written between the parentheses is kept.

- **Element-access fixes preserve the accessor's delimiters.** `['toHaveBeenCalledOnce']` becomes `['toHaveBeenCalledTimes']` rather than being normalized to a single-quoted string, matching the rest of the plugin.

- **A parenthesized optional chain is still reported.** ESTree wraps an optional chain in a `ChainExpression`, and upstream's chain walk does not enter it, so `(expect(fn)?.toHaveBeenCalledOnce)()` goes unreported there. The matcher and its receiver are unambiguous in that shape and the rewrite is exact, so this port reports and fixes it.

## Related Links

Related #935

## Checklist

### DIFF
--- a/internal/plugins/rstest/rules/prefer_called_times/prefer_called_times.go
+++ b/internal/plugins/rstest/rules/prefer_called_times/prefer_called_times.go
@@ -34,7 +34,7 @@ var PreferCalledTimesRule = rule.Rule{
 		analysis := rstestUtils.GetRstestCallAnalysis(ctx)
 		return rule.RuleListeners{
 			ast.KindCallExpression: func(node *ast.Node) {
-				matcher, ok := calledOnceMatcherEntry(analysis.ParseExpectCall(node))
+				matcher, matcherCall, ok := calledOnceMatcherEntry(analysis.ParseExpectCall(node))
 				if !ok {
 					return
 				}
@@ -49,7 +49,7 @@ var PreferCalledTimesRule = rule.Rule{
 						return nil
 					}
 
-					argumentStart, ok := argumentListStart(ctx.SourceFile, matcher.Call)
+					argumentStart, ok := argumentListStart(ctx.SourceFile, matcherCall)
 					if !ok {
 						return nil
 					}
@@ -65,7 +65,7 @@ var PreferCalledTimesRule = rule.Rule{
 }
 
 // calledOnceMatcherEntry returns the matcher entry of a `toHaveBeenCalledOnce()`
-// assertion.
+// assertion together with the call expression that invokes that matcher.
 //
 // The matcher must be invoked: `expect(fn).toHaveBeenCalledOnce` asserts
 // nothing, and Chai's `calledOnce` is a property getter whose call-count
@@ -76,7 +76,9 @@ var PreferCalledTimesRule = rule.Rule{
 // and friends, where the author never passed a value to assert on; the
 // assertion is broken whichever call-count matcher it names, so rewriting the
 // matcher would only make a wrong assertion look right.
-func calledOnceMatcherEntry(parsed *rstestUtils.ParsedRstestExpectCall) (rstestUtils.ParsedRstestFnMemberEntry, bool) {
+func calledOnceMatcherEntry(
+	parsed *rstestUtils.ParsedRstestExpectCall,
+) (rstestUtils.ParsedRstestFnMemberEntry, *ast.Node, bool) {
 	if parsed == nil ||
 		parsed.Reason != rstestUtils.RstestExpectParseReasonNone ||
 		parsed.Head == nil ||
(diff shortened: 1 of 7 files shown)


internal/plugins/rstest/rules/prefer_called_times/prefer_called_times.go
```diff
@@ -0,0 +1,135 @@
...
+					nameRange, nameText, ok := testFramework.AccessorReplacement(
+						ctx.SourceFile,
+						matcher.Node,
+						calledTimesMatcher,
+					)
+					if !ok {
+						return nil
+					}
+
+					argumentStart, ok := argumentListStart(ctx.SourceFile, matcher.Call)
```
swwind: [P2] Please insert the count into the matcher call, not a trailing call

A trailing direct call causes `matcher.Call` to refer to the outer call:

```ts
expect(fn).toHaveBeenCalledOnce()();
```

The pinned upstream rule fixes this to:

```ts
expect(fn).toHaveBeenCalledTimes(1)();
```

Rslint currently produces:

```ts
expect(fn).toHaveBeenCalledTimes()(1);
```

The same problem repeats with additional trailing calls. Could we ensure that `argumentListStart` receives the call expression that directly invokes the matcher, and add a regression case for this shape?

internal/plugins/rstest/rules/prefer_called_times/prefer_called_times.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+var PreferCalledTimesRule = rule.Rule{
+	Name:   "rstest/prefer-called-times",
+	Schema: rule.EmptyArraySchema,
+	Run: func(ctx rule.RuleContext, options []any) rule.RuleListeners {
+		analysis := rstestUtils.GetRstestCallAnalysis(ctx)
+		return rule.RuleListeners{
+			ast.KindCallExpression: func(node *ast.Node) {
+				matcher, ok := calledOnceMatcherEntry(analysis.ParseExpectCall(node))
```
swwind: [P2] Preserve the upstream optional-chain boundary through parentheses

The shared parser crosses the parenthesized optional-chain boundary here:

```ts
(expect(fn)?.toHaveBeenCalledOnce)();
```

`@vitest/eslint-plugin@1.6.27` reports no diagnostic because ESTree keeps the optional chain inside a `ChainExpression`, while Rslint reports and fixes it. The same mismatch occurs with double parentheses and an optional outer call:

```ts
((expect(fn)?.toHaveBeenCalledOnce))();
(expect(fn)?.toHaveBeenCalledOnce)?.();
```

Could we stop the matcher walk at this optional-chain boundary, or explicitly document and test this as an intentional upstream divergence?

internal/plugins/rstest/rules/prefer_called_times/prefer_called_times.go
elecmonkey: Fixed in faaa67660e.

The root cause was `MemberEntry.Call`. `GetMemberEntries` marks the last entry of a callee chain as invoked, so every enclosing call overwrites the field — with a trailing call the matcher entry ended up carrying the outer call. That also affected the guard that withholds the report when the matcher already has arguments, so `expect(fn).toHaveBeenCalledOnce(1)()` was being reported.

The rule now derives the matcher's own call by walking up from the accessor and stopping at the call the matcher name is the callee of, instead of reading `MemberEntry.Call`. Regression cases added for `toHaveBeenCalledOnce()()`, `()()()`, `()(1)`, `['toHaveBeenCalledOnce']()()`, plus a valid case for `toHaveBeenCalledOnce(1)()`.

internal/plugins/rstest/rules/prefer_called_times/prefer_called_times.go
elecmonkey: Kept as an intentional divergence, now documented and tested.

`(expect(fn)?.toHaveBeenCalledOnce)()` is a real `toHaveBeenCalledOnce` assertion, and the fix produces `(expect(fn)?.toHaveBeenCalledTimes)(1)`, which is exact — so this is a coverage difference rather than a correctness one. The reference implementation misses it because ESTree wraps an optional chain in a `ChainExpression` that its chain walk does not enter, which reads more like an oversight there than a deliberate boundary.

Added invalid cases for `(...)()`, `((...))()` and `(...)?.()`, noted in the rule page that parentheses are transparent including around an optional chain, and recorded it in the PR description under the intentional differences.

swwind (approved):
LGTM

PR #1894, opened 2026-08-26
feat: port rule default-param-last

swwind:
## Summary

Port the `default-param-last` rule from ESLint to rslint.

Enforce default and optional parameters to appear after required parameters, including TypeScript parameter properties.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/default-param-last
- Source code: https://github.com/eslint/eslint/blob/v10.9.1/lib/rules/default-param-last.js

## Checklist

### DIFF
--- a/internal/rules/default_param_last/default_param_last.go
+++ b/internal/rules/default_param_last/default_param_last.go
@@ -44,7 +44,11 @@ var DefaultParamLastRule = rule.Rule{
 			}
 			for i := range lastRequired {
 				if !isRequiredParameter(params[i]) {
-					ctx.ReportRange(utils.NodeTextRangeSkippingDecorators(ctx.SourceFile, params[i]), shouldBeLastMessage)
+					reportRange := utils.NodeTextRangeSkippingDecorators(ctx.SourceFile, params[i])
+					if ast.IsParameterPropertyDeclaration(params[i], node) {
+						reportRange = utils.TrimNodeTextRange(ctx.SourceFile, params[i])
+					}
+					ctx.ReportRange(reportRange, shouldBeLastMessage)
 				}
 			}
 		}
--- a/internal/rules/default_param_last/default_param_last.go
+++ b/internal/rules/default_param_last/default_param_last.go
@@ -17,6 +17,9 @@ var DefaultParamLastRule = rule.Rule{
 	Schema: rule.EmptyArraySchema,
 	Run: func(ctx rule.RuleContext, options []any) rule.RuleListeners {
 		check := func(node *ast.Node) {
+			if node.Body() == nil {
+				return
+			}
 			params := utils.ESTreeParameters(node)
 			isRequiredParameter := func(current *ast.Node) bool {
 				if current == nil {
--- a/internal/rules/default_param_last/default_param_last.go
+++ b/internal/rules/default_param_last/default_param_last.go
@@ -1,0 +1,61 @@
+package default_param_last
+
+import (
+	"github.com/microsoft/typescript-go/shim/ast"
+	"github.com/web-infra-dev/rslint/internal/rule"
+	"github.com/web-infra-dev/rslint/internal/utils"
+)
+
+var shouldBeLastMessage = rule.RuleMessage{
+	Id:          "shouldBeLast",
+	Description: "Default parameters should be last.",
+}
+
+// DefaultParamLastRule enforces default parameters to be last.
+var DefaultParamLastRule = rule.Rule{
+	Name:   "default-param-last",
+	Schema: rule.EmptyArraySchema,
+	Run: func(ctx rule.RuleContext, options []any) rule.RuleListeners {
+		check := func(node *ast.Node) {
+			params := utils.ESTreeParameters(node)
+			isRequiredParameter := func(current *ast.Node) bool {
+				if current == nil {
+					return false
+				}
+				param := current.AsParameterDeclaration()
+				questionToken := param.QuestionToken
+				if utils.IsJSDocSyntaxNode(questionToken) {
+					questionToken = nil
+				}
+				return param.Initializer == nil &&
+					param.DotDotDotToken == nil &&
+					questionToken == nil
+			}
+
+			lastRequired := -1
+			for i := len(params) - 1; i >= 0; i-- {
+				if isRequiredParameter(params[i]) {
+					lastRequired = i
+					break
+				}
+			}
+			for i := range lastRequired {
+				if !isRequiredParameter(params[i]) {
+					ctx.ReportNode(params[i], shouldBeLastMessage)
+				}
+			}
+		}
+
+		// ESLint's FunctionExpression listener also reaches methods and
+		// constructors through ESTree wrapper nodes. tsgo models them directly.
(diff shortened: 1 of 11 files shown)


internal/rules/default_param_last/default_param_last.go
```diff
@@ -0,0 +1,64 @@
...
+			lastRequired := -1
+			for i := len(params) - 1; i >= 0; i-- {
+				if isRequiredParameter(params[i]) {
+					lastRequired = i
+					break
+				}
+			}
+			for i := range lastRequired {
+				if !isRequiredParameter(params[i]) {
+					ctx.ReportNode(params[i], shouldBeLastMessage)
```
elecmonkey: [P2] Report range includes parameter decorators when the parameter is not a parameter property

Reporting the whole `ParameterDeclaration` matches upstream for parameter properties, since ESTree's `TSParameterProperty` spans decorators and modifiers. For a decorated parameter with no modifier, upstream reports the bare pattern node instead, whose range starts after the decorators. For `class Controller { list(@Query('page') page = 1, filter: string) {} }`, eslint@10.9.1 with @typescript-eslint/parser reports `3:23-3:31` (`page = 1`) while this rule reports `3:8-3:31`, starting at the `@`. Same messageId and count, only the highlighted span differs; it reproduces for `@D() a = 1`, `@D() a?: number` and `@D() {a} = {a:1}`, and the span grows to cover every decorator when there are several. Parameter properties such as `@D() readonly a = 1` are unaffected. Starting the range at the binding name when the parameter has no modifiers would restore alignment.

internal/rules/default_param_last/default_param_last_extras_test.go
```diff
@@ -0,0 +1,138 @@
...
+
+			// ---- Dimension 4: graceful degradation ----
+			{Code: "function empty() {}"},
+			{Code: "function destructured({value}, [fallback] = []) {}"},
+			{Code: "function rest(defaulted = 1, ...values) {}"},
+			// typescript-eslint exposes bodyless declarations as nodes outside the upstream listeners.
+			{Code: "declare function overloaded(optional?: number, required: number): void;"},
+			{Code: "abstract class C { abstract method(optional?: number, required: number): void; }"},
+			{Code: "class C { constructor(optional?: number, required: number); constructor(required: number) {} }"},
+			{Code: "interface I { method(optional?: number, required?: number): void; }"},
```
elecmonkey: [P3] This case doesn't lock in the interface branch

Both parameters here are optional, so there is no required parameter to the right and the case would pass even if method signatures were visited. Making the second one `required: number` turns it into an actual lock-in — that form reports nothing on either side today. Separately, the rest case on line 47 duplicates the shape already covered on line 40.

elecmonkey (commented):
Verified against eslint@10.9.1 + @typescript-eslint/parser@8.68.0 by building this branch and running 72 differential cases through both engines, plus the Go suite, the JS snapshot test, and one end-to-end CLI repro. 70/72 match exactly on count, messageId, line, column and endColumn — including optional parameters, parameter properties, rest-before-required, `this` parameters, accessors, computed/private/static methods, namespace and export forms, destructured defaults, `as`/`satisfies` defaults, non-BMP defaults, comment-adjacent parameters, and all four bodyless shapes (`declare function`, abstract method, function and constructor overloads, interface methods), which both sides correctly skip. The 41 upstream invalid cases and the full upstream valid list are migrated 1:1 and the JS snapshot columns match ESLint's own output.

Two notes below: one report-range divergence for decorated parameters, and one test that doesn't lock in the branch its comment names.

internal/rules/default_param_last/default_param_last.go
swwind: Fixed in `b8f6b27b1`. Ordinary decorated parameters now exclude decorators from the report range, while TypeScript parameter properties preserve the full decorator and modifier range. Focused range assertions cover both forms.

internal/rules/default_param_last/default_param_last_extras_test.go
swwind: Fixed in `b44b5cbc`. The interface method case now places a required parameter after the optional parameter, so it fails if bodyless signatures are visited. I also removed the duplicate rest-parameter case.

internal/rules/default_param_last/default_param_last.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/web-infra-dev/rslint/internal/utils"
+)
+
+var shouldBeLastMessage = rule.RuleMessage{
+	Id:          "shouldBeLast",
+	Description: "Default parameters should be last.",
+}
+
+// DefaultParamLastRule enforces default parameters to be last.
+var DefaultParamLastRule = rule.Rule{
```
elecmonkey, line 15: [P2] Reuse the core implementation for `@typescript-eslint/default-param-last`

`internal/plugins/typescript/rules/default_param_last/` already implements this rule, and the repo's extension rules delegate to their core counterpart (`dot_notation`, `no_empty_function`, `no_shadow`, `no_redeclare`, `no_magic_numbers`, `no_loop_func`, `no_restricted_imports`). This PR adds a second independent implementation, and the two already disagree on exactly the shapes fixed in this review: for `declare function overloaded(optional?: number, required: number): void;`, for `abstract class C { abstract method(optional?: number, required: number): void; }` and for a constructor overload signature, the plugin copy reports while this rule reports nothing; and for `class C { method(@first @second optional?: number, required: number) {} }` it reports `1:18-1:50` against this rule's `1:33-1:50`. typescript-eslint's own `default-param-last` uses the same three listeners as eslint@10.9.1 core, so wiring the plugin rule through `internal/rules/default_param_last` would fix those and keep the copies from drifting further.

elecmonkey (commented):
Re-reviewed at b44b5cbc.

Both earlier findings are fixed and verified against eslint@10.9.1 with @typescript-eslint/parser@8.68.0: the decorator range now matches upstream for ordinary parameters (`1:33-1:50`) and for parameter properties (`1:23-1:61`), and the interface case now fails if bodyless signatures are visited.

Differential coverage: 70 inputs run through both upstream and this branch — decorators (single/multi/factory/multiline), parameter properties, bodyless forms (declare function, ambient class, abstract method, function/constructor/accessor overloads, interface and construct signatures, type aliases, labeled tuples), destructuring, rest, `this` parameters, generics, .tsx, IIFE, export default, object-literal methods and accessors, nested functions inside default values, comments/parens/trailing commas, CRLF, emoji (UTF-16 columns), and .mjs with JSDoc `[a]`, `[a=1]` and `...a`. No unexplained differences.

One reuse issue below.

internal/rules/default_param_last/default_param_last.go
swwind: Fixed in `935179003`. `@typescript-eslint/default-param-last` now delegates to the core rule, removing the duplicate implementation. Focused plugin regressions cover bodyless function declarations, abstract methods, constructor overload signatures, and the decorator-excluding report range for ordinary parameters.

elecmonkey (approved):
congratulations!

PR #1895, opened 2026-08-26
feat: port rule camelcase

swwind:
## Summary

Port the `camelcase` rule from ESLint to rslint.

Enforce camelcase naming for identifiers and properties, including ESLint-compatible handling for destructuring, imports, globals, property checks, and allow patterns.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/camelcase
- Source code: https://github.com/eslint/eslint/blob/v10.9.1/lib/rules/camelcase.js

## Checklist


internal/rules/camelcase/camelcase.go
```diff
@@ -0,0 +1,443 @@
...
+		return owner != nil && (owner.Kind == ast.KindClassDeclaration || owner.Kind == ast.KindClassExpression) &&
+			utils.IsPlainClassMember(node.Parent)
+	}
+	return false
+}
+
+func isAssignedPropertyAccessName(node *ast.Node) bool {
+	parent := node.Parent
+	return parent != nil && parent.Kind == ast.KindPropertyAccessExpression &&
+		parent.AsPropertyAccessExpression().Name() == node && utils.IsWriteReference(parent)
```
elecmonkey: [P2] Restrict property checks to ESLint assignment targets

ESLint v10.9.1 only reports a non-computed member property when the `MemberExpression` is the left side of an assignment/assignment-pattern. Because this delegates to `utils.IsWriteReference`, Rslint also treats update expressions and `for-in`/`for-of` left sides as property writes. For example, upstream reports no `camelcase` diagnostics for `obj.snake_case++`, `++obj.snake_case`, `for (obj.snake_case in source);`, or `for (obj.snake_case of source);`, but a temporary Rslint rule-tester case reports each `snake_case` property. This creates false positives with the default `properties: "always"` option.

internal/rules/camelcase/camelcase.schema.json
```diff
@@ -0,0 +1,23 @@
...
+    {
+      "type": "object",
+      "properties": {
+        "ignoreDestructuring": { "type": "boolean" },
+        "ignoreImports": { "type": "boolean" },
+        "ignoreGlobals": { "type": "boolean" },
+        "properties": { "enum": ["always", "never"] },
+        "allow": {
+          "type": "array",
+          "items": { "type": "string" },
```
elecmonkey: [P2] Reject invalid `allow` regex patterns

The `allow` entries are later compiled as JavaScript regexps, but the schema only validates them as strings and `parseOptions` silently skips compile failures. Upstream constructs `new RegExp(entry, "u")`, so `allow: ["["]` throws once `foo_bar` is checked; this implementation accepts the config and just reports a normal camelcase diagnostic. Other ported regexp options in this repo use `"format": "regex"` so invalid user patterns fail during config validation instead of being ignored.

internal/rules/camelcase/camelcase.go
swwind: Fixed in 54d3f43b. `camelcase` now uses a rule-local assignment-target classifier that matches ESLint: parentheses remain transparent, while update expressions, direct for-in/for-of left sides, and TypeScript assertion/non-null wrappers are excluded. Added focused regressions for these boundaries.

internal/rules/camelcase/camelcase.schema.json
swwind: Fixed in 54d3f43b. The `allow` schema now marks entries with `format: "regex"`, so malformed patterns are rejected during option validation. Added a schema regression covering invalid `[` and a valid JavaScript lookahead.

internal/rules/camelcase/camelcase.go
```diff
@@ -0,0 +1,516 @@
...
+		return true
+	}
+	return false
+}
+
+func isRuntimeReference(node *ast.Node) bool {
+	if isJsxIntrinsicTagName(node) {
+		return false
+	}
+	return node != nil && !ast.IsPartOfTypeNode(node) && !ast.IsPartOfTypeQuery(node) &&
```
elecmonkey: [P2] Report unresolved names inside type queries

This filter drops every identifier inside a TypeScript `typeof` type query before the unresolved/global-reference path can run. ESLint 10.9.1 still reports `let x: typeof snake_case;` as `Identifier 'snake_case' is not in camel case.` (and similarly reports the leftmost value name in `typeof ns_snake.value`) unless that name is a configured global and `ignoreGlobals: true`; a temporary Rslint RuleTester case for `let x: typeof snake_case;` reported 0 diagnostics. This makes the port under-report non-camelcase value references used from type queries.

swwind: Fixed in 31efe3b8. Type-query roots now remain eligible runtime references, while qualified member segments are excluded. Added regressions for bare and qualified `typeof` queries and `ignoreGlobals`.

internal/rules/camelcase/camelcase.go
```diff
@@ -0,0 +1,533 @@
...
+		}
+	}
+
+	for _, identifier := range s.identifiers {
+		name := identifier.AsIdentifier().Text
+		if s.opts.isGoodName(name) || utils.IsImportAttributeKey(identifier) {
+			continue
+		}
+
+		switch runtimeBindingKind(identifier) {
```
elecmonkey, line 153: [P2] De-duplicate merged binding declarations

This reports every syntactic declaration name, while ESLint reports `variable.identifiers[0]` once per scope variable. For `var snake_case; var snake_case; snake_case;`, ESLint 10.9.1 reports the first declaration and final reference, but Rslint additionally reports the second declaration; the same mismatch occurs for duplicate parameters and merged `var`/function declarations. Track/report each declared variable once instead of treating every binding identifier as a separate declaration.

internal/rules/camelcase/camelcase.go
```diff
@@ -0,0 +1,533 @@
...
+}
+
+func (s *camelcaseState) checkFile() {
+	badReferences := make(map[*ast.Node]struct{})
+	for _, identifier := range s.identifiers {
+		kind := runtimeBindingKind(identifier)
+		if kind == bindingNone || s.opts.isGoodName(identifier.AsIdentifier().Text) {
+			continue
+		}
+		if sym := utils.BindingNameSymbol(identifier); sym != nil && s.ctx.Refs != nil {
```
elecmonkey: [P2] Track the local symbol of parameter properties

A TypeScript parameter property has separate field and local-parameter symbols, and `BindingNameSymbol` selects the former here. Consequently `class C { constructor(public snake_case: string) { snake_case; } }` produces two diagnostics in ESLint 10.9.1 but only the declaration diagnostic in Rslint; the body reference is omitted in both checker-backed and source-only programs. Resolve or reconcile the parameter's local symbol when collecting references.

internal/rules/camelcase/camelcase.go
```diff
@@ -0,0 +1,533 @@
...
+		return false
+	}
+	switch node.Parent.Kind {
+	case ast.KindLabeledStatement, ast.KindBreakStatement, ast.KindContinueStatement:
+		return true
+	}
+	return false
+}
+
+func isRuntimeReference(node *ast.Node) bool {
```
elecmonkey, line 551: [P2] Exclude namespaced JSX attributes from references

Identifiers inside a JSX namespaced attribute are syntactic names, not variable references. With `const a = <div attr_name:value_name='x' />`, ESLint 10.9.1 reports nothing, whereas Rslint reports both `attr_name` and `value_name`. The runtime-reference filter should exclude identifiers whose `JsxNamespacedName` belongs to a `JsxAttribute`, while preserving the current behavior for namespaced tag names.

internal/utils/ast_helpers.go
```diff
@@ -37,6 +37,59 @@ func OutermostParenthesizedExpression(node *ast.Node) *ast.Node {
...
+	container := outer.Parent
+	if container == nil {
+		return false
+	}
+	if container.Kind == ast.KindCallExpression {
+		call := container.AsCallExpression()
+		return call.Expression != nil && call.Expression.Kind == ast.KindImportKeyword &&
+			call.Arguments != nil && len(call.Arguments.Nodes) > 1 && call.Arguments.Nodes[1] == outer
+	}
+	if container.Kind == ast.KindPropertyAssignment && container.AsPropertyAssignment().Initializer == outer {
```
elecmonkey: [P2] Stop import-attribute recursion at computed properties

The recursive dynamic-import check does not verify that the containing property is non-computed. For `import('x', { [my_with]: { my_type: my_value } })`, ESLint 10.9.1 reports `my_with`, `my_type`, and `my_value`, but Rslint incorrectly exempts `my_type` as an import-attribute key. Match upstream's `!parent.computed` guard before following the outer property.

internal/rules/camelcase/camelcase.schema.json
```diff
@@ -0,0 +1,23 @@
...
+    {
+      "type": "object",
+      "properties": {
+        "ignoreDestructuring": { "type": "boolean" },
+        "ignoreImports": { "type": "boolean" },
+        "ignoreGlobals": { "type": "boolean" },
+        "properties": { "enum": ["always", "never"] },
+        "allow": {
+          "type": "array",
+          "items": { "type": "string", "format": "regex" },
```
elecmonkey: [P2] Validate allow patterns in Unicode mode

The rule compiles every `allow` entry as `new RegExp(pattern, "u")`, but the schema's `regex` format validates without the `u` flag. For example, `allow: ["\\a"]` passes this schema and is then silently discarded by `parseOptions`, while ESLint 10.9.1 throws `Invalid regular expression: /\a/u`. Use the repository's `regex-u` format so configuration validation matches the runtime compilation mode.

internal/rules/camelcase/camelcase.go
swwind: Fixed in ba6fa4c5b. Binding diagnostics are now de-duplicated by scope-manager variable (binding scope plus name), covering repeated var declarations, merged var/function declarations, and duplicate parameters. Added focused regressions for all three cases.

internal/rules/camelcase/camelcase.schema.json
swwind: Fixed in ba6fa4c5b. The allow schema now uses regex-u, matching the runtime Unicode-mode compilation. Added coverage showing that \a is rejected while a valid lookahead remains accepted.

internal/utils/ast_helpers.go
swwind: Fixed in ba6fa4c5b. Dynamic import-attribute key recursion now stops when the containing property has a computed name. Added a regression confirming my_with, my_type, and my_value are all checked.

internal/rules/camelcase/camelcase.go
swwind: Fixed in ba6fa4c5b. Identifiers inside a JsxNamespacedName used as a JSX attribute name are now excluded, while namespaced tag names retain their existing reference behavior. Added a TSX regression.

internal/rules/camelcase/camelcase.go
swwind: Fixed in ba6fa4c5b. Parameter properties now query constructor-body references with the constructor's local parameter symbol instead of the class property symbol left on the declaration node. Added a regression covering the declaration and body reference.

internal/rules/camelcase/camelcase.go
```diff
@@ -0,0 +1,592 @@
...
+	symbol := utils.BindingNameSymbol(node)
+	if symbol != nil {
+		for current := node.Parent; current != nil; current = current.Parent {
+			if locals := current.Locals(); locals != nil && locals[name] == symbol {
+				return current
+			}
+		}
+	}
+	for current := node.Parent; current != nil; current = current.Parent {
+		if locals := current.Locals(); locals != nil && locals[name] != nil {
```
elecmonkey, line 282: [P2] Keep function-expression-name bindings in a separate scope

For `const f = function snake_case(snake_case) { return snake_case; }`, ESLint 10.9.1 reports the function name, parameter declaration, and parameter reference, while Rslint reports only the function name and reference. This fallback maps the function-expression self-binding and the same-named parameter/`var`/`let` binding to one `(scope, name)` key, so `reportBinding` suppresses a distinct declaration. Preserve ESLint's separate function-expression-name scope when deduplicating bindings.

internal/rules/camelcase/camelcase.go
```diff
@@ -0,0 +1,592 @@
...
+}
+
+func isRuntimeReference(node *ast.Node) bool {
+	if isJsxIntrinsicTagName(node) || isJsxNamespacedAttributeName(node) {
+		return false
+	}
+	if ast.IsPartOfTypeQuery(node) {
+		return isTypeQueryValueReference(node)
+	}
+	return node != nil && !ast.IsPartOfTypeNode(node) && !utils.IsNonReferenceIdentifier(node)
```
elecmonkey, line 567: [P2] Check unresolved identifiers in ordinary TypeScript type positions

This blanket type-node exclusion drops unresolved references that typescript-eslint exposes through the Program scope. With `type X = type_ref;`, ESLint 10.9.1 plus `@typescript-eslint/parser` 8.65.0 reports `type_ref`, including with `ignoreGlobals: true`, whereas Rslint reports nothing in both checker-backed and source-only Programs. The same under-reporting affects generic arguments, `implements`/interface `extends`, `keyof`, type predicates, and assertion types; locally declared type names should remain exempt through the existing reference-resolution check.

internal/rules/camelcase/camelcase.go
```diff
@@ -0,0 +1,592 @@
...
+func isTypeQueryValueReference(node *ast.Node) bool {
+	current := node
+	for current != nil && current.Parent != nil && current.Parent.Kind == ast.KindQualifiedName {
+		qualified := current.Parent.AsQualifiedName()
+		if qualified.Left != current {
+			return false
+		}
+		current = current.Parent
+	}
+	return current != nil && current.Parent != nil && current.Parent.Kind == ast.KindTypeQuery
```
elecmonkey, line 603: [P2] Exclude import-type qualifiers from type-query references

For `type X = typeof import("pkg").value_name;`, ESLint 10.9.1 reports no diagnostic because `value_name` names an export of the imported module rather than a lexical binding, but Rslint reports it as non-camelcase. Nested import-type qualifiers show the same false positive. Before accepting a type-query root, exclude identifiers belonging to an `ImportType` qualifier, matching the repository's existing scope-reference classification.

internal/rules/camelcase/camelcase.go
swwind: Fixed in eff68618b. Named function-expression self-bindings now use a distinct scope key from same-named parameters and body bindings. Added a regression covering the declaration, parameter, and reference diagnostics.

internal/rules/camelcase/camelcase.go
swwind: Fixed in eff68618b. Unresolved identifiers in ordinary TypeScript type positions now follow the scope-reference classification, while locally resolved type names remain exempt. Added a regression for `type X = type_ref;`.

internal/rules/camelcase/camelcase.go
swwind: Fixed in eff68618b. Import-type syntax is now excluded before type-query root classification, so `typeof import("pkg").value_name` and nested import-type qualifiers are not treated as lexical references. Added a regression.

elecmonkey (commented):
Please see the inline comments.

internal/rules/camelcase/camelcase.go
```diff
@@ -0,0 +1,606 @@
...
+// bindingScope identifies the scope-manager variable that owns a declaration.
+// Looking up the first enclosing table by name also covers declarations that
+// TypeScript keeps as distinct error symbols, such as duplicate parameters or
+// conflicting var/function declarations.
+func bindingScope(node *ast.Node, name string) *ast.Node {
+	// ESLint gives a named function expression's self-binding its own scope,
+	// separate from the function scope that owns its parameters and body
+	// bindings. Keep the two scope-manager variables distinct when they share a
+	// name.
+	if node != nil && node.Parent != nil && node.Parent.Kind == ast.KindFunctionExpression && node.Parent.Name() == node {
```
elecmonkey: [P2] Keep class-expression self-bindings in a separate scope

When a named class expression has the same name as its outer binding, this deduplication merges two distinct scope-manager variables. For `const snake_case = class snake_case {};`, ESLint 10.9.1 with `@typescript-eslint/parser` 8.65.0 reports both identifiers, while Rslint reports only the outer variable; `const Good = class snake_case {}` correctly reports the class name. Please give named class-expression self-bindings their own scope key, as is already done for named function expressions.

internal/rules/camelcase/camelcase.go
```diff
@@ -0,0 +1,606 @@
...
+	}
+	return false
+}
+
+func isExportedName(node *ast.Node) bool {
+	if node == nil || node.Parent == nil {
+		return false
+	}
+	switch node.Parent.Kind {
+	case ast.KindExportSpecifier, ast.KindNamespaceExport:
```
elecmonkey: [P2] Preserve source diagnostics in aliased type exports

For `export type { type_ref as exportedName };`, ESLint 10.9.1 with `@typescript-eslint/parser` 8.65.0 reports `type_ref`, while Rslint emits no diagnostic in both checker-backed and source-only runs. The source identifier is neither the exported name recognized here nor considered a runtime reference afterward. Please keep the local/source side of aliased type export specifiers eligible for Program-scope reference handling.

elecmonkey: [P2] Recognize TypeScript namespace-export declarations

For the valid external-module declaration `export {}; export as namespace namespace_name;`, ESLint 10.9.1 reports `namespace_name`, while Rslint emits no diagnostic, including when the declaration is in a `.d.ts` file. ts-go represents this syntax with `KindNamespaceExportDeclaration`, whereas this switch only handles `KindNamespaceExport`, which represents `export * as name from`. Please include the namespace-export declaration form.

internal/rules/camelcase/camelcase.go
```diff
@@ -0,0 +1,624 @@
...
+
+// bindingScope identifies the scope-manager variable that owns a declaration.
+// Looking up the first enclosing table by name also covers declarations that
+// TypeScript keeps as distinct error symbols, such as duplicate parameters or
+// conflicting var/function declarations.
+func bindingScope(node *ast.Node, name string) *ast.Node {
+	// ESLint gives a named function or class expression's self-binding its own
+	// scope. Keep it distinct from same-named outer, parameter, and body bindings.
+	if node != nil && node.Parent != nil &&
+		(node.Parent.Kind == ast.KindFunctionExpression || node.Parent.Kind == ast.KindClassExpression) &&
```
swwind, line 269: Fixed in 89bd7c514. Named class-expression self-bindings now use their own deduplication scope key, separate from a same-named outer binding. Added regressions for the expression name and a class-body self-reference.

internal/rules/camelcase/camelcase.go
```diff
@@ -0,0 +1,624 @@
...
+// non-reference helper excludes type-only export specifiers.
+func isAliasedTypeExportSource(node *ast.Node) bool {
+	if node == nil || node.Parent == nil || node.Parent.Kind != ast.KindExportSpecifier {
+		return false
+	}
+	specifier := node.Parent.AsExportSpecifier()
+	if specifier == nil || specifier.PropertyName != node || specifier.Name() == node {
+		return false
+	}
+	return ast.IsTypeOnlyImportOrExportDeclaration(node.Parent) && !utils.IsReExportSpecifier(node.Parent)
```
swwind, line 582: Fixed in 89bd7c514. The local side of an aliased type-only export now remains eligible for Program-scope reference handling, while type-only re-exports remain excluded. Added regressions for both `export type { type_ref as exportedName };` and its `from "pkg"` form.

internal/rules/camelcase/camelcase.go
```diff
@@ -0,0 +1,624 @@
...
+	}
+	return false
+}
+
+func isExportedName(node *ast.Node) bool {
+	if node == nil || node.Parent == nil {
+		return false
+	}
+	switch node.Parent.Kind {
+	case ast.KindExportSpecifier, ast.KindNamespaceExport, ast.KindNamespaceExportDeclaration:
```
swwind, line 534: Fixed in 89bd7c514. `isExportedName` now recognizes `KindNamespaceExportDeclaration`, covering `export as namespace namespace_name`, including inside ambient modules. Added regressions for both top-level and ambient-module forms.

elecmonkey (dismissed):
Verified on commit `89bd7c514`: all three reported issues are fixed, and the targeted tests and ESLint differential checks now pass.

fansenze (approved):

PR #1896, opened 2026-08-26
feat: port rule vars-on-top

swwind:
## Summary

Port the `vars-on-top` rule from ESLint to rslint.

Require `var` declarations to be placed at the top of their containing function or program scope, while allowing directive strings and imports before them and supporting class static blocks.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/vars-on-top
- Source code: https://github.com/eslint/eslint/blob/main/lib/rules/vars-on-top.js

## Checklist


internal/rules/vars_on_top/vars_on_top.go
```diff
@@ -0,0 +1,91 @@
...
+func isAtTopOfAllowedScope(statement *ast.Node) bool {
+	if statement == nil || statement.Parent == nil {
+		return false
+	}
+
+	container := statement.Parent
+	switch container.Kind {
+	case ast.KindSourceFile:
+		return isVarOnTop(statement, container.AsSourceFile().Statements.Nodes, true)
+	case ast.KindBlock:
```
elecmonkey, line 54: **[P2] `export var` inside a namespace or ambient module block is a false positive**

`isAtTopOfAllowedScope` only accepts `SourceFile`, a function-body `Block`, and a class static block, so a `var` whose parent is a `ModuleBlock` falls straight through to `return false`. Upstream reaches that container through its `node.parent.type === "ExportNamedDeclaration"` branch, which calls `globalVarCheck(node.parent, node.parent.parent)` and therefore validates against the enclosing `TSModuleBlock`'s `body`.

Repro (same source and config, upstream `eslint@10.8.0` + `@typescript-eslint/parser@8.65.0` vs this PR head):

```ts
namespace N { export var a = 1; }
declare module 'x' { export var a: number; }
```

Upstream: no diagnostics. rslint: `top` at 1:15 and 2:22. `namespace N { var a = 1; }` (without `export`) is reported by both, so the divergence is specific to the exported form — an inconsistency in upstream itself, but one the port needs to follow. `export var` in an ambient `declare module` block is valid, common in `.d.ts` files, and cannot be reordered away, so this is a pure false positive.

Suggested direction: add an `ast.KindModuleBlock` arm to the switch; to mirror upstream exactly, take it only when the statement carries an `export` modifier and otherwise keep returning `false`. Until this is fixed, the "Differences from ESLint" section in `vars_on_top.md` claiming parity with ESLint does not match actual behavior.

internal/rules/vars_on_top/vars_on_top.go
```diff
@@ -0,0 +1,91 @@
...
+		block := container.AsBlock()
+		if block == nil || block.Statements == nil {
+			return false
+		}
+		parent := container.Parent
+		if parent != nil && ast.IsFunctionLike(parent) {
+			return isVarOnTop(statement, block.Statements.Nodes, true)
+		}
+		if parent != nil && parent.Kind == ast.KindClassStaticBlockDeclaration {
+			return isVarOnTop(statement, block.Statements.Nodes, false)
```
elecmonkey, line 64: **[P2] The static-block directive-skip flag has no test coverage**

Passing `false` here is a correct port of upstream's `if (node.parent.type !== "StaticBlock")` guard, but nothing pins it down: flipping this argument to `true` leaves `go test ./internal/rules/vars_on_top/...` fully green. The existing `class C { static { foo(); var x; } }` case cannot distinguish the two values, because `foo();` is not directive-shaped to begin with. Upstream's lock-in case is the invalid `class C { static { 'use strict'; var x; } }`, which this PR does not migrate.

More broadly, the upstream test migration is partial: upstream has 22 valid / 27 invalid, `vars_on_top_upstream_test.go` has 19 / 11. All six upstream static-block valid cases, four of the five static-block invalid cases (including the `'use strict'` one, the `var x; foo(); var y;` case with its line 5 assertion, and both `if (foo) var x;` shapes), and `export * from 'foo'; var test = 1;` are missing. All of these behave correctly today, so this is regression protection rather than a live bug.

packages/rslint-test-tools/rstest.config.mts
```diff
@@ -135,6 +135,7 @@ export default defineConfig({
     './tests/eslint/rules/no-unused-private-class-members.test.ts',
     './tests/eslint/rules/no-use-before-define.test.ts',
     './tests/eslint/rules/no-var.test.ts',
+    './tests/eslint/rules/vars-on-top.test.ts',
```
elecmonkey: **[P3] Breaks the alphabetical order of this list**

The rest of this array is sorted alphabetically; `vars-on-top.test.ts` is inserted between `no-var` and `one-var`. It belongs at the end, around `use-isnan` / `valid-typeof`.

elecmonkey (commented):
Differential review notes: built the PR head and ran 71 cases over the same sources and the same config against upstream `eslint@10.8.0` + `@typescript-eslint/parser@8.65.0` (`vars-on-top` is marked `frozen: true` and matches the source on `eslint/main`). **68 matched exactly** on diagnostic count, `messageId`, and start/end line-column — covering class static blocks, `for` heads, `switch case`, comments/trivia adjacency, `import type`, `import x = require`, `using` / `await using`, decorators, `declare global`, function overloads, `export * from`, and non-BMP characters. Port quality is high overall: directive/import skipping, upstream's permissive `isVariableDeclaration` treatment of `let`/`const`/`export var`, `ast.IsFunctionLike` excluding static blocks, and the `NodeFlagsBlockScoped` filter for `let`/`const`/`using`/`await using` all line up with upstream branch by branch.

The 3 remaining divergences share one root cause, and there is one deliberately implemented branch with zero test coverage. Details in the inline comments.

internal/rules/vars_on_top/vars_on_top.go
swwind: Fixed in 5219fb033. `isAtTopOfAllowedScope` now handles `KindModuleBlock` only when the variable statement has an `export` modifier, mirroring upstream's `ExportNamedDeclaration` path. Added regressions for namespaces, dotted namespaces, ambient modules, and global augmentations, plus a negative case confirming that an unexported namespace `var` is still reported.

internal/rules/vars_on_top/vars_on_top.go
swwind: Fixed in 5219fb033. The upstream migration now contains all 34 valid and 27 invalid cases, including all static-block cases, the directive-shaped static-block prefix, the line-specific late declaration case, both nested/single-statement `if` forms, and `export *` before `var`. `go test ./internal/rules/vars_on_top` passes.

packages/rslint-test-tools/rstest.config.mts
swwind: Fixed in 5219fb033. Moved `vars-on-top.test.ts` to the end of the relevant core-rule section, immediately after `valid-typeof.test.ts`.

elecmonkey (approved):
fine

PR #1897, opened 2026-08-26
fix(typescript-eslint): implement unified-signatures rule

swwind:
## Summary

Port the `@typescript-eslint/unified-signatures` rule from ESLint to rslint.

Disallow overloads that can be combined into one signature with a union, optional parameter, or rest parameter.

## Related Links

- ESLint rule: https://typescript-eslint.io/rules/unified-signatures
- Source code: https://github.com/typescript-eslint/typescript-eslint/blob/v8.68.0/packages/eslint-plugin/src/rules/unified-signatures.ts

## Checklist


internal/plugins/typescript/rules/unified_signatures/unified_signatures.go
```diff
@@ -2,74 +2,546 @@ package unified_signatures
...
+func (a *overloadAnalyzer) overloadKey(node *ast.Node) (string, bool) {
+	switch node.Kind {
+	case ast.KindFunctionDeclaration:
+		if name := node.Name(); name != nil {
+			return "function:name:" + name.Text(), true
 		}
-	},
-})
+		if ast.HasSyntacticModifier(node, ast.ModifierFlagsDefault) {
+			return "function:export-default", true
```
elecmonkey: [P2] This keying prevents a named overload called `ExportDefaultDeclaration` from being grouped with an anonymous default-export overload, but upstream uses `node.id?.name ?? exportingNode?.type`, so both signatures share the key `ExportDefaultDeclaration`. I verified with `@typescript-eslint/eslint-plugin@8.68.0` that `declare function ExportDefaultDeclaration(x: string): void; export default function (x: number): void;` in a `.d.ts` reports `singleParameterDifference`, while this PR's Rslint build reports no `unified-signatures` diagnostic. If the rule is meant to be 1:1 compatible, the function overload keys need to match upstream's raw fallback behavior.

swwind: Fixed in `b7f70ea1`. Function declaration overload keys now mirror upstream's raw `node.id?.name ?? exportingNode?.type` behavior, so a named `ExportDefaultDeclaration` overload and an anonymous default-export overload are grouped together. Added the reported regression, with adjacent ordering, module-scope, multi-overload, scope-isolation, return-type, parameter-name, and JSDoc coverage in `1f3df530`.

internal/plugins/typescript/rules/unified_signatures/unified_signatures.go
```diff
@@ -2,74 +2,547 @@ package unified_signatures
...
+		return a.signaturesHaveSameAmountOfParameters(left, right)
+	}
+	return signaturesDifferByOptionalOrRestParameter(a.ctx.SourceFile, left, right)
+}
+
+func (a *overloadAnalyzer) signaturesCanBeUnified(left, right *ast.Node, outerTypeParameters []*ast.Node) bool {
+	leftParams, rightParams := left.Parameters(), right.Parameters()
+	if a.options.ignoreDifferentlyNamedParameters {
+		for i := range min(len(leftParams), len(rightParams)) {
+			if parametersHaveSameESTreeShape(leftParams[i], rightParams[i]) && staticParameterName(leftParams[i]) != staticParameterName(rightParams[i]) {
```
elecmonkey, line 246: [P2] With `ignoreDifferentlyNamedParameters: true`, this check compares the inner ts-go binding-name kinds after confirming both parameters are rest parameters. For valid TypeScript such as `method(...[value]: [string]): void; method(...values: [string]): void;`, this returns false because the names are an array binding pattern and an identifier, so the static-name inequality is skipped. `@typescript-eslint/eslint-plugin@8.68.0` emits no diagnostic because both ESTree parameter nodes are `RestElement` and their recursively extracted names differ (`undefined` versus `values`), while this PR emits `allParametersAreSame`; unwrap rest arguments when reproducing the upstream shape/name checks.

internal/plugins/typescript/rules/unified_signatures/unified_signatures.go
```diff
@@ -2,74 +2,547 @@ package unified_signatures
...
+		}
+		a.ctx.ReportRange(signatureReportRange(a.ctx.SourceFile, result.signature1), rule.RuleMessage{Id: "allParametersAreSame", Description: failureStringStart(otherLine) + " with identical parameters."})
+	}
+}
+
+func signatureReportRange(sourceFile *ast.SourceFile, signature *ast.Node) core.TextRange {
+	trimmed := utils.TrimNodeTextRange(sourceFile, signature)
+	if signature.Kind == ast.KindMethodDeclaration || signature.Kind == ast.KindConstructor {
+		if parameters := signature.ParameterList(); parameters != nil && parameters.Pos() > 0 {
+			return core.NewTextRange(parameters.Pos()-1, trimmed.End())
```
elecmonkey: [P2] `parameters.Pos()-1` does not reproduce the ESTree node reported upstream. For a class overload split as `method<T>\n(x: T): void`, upstream starts the diagnostic at `<T>` while Rslint starts at `(` on the following line; therefore an `eslint-disable-next-line` placed before `method<T>` suppresses upstream but not Rslint. Exported declarations have the inverse mismatch because the fallback full ts-go range starts at `export`, whereas upstream reports the inner declaration starting at `declare` or `function`. Compute the range corresponding to the upstream signature/value node rather than using the opening parenthesis or entire ts-go declaration.

internal/plugins/typescript/rules/unified_signatures/unified_signatures.go
```diff
@@ -2,74 +2,564 @@ package unified_signatures
...
+	trimmed := utils.TrimNodeTextRange(sourceFile, signature)
+	start := trimmed.Pos()
+	// ESTree reports the inner declaration, not its ExportNamedDeclaration
+	// wrapper. tsgo includes `export` in the declaration range.
+	for _, modifier := range signature.ModifierNodes() {
+		if modifier.Kind == ast.KindExportKeyword {
+			start = scanner.SkipTrivia(sourceFile.Text(), modifier.End())
+		}
+	}
+	if (signature.Kind == ast.KindMethodDeclaration || signature.Kind == ast.KindMethodSignature || signature.Kind == ast.KindConstructor) && signature.ParameterList() != nil {
```
elecmonkey: [P2] Preserve ESTree ranges for method signatures

`signatureReportRange` treats ts-go `KindMethodSignature` like a class method value and then overrides generic signatures with the first type-parameter node. For `interface I { f(x: string): void; f(x: string): void; }`, `@typescript-eslint/eslint-plugin@8.68.0` starts the second diagnostic at the method name, while Rslint starts at `(`; when the name and parameters are on separate lines, an `eslint-disable-next-line` before the name suppresses upstream but Rslint still reports on the parameter line. Generic class methods also start at `T` instead of `<`. Keep interface/type-literal signatures on their full node range and use the type-parameter-list opening token only for class method value ranges.

internal/plugins/typescript/rules/unified_signatures/unified_signatures.go
```diff
@@ -2,74 +2,564 @@ package unified_signatures
...
+	}
+}
+
+func signatureReportRange(sourceFile *ast.SourceFile, signature *ast.Node) core.TextRange {
+	trimmed := utils.TrimNodeTextRange(sourceFile, signature)
+	start := trimmed.Pos()
+	// ESTree reports the inner declaration, not its ExportNamedDeclaration
+	// wrapper. tsgo includes `export` in the declaration range.
+	for _, modifier := range signature.ModifierNodes() {
+		if modifier.Kind == ast.KindExportKeyword {
```
elecmonkey: [P2] Advance past `default` in anonymous exports

For an anonymous `export default function`, upstream reports the inner `TSDeclareFunction` beginning at `function`, but this loop advances only past `export`, leaving the Rslint range at `default`. With `export default` and `function` on separate lines, an `eslint-disable-next-line` before `export` suppresses Rslint while upstream still reports on the following `function` line. Skip the complete export wrapper, including `default`, when reconstructing the inner declaration range.

internal/plugins/typescript/rules/unified_signatures/unified_signatures.go
```diff
@@ -2,74 +2,564 @@ package unified_signatures
...
+		if !onlyTwo {
+			otherLine = lineOf(a.ctx.SourceFile, result.parameter0)
+		}
+		a.ctx.ReportNode(result.parameter1, rule.RuleMessage{
+			Id:          "singleParameterDifference",
+			Description: fmt.Sprintf("%s taking `%s`.", failureStringStart(otherLine), a.unifiedTypeText(parameterTypeAnnotation(result.parameter0), parameterTypeAnnotation(result.parameter1))),
+		})
+	case unifyExtraParameter:
+		if !onlyTwo {
+			otherLine = lineOf(a.ctx.SourceFile, result.otherSignature)
```
elecmonkey: [P2] Report the line of the compared ESTree signature

For groups of three or more overloads, `otherLine` comes from the full ts-go declaration rather than the ESTree signature node being compared. With a class overload written as `f` on line 2 and `(): void` on line 3, upstream says the other overload is on line 3 because `MethodDefinition.value` begins at `(`, while Rslint says line 2; exported declarations split after `export` have the same mismatch. Derive this line from the same ESTree-shaped signature range used for reporting, including the identical-signature branch below.

internal/plugins/typescript/rules/unified_signatures/unified_signatures.go
```diff
@@ -2,74 +2,564 @@ package unified_signatures
...
+		if left[i].Name().Text() != right[i].Name().Text() || !constraintsAreEqual(leftParam.Constraint, rightParam.Constraint) {
+			return false
+		}
+	}
+	return true
+}
+
+func constraintsAreEqual(left, right *ast.Node) bool {
+	left, right = skipParenthesizedType(left), skipParenthesizedType(right)
+	return left == right || left != nil && right != nil && left.Kind == right.Kind
```
elecmonkey: [P2] Distinguish `null` from literal constraints

Comparing raw ts-go kinds conflates a `null` constraint with string, numeric, and boolean literal constraints because ts-go wraps all of them in `KindLiteralType`, whereas typescript-estree emits `null` as `TSNullKeyword`. For `f<T extends null>(x: string)` and `f<T extends 'x'>(x: number)`, upstream 8.68.0 emits no diagnostic but this PR emits `singleParameterDifference`. Map constraints to their ESTree node kinds before comparing them.

internal/plugins/typescript/rules/unified_signatures/unified_signatures.go
swwind: Fixed in `83f89c8a2`, with additional range-boundary coverage in `4f7338f1e`. `MethodSignature` now preserves its full ESTree range from the method name, while generic class method values start at `<`.

internal/plugins/typescript/rules/unified_signatures/unified_signatures.go
swwind: Fixed in `83f89c8a2`, with additional export-wrapper coverage in `4f7338f1e`. `signatureReportRange` now advances past the complete `export default` wrapper for anonymous exports.

internal/plugins/typescript/rules/unified_signatures/unified_signatures.go
swwind: Fixed in `83f89c8a2`. Line-qualified messages now derive the compared overload's line from the same ESTree-shaped signature range used for reporting; `4f7338f1e` adds further range-boundary coverage.

internal/plugins/typescript/rules/unified_signatures/unified_signatures.go
swwind: Fixed in `83f89c8a2`. A `null` literal constraint now maps to ESTree's `TSNullKeyword` shape instead of sharing `LiteralType` with other literals.

internal/plugins/typescript/rules/unified_signatures/unified_signatures.go
```diff
@@ -2,74 +2,583 @@ package unified_signatures
...
+	}
+	return unifyResult{kind: unifyExtraParameter, extraParameter: longer[len(longer)-1], otherSignature: shorterSignature}, true
+}
+
+func (a *overloadAnalyzer) report(result unifyResult, onlyTwo bool) {
+	otherLine := 0
+	switch result.kind {
+	case unifySingleParameterDifference:
+		if !onlyTwo {
+			otherLine = lineOfRange(a.ctx.SourceFile, signatureReportRange(a.ctx.SourceFile, result.signature0))
```
elecmonkey: [P2] Use the differing parameter’s line in multi-overload messages

For groups with three or more overloads, upstream uses `p0.loc.start.line`, but this derives `otherLine` from the whole signature. With `declare function f(
  x: string): void; declare function f(x: number): void; declare function f(x: boolean): string;`, typescript-eslint 8.68.0 reports “the one on line 2”, while Rslint reports line 1. Use the first differing parameter’s range for `unifySingleParameterDifference`; signature ranges remain appropriate for the other result kinds.

internal/plugins/typescript/rules/unified_signatures/unified_signatures.go
```diff
@@ -2,74 +2,583 @@ package unified_signatures
...
+
+func failureStringStart(otherLine int) string {
+	if otherLine == 0 {
+		return "These overloads can be combined into one signature"
+	}
+	return fmt.Sprintf("This overload and the one on line %d can be combined into one signature", otherLine)
+}
+
+func (a *overloadAnalyzer) unifiedTypeText(left, right *ast.Node) string {
+	members := append(unionMembers(left), unionMembers(right)...)
```
elecmonkey, line 381: [P2] Preserve the annotated type text when its peer is untyped

Upstream special-cases a missing parameter annotation and returns the other type node’s exact source text without flattening or deduplicating it. For `declare function plain(value): void; declare function plain(value: string|number): void;`, typescript-eslint 8.68.0 says ``taking `string|number` ``, while Rslint says ``taking `string | number` ``; duplicate union members are likewise removed only by Rslint. Return the non-nil type through the single-member formatting path before constructing and deduplicating a combined union.

internal/plugins/typescript/rules/unified_signatures/unified_signatures.go
swwind: Fixed in `3d8677b33`. For groups of three or more overloads, `singleParameterDifference` now derives the referenced line from the first differing parameter, matching upstream's `p0.loc.start.line`. Added a multiline regression where the signature starts on line 1 and the parameter starts on line 2.

internal/plugins/typescript/rules/unified_signatures/unified_signatures.go
swwind: Fixed in `3d8677b33`. When exactly one parameter lacks a type annotation, the rule now preserves the annotated peer's exact authored type text before union flattening and deduplication. Added coverage for `string|number`.

internal/plugins/typescript/rules/unified_signatures/unified_signatures.go
```diff
@@ -2,74 +2,591 @@ package unified_signatures
...
+	if otherLine == 0 {
+		return "These overloads can be combined into one signature"
+	}
+	return fmt.Sprintf("This overload and the one on line %d can be combined into one signature", otherLine)
+}
+
+func (a *overloadAnalyzer) unifiedTypeText(left, right *ast.Node) string {
+	// typescript-eslint returns the sole annotated type node's authored text
+	// unchanged when its peer parameter has no annotation.
+	if left == nil {
```
elecmonkey, line 375: [P2] Format the sole annotated type like a union member

When one overload parameter is untyped, these early returns bypass the formatting applied by upstream's `getUnionMemberText`. For `declare function f(value): void; declare function f(value: () => void): void;`, `@typescript-eslint/eslint-plugin@8.68.0` reports ``taking `(() => void)` ``, while Rslint reports ``taking `() => void` ``; similarly, an authored `(string)` is normalized to `string` upstream but retained by Rslint. Pass the sole non-nil type through the same parenthesis-stripping and function/constructor/conditional-type wrapping path used for ordinary union members.

swwind: Fixed in `a96060e4a`. The single-annotation path now uses the same union-member formatter as the normal path, so transparent parenthesized types are removed and function, constructor, and conditional types receive the necessary parentheses. Added regressions for `() => void` → `(() => void)` and `(string)` → `string`.

internal/plugins/typescript/rules/unified_signatures/unified_signatures.go
```diff
@@ -2,74 +2,596 @@ package unified_signatures
...
+	}
+	if a.tokens == nil {
+		a.tokens = scanner.NewScanner()
+		a.tokens.SetText(a.ctx.SourceFile.Text())
+		a.tokens.SetLanguageVariant(a.ctx.SourceFile.LanguageVariant)
+	}
+	a.tokens.ResetTokenState(utils.TrimNodeTextRange(a.ctx.SourceFile, node).Pos())
+	var nameKind ast.Kind
+	var nameText string
+	for a.tokens.Scan(); a.tokens.Token() != ast.KindEndOfFile && a.tokens.TokenEnd() <= openParen; a.tokens.Scan() {
```
elecmonkey, line 172: [P2] Stop constructor key scanning before type parameters

For overloads such as `constructor<T>(x: string); constructor<T>(x: number);`, this loop scans through `<T>` before reaching `(`, leaving `nameKind` as `GreaterThanToken`; `constructorOverloadKey` then returns false and drops both declarations. `@typescript-eslint/eslint-plugin@8.68.0` reports `singleParameterDifference`, while Rslint reports nothing. Preserve the constructor token as the key when an optional type-parameter list follows it.

elecmonkey (commented):
I found two parity issues in the unified-signatures implementation.

internal/plugins/typescript/rules/unified_signatures/unified_signatures.go
```diff
@@ -2,74 +2,596 @@ package unified_signatures
...
+		return false
+	}
+	switch typeNode.Kind {
+	case ast.KindTypeReference:
+		typeName := typeNode.AsTypeReferenceNode().TypeName
+		if typeName.Kind == ast.KindIdentifier {
+			_, ok := names[typeName.Text()]
+			return ok
+		}
+	case ast.KindArrayType:
```
elecmonkey, line 594: [P2] Recurse through mapped-type value annotations

For valid code such as `interface I<T> { f(x: { [K in keyof T]: T }): void; f(x: string): void; }`, upstream emits no diagnostic because its recursive check follows `TSMappedType.typeAnnotation` and detects the outer `T`. This switch does not handle `KindMappedType`, so both signatures appear not to use the outer type parameter and Rslint incorrectly reports `singleParameterDifference`; the same false positive occurs when the mapped value is `T[]`.

elecmonkey:
conflict

internal/plugins/typescript/rules/unified_signatures/unified_signatures.go
swwind: Fixed in `bd7a858a3`. Constructor overload key scanning now stops after the constructor name, so a following `<T>` cannot replace the key with `GreaterThanToken`. Added a regression covering generic constructor overloads.

internal/plugins/typescript/rules/unified_signatures/unified_signatures.go
swwind: Fixed in `bd7a858a3`. Outer-type-parameter detection now recurses through mapped-type value annotations, including nested array element types. Added regressions for `{ [K in keyof T]: T }` and `{ [K in keyof T]: T[] }`.

elecmonkey (approved):

PR #1898, opened 2026-08-26
feat: port rule grouped-accessor-pairs

swwind:
## Summary

Port the `grouped-accessor-pairs` rule from ESLint to rslint.

Require getter and setter definitions for the same property to be adjacent, with options for accessor order and TypeScript type members.

## Related Links

- ESLint rule: https://eslint.org/docs/latest/rules/grouped-accessor-pairs
- Source code: https://github.com/eslint/eslint/blob/v10.9.1/lib/rules/grouped-accessor-pairs.js

## Checklist


internal/rules/accessorutil/accessor_key.go
```diff
@@ -0,0 +1,63 @@
...
+		return false
+	}
+}
+
+// computedKeysEqual mirrors ESLint's token-list comparison. HasSameTokens
+// removes only transparent outer parentheses, preserves nested parentheses,
+// and compares punctuation and operators that tsgo does not expose as child
+// nodes.
+func computedKeysEqual(sourceFile *ast.SourceFile, left *ast.Node, right *ast.Node) bool {
+	return utils.HasSameTokens(sourceFile, left, right)
```
elecmonkey, line 62: [P2] Normalize identifier escapes when comparing computed keys

This uses `HasSameTokens` for dynamic accessor keys, but that helper compares identifier leaves by raw source text, while ESLint's `sourceCode.getTokens(node.key)` compares token `value`. For `({ get [foo](){}, middle: true, set [\u0066oo](value){} })`, ESLint v10.9.1 reports `notGrouped` because both identifier tokens have value `foo`; a temporary Rslint rule-tester case against this PR reported 0 diagnostics, so escaped identifiers make the port miss accessor pairs.

elecmonkey (commented):
One issue.

internal/rules/accessorutil/accessor_key.go
swwind: Fixed. `HasSameTokens` now compares identifier leaves by ts-go's decoded `Identifier.Text`, while keeping raw source comparison for literal tokens. Added regression coverage for `[foo]` and `[\u0066oo]` in `grouped-accessor-pairs`, plus shared helper coverage. The affected rule and all existing `HasSameTokens` consumer packages pass.

internal/utils/ts_eslint.go
```diff
@@ -2588,9 +2589,14 @@ func hasSameTokens(sf *ast.SourceFile, a, b *ast.Node) bool {
 	if a.Kind != b.Kind {
 		return false
 	}
+	// ESLint identifier tokens expose their decoded value, so `foo` and
+	// `\u0066oo` are the same token even though their source text differs.
+	if a.Kind == ast.KindIdentifier {
```
elecmonkey: **[P2] Decode private identifier escapes when comparing token values**

When a dynamic key contains an escaped private identifier, e.g. `class C { #foo; get [this.#foo](){} middle(){} set [this.#\u0066oo](v){} }`, ESLint 10.9.1 reports `notGrouped` at columns 48–68, while this PR reports nothing. `SourceCode#getTokens()` decodes both `PrivateIdentifier` tokens to `foo`, but `hasSameTokens` only normalizes `ast.KindIdentifier`, leaving private identifiers on raw-source comparison. Please compare `ast.KindPrivateIdentifier` via `AsPrivateIdentifier().Text` as well; I verified that this makes the 40-case differential matrix pass and restores `no-self-compare` parity for the same token shape.

swwind: fixed

fansenze (approved):

PR #1899, opened 2026-08-26
fix(typescript-eslint): align consistent-type-imports rule

swwind:
## Summary

Port the `typescript-eslint/consistent-type-imports` rule to rslint.

Enforce consistent type-only import syntax, including reference classification, all upstream options, decorator metadata safeguards, JSX factory handling, and autofixes for default, named, namespace, inline, and split imports.

## Related Links

- ESLint rule: https://typescript-eslint.io/rules/consistent-type-imports
- Source code: https://github.com/typescript-eslint/typescript-eslint/blob/v8.67.0/packages/eslint-plugin/src/rules/consistent-type-imports.ts

## Checklist

- [x] Tests updated.
- [x] Documentation updated.


internal/plugins/typescript/rules/consistent_type_imports/consistent_type_imports.go
```diff
@@ -2,103 +2,687 @@ package consistent_type_imports
...
+	if defaultSpecifier != nil && containsSpecifier(report.typeSpecifiers, defaultSpecifier) {
+		if len(report.typeSpecifiers) == len(report.allSpecifiers) {
+			fixes = append(fixes, insertTypeAfterImportFix(ctx, report.node))
+		} else {
+			sourceText := utils.TrimmedNodeText(ctx.SourceFile, report.node.AsImportDeclaration().ModuleSpecifier)
+			position := utils.TrimNodeTextRange(ctx.SourceFile, report.node).Pos()
+			defaultRange := utils.TrimNodeTextRange(ctx.SourceFile, defaultSpecifier.name)
+			comma, ok := utils.TokenAtOrAfter(ctx.SourceFile, defaultRange.End())
+			if ok && comma.Kind == ast.KindCommaToken {
+				defaultText := strings.TrimSpace(ctx.SourceFile.Text()[defaultRange.Pos():comma.Start])
```
elecmonkey: [P1] Use `ecmascript.StringTrim` instead of `strings.TrimSpace`

`golangci-lint`'s `forbidigo` rule bans `strings.TrimSpace` under `internal/plugins/**`, so `npm run lint:go` (run by `.github/workflows/ci.yml`) fails on this line: ``use of `strings.TrimSpace` forbidden ... Use ecmascript.StringTrim``. It is also a behavior difference, because JavaScript's `String.prototype.trim` strips U+FEFF and `strings.TrimSpace` does not.

Repro: `import Def\uFEFF, { A } from 'foo';` + `A();` + `type T = Def;`
Upstream (@typescript-eslint/eslint-plugin@8.67.0): `import type Def from 'foo';`
Rslint: `import type Def\uFEFF from 'foo';`

internal/plugins/typescript/rules/consistent_type_imports/consistent_type_imports.go
```diff
@@ -2,103 +2,687 @@ package consistent_type_imports
...
+		for _, imports := range orderedSources {
+			for _, report := range imports.reports {
+				applyImplicitJSXReferences(ctx, report, hasJSX, hasJSXFragment)
+				if len(report.typeSpecifiers) == 0 {
+					continue
+				}
+				allUsedImportsAreTypes := len(report.valueSpecifiers) == 0 && len(report.unusedSpecifiers) == 0
+				if allUsedImportsAreTypes {
+					declaration := report.node.AsImportDeclaration()
+					if declaration.Attributes != nil {
```
elecmonkey: [P2] Empty import attributes suppress the diagnostic entirely

Upstream checks `report.node.attributes.length === 0`; this checks `Attributes != nil`. For `with {}` tsgo produces a non-nil `Attributes` with zero elements, so both the report and the fix are skipped. `consistent_type_imports.go:454` has the same check for the namespace branch. Non-empty attributes (`with { type: 'json' }`) agree on both sides.

Repro, default options — upstream reports `typeOverValue` and fixes, Rslint reports nothing, for all three of:
`import Foo from 'foo' with {};` + `type T = Foo;`
`import * as N from 'foo' with {};` + `type T = N.Foo;`
`import { A } from 'foo' with {};` + `type T = A;`

internal/plugins/typescript/rules/consistent_type_imports/consistent_type_imports.go
```diff
@@ -2,103 +2,687 @@ package consistent_type_imports
...
+		}
+		return insertTopLevelTypeFixes(ctx, report.node, false)
+	}
+
+	if defaultSpecifier != nil {
+		if containsSpecifier(report.typeSpecifiers, defaultSpecifier) && len(namedSpecifiers) == 0 && namespaceSpecifier == nil {
+			return insertTopLevelTypeFixes(ctx, report.node, true)
+		}
+		if fixStyle == "inline-type-imports" && !containsSpecifier(report.typeSpecifiers, defaultSpecifier) && len(namedSpecifiers) > 0 && namespaceSpecifier == nil {
+			return insertInlineTypeFixes(ctx, namedTypeSpecifiers(namedSpecifiers, report.typeSpecifiers))
```
elecmonkey, line 496: [P2] `inline-type-imports` emits a fix upstream does not produce

Upstream's `fixInlineTypeImportDeclaration` only yields the `type` modifier when `sourceImports.valueImport` is set *and* that declaration itself has named specifiers (or `valueOnlyNamedImport` is set). This port has no `valueImport`/`valueOnlyNamedImport` state, so it always emits.

Repro with `{ "fixStyle": "inline-type-imports" }`:

```ts
import Def from 'foo';
import Def2, { B } from 'foo';
const a = Def;
const b = Def2;
type T = B;
```

Upstream reports `someImportsAreOnlyTypes` and produces **no** fix, because `valueImport` is pinned to the first declaration, which has no named specifiers. Rslint reports the same diagnostic and rewrites to `import Def2, { type B } from 'foo';`. The same shape occurs with a leading `import Def, * as N from 'foo';`.

Rslint's output is valid and arguably better, so this may be worth keeping — but it is currently undocumented. Either add the gate or list it under "Differences from ESLint" in `consistent_type_imports.md`.

packages/rslint-test-tools/tests/typescript-eslint/tsconfig.files.json
```diff
@@ -0,0 +1,4 @@
+{
+  "extends": "./tsconfig.virtual.json",
+  "include": ["fixtures/**/*.ts", "fixtures/**/*.tsx"]
```
elecmonkey: [P3] This file appears to be unused

The new `packages/rslint-test-tools/tests/typescript-eslint/rslint.config.mjs` references only `./tsconfig.virtual.json`. The other `tsconfig.files.json` references in the repo are each directory's own copy (`tests/rstest/`, `tests/eslint-plugin-import/`, ...), so nothing resolves to this one.

elecmonkey (commented):
Reviewed the port against `@typescript-eslint/eslint-plugin@8.67.0` by running both implementations over the same sources: all 381 cases extracted from the upstream `consistent-type-imports` test file at v8.67.0 (all three `experimentalDecorators`/`emitDecoratorMetadata` combinations plus the `jsxPragma`/`jsxFragmentName` cases), comparing messageId, message text, start/end line/column, and the multi-pass autofix result. Those 381 cases match exactly on every field. Another 85 hand-written cases (import attributes, comments, non-BMP characters, malformed input, cross-declaration merging, each `inline-type-imports` branch) surfaced the three divergences below. No panics on syntactically broken input.

One note on coverage: `filterSnapshot` in `packages/rule-tester/src/index.ts` overwrites the response's `output` with the test case's *expected* output and deletes `fixes`, so `packages/rslint-test-tools/tests/typescript-eslint/rules/__snapshots__/consistent-type-imports.test.ts.snap` does not verify autofix at all. The fix-parity result above comes from a separate side-by-side run, not from that snapshot.

internal/plugins/typescript/rules/consistent_type_imports/consistent_type_imports.go
swwind: The implementation already uses `ecmascript.StringTrim`; I added focused U+FEFF regression coverage to lock in the ECMAScript trimming behavior. The case now produces `import type Def from 'foo';`.

internal/plugins/typescript/rules/consistent_type_imports/consistent_type_imports.go
swwind: Fixed in `14dcb3e26`. Import attributes now block the diagnostic and fix only when the attribute list is non-empty. I added focused default, namespace, and named-import cases for `with {}`; the existing non-empty-attributes case remains valid.

internal/plugins/typescript/rules/consistent_type_imports/consistent_type_imports.go
swwind: Kept as a deliberate rslint divergence in `14dcb3e26`. The generated inline fix is valid, so I added focused regression coverage and documented the behavior under “Differences from ESLint”.

packages/rslint-test-tools/tests/typescript-eslint/tsconfig.files.json
swwind: Removed the unused `packages/rslint-test-tools/tests/typescript-eslint/tsconfig.files.json` in `14dcb3e26`.

internal/plugins/typescript/rules/consistent_type_imports/consistent_type_imports.go
```diff
@@ -2,103 +2,696 @@ package consistent_type_imports
...
+
+	bySource := map[string]*sourceImports{}
+	orderedSources := []*sourceImports{}
+	hasDecorator := false
+	hasJSX := false
+	hasJSXFragment := false
+	guardDecoratorMetadata := false
+	if program := ctx.Program(); program != nil && program.Options() != nil {
+		compilerOptions := program.Options()
+		guardDecoratorMetadata = compilerOptions.ExperimentalDecorators.IsTrue() && compilerOptions.EmitDecoratorMetadata.IsTrue()
```
elecmonkey, line 175: [P1] Read decorator metadata flags from effective parser options

When `experimentalDecorators` and `emitDecoratorMetadata` are enabled through `languageOptions.parserOptions` without a matching tsconfig, this guard reads both flags as false from `Program.Options()`. For `import Foo from 'foo'; @deco class A { constructor(foo: Foo) {} }`, upstream 8.67.0 emits no diagnostic, while Rslint reports `typeOverValue` and rewrites the import to `import type`, defeating the runtime-metadata safeguard. The migrated tests mask this path by adding dedicated tsconfig files; the rule needs the effective parser-service flags as well.

internal/plugins/typescript/rules/consistent_type_imports/consistent_type_imports.go
```diff
@@ -2,103 +2,696 @@ package consistent_type_imports
...
+		return
+	}
+	options := ctx.Program().Options()
+	factory := "React"
+	if options.JsxFactory != "" {
+		factory = typescriptutil.JSXFactoryRoot(options.JsxFactory)
+	}
+	fragmentFactory := ""
+	if options.JsxFragmentFactory != "" {
+		fragmentFactory = typescriptutil.JSXFactoryRoot(options.JsxFragmentFactory)
```
elecmonkey, line 281: [P1] Honor custom JSX pragmas configured outside tsconfig

This derives the implicit JSX binding only from TypeScript compiler options, ignoring `languageOptions.parserOptions.jsxPragma` and `jsxFragmentName`. With `jsxPragma: 'h'` and `import h from 'lib'; type T = h.Node; const C = () => <div />;`, upstream emits no diagnostic, but Rslint reports `typeOverValue` and fixes the runtime factory import to `import type h`. The migrated cases hide the mismatch by adding tsconfigs with `jsxFactory`/`jsxFragmentFactory`; the native rule must receive the effective parser options or preserve the import.

internal/plugins/typescript/rules/consistent_type_imports/consistent_type_imports.go
```diff
@@ -2,103 +2,696 @@ package consistent_type_imports
...
+	}
+	remaining := report.typeSpecifiers[:0]
+	for _, specifier := range report.typeSpecifiers {
+		used := hasJSX && specifier.name.Text() == factory
+		used = used || hasFragment && fragmentFactory != "" && specifier.name.Text() == fragmentFactory
```
elecmonkey: [P2] Resolve implicit JSX references in the JSX scope

The file-wide `hasJSX` flag marks every same-named import as a value use without checking which lexical binding the JSX expression resolves to. For `import React from 'react'; type T = React.FC; function f(React: unknown) { return <div />; }`, upstream resolves the implicit JSX reference to the parameter, reports `typeOverValue` on the import, and fixes it to `import type`; Rslint emits nothing in both tsconfig-backed and source-only runs. Track the JSX site and resolve the factory binding in that scope before reclassifying the import.

internal/plugins/typescript/rules/consistent_type_imports/consistent_type_imports.go
swwind: Not changed in this patch. `RuleContext` currently exposes the compiler options but not the effective parser-service decorator flags. Per the scope of this fix, the existing framework limitation is left unchanged.

internal/plugins/typescript/rules/consistent_type_imports/consistent_type_imports.go
swwind: Not changed in this patch. The native rule currently receives JSX factory settings from compiler options, but not `jsxPragma` or `jsxFragmentName` authored only in parser options. Per the scope of this fix, the existing framework limitation is left unchanged.

internal/plugins/typescript/rules/consistent_type_imports/consistent_type_imports.go
swwind: Fixed in `fb066f2c4`. The rule now records individual JSX and fragment sites and only treats the configured factory import as a value use when at least one site reaches it without lexical shadowing. Added focused coverage for React element/fragment parameter shadowing and a tsconfig-defined preact factory. `go test ./internal/plugins/typescript/rules/consistent_type_imports/` passes.

elecmonkey (approved):
👍

PR #1901, opened 2026-08-26
docs: rewrite `rstest/` rules' documents

elecmonkey:
## Summary

Rewrites all 23 rule pages of the `rstest` plugin so each page describes the rule on its own terms.

## Changes

- **Rule Details** now states the behavior a user needs to know, including things that were previously only implied by an example: `valid-title`'s whitespace, non-string-title and duplicate-prefix checks, and the syntax forms `no-conditional-in-test` reports (`if`, `switch`, conditional expressions, `&&` / `||` / `??`, and optional chaining under `allowOptionalChaining`).
- **Examples** show one incorrect and one correct snippet that match what the rule is for, rather than edge cases. `valid-title` no longer leads with an invalid `.each` format specifier.
- **Options** sections follow one format everywhere: a multi-line JSON example that overrides the defaults, and a table with the option's type, default and a one-line description. Shared options such as `additionalTestBlockFunctions` are described identically across pages.
- **Autofix / Suggestions** sections state when a report is fixed and when it is only reported, instead of hedging with "supported" or "when behavior is preserved". `prefer-todo`, `prefer-called-exactly-once-with`, `prefer-expect-type-of` and `valid-expect` all name their real conditions.

No rule behavior changes; documentation only.

### DIFF
--- a/internal/plugins/rstest/rules/expect_expect/expect_expect.md
+++ b/internal/plugins/rstest/rules/expect_expect/expect_expect.md
@@ -2,94 +2,43 @@
 
 ## Rule Details
 
-Enforce that every Rstest test registration contains at least one assertion. A
-test with no assertion can pass without verifying anything, hiding the fact that
-its intent was never checked.
+Requires every Rstest `test` or `it` callback to contain an assertion. This prevents tests from passing after exercising code without verifying its result.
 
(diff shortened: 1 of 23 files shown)


swwind (approved):
Looks like no behavior changes.

PR #1902, opened 2026-08-26
fix(typescript-eslint): align prefer-readonly-parameter-types rule

swwind:
## Summary

Port the `@typescript-eslint/prefer-readonly-parameter-types` rule from typescript-eslint to rslint.

Require function parameters to use deeply readonly types, including nested objects, arrays, tuples, unions, intersections, conditional types, recursive types, and configurable allowlists.

## Related Links

- ESLint rule: https://typescript-eslint.io/rules/prefer-readonly-parameter-types
- Source code: https://github.com/typescript-eslint/typescript-eslint/blob/v8.68.0/packages/eslint-plugin/src/rules/prefer-readonly-parameter-types.ts

## Checklist

- [x] Tests updated.
- [x] Documentation updated.


internal/plugins/typescript/typescriptutil/readonly.go
```diff
@@ -0,0 +1,319 @@
...
+			return readonlynessMutable
+		}
+	}
+
+	for _, property := range properties {
+		// tsgo materializes the array's well-known symbol property as a mapped
+		// object whose inherited method keys appear mutable. Upstream treats the same
+		// well-known-symbol member as readonly; it cannot be assigned through a
+		// normal property name, so keep it out of the deep value walk.
+		if isWellKnownSymbolName(property.Name) {
```
elecmonkey: [P2] Skip private members during the deep-value pass

For `class State { #items: string[] = [] } function consume(value: State) {}`, `@typescript-eslint/eslint-plugin@8.68.0` emits no diagnostic because private identifiers are exempt, while Rslint reports `value`. The first property loop skips `#items`, but this second loop recursively checks its mutable array type again; skip private properties here as well.

internal/plugins/typescript/typescriptutil/readonly.go
```diff
@@ -0,0 +1,319 @@
...
+		}
+		if _, ok := seen[propertyType]; ok {
+			continue
+		}
+		if isTypeReadonly(typeChecker, propertyType, options, sourceProgram, seen) == readonlynessMutable {
+			return readonlynessMutable
+		}
+	}
+
+	for _, indexInfo := range typeChecker.GetIndexInfosOfType(t) {
```
elecmonkey, line 209: [P2] Restrict index checks to string and number keys

For `type T = { [key: symbol]: string }; function f(value: T) {}`, upstream 8.68.0 emits no diagnostic while Rslint reports one; the same mismatch occurs for symbol/template-literal index signatures with mutable values. Upstream explicitly queries only the string and number index signatures, whereas this loop checks every index info returned by ts-go, so it should filter or query those two key types only.

internal/plugins/typescript/typescriptutil/readonly.go
```diff
@@ -0,0 +1,319 @@
...
+func isReadonlyObject(
+	typeChecker *checker.Checker,
+	t *checker.Type,
+	options ReadonlynessOptions,
+	sourceProgram *program.Program,
+	seen map[*checker.Type]struct{},
+) readonlyness {
+	properties := typeChecker.GetPropertiesOfType(t)
+	for _, property := range properties {
+		if options.TreatMethodsAsReadonly && property.Flags&ast.SymbolFlagsMethod != 0 {
```
elecmonkey: [P2] Detect methods behind mapped property symbols

With `treatMethodsAsReadonly: true`, `interface S { method(): void } type M<T> = { [K in keyof T]: T[K] }; function f(value: M<S>) {}` is accepted upstream but reported by Rslint. A mapped property's synthetic symbol does not retain `SymbolFlagsMethod`; upstream also checks the value declaration and final declaration symbols, so relying only on `property.Flags` loses the option's behavior for mapped types.

internal/plugins/typescript/typescriptutil/readonly.go
```diff
@@ -0,0 +1,319 @@
...
+			}
+		}
+		return false
+	}
+
+	// Mapped types can synthesize transient property symbols whose declarations
+	// point back to the original (mutable) member. Read the mapped declaration's
+	// readonly modifier before falling back to those declarations. As upstream
+	// does, leave well-known symbol properties to their original declarations.
+	if checker.Type_objectFlags(t)&checker.ObjectFlagsMapped != 0 && !isWellKnownSymbolName(name) {
```
elecmonkey: [P2] Apply mapped readonly modifiers to unique-symbol keys

For `declare const tag: unique symbol; interface S { readonly [tag]: string } type M<T> = { -readonly [K in keyof T]: T[K] }; function f(value: M<S>) {}`, upstream reports the mutable mapped property but Rslint emits nothing. The `isWellKnownSymbolName` guard classifies ts-go's custom unique-symbol name (which contains a second `@`) as well-known, bypassing the mapped `-readonly` modifier and falling back to the original readonly declaration.

internal/plugins/typescript/typescriptutil/readonly.go
```diff
@@ -0,0 +1,319 @@
...
+	if property.CheckFlags&ast.CheckFlagsReadonly != 0 || property.Flags&ast.SymbolFlagsValueModule != 0 {
+		return true
+	}
+	if property.Flags&ast.SymbolFlagsGetAccessor != 0 && property.Flags&ast.SymbolFlagsSetAccessor == 0 {
+		return true
+	}
+	if property.Flags&ast.SymbolFlagsEnumMember != 0 {
+		return true
+	}
+	for _, declaration := range property.Declarations {
```
elecmonkey, line 305: [P2] Honor all upstream readonly declaration forms

For `namespace Constants { export const version = 1 } function consume(value: typeof Constants) {}`, upstream 8.68.0 emits no diagnostic while Rslint reports one; a JS property created by `Object.defineProperty(..., { writable: false })` has the same mismatch. Upstream's `symbolHasReadonlyDeclaration` recognizes const variable declarations and readonly assignment declarations, but this fallback only recognizes explicit `readonly` modifiers, so those declaration forms need equivalent handling.

internal/plugins/typescript/typescriptutil/readonly.go
swwind: Fixed in `4bb2ae0f`. The deep-value pass now skips private properties just like the initial readonly-property pass. I added a focused regression for the reported `State.#items` case.

internal/plugins/typescript/typescriptutil/readonly.go
swwind: Fixed in `4bb2ae0f`. Index traversal now considers only primitive string and number key types, matching upstream’s two index-kind queries. Focused regressions cover both symbol and template-literal index signatures.

internal/plugins/typescript/typescriptutil/readonly.go
swwind: Fixed in `4bb2ae0f`. Method detection now checks the property symbol as well as the value declaration and final declaration symbols, preserving `treatMethodsAsReadonly` for mapped properties. A focused mapped-method regression was added.

internal/plugins/typescript/typescriptutil/readonly.go
swwind: Fixed in `4bb2ae0f`. The well-known-symbol check now distinguishes ts-go’s suffixed standard symbol names from custom unique-symbol names, so mapped `-readonly` modifiers apply to the latter without regressing `ReadonlyArray` and other standard symbol members. A focused unique-symbol mapped-type regression was added.

internal/plugins/typescript/typescriptutil/readonly.go
swwind: Fixed in `4bb2ae0f`. Readonly declaration detection now recognizes const variable declarations and bindable `Object.defineProperty` assignments with readonly descriptors. Focused regressions cover both the namespace const property and checked-JavaScript assignment forms.

internal/plugins/typescript/rules/prefer_readonly_parameter_types/prefer_readonly_parameter_types.go
```diff
@@ -118,15 +65,27 @@ func checkParameter(ctx rule.RuleContext, param *ast.Node, opts PreferReadonlyPa
 		return
 	}
 
-	// Get the type of the parameter
+	// Reading an authored annotation through GetTypeFromTypeNode preserves its
+	// alias symbol, which is required for allowlist matching around unions.
 	paramType := ctx.TypeChecker.GetTypeAtLocation(param)
+	if paramDecl.Type != nil {
```
elecmonkey, line 73: [P2] Match ignoreInferredTypes to AssignmentPattern

When `ignoreInferredTypes: true` is used with a typed default parameter such as `function f(value: string[] = []) {}`, upstream 8.68.0 emits no diagnostic because it checks whether the ESTree `AssignmentPattern` itself has a type annotation. This implementation instead reads `ParameterDeclaration.Type`, so Rslint still reports it. Please preserve upstream’s actual-parameter-node semantics so this option does not produce extra diagnostics.

internal/plugins/typescript/typescriptutil/readonly.go
```diff
@@ -0,0 +1,382 @@
...
+			return readonlynessReadonly
+		}
+		if result := isReadonlyObject(typeChecker, t, options, sourceProgram, seen); result != readonlynessUnknown {
+			return result
+		}
+	}
+
+	if flags&checker.TypeFlagsConditional != 0 {
+		for _, part := range []*checker.Type{
+			typeChecker.GetTrueTypeOfConditionalType(t),
```
elecmonkey: [P2] Inspect uninstantiated conditional branches

For `type C<T, U> = T extends string ? U : number; function f<T>(value: C<T, string[]>) {}`, upstream 8.68.0 emits no diagnostic while Rslint reports one. Upstream reads the uninstantiated true and false type nodes from the conditional root, so it sees the type parameter `U`; these ts-go APIs apply the mapper and turn `U` into mutable `string[]`. Please inspect the uninstantiated branch types to preserve upstream behavior.

internal/plugins/typescript/typescriptutil/readonly.go
```diff
@@ -0,0 +1,382 @@
...
+			continue
+		}
+		// tsgo materializes the array's well-known symbol property as a mapped
+		// object whose inherited method keys appear mutable. Upstream treats the same
+		// well-known-symbol member as readonly; it cannot be assigned through a
+		// normal property name, so keep it out of the deep value walk.
+		if isWellKnownSymbolName(property.Name) {
+			continue
+		}
+		propertyType := typeChecker.GetTypeOfPropertyOfType(t, property.Name)
```
elecmonkey, line 197: [P2] Read computed-property types by symbol

For `declare const tag: unique symbol; interface S { readonly [tag]: string[] } function f(value: S) {}`, upstream 8.68.0 emits no diagnostic while Rslint reports one. `GetTypeOfPropertyOfType` cannot resolve ts-go’s internal computed-symbol name as a normal string and returns nil, which this code treats as mutable. Upstream resolves the escaped property symbol and reads its declared type, so this path needs an equivalent symbol-aware fallback.

internal/plugins/typescript/typescriptutil/readonly.go
```diff
@@ -0,0 +1,382 @@
...
+
+	// Mapped types can synthesize transient property symbols whose declarations
+	// point back to the original (mutable) member. Read the mapped declaration's
+	// readonly modifier before falling back to those declarations. As upstream
+	// does, leave well-known symbol properties to their original declarations.
+	if checker.Type_objectFlags(t)&checker.ObjectFlagsMapped != 0 && !isWellKnownSymbolName(name) {
+		if symbol := checker.Type_symbol(t); symbol != nil && len(symbol.Declarations) > 0 {
+			declaration := symbol.Declarations[0]
+			if declaration.Kind == ast.KindMappedType {
+				readonlyToken := declaration.AsMappedTypeNode().ReadonlyToken
```
elecmonkey: [P2] Do not inherit readonly through key remapping

For `interface S { readonly item: string } type M<T> = { [K in keyof T as `get${Capitalize<K & string>}`]: T[K] }; function f(value: M<S>) {}`, upstream 8.68.0 reports a diagnostic while Rslint does not. This mapped type has no readonly token, but the fallback inherits readonly from the original `item` declaration even though the property was remapped to `getItem`. Key-remapped properties must not automatically inherit the source property’s modifier.

internal/plugins/typescript/typescriptutil/readonly.go
```diff
@@ -0,0 +1,382 @@
...
+			}
+		}
+		return false
+	}
+
+	// Mapped types can synthesize transient property symbols whose declarations
+	// point back to the original (mutable) member. Read the mapped declaration's
+	// readonly modifier before falling back to those declarations. As upstream
+	// does, leave well-known symbol properties to their original declarations.
+	if checker.Type_objectFlags(t)&checker.ObjectFlagsMapped != 0 && !isWellKnownSymbolName(name) {
```
elecmonkey: [P2] Apply mapped readonly modifiers to well-known symbol keys

For `interface S { readonly [Symbol.iterator]: string } type M<T> = { -readonly [K in keyof T]: T[K] }; function f(value: M<S>) {}`, upstream 8.68.0 reports a diagnostic while Rslint does not. The `isWellKnownSymbolName` guard bypasses mapped-type handling for `Symbol.iterator`, so the explicit `-readonly` modifier is ignored and the original readonly declaration wins. Mapped readonly modifiers must also apply to well-known symbol keys.

internal/plugins/typescript/rules/prefer_readonly_parameter_types/prefer_readonly_parameter_types.go
swwind: Fixed in `96fba929`. `ignoreInferredTypes` now checks the ESTree-equivalent actual parameter shape: any defaulted parameter maps to an `AssignmentPattern`, which has no direct type annotation. Added the reported regression, with adjacent destructuring, parameter-property, and typed-rest coverage in `68646313`.

internal/plugins/typescript/typescriptutil/readonly.go
swwind: Fixed in `96fba929`. Conditional readonly analysis now reads the true and false type nodes from the uninstantiated conditional root through generated shim accessors, instead of ts-go's mapper-applied resolved branch APIs. Added the reported regression, with inline, nested, and mutable-branch coverage in `68646313`.

internal/plugins/typescript/typescriptutil/readonly.go
swwind: Fixed in `96fba929`. The computed-symbol path now matches upstream's effective declared-type result instead of treating ts-go's failed string lookup as mutable, while still checking the property's readonly modifier separately. Added unique-symbol and well-known-symbol coverage, including a mutable-property counterexample.

internal/plugins/typescript/typescriptutil/readonly.go
swwind: Fixed in `96fba929`. A mapped type with key remapping and no readonly token now treats the remapped property as mutable instead of inheriting the source property's modifier. Additional tests cover ordinary homomorphic inheritance and explicit `+readonly` remapping.

internal/plugins/typescript/typescriptutil/readonly.go
swwind: Fixed in `96fba929`. Explicit mapped readonly modifiers are now applied before falling back to source declarations for every key spelling, including well-known symbol keys. Added the reported `Symbol.iterator` regression and adjacent unique-symbol modifier coverage.

internal/plugins/typescript/typescriptutil/readonly.go
```diff
@@ -0,0 +1,373 @@
...
+		if symbol := checker.Type_symbol(t); symbol != nil && len(symbol.Declarations) > 0 {
+			declaration := symbol.Declarations[0]
+			if declaration.Kind == ast.KindMappedType {
+				mapped := declaration.AsMappedTypeNode()
+				readonlyToken := mapped.ReadonlyToken
+				if readonlyToken != nil {
+					return readonlyToken.Kind != ast.KindMinusToken
+				}
+				if mapped.NameType != nil {
+					return false
```
elecmonkey, line 286: [P2] Preserve inherited readonly in filtering mapped types

For `interface S { readonly x: string } type I<T> = { [K in keyof T as K]: T[K] }; function f(value: I<S>) {}`, `@typescript-eslint/eslint-plugin@8.68.0` emits no diagnostic while Rslint reports `value`. A non-nil `NameType` may represent identity or conditional filtering rather than key remapping, and those forms retain source-property modifiers. Distinguish ts-go's filtering and remapping mapped-type kinds before returning mutable here.

internal/plugins/typescript/rules/prefer_readonly_parameter_types/prefer_readonly_parameter_types.go
```diff
@@ -45,88 +47,47 @@ func parseOptions(options []any) PreferReadonlyParameterTypesOptions {
...
 
-	// Check if the parameter type is readonly
-	if !isReadonlyType(paramType, opts) {
-		ctx.ReportNode(param, buildShouldBeReadonlyMessage())
+	readonly := typescriptutil.IsTypeReadonly(ctx.TypeChecker, paramType, typescriptutil.ReadonlynessOptions{
+		Allow:                  opts.Allow,
+		TreatMethodsAsReadonly: opts.TreatMethodsAsReadonly,
+	}, ctx.Program())
+	if !readonly && !typescriptutil.IsTypeBrandedLiteralLike(ctx.TypeChecker, paramType) {
+		start := param.Pos()
```
elecmonkey: [P2] Exclude parameter decorators from the diagnostic range

For `class Consumer { method(@dec value: string[]) {} }`, upstream 8.68.0 reports columns 30-45, covering `value: string[]`, while Rslint reports columns 25-45 and includes `@dec`. `param.Pos()` begins before the decorator in ts-go, whereas upstream reports the unwrapped ESTree parameter node; derive the start from the actual parameter name or binding pattern so decorated parameters have matching locations.

internal/plugins/typescript/typescriptutil/readonly.go
swwind: Fixed in `292a3ae1`. Mapped readonly detection now distinguishes name-type filtering from remapping using ts-go's own assignability criterion: identity and conditional filtering inherit source modifiers, while real remaps remain mutable without explicit `readonly`. Added focused regressions for both `as K` and conditional filtering; the existing template-remapping regression remains in place.

internal/plugins/typescript/rules/prefer_readonly_parameter_types/prefer_readonly_parameter_types.go
swwind: Fixed in `292a3ae1`. Diagnostic ranges now start at the ESTree-equivalent parameter shape: the parameter name or binding pattern, or the `...` token for a rest element. This excludes decorators and parameter-property modifiers while preserving rest-element ranges. Added the reported decorated-parameter regression, and the existing rest and parameter-property range cases continue to pass.

elecmonkey (approved):
LGTM

PR #1903, opened 2026-08-26
feat: port rule unicorn/prefer-ternary

hebeiqing:
## Summary

Follow-up to #1903. Addresses the five review comments and the CI issues (`fmt`, `cspell`) called out in the discussion.

## Fixes

| # | File | Change |
|---|---|---|
| **P1 #1** | `prefer_ternary.go` — `getNodeBody` | Use `node.AsParenthesizedExpression().Expression` directly instead of `WalkUpParenthesizedExpressions(...).AsParenthesizedExpression().Expression`, which can walk to the surrounding `ExpressionStatement` and panic. |
| **P1 #2** | `prefer_ternary.go` — `hasSideEffect` | Port the recursive visitor from `@eslint-community/eslint-utils` (default options). Only the leaves that are unconditionally side-effect-bearing trip the predicate; everything else recurses into its children. Compound assignments and other binary operators visit children too, so `a + b` is pure while `a + sideEffect()` is not. |
| **P1 #3** | `prefer_ternary.go` — `checkIfStatement` / `checkLetPlusIfProblem` | Move the `only-single-line` gate to operate on the unwrapped `consequent` / `alternate` (after `getNodeBody`), and walk parens off the test expression to mirror ESTree's transparent-paren range. Also unwrap parens in the `isConditionalExpression` test gate. |
| **P1 #4** | `prefer_ternary.go` — new `staticAccessName` / `isSameAssignmentTarget`; plumbed `*utils.StaticStringEvaluator` from `Run` through `checkIfStatement` → `buildMerge` → `isMergeableAssignmentExpression` | Use the static evaluator for the LHS-comparison fast path so `(x)['b' + 'ar']` and `x.bar` collapse to the same property name. Graceful nil-evaluator fallback on JS / gap files. |
| **P2** | `prefer_ternary.md` | Drop the "Differences from ESLint" notes about the computed-key divergence and the alleged nested-if double-report — both are no longer accurate after the fixes above. |

## Tests added (`prefer_ternary_extras_test.go`)

- `if (t) { (x = a); } else { (x = b); }` → `x = t ? a : b;` — regression for the parenthesized-body panic.
- `if (t) { ({x} = a); } else { ({x} = b); }` — destructuring form must not crash and must not report.
- `let x = a + b; if (test) { x = c; }` → `const x = test ? c : (a + b);` — pure binary initializer is now allowed.
- `let x = () => a; if (test) { x = b; }` → `const x = test ? b : () => a;` — function/arrow initializers are pure.
- `let x = {a: sideEffect()};`, `let x = [sideEffect()];`, `` let x = `${sideEffect()}`; ``, `let x = (y = 1);` — all must NOT be collapsed (nested effects are now correctly detected).
- `if (t) {\n  x=a;\n} else {\n  x=b;\n}` with `only-single-line` — unwrapped bodies now report.
- `if (\n\tt\n) { ... }` with `only-single-line` — multi-line parenthesized test now reports.
- `if (test) { (\n  a = foo\n); } else { a = bar; }` with `only-single-line` — parenthesized body expression now reports.
- The upstream pinned v73.0.0 case `(foo)['b' + 'ar'] = a; foo.bar = b;` → `;(foo)['b' + 'ar'] = test ? a : b;` — computed-key static evaluation now works.

## Verification

- `go test ./internal/plugins/unicorn/rules/prefer_ternary/...` — all tests pass.
- `gofmt -l` — clean.
- `cspell` — 0 issues.


elecmonkey:
Could you use English in the PR discussion?

internal/plugins/unicorn/rules/prefer_ternary/prefer_ternary.go
```diff
@@ -0,0 +1,858 @@
...
+	}
+	switch node.Kind {
+	case ast.KindExpressionStatement:
+		return getNodeBody(node.AsExpressionStatement().Expression)
+	case ast.KindParenthesizedExpression:
+		// Walk through any chain of parenthesized wrappers, mirroring how
+		// ESTree flattens parens so the parent is the semantic construct.
+		// WalkUpParenthesizedExpressions returns the innermost expression
+		// (skipping wrappers), so recurse on that.
+		return getNodeBody(ast.WalkUpParenthesizedExpressions(node).AsParenthesizedExpression().Expression)
```
swwind: [P1] Avoid panicking while unwrapping parenthesized expression statements

`ast.WalkUpParenthesizedExpressions(node)` walks toward the parent and can return the surrounding `ExpressionStatement`, so the following `AsParenthesizedExpression()` panics. For example, `if(t){(x=a)}else{(x=b)}` and `if(t){({x}=a)}else{({x}=b)}` are accepted by upstream, but rslint crashes while visiting them. Please unwrap the current `ParenthesizedExpression` through its own expression field and add regression coverage for both forms.

internal/plugins/unicorn/rules/prefer_ternary/prefer_ternary.go
```diff
@@ -0,0 +1,858 @@
...
+		}}
+	})
+}
+
+// hasSideEffect is a small subset of hasSideEffect from eslint-community, used
+// only to reject `let x = fn();` / `let x = new Foo();` where the init cannot
+// be inlined into the alternate position of the ternary. It deliberately
+// stays narrow — a more general side-effect classifier would re-evaluate
+// arbitrary expressions and over-approximate the cases upstream rules out.
+func hasSideEffect(node *ast.Node) bool {
```
swwind, line 603: [P1] Match upstream's recursive side-effect classification

This top-level kind switch produces mismatches in both directions. `let x = a + b; if(t){x=c}` and function/arrow initializers are pure according to upstream but are rejected here. Conversely, nested effects such as `let x={a:sideEffect()};if(t){x=c}` are missed, causing rslint to suggest `const x = t ? c : {a:sideEffect()}`. That changes an unconditional call into one skipped when `t` is true. Arrays, templates, and parenthesized assignments have the same problem. Please port the recursive upstream `hasSideEffect` semantics, distinguishing ordinary binary expressions from assignments.

internal/plugins/unicorn/rules/prefer_ternary/prefer_ternary.go
```diff
@@ -0,0 +1,858 @@
...
+		}
+	}
+
+	// The test cannot already be a ternary.
+	if isConditionalExpression(ifStmt.Expression) {
+		return
+	}
+
+	lineStarts := ctx.SourceFile.ECMALineMap()
+	if isOnlySingleLine {
```
swwind: [P1] Apply `only-single-line` to the unwrapped bodies

Upstream calls `getNodeBody` first and checks the resulting expressions, while this checks the complete `ThenStatement` and `ElseStatement`. Consequently, `if (t) {\n x=a;\n} else {\n x=b;\n}` is reported and fixed upstream because `t`, `x=a`, and `x=b` are individually single-line, but rslint emits nothing. Parentheses excluded from ESTree ranges cause a related miss for a multiline-parenthesized `t`. Please perform the check on the unwrapped consequent/alternate and mirror ESTree's transparent-parenthesis range.

internal/plugins/unicorn/rules/prefer_ternary/prefer_ternary.go
```diff
@@ -0,0 +1,858 @@
...
+	if c.OperatorToken.Kind != a.OperatorToken.Kind {
+		return false
+	}
+	if isConditionalExpression(c.Left) || isConditionalExpression(a.Left) {
+		return false
+	}
+	if isConditionalExpression(c.Right) || isConditionalExpression(a.Right) {
+		return false
+	}
+	return utils.IsSameReference(c.Left, a.Left, false)
```
swwind: [P1] Preserve upstream static computed-key equivalence

`if(t){(x)['y'+'']=a}else{x.y=b}` produces one diagnostic and fixes to `(x)['y'+''] = t ? a : b;` under Unicorn v73.0.0, but rslint emits nothing because the shared static-name lookup does not evaluate the computed constant. The pinned upstream suite contains the equivalent `'b' + 'ar'` case. Please support the same statically evaluated keys instead of documenting this as an intentional omission.

internal/plugins/unicorn/rules/prefer_ternary/prefer_ternary.md
```diff
@@ -0,0 +1,94 @@
...
+The upstream `unicorn/prefer-ternary` rule is reported under
+`eslint-plugin-unicorn`'s `recommended` and `unopinionated` configs and
+auto-fixes via `--fix`. The current port reproduces that contract:
+
+- The simple `if/else` form is rewritten to a ternary and the result is
+  reported with a single diagnostic whose `fix` rewrites the source.
+- The `let x = a; if (test) x = b;` form is reported with a diagnostic
+  whose `suggest` rewrites the source to `const x = test ? b : a;`
+  (or `let` when the variable receives a later write).
+- Two upstream test cases are not ported: the
```
swwind: [P2] Correct the claimed nested-if divergence

The pinned upstream listener skips an `IfStatement` when it is its parent's alternate, and the differential matrix showed matching behavior for nested-if consequent and alternate forms: only the mergeable inner `if` is reported. The statement that upstream reports both is therefore inaccurate. Please remove it, and remove the computed-key divergence once that parity issue is fixed.

swwind:
Thanks for your help, would you like to fix those problems above, and maybe also the CI failing issues? (like fmt, cspell). After that we can merge this rule.

hebeiqing:
> Could you use English in the PR discussion?

Sorry about the Chinese description. I have addressed the review feedback and updated the PR description in English. Thank you for the reminder.

hebeiqing:
> Thanks for your help, would you like to fix those problems above, and maybe also the CI failing issues? (like fmt, cspell). After that we can merge this rule.

Thanks for the review. I've addressed the issues above, including the fmt and cspell CI issues, in dd0a4bb. The PR description has also been updated in English. Could you please take another look?

internal/plugins/unicorn/rules/prefer_ternary/prefer_ternary.go
```diff
@@ -0,0 +1,1005 @@
...
+	if left == nil || right == nil {
+		return false
+	}
+	left = ast.SkipOuterExpressions(left, ast.OEKParentheses|ast.OEKAssertions)
+	right = ast.SkipOuterExpressions(right, ast.OEKParentheses|ast.OEKAssertions)
+	if left == nil || right == nil {
+		return false
+	}
+
+	if ast.IsAccessExpression(left) && ast.IsAccessExpression(right) && staticStrings != nil {
```
swwind, line 299: [P1] Keep private fields out of static string-key equivalence

`class C { #x; m(){ if (t) { this.#x = a; } else { this["#x"] = b; } } }` produces no diagnostic under Unicorn v73.0.0 because the private field and the public string-keyed property are different assignment targets. Rslint currently treats both names as `"#x"`, reports the `if`, and fixes it to `this.#x = t ? a : b;`, changing the else branch’s target.

Please exclude private property accesses from this static-name fast path and preserve private-to-private structural comparison. Add coverage for both `this.#x` versus `this["#x"]` and two matching `this.#x` accesses.

internal/plugins/unicorn/rules/prefer_ternary/prefer_ternary.go
```diff
@@ -0,0 +1,1005 @@
...
+// still caught. ArrowFunction and FunctionExpression short-circuit to false:
+// their bodies do not run until invoked.
+//
+// The non-assignment BinaryExpression case intentionally visits children
+// rather than returning true: upstream considers `a + b` and `a ? b : c`
+// pure, and the only assignment-shaped binary with a side effect is the
+// plain `=` (compound assignments like `+=` also visit children — the
+// right-hand side carries the effect, and the LHS write is recursed into
+// separately).
+func hasSideEffect(node *ast.Node) bool {
```
swwind, line 603: [P1] Complete the upstream `hasSideEffect` node mapping

The real `@eslint-community/eslint-utils` visitor treats every `AssignmentExpression` as side-effectful and treats `delete` as side-effectful. It also stops at function bodies. The current port differs in both directions:

- `let x = (y += 1); if (test) { x = next; }`, `y ||= 1`, `y ??= 1`, nested compound assignments, and `delete object.x` are silent upstream, but Rslint suggests moving the effect into the ternary’s alternate branch.
- `let x = ({m(){ sideEffect(); }}); if (test) { x = next; }`, the getter equivalent, and a class with such a method are reported upstream, but Rslint walks the deferred method body and emits nothing.

Please classify every assignment operator as side-effectful, handle the actual ts-go node shape for `delete`, and stop traversal at method/accessor bodies while retaining upstream traversal of immediately evaluated keys and initializers. These cases need regression coverage because the first group can produce semantics-changing suggestions.

internal/plugins/unicorn/rules/prefer_ternary/prefer_ternary.go
```diff
@@ -0,0 +1,1005 @@
...
+			}
+		}
+
+		// The static evaluator is optional: a nil evaluator just means the
+		// computed-key fast path falls back to literal-only static names.
+		// The rule stays syntactic on JS/gap files where no TypeChecker is
+		// available — literal concat (`x['b' + 'ar']`) still folds through
+		// the evaluator's literal handling, but identifier resolution
+		// degrades to "unknown".
+		var staticStrings *utils.StaticStringEvaluator
```
swwind: [P1] Preserve computed-key evaluation without a TypeChecker

When `ctx.TypeChecker` is nil, `staticStrings` remains nil, so the new fast path is skipped entirely. Consequently, the previously reported `(x)['b' + 'ar']` versus `x.bar` case still produces no diagnostic on JS/gap programs without a checker, despite the comment saying literal concatenation continues to fold.

Please construct a literal-only evaluator for this path, for example with `NewStaticStringEvaluatorWithoutScope()`, and add a no-TypeChecker regression test. Identifier resolution may remain unavailable, but literal expressions must retain upstream parity.

internal/plugins/unicorn/rules/prefer_ternary/prefer_ternary.md
```diff
@@ -0,0 +1,85 @@
...
+        'multiple line array'
+    ];
+} else {
+    foo = bar;
+}
+```
+
+## Differences from ESLint
+
+The upstream `unicorn/prefer-ternary` rule is reported under
```
swwind, line 72: [P1] Enable the implemented rule in the Unicorn preset

`packages/rslint/src/config/presets/unicorn.ts` still leaves `unicorn/prefer-ternary` commented out as “not implemented”. This contradicts the contract described here and means users of Rslint’s Unicorn preset do not receive this upstream-recommended rule.

Please enable the preset entry as part of the port, as is done for other newly implemented Unicorn rules.

packages/rslint-test-tools/tests/eslint-plugin-unicorn/rules/prefer-ternary.test.ts
```diff
@@ -0,0 +1,173 @@
+import { RuleTester } from '../rule-tester';
```
swwind, line 1: [P1] Register this test file in the Rstest configuration

This new test file is not listed in `packages/rslint-test-tools/rstest.config.mts`, whose `include` array enumerates rule tests explicitly. As a result, the normal JS test command does not execute this suite.

Please add the file to the include list so these integration tests run in CI.

hebeiqing:
Addressed all review comments in the latest commits.

- Fixed parenthesized expression unwrapping and added regression coverage.
- Aligned recursive side-effect detection with upstream, including assignments, `delete`, nested expressions, and deferred method/accessor bodies.
- Updated `only-single-line` handling for unwrapped bodies and parenthesized tests.
- Added static computed-key comparison, including support for JS files without a TypeChecker.
- Fixed private-field assignment-target comparison.
- Enabled `unicorn/prefer-ternary` in the Unicorn preset.
- Registered the JS integration test in the Rstest configuration.

Verified with:

- `go test ./internal/plugins/unicorn/rules/prefer_ternary/...`
- `pnpm --filter @rslint/test-tools test -- tests/eslint-plugin-unicorn/rules/prefer-ternary.test.ts`
- `pnpm run check-spell`

internal/plugins/unicorn/rules/prefer_ternary/prefer_ternary.go
```diff
@@ -0,0 +1,1159 @@
...
+		}
+		if len(filtered) == 1 {
+			return getNodeBody(filtered[0])
+		}
+	}
+	return node
+}
+
+// isConditionalExpression mirrors upstream's `isTernary` helper.
+func isConditionalExpression(node *ast.Node) bool {
```
swwind, line 179: **[P1] Treat source parentheses as transparent in every ternary gate**

ESTree removes these parentheses, but `isConditionalExpression` and `isBooleanLiteral` inspect the direct ts-go node. Consequently, `function f(){if(t){return (a ? b : c)}else{return d}}`, `if(t){x=(a ? b : c)}else{x=d}`, and `let x=(a ? b : c);if(t){x=d}` are all silent under Unicorn v73.0.0, while Rslint reports and rewrites them. `return (true)` versus `return ((false))` has the same issue with the boolean-literal exclusion.

Please unwrap transparent parentheses before all conditional/boolean classification checks, including return arguments, assignment operands, and the let-plus-if initializer/right side, and add regressions for each affected branch.

internal/plugins/unicorn/rules/prefer_ternary/prefer_ternary.go
```diff
@@ -0,0 +1,1159 @@
...
+	switch node.Kind {
+	case ast.KindMethodDeclaration:
+		method := node.AsMethodDeclaration()
+		if method == nil {
+			return false
+		}
+		if hasSideEffect(method.Name()) {
+			return true
+		}
+		if hasSideEffectInParameters(method.Parameters) {
```
swwind: **[P1] Keep method parameters and constructors behind the deferred function boundary**

The upstream `hasSideEffect` visitor reaches a method key and then stops at its `FunctionExpression` value, so neither its parameter defaults nor its body are evaluated while creating the object/class. For example, Unicorn reports and suggests a collapse for `let x={m(v=sideEffect()) {}};if(t){x=next}`, `let x={set m(v=sideEffect()) {}};if(t){x=next}`, and `let x=class {constructor(){sideEffect()}};if(t){x=next}`, while Rslint emits nothing.

The current helper explicitly walks method/accessor parameter initializers, and constructors are not treated as deferred at all, so their complete bodies are traversed. Please visit only the immediately evaluated computed key and skip parameters/body for methods, accessors, and constructors. The existing “parameter default values are walked” valid test asserts the opposite of upstream behavior, while the later invalid test described as a parameter-default case actually contains `m() {}` without a parameter.

internal/plugins/unicorn/rules/prefer_ternary/prefer_ternary.go
```diff
@@ -0,0 +1,1159 @@
...
+		}
+
+		// The static evaluator is optional: a nil evaluator just means the
+		// computed-key fast path falls back to literal-only static names.
+		// On JS/gap files where no TypeChecker is available we still
+		// construct a no-scope evaluator so literal concatenation
+		// (`x['b' + 'ar']`) keeps folding — identifier resolution is
+		// unavailable in that mode, but literal expressions retain
+		// upstream parity.
+		staticStrings := utils.NewStaticStringEvaluatorWithoutScope()
```
swwind, line 55: **[P1] Keep static property-name evaluation scope-free**

Upstream calls `getStaticValue(property)` without a scope when comparing assignment targets. Switching to a reference-resolving evaluator when a TypeChecker exists broadens the rule: `const key='bar';if(t){x[key]=a}else{x.bar=b}` and the equivalent `let key='bar'` case are silent upstream, but Rslint reports and fixes both. Numeric and concatenated local constants behave the same way.

Please use the no-scope evaluator for this comparison regardless of checker availability. Add separate checker-backed and actual source-only regressions: the current `file.js` RuleTester case still receives a TypeChecker and therefore does not exercise the stated nil-checker path.

internal/plugins/unicorn/rules/prefer_ternary/prefer_ternary.go
```diff
@@ -0,0 +1,1159 @@
...
+	consequentText := renderMergeOperand(ctx, result.consequent)
+	alternateText := renderMergeOperand(ctx, result.alternate)
+
+	fixed := result.before + testText + " ? " + consequentText + " : " + alternateText + result.after
+
+	// ASI safety: prepend `;` when the previous token would absorb the
+	// replacement into the prior statement.
+	ifPrev := utils.TrimNodeTextRange(ctx.SourceFile, ifNode).Pos()
+	if prev, ok := utils.TokenBeforePosition(ctx.SourceFile, ifPrev); ok {
+		if needsSemicolonBeforeAfterText(prev, fixed) {
```
swwind: **[P1] Do not insert an ASI semicolon after a control-statement header**

Upstream’s `needsSemicolon` recognizes when the preceding `)` belongs to a `while`, `if`, or `for` header and returns false. Rslint treats every closing parenthesis as an ASI hazard. For `while(keep) if(t){(x).y=a}else{(x).y=b}`, upstream fixes to `while(keep) (x).y = t ? a : b;`, while Rslint produces `while(keep) ;(x).y = t ? a : b;`, moving the assignment outside the loop. The same failure occurs for direct `if` and `for` bodies.

Please include the preceding token’s AST context in this decision, matching the upstream control-body exceptions, and add focused fix-output regressions for these three forms.

internal/plugins/unicorn/rules/prefer_ternary/prefer_ternary.go
```diff
@@ -0,0 +1,1140 @@
...
+	return []rule.RuleFix{rule.RuleFixReplace(ctx.SourceFile, ifNode, fixed)}
+}
+
+// isControlStatementBody reports whether node is an unbraced control-flow
+// body. A leading semicolon there would turn the body into an empty statement.
+func isControlStatementBody(node *ast.Node) bool {
+	if node == nil || node.Parent == nil {
+		return false
+	}
+	switch node.Parent.Kind {
```
swwind, line 1047: **[P1] Preserve for-in, for-of, and with statement bodies**

`isControlStatementBody` does not recognize `ForInStatement`, `ForOfStatement`, or `WithStatement`. For example, upstream fixes `for(k in object) if(t){(x).y=a}else{(x).y=b}` to `for(k in object) (x).y = t ? a : b;`, while this implementation inserts a semicolon after the loop header: `for(k in object) ;(x).y = t ? a : b;`. That turns the loop body into an empty statement and executes the assignment only once after the loop. The same semantic change occurs for `for...of` and `with`.

internal/plugins/unicorn/rules/prefer_ternary/prefer_ternary.go
```diff
@@ -0,0 +1,1140 @@
...
+	if replacement == "" {
+		return false
+	}
+	switch replacement[0] {
+	case '[', '(', '`', '/', '+', '-', ',', '.':
+		// ok, fall through
+	default:
+		return false
+	}
+	switch prev.Kind {
```
swwind, line 1090: **[P1] Handle an ObjectExpression before an ASI-sensitive replacement**

For `const value={}\nif(t){(x).y=a}else{(x).y=b}`, upstream inserts a leading semicolon and produces `const value={}\n;(x).y = t ? a : b;`. The current token-kind check does not recognize the closing brace of an `ObjectExpression`, so it produces `const value={}\n(x).y = t ? a : b;`, allowing the replacement to continue the initializer as a call/member expression. Upstream resolves the AST node at the previous token specifically to distinguish an object expression from other closing braces.

internal/plugins/unicorn/rules/prefer_ternary/prefer_ternary.go
```diff
@@ -0,0 +1,1140 @@
...
+	if node == nil {
+		return false
+	}
+	switch node.Kind {
+	case ast.KindMethodDeclaration:
+		method := node.AsMethodDeclaration()
+		if method == nil {
+			return false
+		}
+		if hasSideEffect(method.Name()) {
```
swwind, line 678: **[P1] Include method decorators in the immediate side-effect walk**

Upstream does not report `let x=class{@dec(call()) m(){}};if(t){x=next}`, because evaluating the class immediately evaluates the method decorator and therefore the initializer has a side effect. `hasSideEffectInDeferredBody` currently checks only the method name and drops its decorators, so the rule reports this case and offers a suggestion that moves the decorator call into one branch of a conditional. Please walk decorators and other immediately evaluated metadata while continuing to skip the deferred parameters and body.

internal/plugins/unicorn/rules/prefer_ternary/prefer_ternary.go
```diff
@@ -0,0 +1,1140 @@
...
+// listed node kinds all bind more loosely than `?:`, so a child of a
+// ConditionalExpression must be wrapped in parens to preserve semantics.
+func shouldAddParenthesesToConditionalExpressionChild(node *ast.Node) bool {
+	if node == nil {
+		return false
+	}
+	switch node.Kind {
+	case ast.KindAwaitExpression,
+		ast.KindYieldExpression,
+		ast.KindBinaryExpression, // AssignmentExpression
```
swwind: **[P2] Do not treat every ts-go BinaryExpression as an AssignmentExpression**

ts-go represents ordinary binary operators with `BinaryExpression` too. For `let x=a+b;if(t){x=c}`, upstream suggests `const x=t ? c : a+b;`, while this implementation produces `const x=t ? c : (a+b);`. Only assignment and sequence operators have the lower-precedence behavior covered by the upstream helper; additive and other higher-precedence binary expressions should not be parenthesized.

internal/plugins/unicorn/all.go
```diff
@@ -50,6 +51,7 @@ func GetAllRules() []rule.Rule {
 		prefer_node_protocol.PreferNodeProtocolRule,
 		prefer_number_properties.PreferNumberPropertiesRule,
 		prefer_set_has.PreferSetHasRule,
+	prefer_ternary.PreferTernaryRule,
```
swwind: **[P2] Run gofmt on the rule registration**

The `prefer_ternary.PreferTernaryRule` entry is indented one level less than the surrounding entries, and `gofmt -d internal/plugins/unicorn/all.go` reports this line.

internal/plugins/unicorn/rules/prefer_ternary/prefer_ternary.go
```diff
@@ -0,0 +1,1155 @@
...
+		// the outermost operator as `|=`, but the upstream test expects the
+		// LHS to be the entire chain and the operator to be the first `=`
+		// that follows it. Walk down the right side of the chain to the
+		// first `=` (or any plain assignment operator) and use its LHS
+		// and right operand instead.
+		leftText := getParenthesizedText(ctx, c.Left)
+		rightNode := c.Right
+		opNode := c.OperatorToken
+		cur := c
+		for cur.Right != nil && cur.Right.Kind == ast.KindBinaryExpression {
```
swwind: [P1] Preserve each branch's inner assignment chain

For `if (t) x = y = a; else x = z = b;`, Unicorn v73 fixes this to `x = t ? (y = a) : (z = b);`, but this walker stops at the first nested `=` and Rslint produces `x = y = t ? a : b;`. On the false branch that writes `y` instead of `z`, so the autofix changes runtime behavior. The same problem reproduces with compound, deeper, and property assignment chains. Please keep only the common outer assignment in `before` and preserve each branch's complete right-hand assignment as the conditional arm.

internal/plugins/unicorn/rules/prefer_ternary/prefer_ternary.go
```diff
@@ -0,0 +1,1155 @@
...
+
+		// 2. Replace the init expression with the ternary.
+		initRange := utils.TrimNodeTextRange(ctx.SourceFile, p.init)
+		// 3. Drop the if statement (and any whitespace / trailing semicolon
+		//    that bridged the two). A `;` is added only when the original
+		//    would have needed one for ASI — that is, when the next token
+		//    begins with `[` / `(` / `` ` `` / `/` / `+` / `-` / `,` / `.`.
+		declEnd := utils.TrimNodeTextRange(ctx.SourceFile, p.declStmt).End()
+		ifEnd := utils.TrimNodeTextRange(ctx.SourceFile, p.ifNode).End()
+		dropText := ";"
```
swwind, line 566: [P2] Avoid inserting a second semicolon after the declaration

For `let x = a; if (t) { x = b; } (foo)();`, upstream suggests `const x = t ? b : a; (foo)();`, while this code chooses `dropText = ";"` solely from the next token and produces `const x = t ? b : a;; (foo)();`. Upstream's `needsSemicolon` also checks whether the previous declaration already ends with `;`; no additional separator is needed in that case. This also reproduces before `[`, template literals, regexps, `+`, `-`, and `.` and inside blocks, case clauses, and static blocks.

internal/plugins/unicorn/rules/prefer_ternary/prefer_ternary.go
```diff
@@ -0,0 +1,1142 @@
...
+		// the shared outer operation. Plain `x = y = value` must not enter
+		// this path: its inner assignments belong to the conditional arms.
+		if c.OperatorToken.Kind != ast.KindEqualsToken {
+			cur := c
+			for cur.Right != nil && cur.Right.Kind == ast.KindBinaryExpression {
+				rb := cur.Right.AsBinaryExpression()
+				if rb.OperatorToken == nil || !isAssignmentOperatorKind(rb.OperatorToken.Kind) {
+					break
+				}
+				if rb.OperatorToken.Kind == ast.KindEqualsToken {
```
swwind: **[P1] Compare both assignment chains before hoisting an inner assignment**

This loop finds the first nested `=` by walking only the consequent, while `findMatchingRight` selects an `=` from the alternate solely by operator kind. It never verifies that the targets and operators along both prefixes are the same. For `if (t) { x += y = a; } else { x += z = b; }`, upstream stops after the common `x +=` and fixes this to `x += t ? (y = a) : (z = b);`, but Rslint produces `x+=y = t ? a : b;`. The false branch now writes `y` instead of `z`, changing runtime behavior. Deeper and property assignment chains fail the same way. Please mirror the recursive merge: descend only while both current assignments remain mergeable, accumulating their common prefix; otherwise preserve each complete right-hand assignment as a conditional arm. Add regressions with matching and differing inner targets and operators.

hebeiqing: Fixed in 6a7b994. The merge now descends through assignment chains only while both sides have matching targets and operators. Added regression coverage for matching and differing inner targets and operators.

internal/plugins/unicorn/rules/prefer_ternary/prefer_ternary.go
```diff
@@ -0,0 +1,1108 @@
...
+// CallExpression, NewExpression, AwaitExpression, YieldExpression, ++/--,
+// `delete`, and any assignment (`=`, `+=`, `-=`, `||=`, `&&=`, `??=`, …) —
+// trip the predicate. Everything else recurses into its children so nested
+// effects (a call inside an object literal, an assignment inside
+// parentheses, a `new` inside a binary expression) are still caught.
+// ArrowFunction, FunctionExpression, FunctionDeclaration, and
+// method/accessor bodies short-circuit to false: their bodies do not run
+// until invoked. The method/accessor case still walks the immediately
+// evaluated name (computed keys) and parameter initializers, so an effect
+// inside a computed key or default value is still surfaced.
```
elecmonkey: LGTM overall. This is only a comment drift: the current implementation skips method/accessor parameters, matching upstream, so saying "parameter initializers" are walked is no longer accurate. If you'd like, could you apply this small cleanup?

```suggestion
// ArrowFunction, FunctionExpression, FunctionDeclaration, and
// method/accessor bodies short-circuit to false: their bodies do not run
// until invoked. The method/accessor case still walks immediately evaluated
// metadata such as decorators and computed names, while skipping parameters
// and body.
```

hebeiqing: Applied in 8729771. Thanks!

elecmonkey (approved):

PR #1907, opened 2026-08-26
feat: port rule unicorn/no-xor-as-exponentiation

rockm3:
## Summary

Port the `unicorn/no-xor-as-exponentiation` rule from ESLint to rslint.

Flags `^` between two decimal integer literals (a common typo when developers expect exponentiation, but JavaScript `^` is bitwise XOR). Suggests replacing `^` with `**`.

## Related Links

- ESLint rule: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-xor-as-exponentiation.md
- Source code: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/rules/no-xor-as-exponentiation.js
- Test cases: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/test/no-xor-as-exponentiation.js

## Implementation Notes

- **Raw text match**: tsgo's `NumericLiteral.Text` is decimal-normalized (e.g. `0xFF` → `255`), which would fold the upstream regex's literal-shape classification. Uses `scanner.GetSourceTextOfNodeFromSourceFile` to read the raw source form before testing the decimal-integer regex.
- **Single suggestion**: only the `^` token is replaced; surrounding whitespace and comments are preserved (covered by the `2 /* comment */ ^ 8` test).
- **Suggestions are deferred**: detection, message construction, and the report range are eager; the suggestion's fix is built lazily via `ReportNodeWithDeferredSuggestions`.

## Checklist

### DIFF
--- a/internal/plugins/unicorn/rules/no_xor_as_exponentiation/no_xor_as_exponentiation.md
+++ b/internal/plugins/unicorn/rules/no_xor_as_exponentiation/no_xor_as_exponentiation.md
@@ -1,0 +1,21 @@
+# no-xor-as-exponentiation
+
+## Rule Details
+
+In JavaScript, `^` is the [bitwise XOR](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_XOR) operator, not exponentiation. Developers coming from languages like Lua, Julia, R, or MATLAB, or from math notation, often expect `^` to mean "to the power of", so `2 ^ 32` silently evaluates to `34` instead of `4294967296`. The actual exponentiation operator is [`**`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Exponentiation).
+
+This rule flags `^` between two decimal integer literals, which is almost always this mistake. Hexadecimal, octal, and binary literals (such as `0xFF ^ 8`) and any non-literal operands (such as `flags ^ MASK`) are ignored, since those are far more likely to be intentional bitwise XOR.
+
+Examples of **incorrect** code for this rule:
+
+```javascript
+const bytes = 2 ^ 10; // 8, not 1024
+const cube = 3 ^ 3; // 0, not 27
+```
+
+Examples of **correct** code for this rule:
+
+```javascript
+const bytes = 2 ** 10;
+const cube = 3 ** 3;
+```
--- a/internal/plugins/unicorn/rules/no_xor_as_exponentiation/no_xor_as_exponentiation.md
+++ b/internal/plugins/unicorn/rules/no_xor_as_exponentiation/no_xor_as_exponentiation.md
@@ -19,3 +19,8 @@ Examples of **correct** code for this rule:
 const bytes = 2 ** 10;
 const cube = 3 ** 3;
 ```
+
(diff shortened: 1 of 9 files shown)


internal/plugins/unicorn/rules/no_xor_as_exponentiation/no_xor_as_exponentiation_upstream_test.go
```diff
@@ -0,0 +1,90 @@
...
+			// ---- Other operators ----
+			jsValid(`2 | 8`),
+			jsValid(`2 & 8`),
+			jsValid(`2 << 8`),
+			jsValid(`2 ** 8`),
+
+			// ---- TypeScript: operand is a `TSAsExpression`, not a literal ----
+			// rslint does not differentiate AsExpression from its operand at the
+			// rule layer; Skip: the upstream parser boundary is not relevant here.
+			{Code: `(2 as number) ^ 8`, Skip: true},
```
swwind: This upstream TypeScript case does not need to be skipped. The current implementation preserves the `AsExpression` wrapper after `SkipParentheses`, so `(2 as number) ^ 8` is correctly treated as valid. Please run it as a `.ts` test instead, for example with `FileName: "file.ts"`, and remove the inaccurate comment. Otherwise the only upstream TS wrapper regression case is not actually covered.

rockm3: Thanks for catching this! You were right — the current implementation does preserve the `AsExpression` wrapper after `SkipParentheses`, so `(2 as number) ^ 8` is correctly treated as valid. Fixed in 2edbe019: removed the `Skip: true` and the misleading comment, and now runs as `FileName: "file.ts"` to actually exercise the TypeScript parser path.

internal/plugins/unicorn/rules/no_xor_as_exponentiation/no_xor_as_exponentiation.md
```diff
@@ -0,0 +1,21 @@
...
+const bytes = 2 ^ 10; // 8, not 1024
+const cube = 3 ^ 3; // 0, not 27
+```
+
+Examples of **correct** code for this rule:
+
+```javascript
+const bytes = 2 ** 10;
+const cube = 3 ** 3;
+```
```
swwind, line 21: Missing *Original Documentation* chapter here.
````suggestion
```

## Original Documentation

- [eslint-plugin-unicorn: no-xor-as-exponentiation](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v73.0.0/docs/rules/no-xor-as-exponentiation.md)
- [Source code](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v73.0.0/rules/no-xor-as-exponentiation.js)
````

rockm3: Added in 8524fdda. Used v73.0.0 to match the other recently ported rules (error-message, filename-case, etc.).

swwind (approved):
LGTM, thanks for help.

PR #1909, opened 2026-08-26
fix(website): keep playground editor hover from being clipped

elecmonkey:
## Summary

Hovering a symbol on the first lines of the playground editor showed a hover card that was cut off at the top and painted underneath the site nav bar; Monaco renders hover and suggest widgets inside the editor's own DOM by default, so they are clipped by the editor's overflow guard and trapped in the playground container's stacking context — this hosts them in a body-level node instead, kept above the nav.

Before:

image

This PR:

image

### DIFF
--- a/website/theme/components/Playground/EditorTabs.css
+++ b/website/theme/components/Playground/EditorTabs.css
@@ -1,4 +1,15 @@
 /* Monaco Editor specific styles - cannot be done with Tailwind */
+
+/* Body-level host for Monaco hover/suggest widgets, above the site nav bar */
+.playground-overflow-widgets {
+  position: absolute;
+  top: 0;
+  left: 0;
+  width: 0;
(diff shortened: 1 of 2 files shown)


swwind:
Tested for a while, I found that only adding the following css is sufficient enough.

```css
.monaco-editor .overflowingContentWidgets {
  position: absolute;
  z-index: 100000;
}
```

What do you think?

swwind:
I felt sorry with previous comment. I found the key here.

The hover element `.monaco-resizable-hover` has inline style `z-index: 50`. And the rspress's header has default `z-index: 60`. So header is always above the hover box.

So the minimal fix should be

```css
.monaco-resizable-hover {
  z-index: 1000000 !important;
}
```

swwind (approved):
LGTM

PR #1912, opened 2026-08-27
fix: update LintSingleFile test call sites to rule.ConfiguredRule

elecmonkey:
## Summary

#1893 moved `ConfiguredRule` into the `rule` package and dropped `ExcludePaths` from `LintSingleFileOptions`. Two rule test files merged around the same time still use the old API, so `main` no longer builds. This updates those call sites; behavior is unchanged.

## Related Links

- #1893
- #1890
- #1868

## Checklist


elecmonkey:
duplicated with https://github.com/web-infra-dev/rslint/pull/1911

PR #1913, opened 2026-08-27
feat: port rule unicorn/no-exports-in-scripts

guihua-c:
## Summary

Port the `unicorn/no-exports-in-scripts` rule from ESLint to rslint.

The rule disallows export syntax in files treated as scripts while allowing exports in modules.

## Related Links

- ESLint rule: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-exports-in-scripts.md
- Source code: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/rules/no-exports-in-scripts.js

## Checklist

- [x] Tests updated.
- [x] Documentation updated.

### DIFF
--- a/internal/plugins/unicorn/all.go
+++ b/internal/plugins/unicorn/all.go
@@ -7,6 +7,7 @@ import (
 	"github.com/web-infra-dev/rslint/internal/plugins/unicorn/rules/new_for_builtins"
 	"github.com/web-infra-dev/rslint/internal/plugins/unicorn/rules/no_array_fill_with_reference_type"
 	"github.com/web-infra-dev/rslint/internal/plugins/unicorn/rules/no_await_in_promise_methods"
+	"github.com/web-infra-dev/rslint/internal/plugins/unicorn/rules/no_exports_in_scripts"
 	"github.com/web-infra-dev/rslint/internal/plugins/unicorn/rules/no_instanceof_builtins"
 	"github.com/web-infra-dev/rslint/internal/plugins/unicorn/rules/no_invalid_fetch_options"
 	"github.com/web-infra-dev/rslint/internal/plugins/unicorn/rules/no_invalid_remove_event_listener"
@@ -33,6 +34,7 @@ func GetAllRules() []rule.Rule {
 		new_for_builtins.NewForBuiltinsRule,
 		no_array_fill_with_reference_type.NoArrayFillWithReferenceTypeRule,
 		no_await_in_promise_methods.NoAwaitInPromiseMethodsRule,
+		no_exports_in_scripts.NoExportsInScriptsRule,
 		no_instanceof_builtins.NoInstanceofBuiltinsRule,
 		no_invalid_fetch_options.NoInvalidFetchOptionsRule,
 		no_invalid_remove_event_listener.NoInvalidRemoveEventListenerRule,
--- a/internal/plugins/unicorn/rules/no_exports_in_scripts/no_exports_in_scripts.go
+++ b/internal/plugins/unicorn/rules/no_exports_in_scripts/no_exports_in_scripts.go
@@ -8,8 +8,10 @@ package no_exports_in_scripts
 
 import (
 	"github.com/microsoft/typescript-go/shim/ast"
+	"github.com/microsoft/typescript-go/shim/core"
 	"github.com/microsoft/typescript-go/shim/scanner"
 	"github.com/web-infra-dev/rslint/internal/rule"
+	"github.com/web-infra-dev/rslint/internal/utils"
 )
 
 const messageID = "no-exports-in-scripts"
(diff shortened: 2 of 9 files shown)


internal/plugins/unicorn/rules/no_exports_in_scripts/no_exports_in_scripts.go
```diff
@@ -0,0 +1,102 @@
...
+			// and the empty `export {};`. Each of these fires the
+			// listener and is reported directly.
+			ast.KindExportDeclaration: func(node *ast.Node) {
+				ctx.ReportNode(node, message())
+			},
+
+			// ---- tsgo divergence: ExportAssignment covers `export
+			// default ...` (ESTree ExportDefaultDeclaration) and
+			// `export = ...` (TypeScript's legacy CommonJS-interop form).
+			ast.KindExportAssignment: func(node *ast.Node) {
```
swwind, line 103: **[P2] Preserve the upstream TypeScript export node set**

With a shebang, unicorn v72 reports both `export import Alias = require('pkg')` and `export import Alias = Namespace.Value`, because typescript-estree wraps them in `ExportNamedDeclaration`; rslint emits no diagnostics because `KindImportEqualsDeclaration` is not registered. Conversely, upstream does not report `export = value` (`TSExportAssignment`), while the unconditional `KindExportAssignment` listener does. The same mismatch reproduces inside a namespace or ambient module. Please add `KindImportEqualsDeclaration` through `reportIfExported`, and report `KindExportAssignment` only when `!node.AsExportAssignment().IsExportEquals`.

internal/plugins/unicorn/rules/no_exports_in_scripts/no_exports_in_scripts.go
```diff
@@ -0,0 +1,102 @@
...
+	Schema: rule.EmptyArraySchema,
+	Run: func(ctx rule.RuleContext, options []any) rule.RuleListeners {
+		// Mirror upstream's early-out: only run when the first line of the
+		// file starts with `#!`. `scanner.GetShebang` returns the shebang
+		// substring (consumed up to and including the newline) or empty when
+		// no shebang is present; this matches ESLint's `sourceCode.lines[0]
+		// .startsWith('#!')` exactly — a shebang in a comment or string
+		// never matches because the file does not literally start with `#!`.
+		//
+		// tsgo divergence: tsgo's `SourceFile.Text()` strips a leading BOM
```
swwind: **[P3] Remove the incorrect BOM divergence**

ESLint strips a leading BOM before constructing `SourceCode.text` and `SourceCode.lines`, and rslint already provides the same BOM-transparent source-text abstraction before `scanner.GetShebang` runs. There is no tsgo divergence here. Please remove this entire BOM-specific discussion and the corresponding rule-level extra test.

internal/plugins/unicorn/rules/no_exports_in_scripts/no_exports_in_scripts.md
```diff
@@ -0,0 +1,59 @@
...
+
+A file without a shebang is a module, and modules are free to use exports:
+
+```javascript
+export const foo = 1;
+```
+
+## Original Documentation
+
+- [eslint-plugin-unicorn: no-exports-in-scripts](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-exports-in-scripts.md)
```
swwind: **[P3] Pin upstream references to the ported version**

The repository dependency is eslint-plugin-unicorn 72.0.0, but both references point to `main`, which can change independently. Please link the documentation and source to `v72.0.0`, including the reference in the Go source, so the implementation and migrated tests have a reproducible parity baseline.

swwind (changes requested):
Found one behavioral parity issue and two documentation accuracy issues; see the inline comments.

internal/plugins/unicorn/rules/no_exports_in_scripts/no_exports_in_scripts.md
guihua-c: Thanks for the review! Fixed all three issues in 1fcd9b7:
• Added support for exported TypeScript import-equals declarations.
• Excluded export = value to match upstream behavior.
• Removed the incorrect BOM divergence and its extra test.
• Pinned the upstream source and documentation links to v72.0.0.

internal/plugins/unicorn/rules/no_exports_in_scripts/no_exports_in_scripts.go
```diff
@@ -0,0 +1,96 @@
...
+		// .startsWith('#!')` exactly — a shebang in a comment or string
+		// never matches because the file does not literally start with `#!`.
+		text := ctx.SourceFile.Text()
+		if scanner.GetShebang(text) == "" {
+			return rule.RuleListeners{}
+		}
+
+		reportIfExported := func(node *ast.Node) {
+			if isExported(node) {
+				ctx.ReportNode(node, message())
```
swwind: [P2] Anchor decorated exports at the export wrapper

With a shebang, `@dec\n// eslint-disable-next-line\nexport class C {}` is suppressed by unicorn v72 because the reported ExportNamedDeclaration starts on the `export` line, but Rslint reports the flattened ClassDeclaration starting at `@dec`, so the directive does not suppress it. Moving the directive before `@dec` reverses the mismatch: Rslint suppresses the diagnostic while upstream reports it. The plain `@dec\nexport class C {}` and default-class forms likewise start one line too early in Rslint.

Please report from the authored `export` modifier through the declaration end instead of using the full decorated class node range, while retaining `export @dec class C {}` as starting at `export`, and add regressions for both disable-directive placements.

internal/plugins/unicorn/all.go
```diff
@@ -33,6 +34,7 @@ func GetAllRules() []rule.Rule {
 		new_for_builtins.NewForBuiltinsRule,
 		no_array_fill_with_reference_type.NoArrayFillWithReferenceTypeRule,
 		no_await_in_promise_methods.NoAwaitInPromiseMethodsRule,
+		no_exports_in_scripts.NoExportsInScriptsRule,
```
swwind, line 37: [P2] Complete the recommended-preset integration

unicorn v72 marks this rule as `recommended: 'unopinionated'`, but `packages/rslint/src/config/presets/unicorn.ts` still leaves `unicorn/no-exports-in-scripts` commented out as “not implemented”. Consequently, users of `unicorn.configs.recommended` do not receive the newly ported rule.

Please enable the existing preset entry and update the relevant preset integration expectation.

guihua-c: Thanks for the review! Both issues have been addressed in 91dbb3a, with regression tests added. All relevant Go/JS tests, type checks, formatting, and lint checks pass. PTAL.

swwind (approved):
LGTM, thanks 👍

PR #1918, opened 2026-08-27
feat(rstest): add require-test-timeout rule

elecmonkey:
## Summary

Report a test whose timeout is left to the project-wide default, so that a slow test is a visible decision rather than a silent one and a hang does not burn the default five seconds before anyone notices. The rule has no fix: what a test's timeout should be is a judgement about the code under test, not something a linter can supply.

A test is exempt when it has a timeout from any of the three places Rstest itself resolves one from. Its own, in either overload — the trailing number of `test(name, fn, 5000)` and the `timeout` property of `test(name, { timeout: 5000 }, fn)`, where `0` counts because Rstest spells "no limit" that way and a negative value does not. An enclosing suite's, because the runner applies suite options as defaults through `test.timeout ??= current.timeout` (`packages/core/src/runtime/runner/runtime.ts`), so `describe('slow', { timeout: 30_000 }, ...)` covers every test nested inside it. Or a configured runtime timeout, `rs.setConfig({ testTimeout })`, covering the tests that follow it in the source until a `resetConfig()` on the same binding puts the default back.

Registrations resolve through the shared `ParseTestCall`, so globals, named and renamed imports, namespace and whole-module `require`, `import.meta.rstest` and `@rstest/playwright` are all followed, and `.each` / `.for` and the modifier chains are included. `test.todo`, `test.skip` and a registration with no callback at all are left alone — the last of those belongs to `rstest/prefer-todo`. Registered in the rstest plugin but left out of the `recommended` preset, matching upstream.

The rule reports only what it can read. A timeout arriving through an expression it cannot follow — an imported constant, a function call, a spread into the options object, a `let` binding — is taken as a timeout the author wrote down. The same holds for a test inside a function that cannot be tied back to a `describe`, such as one handed to `describe` by name: it may well belong to a timed suite, so it is not reported. Attributing those needs a callback ownership index that this rule deliberately does not build.

The `setConfig` recogniser — `rs` and `rstest` under whatever local name they were imported as, a namespace member, a `require` binding, `import.meta.rstest`, and the unshadowed globals — is private to the rule and allocates nothing until such a call appears in the file. No shared Rstest analysis changed.

## Credits

Port from [`eslint-plugin-vitest`](https://github.com/vitest-dev/eslint-plugin-vitest)'s `require-test-timeout`.

- **A timeout on an enclosing `describe` exempts the tests inside it.** Rstest applies suite options as defaults to the tests they contain, so a test in a timed suite already has a definite timeout and reporting it is a false positive. Upstream models no inheritance, because it looks no further than the registration itself.

- **`resetConfig()` cancels a `setConfig({ testTimeout })`.** Placed between the two, it restores the default, and the tests after it are reported again. Upstream has no reset branch at all.

- **A timeout that cannot be read as a number is not reported.** Upstream reports `{ timeout: 'soon' }`, `{ ...options }`, a `let` binding and a value returned by a function through its `Number.NaN` branch. TypeScript already rejects a `timeout` that is not a number, so a lint report there adds nothing but the risk of flagging a test whose timeout is real and merely written somewhere the rule cannot follow.

- **A call with more arguments than either overload takes is left alone.** Neither `(name, fn?, timeout?)` nor `(name, options, fn?)` accepts a fourth argument, so such a call is not a registration this rule can reason about. Upstream scans every argument for a timeout regardless of position or count.

- **A callback passed by name is reported.** Upstream looks for an inline function argument and skips the rest, though `const handler = () => {}; test('a', handler)` needs a timeout no less than an inline callback does. The resolved function is recognised as the callback in whichever argument slot it occupies, so it is never mistaken for the options object or the timeout.

- **No `x`-prefixed branch.** Upstream skips a registration whose name starts with `x`. Rstest exports no `xit`, `xtest` or `xdescribe`, so there is nothing to skip.

## Related Links

Related #935

## Checklist

### DIFF
--- a/internal/plugins/rstest/rules/require_test_timeout/require_test_timeout.go
+++ b/internal/plugins/rstest/rules/require_test_timeout/require_test_timeout.go
@@ -20,8 +20,9 @@ var missingTimeoutMessage = rule.RuleMessage{
 //
 // The rule only reports on timeoutAbsent and timeoutNegative: everything it
 // cannot read is timeoutUnknown, which exempts. Upstream instead reports the
-// unreadable shapes (its `Number.NaN` branch), and this port deliberately
-// diverges — see require_test_timeout.md.
+// unreadable shapes through its `Number.NaN` branch; reporting a timeout that
+// is merely written somewhere unreadable is a false positive, and TypeScript
+// already rejects a `timeout` that is not a number.
 type timeoutFinding uint8
 
 const (
--- a/internal/plugins/rstest/rules/require_test_timeout/require_test_timeout.go
+++ b/internal/plugins/rstest/rules/require_test_timeout/require_test_timeout.go
@@ -1,0 +1,564 @@
+package require_test_timeout
+
+import (
+	"strconv"
+
+	"github.com/microsoft/typescript-go/shim/ast"
+	rstestUtils "github.com/web-infra-dev/rslint/internal/plugins/rstest/utils"
+	"github.com/web-infra-dev/rslint/internal/rule"
+	internalUtils "github.com/web-infra-dev/rslint/internal/utils"
+	testFramework "github.com/web-infra-dev/rslint/internal/utils/test_framework"
+)
+
+var missingTimeoutMessage = rule.RuleMessage{
+	Id:          "missingTimeout",
+	Description: "Test is missing a timeout. Add an explicit timeout.",
+}
+
+// timeoutFinding is what a static read of one registration argument, one
+// options object or one runtime-config call was able to conclude.
+//
+// The rule only reports on timeoutAbsent and timeoutNegative: everything it
+// cannot read is timeoutUnknown, which exempts. Upstream instead reports the
+// unreadable shapes (its `Number.NaN` branch), and this port deliberately
+// diverges — see require_test_timeout.md.
+type timeoutFinding uint8
+
+const (
+	// timeoutAbsent means the shape was fully readable and declares no timeout.
+	timeoutAbsent timeoutFinding = iota
+	// timeoutDeclared means a timeout of zero or more was written down.
+	timeoutDeclared
+	// timeoutNegative means a timeout below zero was written down. Rstest has
+	// no meaning for it, and `0` already spells "no limit", so the author
+	// still owes the test a real timeout.
+	timeoutNegative
+	// timeoutUnknown means the shape could not be read statically.
+	timeoutUnknown
+)
+
+var RequireTestTimeoutRule = rule.Rule{
+	Name:   "rstest/require-test-timeout",
+	Schema: rule.EmptyArraySchema,
+	Run: func(ctx rule.RuleContext, options []any) rule.RuleListeners {
+		analysis := rstestUtils.GetRstestCallAnalysis(ctx)
+		runtimeConfig := runtimeConfigTracker{ctx: ctx}
+
+		return rule.RuleListeners{
+			ast.KindCallExpression: func(node *ast.Node) {
+				// Every `setConfig` / `resetConfig` call has to be seen before
+				// the tests that follow it, which the source-order walk already
+				// guarantees; the tracker itself allocates nothing until one of
+				// those calls actually appears in the file.
+				runtimeConfig.observe(node)
+
+				parsed := analysis.ParseTestCall(node)
+				if parsed == nil || parsed.Todo || parsed.Skipped {
+					return
+				}
+				// A registration with no callback declares no test body to time
+				// out. Reporting it belongs to `rstest/prefer-todo`.
+				if callback, name := analysis.TestCallback(node); callback == nil && name == "" {
+					return
+				}
+				if runtimeConfig.exemptsAt(node.Pos()) {
+					return
+				}
+				if inheritsSuiteTimeout(ctx, analysis, node) {
+					return
+				}
+				switch registrationTimeout(ctx, node) {
+				case timeoutDeclared, timeoutUnknown:
+					return
+				}
+
+				ctx.ReportNode(node, missingTimeoutMessage)
+			},
+		}
+	},
+}
+
+// registrationTimeout reads the timeout a `test` or `describe` registration
+// declares for itself.
+//
+// Rstest has exactly two overloads — `(name, fn?, timeout?)` and
+// `(name, options, fn?)` (packages/core/src/types/api.ts:91-94) — so only the
+// second and third arguments can carry a timeout, and a call with more
+// arguments than that matches neither overload and is left alone.
+func registrationTimeout(ctx rule.RuleContext, node *ast.Node) timeoutFinding {
+	call := node.AsCallExpression()
+	if call == nil || call.Arguments == nil {
+		return timeoutAbsent
+	}
+	arguments := call.Arguments.Nodes
+	if len(arguments) > 3 {
+		return timeoutUnknown
+	}
+
+	finding := timeoutAbsent
+	for index := 1; index < len(arguments) && index <= 2; index++ {
+		switch argument := argumentTimeout(ctx, arguments[index]); argument {
+		case timeoutNegative:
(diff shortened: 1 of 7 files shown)


internal/plugins/rstest/rules/require_test_timeout/require_test_timeout.go
```diff
@@ -0,0 +1,564 @@
...
+// runtimeConfigTracker follows the runtime timeout configured through the
+// Rstest utility object. Positions are compared rather than scopes because
+// `setConfig` takes effect from the moment it runs, which for a file of
+// top-level registrations is exactly "from this point in the source down".
+type runtimeConfigTracker struct {
+	ctx    rule.RuleContext
+	events []runtimeConfigEvent
+}
+
+func (tracker *runtimeConfigTracker) observe(node *ast.Node) {
```
swwind, line 413: `observe` treats every syntactically earlier `setConfig` call as executed, including calls inside functions that are never invoked and test callbacks that only run after collection. For example, `function configure() { rs.setConfig({ testTimeout: 5000 }) } test("a", () => {})` produces no diagnostic even though `configure` never runs. Please only apply configuration changes from execution contexts that run during test collection.

internal/plugins/rstest/rules/require_test_timeout/require_test_timeout.go
```diff
@@ -0,0 +1,564 @@
...
+			local,
+			receiver,
+			ctx.TypeChecker,
+			ctx.SourceFile,
+			rstestUtils.RstestImportModule,
+		)
+		if name != "rs" && name != "rstest" {
+			return "", false
+		}
+		return "identifier:" + local, true
```
swwind: `rs` and `rstest` are aliases of the same runtime utilities instance (`api/index.ts` assigns `rs: rstest`), but this key uses the local identifier name. Consequently, `rs.setConfig(...); rstest.resetConfig(); test(...)` leaves the rule's `rs` state enabled and misses the test. Please canonicalize all recognized aliases of this singleton to the same configuration state.

internal/plugins/rstest/rules/require_test_timeout/require_test_timeout.go
```diff
@@ -0,0 +1,564 @@
...
+			continue
+		}
+		call := enclosingCallOfArgument(node)
+		if call == nil {
+			return true
+		}
+		parsed := analysis.ParseFnCall(call)
+		if parsed != nil &&
+			parsed.Kind == rstestUtils.RstestFnTypeDescribe &&
+			registrationTimeout(ctx, call) != timeoutAbsent {
```
swwind: This treats `timeoutNegative` as an inherited timeout because every result except `timeoutAbsent` returns true. Both `describe("s", { timeout: -1 }, () => { test("a", () => {}) })` and the trailing `-1` form suppress the diagnostic, although negative timeouts are unusable according to this rule's own semantics. Please exempt only `timeoutDeclared` or `timeoutUnknown`.

internal/plugins/rstest/rules/require_test_timeout/require_test_timeout.go
```diff
@@ -0,0 +1,564 @@
...
+		key, ok := internalUtils.GetStaticPropertyName(name)
+		if !ok {
+			return timeoutUnknown
+		}
+		if key != property {
+			continue
+		}
+		switch member.Kind {
+		case ast.KindPropertyAssignment:
+			return scalarTimeout(ctx, member.AsPropertyAssignment().Initializer)
```
swwind: `objectTimeout` returns after the first `timeout` property, so later members never affect the result. For example, `test("a", { timeout: -1, ...options }, () => {})` is reported even though the later spread may provide a usable timeout and should make the object unreadable under the documented conservative policy. Later explicit overrides are also ignored. Please scan the complete object and combine members in source order.

internal/plugins/rstest/rules/require_test_timeout/require_test_timeout.go
```diff
@@ -0,0 +1,564 @@
...
+		offset: node.End(),
+		sets:   true,
+	})
+}
+
+// exemptsAt reports whether a test starting at offset runs under a configured
+// test timeout. Events are recorded in source order by the rule's own walk, so
+// replaying the ones that end before the test reproduces the state the runtime
+// would be in when the test is registered.
+func (tracker *runtimeConfigTracker) exemptsAt(offset int) bool {
```
swwind, line 438: `exemptsAt` allocates a new map and replays every recorded configuration event for every test. With N configuration events followed by N tests, this becomes O(N²) work plus N map allocations. Please build the effective configuration state incrementally or precompute prefix state so each test lookup does not replay the complete history.

internal/plugins/rstest/rules/require_test_timeout/require_test_timeout.go
elecmonkey: Fixed in 71c69ea. `objectTimeout` now scans the whole object and combines members in source order, so a spread or a computed key after an explicit `timeout` erases what came before it. That also fixes the mirror case `{ timeout: -1, timeout: 500 }`, where the effective value is 500. `{ ...options, timeout: -1 }` is still reported, since the explicit member has the last word there. Both orderings and the duplicate-key case are locked in as tests.

internal/plugins/rstest/rules/require_test_timeout/require_test_timeout.go
elecmonkey: Fixed in 71c69ea. Suite inheritance now exempts only `timeoutDeclared` and `timeoutUnknown`. You are right that this was inconsistent with the rule’s own semantics: a negative value is refused on the test itself and documented as not counting, so inheriting one should not silence the diagnostic. `describe("s", { timeout: -1 }, ...)` and the trailing `-1` form both have tests now.

internal/plugins/rstest/rules/require_test_timeout/require_test_timeout.go
elecmonkey: Fixed in 71c69ea. All the spellings are one instance, so the tracker keeps one state instead of one per binding, and a `resetConfig()` on any of them cancels a `setConfig()` on any other. `rs` / `rstest`, `import.meta.rstest`, and a namespace member are covered by tests.

internal/plugins/rstest/rules/require_test_timeout/require_test_timeout.go
elecmonkey: Fixed in 71c69ea. The state is folded in incrementally now, so there is no per-test map and no replay: a lookup is a boolean and an offset comparison. Collapsing the per-binding keys for the alias issue above reduced the state to a single flag, which made this straightforward. The offset is still kept so a test nested inside the configuration call’s own arguments is not exempted by it, which is the one ordering the source-order walk cannot rule out.

internal/plugins/rstest/rules/require_test_timeout/require_test_timeout.go
elecmonkey: Thanks — the description is accurate, but I would like to keep the current behaviour, so I have left this one open rather than resolving it.

The rule decides exemption by source position: a `setConfig` whose call ends before the test begins exempts it. That is deliberate, and it is also what the equivalent Vitest rule does. Your example does slip through, as does `test("a", () => { rs.setConfig({ testTimeout: 5000 }) }, 5); test("b", () => {})`, which is the same shape.

Both are missed diagnostics rather than false reports, and this rule is built to err that way — a test whose timeout it cannot follow is left alone rather than flagged. Telling which functions actually run during collection needs call-graph analysis, and tightening it without that would mean reporting tests that do have a configured timeout.

The narrower framing is that this rule cannot model execution order, only source order, and I would rather that limitation stay visible here than be half-addressed. Happy to revisit if you would rather see it handled.

fansenze (approved):

PR #1919, opened 2026-08-27
feat(rstest): add no-conditional-tests rule

elecmonkey:
## Summary

Report a test or suite registration that is written inside an `if` statement's `then` or `else` branch, since a registration that only runs on some executions is a test whose coverage silently shrinks instead of failing. No fix is offered.

The call resolves through the shared `ParseFnCall`, so only a real Rstest `test` or `describe` registration is reported — hooks, a same-named export from another test framework, and a local variable that merely shares the name are all left alone. The walk from the registration call up to its enclosing scope stops at the first deferred-execution boundary — a function, arrow, method, constructor, or accessor body, a parameter default, or an instance field initializer — or at the source file. It never follows a function to its call sites, so a registration wrapped in a helper is not reported at all, and a nested pair such as `if (x) { describe('a', () => { test('b', fn) }) }` reports only the outer `describe`. Positions that run while a class is defined — static blocks, static field initializers, computed member names, decorators — are not boundaries, so a registration in one stays conditional on an enclosing `if`. A call sitting in the `if`'s own condition, such as `if (test('a')) {}`, is not reported — it runs unconditionally. Registered in the rstest plugin but left out of the `recommended` preset, matching upstream.

## Credits

Port from [`eslint-plugin-vitest`](https://github.com/vitest-dev/eslint-plugin-vitest)'s `no-conditional-tests`. `eslint-plugin-jest` has no counterpart, so there is no second baseline to cross-check against.

- **The registration is resolved instead of name-matched.** Upstream reports any `Identifier` literally spelled `test`, `it`, or `describe`, so an ordinary variable with one of those names is reported even though it has nothing to do with the test framework. This port only reports a call it can resolve to a Rstest registration through `ParseFnCall`, which also lets it follow imports from `@rstest/core` and `@rstest/playwright`, `require` forms, `import.meta.rstest` members, globals, and local aliases, while staying silent on a shadowed name or a same-named API from Vitest, Jest, or `node:test`.

- **The parent walk follows the enclosing scope instead of a fixed depth.** Upstream checks whether the identifier's parent four levels up is exactly an `IfStatement`, which assumes the registration is a direct expression statement inside a braced block. `if (cond) test('a', fn)`, written without braces, puts the call three levels below the `if` instead of four, and upstream misses it. This port walks up to the first function boundary or the source file instead, so the shape and spacing of the intervening statements no longer matter, and the same walk naturally reports only the outermost registration of a nested pair.

## Related Links

Related #935

## Checklist

### DIFF
--- a/internal/plugins/rstest/rules/no_conditional_tests/no_conditional_tests.go
+++ b/internal/plugins/rstest/rules/no_conditional_tests/no_conditional_tests.go
@@ -86,11 +86,11 @@ func registeredUnderIfBranch(node *ast.Node) bool {
 // controls parent's own evaluation — so a registration on the far side of it
 // is no longer conditional on anything wrapping parent.
 //
-// A function, method, constructor or accessor body, and a parameter's default
-// value, qualify: neither runs when the declaration itself is evaluated, only
-// later when the function is called. An instance field initializer
-// (`p = expr`, no `static`) qualifies too: it runs once per `new`, not when
-// the class declaration is evaluated.
+// A function, method, constructor or accessor body, and the runtime-evaluated
+// parts of a parameter, qualify: none runs when the declaration itself is
+// evaluated, only later when the function is called. An instance field
+// initializer (`p = expr`, no `static`) qualifies too: it runs once per `new`,
+// not when the class declaration is evaluated.
 //
 // A class static block, a static field initializer (`static p = expr`), a
 // computed member name (`[expr]`), and a decorator (`@dec(expr)`) do not
@@ -106,11 +106,18 @@ func isDeferredExecutionBoundary(child *ast.Node, parent *ast.Node) bool {
 	switch parent.Kind {
 	case ast.KindComputedPropertyName, ast.KindDecorator, ast.KindClassStaticBlockDeclaration:
 		return false
+	case ast.KindParameter:
+		// Parameter decorators run while the enclosing class is evaluated.
+		// Defaults and binding patterns run later, when the function is called.
+		return child.Kind != ast.KindDecorator
 	case ast.KindPropertyDeclaration:
-		return !ast.HasStaticModifier(parent)
+		// A non-static field's initializer is deferred until construction, but
+		// its computed name and decorators run while the class is evaluated.
+		property := parent.AsPropertyDeclaration()
+		return !ast.HasStaticModifier(parent) && child == property.Initializer
 	}
 	if !ast.IsFunctionLikeDeclaration(parent) {
 		return false
 	}
-	return child == parent.Body() || child.Kind == ast.KindParameter
+	return child == parent.Body()
 }
--- a/internal/plugins/rstest/rules/no_conditional_tests/no_conditional_tests.go
+++ b/internal/plugins/rstest/rules/no_conditional_tests/no_conditional_tests.go
@@ -50,13 +50,13 @@ var NoConditionalTestsRule = rule.Rule{
 // registeredUnderIfBranch walks up from a registration call and reports whether
 // it is reached through the then or else branch of an `if`.
 //
-// The walk stops at the first function boundary, which is what keeps the rule
-// honest in two directions. Downwards, it means only the registration that the
-// `if` itself decides to run is reported: in
+// The walk stops at the first deferred-execution boundary, which is what keeps
+// the rule honest in two directions. Downwards, it means only the registration
+// that the `if` itself decides to run is reported: in
 // `if (x) { describe('a', () => { test('b', fn) }) }` the inner `test` walk
-// stops at the describe callback, so only the outer `describe` is reported.
-// Upwards, it means a condition inside an unrelated enclosing function never
-// leaks onto a registration that function merely contains.
+// stops at the describe callback's body, so only the outer `describe` is
+// reported. Upwards, it means a condition inside an unrelated enclosing
+// function never leaks onto a registration that function merely contains.
 //
 // Arriving through an `if`'s condition is not a conditional registration —
 // `if (test('a')) {}` runs the call unconditionally — so the walk continues
@@ -68,7 +68,7 @@ func registeredUnderIfBranch(node *ast.Node) bool {
 		if parent == nil || parent.Kind == ast.KindSourceFile {
 			return false
 		}
-		if ast.IsFunctionLikeOrClassStaticBlockDeclaration(parent) {
+		if isDeferredExecutionBoundary(child, parent) {
 			return false
 		}
 		if parent.Kind == ast.KindIfStatement {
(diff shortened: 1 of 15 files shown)


internal/plugins/rstest/rules/no_conditional_tests/no_conditional_tests.go
```diff
@@ -0,0 +1,82 @@
...
+// `if (test('a')) {}` runs the call unconditionally — so the walk continues
+// past it rather than reporting, and an outer `if` can still catch it.
+func registeredUnderIfBranch(node *ast.Node) bool {
+	child := node
+	for {
+		parent := child.Parent
+		if parent == nil || parent.Kind == ast.KindSourceFile {
+			return false
+		}
+		if ast.IsFunctionLikeOrClassStaticBlockDeclaration(parent) {
```
swwind: `ast.IsFunctionLikeOrClassStaticBlockDeclaration` conflates distinct execution phases into a single boundary. Static blocks, computed property names, and method decorators within a class are executed when the class definition is evaluated; thus, the registration within `if (flag) { class C { static { test(...) } [test(...)]() {} @dec(test(...)) m() {} } }` is indeed controlled by the `if` statement, yet the current logic returns early and fails to report it. Conversely, instance fields like `p = test(...)` execute only upon instance construction; the current traversal passes through the field declaration and incorrectly attributes it to the `if` statement enclosing the class declaration. Please redefine the boundaries based on the type of class member and the position of the current child node: traversal should halt for function/method bodies, parameter default values, and instance field initializers, but should continue for static blocks, static initializers, computed property names, and decorators; additionally, please add corresponding regression tests.

elecmonkey: Confirmed — thanks for the precise breakdown. Fixed in 5346b3b3b5: the boundary check now distinguishes deferred-execution positions (function/method/constructor/accessor bodies, parameter default values, instance field initializers) from synchronous class-evaluation-time positions (static blocks, static field initializers, computed member names, decorators). Added regression tests for all four non-boundary shapes plus the instance-field-initializer boundary case. Also corrected the shared batch spec's boundary definition, since it previously listed class static block as a stopping point.

swwind (commented):
isDeferredExecutionBoundary treats every instance PropertyDeclaration and every method Parameter as deferred. That misses registration expressions which execute during class evaluation rather than during instance construction or a later method call:

```js
if (x) {
   class C {
    [test("a", () => {})] = 1;
  }
}

if (x) {
  class C {
    @dec(test("a", () => {}))
    p = 1;
  }
}

if (x) {
  class C {
    method(@dec(test("a", () => {})) value: unknown) {}
  }
}
```

All three registrations are conditional but currently go unreported. Please distinguish a field initializer / parameter default (deferred) from a computed name or decorator (synchronous class evaluation).

----

The shared registration parser does not recognize a comma-wrapped test callee:

```js
if (x) {
  (0, test)("a", () => {});
}
```

This still invokes the real Rstest test registration conditionally, but the rule emits no diagnostic. Consider unwrapping a comma expression whose right operand is the resolved registration reference, or document this intentional parser limitation.

elecmonkey:
@swwind Fixed both points in 99d9169781.

- Made class-member execution boundaries position-sensitive, so computed field names, field decorators, and parameter decorators are treated as class-evaluation-time expressions, while instance initializers and parameter defaults remain deferred.
- Updated the shared Rstest call parser to unwrap comma expressions through their right operand, including imports, aliases, hooks, and `import.meta.rstest` calls.
- Added dedicated class-timing regressions and updated the shared parser and `no-hooks` contract tests.

The affected Go tests, binary build, JS integration test, typecheck, lint, spell check, formatting check, and targeted Go lint all pass.

internal/plugins/rstest/rules/no_conditional_tests/no_conditional_tests.go
```diff
@@ -0,0 +1,123 @@
...
+		Id:          "noConditionalTests",
+		Description: "Avoid using if conditions in a test",
+	}
+}
+
+var NoConditionalTestsRule = rule.Rule{
+	Name:   "rstest/no-conditional-tests",
+	Schema: rule.EmptyArraySchema,
+	Run: func(ctx rule.RuleContext, options []any) rule.RuleListeners {
+		analysis := rstestUtils.GetRstestCallAnalysis(ctx)
```
swwind, line 21: [P2] Preserve registration resolution in source-only programs

This rule does not require type information, but `GetRstestCallAnalysis` loses binding resolution when `ctx.TypeChecker` is nil. In a focused source-only `.ts`/`.js` multi-file run, `import { test } from 'vitest'; if (flag) { test(...) }` and a locally declared `test` were incorrectly reported as Rstest globals, while renamed and namespace imports from `@rstest/core` were missed. Please use the file-local reference store as the fallback when resolving the root symbol, or require type information for this rule; otherwise the source discrimination promised by this port reverses in both directions outside checker-backed programs.

internal/plugins/rstest/rules/no_conditional_tests/no_conditional_tests.md
```diff
@@ -0,0 +1,52 @@
...
+
+The rule does not carry an enclosing `if` through code that runs later. A
+registration in a function or method body, a parameter default, or an instance
+field initializer is attributed to the later call or construction rather than
+to an `if` around its declaration. Class static blocks, static field
+initializers, computed member names, and decorators run while the class is
+defined, so registrations in those positions remain conditional on an
+enclosing `if`.
+
+A registration wrapped in a helper is therefore attributed to wherever that
```
swwind: [P3] Do not claim helper call-site attribution

The implementation stops when it crosses a function body and does not inspect where that function is called. Consequently, `function register() { test("a", fn) } if (flag) { register() }` emits no diagnostic, as does an equivalent IIFE inside the branch. Please reword this paragraph to say only that conditions do not cross deferred-execution boundaries, or implement the claimed call-site attribution.

internal/plugins/rstest/rules/no_conditional_tests/no_conditional_tests.go
elecmonkey: Fixed in a62280f0.

`resolveRstestRootSymbol` now falls back to `ctx.Refs.ResolveInFile` when no TypeChecker is available, and `GetRstestCallAnalysis` carries `Refs` into the cached analysis context. The binder scope walk places imports and declarations authored in the file, which covers every discrimination the parser promises; bindings declared outside the file were never registrations.

The degradation lived in the shared call parser, so every rstest rule resolving through `ParseFnCall` picks up the fix, not only this one.

Added `TestNoConditionalTestsResolvesInSourceOnlyProgram`, which lints a checker-free `NewFromRoots` Program and asserts that `import { test } from 'vitest'` and a local `const localTest` are not reported, while a renamed import, a namespace import, and a plain import from `@rstest/core` are.

internal/plugins/rstest/rules/no_conditional_tests/no_conditional_tests.md
elecmonkey: Reworded in a62280f0. The doc no longer claims call-site attribution: it now says the rule does not follow a function to its call sites, so a registration in a helper is never reported, whether or not the helper is only ever called from inside an `if`. The paragraph above it likewise says such registrations are left alone, rather than being "attributed to the later call or construction".

swwind (approved):

PR #1920, opened 2026-08-27
feat(rstest): add require-awaited-expect-poll rule

elecmonkey:
## Summary

Report an `expect.poll(...)` or `expect.element(...)` assertion whose promise is dropped. Both factories are asynchronous — `expect.poll(fn)` re-evaluates `fn` until its matcher passes or the timeout expires, and every matcher `expect.element(locator)` exposes is declared `async` — so a dropped promise lets the test finish before the assertion does, and a failing assertion either surfaces against an unrelated test or does not surface at all.

Both factories are real and independently modeled by the shared Rstest parser (`RstestExpectEntryPoll` / `RstestExpectEntryElement`), so no parser changes were needed to cover them. Assertions resolve through the shared `ParseExpectCall`, so globals, named and renamed imports, namespace and whole-module `require`, `import.meta.rstest`, `@rstest/playwright` and the `expect` a test callback receives through its context are all followed, and an `expect` from Vitest, Jest, Playwright or Chai is not. Registered in the rstest plugin but left out of the `recommended` preset, matching upstream. There is no fix: whether a dropped assertion should be awaited, returned, or collected with the other promises around it is a decision about the test, not a mechanical rewrite.

## Credits

Port from [`@vitest/eslint-plugin`](https://github.com/vitest-dev/eslint-plugin-vitest)'s `require-awaited-expect-poll`. It is Vitest-only; `eslint-plugin-jest` has no counterpart.

- **A promise that is bound or passed on, not just awaited or returned, counts as handled.** Upstream accepts only `await` and `return`. This port also accepts a concise arrow body, a variable initializer, the right-hand side of an assignment, a call argument (covering `Promise.all([...])` and `Promise.allSettled([...])`), an array element, a `yield` operand, and an object property value — every shape where the promise is consumed rather than dropped. The trade-off is deliberate: false positives on these shapes would be worse than the false negatives they trade away.

- **Parenthesized and TypeScript-wrapped assertions are still recognised as awaited.** `await (expect.poll(fn).toBe(1) as Promise<void>)` and `await (expect.poll(fn).toBe(1)!)` are accepted here. tsgo keeps parentheses and TypeScript assertions as real nodes where ESTree has none, so upstream's walk — which stops at the first parenthesis-shaped gap — reports these; this port walks through the wrapper to find the `await` on the other side.

## Related Links

Related #935

## Checklist


internal/plugins/rstest/rules/require_awaited_expect_poll/require_awaited_expect_poll.go
```diff
@@ -0,0 +1,154 @@
...
+	parent := node.Parent
+	if parent == nil {
+		return false
+	}
+	switch parent.Kind {
+	case ast.KindAwaitExpression, ast.KindReturnStatement:
+		return true
+	case ast.KindArrowFunction:
+		return parent.AsArrowFunction().Body == node
+	case ast.KindVariableDeclaration:
```
swwind, line 184: [P2] Recognize promises bound by destructuring defaults

Under this rule's handled-promise contract, `const [p = expect.poll(fn).toBe(1)] = values;` should be accepted just like a direct variable initializer: when the default runs, its promise is bound to `p`. Here ts-go places the initializer under a `BindingElement`, so this switch falls through and reports it. The same false positive occurs for object and nested bindings, `for...of` declarations, and parameter defaults; object shorthand assignment defaults such as `({p = expect.poll(fn).toBe(1)} = values)` are also reported even though the equivalent array assignment default is accepted. Please recognize these initializer positions so the result does not depend on the pattern's AST shape.

elecmonkey: Confirmed — this was a real false positive, not three separate issues: `BindingElement` (array/object destructuring defaults, including nested patterns and `for...of` declarations), `ParameterDeclaration` (plain and destructured parameter defaults), and `ShorthandPropertyAssignment` (the object assignment-pattern default, e.g. `({p = expect.poll(fn).toBe(1)} = values)`) were all missing from the handled-position switch and fell through to "not handled". The array assignment-pattern default (`[p = expect.poll(fn).toBe(1)] = values`) already worked by coincidence, since ts-go parses that one as an ordinary assignment `BinaryExpression` rather than a dedicated node — which is exactly the asymmetry you pointed out.

Fixed in e5a4c761c5: added the three missing cases to `isHandled`, plus lock-in tests for all of the shapes you listed (array/object/nested destructuring declarations, `for...of`, plain and destructured parameter defaults, and both assignment-pattern forms).

swwind (commented):
The widened “handled promise” policy is incomplete. isHandled only recognizes a few immediate-parent forms (and topMostConsumedNode only propagates through comma expressions), so it reports assertions whose promise is still bound or passed onward:

```tsx
class Pending {
  visible = expect.poll(() => el).toBeVisible();
  accessor loaded = expect.element(el).toBeVisible();
}

new Collector(expect.poll(() => el).toBeVisible());

const pending = ready
  ? expect.poll(() => el).toBeVisible()
  : fallback;

const pending = ready && expect.element(el).toBeVisible();

const view = <Widget assertion={expect.element(el).toBeVisible()} />;
```

These are inconsistent with currently accepted direct object-property values, call arguments, and variable initializers. Please either propagate consumption through these enclosing expression/container forms, or narrow the documented intentional difference so the rule does not promise to accept every bound/passed-on promise.

elecmonkey:
Fixed in f26bb1db by propagating handled-value detection through class field/accessor initializers, constructor arguments, conditional and logical expressions, and JSX expression containers. Added regression coverage for all reported shapes, including assertions in discarded branches/operands so only values that can reach the enclosing handled position are accepted.

internal/plugins/rstest/rules/require_awaited_expect_poll/require_awaited_expect_poll.go
```diff
@@ -0,0 +1,209 @@
...
+	case ast.KindCallExpression:
+		return parent.AsCallExpression().Expression != node
+	case ast.KindNewExpression:
+		return parent.AsNewExpression().Expression != node
+	case ast.KindJsxExpression:
+		jsxExpression := parent.AsJsxExpression()
+		return jsxExpression.DotDotDotToken == nil && jsxExpression.Expression == node
+	case ast.KindBinaryExpression:
+		binary := parent.AsBinaryExpression()
+		return binary.OperatorToken != nil &&
```
swwind, line 209: [P2] Do not treat arithmetic compound assignments as handling the assertion promise

`ast.IsAssignmentOperator` includes arithmetic, shift, and bitwise compound assignments, so code such as:

```ts
count += expect.poll(fn).toBe(1);
bits |= expect.element(el).toBeVisible();
```

receives no diagnostic. These operators coerce the Promise and store the operation result; they do not bind, pass onward, or settle the assertion Promise itself. The polling assertion is therefore still dropped.

The pinned upstream rule reports these cases, and focused probes reproduce the false negative for `+=`, `-=`, `*=`, `/=`, `%=`, `**=`, shift assignments, and bitwise assignments. Please restrict this handled position to plain `=` and the logical assignments (`||=`, `&&=`, `??=`), where the evaluated right-hand Promise itself is assigned, and add regression cases for both groups.

elecmonkey: Confirmed — `ast.IsAssignmentOperator` covers all 16 assignment operators, so the arithmetic, shift and bitwise compound assignments fell into the handled position along with `=`. Those operators coerce the promise to a primitive and store the result of the operation, so the assertion promise is dropped exactly as if the assignment were not there; only `=` and the logical assignments store the evaluated right-hand promise itself.

Fixed in 4a8b9215: the handled assignment position now goes through a dedicated operator check that accepts only `=`, `||=`, `&&=` and `??=`. Added regression cases for both groups — valid for the three logical assignments, invalid for `+=`, `-=`, `*=`, `/=`, `%=`, `**=`, `<<=`, `>>=`, `>>>=`, `&=`, `|=` and `^=` — and narrowed the rule doc's wording accordingly.

internal/plugins/rstest/rules/require_awaited_expect_poll/require_awaited_expect_poll.go
```diff
@@ -0,0 +1,226 @@
...
+		return true
+	case ast.KindArrowFunction:
+		return parent.AsArrowFunction().Body == node
+	case ast.KindVariableDeclaration:
+		return parent.AsVariableDeclaration().Initializer == node
+	case ast.KindPropertyDeclaration:
+		return parent.AsPropertyDeclaration().Initializer == node
+	case ast.KindPropertyAssignment:
+		return parent.AsPropertyAssignment().Initializer == node
+	case ast.KindYieldExpression:
```
swwind, line 190: This also accepts `yield*`. Unlike `yield assertion`, delegation obtains an iterator from the assertion's Promise; it neither yields nor awaits that Promise (native Promises are not iterable). Consequently `function* run() { yield* expect.poll(() => el).toBeVisible(); }` gets no diagnostic although the assertion is never awaited. Restrict this branch to non-delegating `yield` expressions and add this as an invalid case.

elecmonkey: Confirmed — `yield` and `yield*` share one node kind in ts-go, separated only by
`AsteriskToken`, so the handled position accepted both. They are not the same:
a plain `yield` hands the promise to whatever drives the generator, which can
await it, while `yield*` looks up an iterator on the promise and a native
promise has neither `Symbol.iterator` nor `Symbol.asyncIterator`, so the
assertion is never passed on or awaited.

Fixed in 2d63682d: the branch now requires the yield to be non-delegating.
Added invalid cases for `yield*` in both a plain and an async generator, and
narrowed the rule doc and the accompanying comment to say non-delegating
`yield`.

swwind:
Looks good to me (abbr. LGTM).

swwind (approved):

PR #1921, opened 2026-08-27
feat(rstest): add prefer-called-once rule

elecmonkey:
## Summary

Report `expect(fn).toHaveBeenCalledTimes(1)` and its deprecated alias `expect(fn).toBeCalledTimes(1)`, and rewrite both to `expect(fn).toHaveBeenCalledOnce()`, so a call-count assertion for exactly one call names "once" in the matcher rather than in an argument the reader has to check. Only a literal `1` counts — `0`, `2`, a variable, `1n`, or `+1` are all left alone. The rule is autofixable.

The matcher has to be called — `expect(fn).toHaveBeenCalledTimes` asserts nothing — and an assertion factory has to have run, so `expect.toHaveBeenCalledTimes(1)`, which never received a value to count calls on, is left alone. Chai's call-count assertion, `expect(fn).callCount(1)`, is not covered: its "once" spelling is the `calledOnce` property rather than a matcher call.

Assertions resolve through the shared `ParseExpectCall`, so globals, named and aliased imports, namespace and whole-module `require`, `import.meta.rstest`, `@rstest/playwright` and the `expect` a test callback receives through its context are all followed. Registered in the rstest plugin but left out of the `recommended` preset, matching upstream.

The fix edits two places and leaves the rest of the assertion exactly as written: it renames the matcher accessor to `toHaveBeenCalledOnce`, and empties the argument list between the parentheses. The expect root as spelled at the call site, `expect.soft`, the optional second `message` argument of `expect(actual, message)`, the modifier chain, the accessor's own quoting and the surrounding line breaks and comments all survive. A comment written between the parentheses would be deleted along with the emptied argument list, so that shape is reported without a fix.

## Credits

Port from [`eslint-plugin-vitest`](https://github.com/vitest-dev/eslint-plugin-vitest)'s `prefer-called-once`.

- **The fix always writes `toHaveBeenCalledOnce`, never `toBeCalledOnce`.** Upstream derives the replacement by substituting `Once` for `Times` in whichever matcher was written, so `expect(fn).toBeCalledTimes(1)` becomes `expect(fn).toBeCalledOnce()`. `@vitest/expect@4.1.10`, which Rstest depends on, declares `toHaveBeenCalledOnce` but has no `toBeCalledOnce` at all — neither `dist/index.d.ts` nor `dist/index.js` mentions it — so upstream's fix would turn a passing assertion into a call to an undefined matcher. This port writes the one matcher name that exists, regardless of which count matcher was read.

- **The whole argument span between the parentheses is removed, not just the argument node.** Upstream removes only the `1` node, so `expect(fn).toHaveBeenCalledTimes(1,)` becomes `expect(fn).toHaveBeenCalledOnce(,)`, which does not parse. This port removes everything between the matcher call's parentheses, including a trailing comma or multi-line whitespace.

- **Element-access fixes preserve the accessor's delimiters.** `['toBeCalledTimes']` becomes `['toHaveBeenCalledOnce']` rather than being normalized to a bare identifier, matching the rest of the plugin.

- **The opening parenthesis is scanned for, not assumed to follow the matcher name.** `expect(fn)['toBeCalledTimes'](1)`, a call split across lines, and a comment between the matcher name and its parentheses each put other characters where upstream's fixer expects the parenthesis to be; this port scans past the callee for the actual token.

- **A parenthesized optional chain is still reported.** ESTree wraps an optional chain in a `ChainExpression`, and upstream's chain walk does not enter it, so `(expect(fn)?.toBeCalledTimes)(1)` goes unreported there. The matcher and its receiver are unambiguous in that shape and the rewrite is exact, so this port reports and fixes it.

- **A comment inside the argument list withholds the fix instead of silently dropping it.** Upstream's `fixer.remove(vitestFnCall.args[0])` deletes only the argument node and leaves a comment written alongside it (`toHaveBeenCalledTimes(/* exactly one */ 1)`) untouched but orphaned. This port detects a comment anywhere in the removed span and reports without a fix.

## Related Links

Related #935

## Checklist


internal/plugins/rstest/rules/prefer_called_once/prefer_called_once.md
```diff
@@ -0,0 +1,56 @@
...
+
+Chai's call-count assertion, `expect(fn).callCount(1)`, is not reported. Its "once" spelling is the `calledOnce` property rather than a matcher call, so rewriting it is a different edit than the one this rule makes.
+
+## Relationship to other rules
+
+This rule is the exact inverse of `rstest/prefer-called-times`, which rewrites `toHaveBeenCalledOnce()` back into `toHaveBeenCalledTimes(1)`. Enable one or the other, never both: with both on, every call-count assertion is reported no matter how it is written, and the two autofixes undo each other.
+
+Enabling this rule together with `rstest/no-alias-methods` is fine. Both rewrite towards the canonical matcher names, so they agree on the result.
+
+## Differences from ESLint
```
swwind: hello?

swwind (approved):

PR #1924, opened 2026-08-27
fix(typescript-eslint): align no-unsafe-enum-comparison rule

swwind:
## Summary

- Add the missing enum replacement suggestions for unsafe enum comparisons.
- Align enum/union comparison detection with typescript-eslint v8.68.0.
- Split upstream and rslint-specific tests into dedicated files.
- Preserve parentheses when applying enum replacement suggestions.

## Related Links

- [typescript-eslint rule](https://typescript-eslint.io/rules/no-unsafe-enum-comparison)
- [Upstream source](https://github.com/typescript-eslint/typescript-eslint/blob/v8.68.0/packages/eslint-plugin/src/rules/no-unsafe-enum-comparison.ts)

## Checklist

- [x] Tests updated.
- [x] Documentation updated.


internal/plugins/typescript/rules/no_unsafe_enum_comparison/no_unsafe_enum_comparison.go
```diff
@@ -22,14 +23,86 @@ func buildMismatchedConditionMessage() rule.RuleMessage {
...
+	return false
+}
+
+func getEnumKeyForLiteral(
+	sourceFile *ast.SourceFile,
+	enumLiterals []*checker.Type,
+	literal *ast.Node,
+	staticEvaluator *utils.StaticStringEvaluator,
+) (string, bool) {
+	staticValue, ok := staticEvaluator.EvalValue(literal)
```
elecmonkey, line 61: [P2] Match upstream static-expression evaluation

For `enum E { A = 2 }; declare const e: E; e === (1, 2)`, typescript-eslint 8.68.0 reports `mismatchedCondition` with a suggestion producing `e === (E.A)`, while Rslint reports the diagnostic without a suggestion. Upstream `getStaticValue` evaluates a sequence expression to its final operand, but this evaluator returns unknown. The same gap affects other upstream-supported constants such as `'ab'.length` and `'ab'.indexOf('b')`; please use an evaluator with equivalent coverage or add the missing expression semantics.

internal/plugins/typescript/rules/no_unsafe_enum_comparison/no_unsafe_enum_comparison.go
```diff
@@ -22,14 +23,86 @@ func buildMismatchedConditionMessage() rule.RuleMessage {
...
+) bool {
+	literalValue := enumLiteral.AsLiteralType().Value()
+	if utils.IsTypeFlagSet(enumLiteral, checker.TypeFlagsStringLiteral) {
+		enumString, enumStringOK := literalValue.(string)
+		staticString, staticStringOK := staticValue.(string)
+		return enumStringOK && staticStringOK && enumString == staticString
+	}
+	if utils.IsTypeFlagSet(enumLiteral, checker.TypeFlagsNumberLiteral) {
+		_, staticNumberOK := staticValue.(staticNumber)
+		return staticNumberOK && checker.ValueToString(literalValue) == staticValueText
```
elecmonkey: [P2] Reject NaN before matching numeric enum values

For `enum E { A = 0 / 0 }; declare const e: E; e === 0 / 0`, typescript-eslint 8.68.0 reports the comparison without a replacement suggestion because `NaN !== NaN`. Rslint stringifies both values as `"NaN"`, treats them as equal, and suggests `E.A`. Please check `IsNaN()` before the textual numeric comparison so NaN values do not produce a false match.

internal/plugins/typescript/rules/no_unsafe_enum_comparison/no_unsafe_enum_comparison.go
swwind: fixed

internal/plugins/typescript/rules/no_unsafe_enum_comparison/no_unsafe_enum_comparison.go
swwind: fixed

fansenze (approved):

PR #1925, opened 2026-08-27
fix(typescript-eslint): align no-unsafe-argument rule

swwind:
## Summary

- Check generic iterable spread element types in `no-unsafe-argument`.
- Align rest parameter handling and upstream tests with typescript-eslint v8.68.0.
- Add focused regression coverage for generic iterable and constrained rest parameters.

## Related Links

- [typescript-eslint: no-unsafe-argument](https://typescript-eslint.io/rules/no-unsafe-argument)
- [Upstream source at v8.68.0](https://github.com/typescript-eslint/typescript-eslint/blob/v8.68.0/packages/eslint-plugin/src/rules/no-unsafe-argument.ts)

## Checklist

- [x] Tests updated.
- [x] Documentation updated.


internal/plugins/typescript/rules/no_unsafe_argument/no_unsafe_argument.go
```diff
@@ -247,12 +250,47 @@ var NoUnsafeArgumentRule = rule.CreateRule(rule.Rule{
...
+							checker.IterationTypeKindYield,
+							spreadArgType,
+							nil,
+						)
+						if spreadElementType == nil {
+							constrainedType := checker.Checker_getBaseConstraintOfType(ctx.TypeChecker, spreadArgType)
+							if constrainedType == nil {
+								constrainedType = spreadArgType
+							}
+							if checker.Checker_isArrayType(ctx.TypeChecker, constrainedType) || checker.IsTupleType(constrainedType) {
```
elecmonkey: **[P2] Handle array unions in the ES5 fallback**

With `lib: ["es5"]`, `getIterationTypeOfIterable` returns `nil` for `declare const values: string[] | any[]; acceptStrings(...values)`, but `Checker_isArrayType` is false for the union, so this guard skips `GetNumberIndexType` and Rslint emits no diagnostic. The same source emits `unsafeArgument` when `Iterable` libs are present, making the new spread check lib-dependent. Calling `GetNumberIndexType` on the constrained type without requiring the whole type to be a single array or tuple restores the diagnostic and keeps the existing ES5 constrained-array case passing.

swwind: fixed

fansenze (approved):

PR #1926, opened 2026-08-27
fix(typescript-eslint): align require-await rule

swwind:
## Summary

- Add the missing `removeAsync` suggestions to `@typescript-eslint/require-await`.
- Align the rule tests with typescript-eslint v8.68.0 and ESLint v10.9.1.
- Split Go tests into upstream and additional regression suites.

## Related Links

- https://typescript-eslint.io/rules/require-await
- https://github.com/typescript-eslint/typescript-eslint/blob/v8.68.0/packages/eslint-plugin/src/rules/require-await.ts

## Checklist


internal/plugins/typescript/rules/require_await/require_await.go
```diff
@@ -195,14 +197,133 @@ func functionOuterName(node *ast.Node) string {
...
+		if !ok {
+			return nil
+		}
+
+		replacement := ""
+		if shouldReplaceAsyncWithSemicolon(ctx.SourceFile, node, asyncInfo) {
+			replacement = ";"
+		}
+		fixes := []rule.RuleFix{
+			rule.RuleFixReplaceRange(asyncInfo.removeRange, replacement),
```
elecmonkey: [P2] Insert the ASI semicolon before the entire class member

When a computed method has modifiers or decorators, this inserts the protective semicolon at the `async` token rather than before the entire member. For example, `static async [key]()` becomes `static ;[key]()`, which parses as a separate `static` field followed by a non-static method, while `@dec
async [key]()` becomes the invalid `@dec
;[key]()`. Applying the `removeAsync` suggestion can therefore change the member’s semantics or produce a syntax error.

internal/plugins/typescript/rules/require_await/require_await.go
```diff
@@ -195,14 +197,133 @@ func functionOuterName(node *ast.Node) string {
...
+		return nil
+	}
+
+	reference, ok := typeReferenceNamed(returnType, "Promise")
+	if !ok || reference.TypeArguments == nil || len(reference.TypeArguments.Nodes) == 0 {
+		return nil
+	}
+	openAngle, hasOpenAngle := utils.TokenAtOrAfter(sourceFile, reference.TypeName.End())
+	returnTypeRange := utils.TrimNodeTextRange(sourceFile, ast.SkipTypeParentheses(returnType))
+	closeAngle, hasCloseAngle := utils.TokenBeforePosition(sourceFile, returnTypeRange.End())
```
elecmonkey: [P2] Avoid rescanning from the start of the file for every suggestion

`TokenBeforePosition` scans from byte zero on every call, so a file containing N reported `async ...: Promise<T>` functions takes quadratic time when API/LSP eagerly materializes suggestions. In a focused benchmark, 250, 1,000, and 2,000 functions took approximately 13 ms, 261 ms, and 938 ms respectively, while diagnostics-only execution remained around 2–3 ms. Please use a token lookup bounded to the return-type range or derive the closing token from the type-argument AST.

internal/plugins/typescript/rules/require_await/require_await.go
swwind: fixed

internal/plugins/typescript/rules/require_await/require_await.go
swwind: fixed

fansenze (approved):

PR #1927, opened 2026-08-27
fix(typescript-eslint): align no-unnecessary-type-assertion rule

swwind:
## Summary

- Add the missing contextually unnecessary assertion checks.
- Align rule behavior and the complete upstream test suite with typescript-eslint v8.68.0.
- Split Go tests into upstream and rslint-specific regression suites.
- Add guards for recursive generic types and object spread assertions.

## Related Links

- [typescript-eslint: no-unnecessary-type-assertion](https://typescript-eslint.io/rules/no-unnecessary-type-assertion)
- [Upstream source at v8.68.0](https://github.com/typescript-eslint/typescript-eslint/blob/v8.68.0/packages/eslint-plugin/src/rules/no-unnecessary-type-assertion.ts)

## Checklist

- [x] Tests updated.
- [x] Documentation updated.


internal/plugins/typescript/rules/no_unnecessary_type_assertion/no_unnecessary_type_assertion.go
```diff
@@ -224,110 +698,242 @@ var NoUnnecessaryTypeAssertionRule = rule.CreateRule(rule.Rule{
...
-					uncastParts.Add(part)
-				}
+			if getUnionTypeFlags(uncast)&checker.TypeFlagsUndefined != 0 &&
+				getUnionTypeFlags(cast)&checker.TypeFlagsUndefined != 0 &&
+				compilerOptions.ExactOptionalPropertyTypes.IsTrue() {
+				return areUnionPartsEquivalentIgnoringUndefined(uncast, cast)
 			}
 
-			if !uncastHasUndefined {
+			if (utils.IsTypeFlagSet(uncast, checker.TypeFlagsNonPrimitive) && !utils.IsTypeFlagSet(cast, checker.TypeFlagsNonPrimitive)) ||
```
elecmonkey: [P1] Aggregate NonPrimitive across union constituents

`utils.IsTypeFlagSet` only examines the top-level type flags, unlike upstream's union-aware `isTypeFlagSet`. With `declare const x: object | string; let y = x as {} | string; y = 1;`, typescript-eslint 8.68.0 reports nothing, while Rslint reports `unnecessaryAssertion` and fixes the declaration to `let y = x`; that narrows `y` to `object | string` and makes the previously valid `y = 1` fail with TS2322. Use the union-aggregated flags here so the assertion is preserved.

internal/plugins/typescript/rules/no_unnecessary_type_assertion/no_unnecessary_type_assertion.go
```diff
@@ -224,110 +698,242 @@ var NoUnnecessaryTypeAssertionRule = rule.CreateRule(rule.Rule{
...
+			typeIsUnchanged := isTypeUnchanged(node, expression, uncastType, castType)
+			wouldSameTypeBeInferred := !typeAnnotationIsConstAssertion
+			if castTypeIsLiteral {
+				wouldSameTypeBeInferred = isImplicitlyNarrowedLiteralDeclaration(node)
+			}
+			isControlFlowSensitiveUndefinedUnion :=
+				getUnionTypeFlags(uncastType)&checker.TypeFlagsUndefined != 0 &&
+					getUnionTypeFlags(castType)&checker.TypeFlagsUndefined != 0 &&
+					assertionIsInOptionalOrLogicalContext(node)
+			if typeIsUnchanged && wouldSameTypeBeInferred && !isControlFlowSensitiveUndefinedUnion {
```
elecmonkey: [P2] Keep reporting unchanged nullable assertions

This guard suppresses every unchanged `undefined` union inside an optional chain or logical expression, although upstream still reports those assertions. For `declare const x: string | undefined; (x as string | undefined)?.trim();`, typescript-eslint 8.68.0 emits `unnecessaryAssertion` and fixes it to `(x)?.trim()`, while Rslint emits nothing; the same false negative occurs with `||`, `&&`, `??`, optional element access, and optional calls. The protection needs to be limited to the actual ts-go control-flow mismatch instead of bypassing an already-proven unchanged type.

internal/plugins/typescript/rules/no_unnecessary_type_assertion/no_unnecessary_type_assertion.go
```diff
@@ -68,6 +69,479 @@ func parseOptions(options []any) NoUnnecessaryTypeAssertionOptions {
...
+	}
+	return false
+}
+
+func assertionTypeArguments(typeChecker *checker.Checker, t *checker.Type) []*checker.Type {
+	if alias := checker.Type_alias(t); alias != nil {
+		return alias.TypeArguments()
+	}
+	if utils.IsTypeReference(t) {
+		return checker.Checker_getTypeArguments(typeChecker, t)
```
elecmonkey, line 113: [P2] Keep reference arguments when an alias has none

Returning `alias.TypeArguments()` unconditionally loses the underlying reference arguments for non-generic aliases. For `type T = [any]; declare const x: T; declare function f(x: object): void; f(x as object);`, upstream 8.68.0 finds the nested `any` and reports nothing, whereas Rslint reports `contextuallyUnnecessary` and removes the assertion. Match upstream's nullish fallback by consulting the referenced type arguments when the alias has none.

internal/plugins/typescript/rules/no_unnecessary_type_assertion/no_unnecessary_type_assertion.go
```diff
@@ -68,6 +69,479 @@ func parseOptions(options []any) NoUnnecessaryTypeAssertionOptions {
...
+			return true
+		}
+		if t.IsUnion() || t.IsIntersection() {
+			stack = append(stack, t.Types()...)
+			continue
+		}
+		stack = append(stack, assertionTypeArguments(typeChecker, t)...)
+		for _, signature := range utils.GetCallSignatures(typeChecker, t) {
+			if !skipRecursiveCallReturns || !assertionSignatureReturnsContainingAlias(signature) {
+				stack = append(stack, checker.Checker_getReturnTypeOfSignature(typeChecker, signature))
```
elecmonkey, line 176: [P2] Inspect non-recursive parts of recursive return types

The recursion workaround drops an entire call return type whenever its syntax contains the enclosing alias, including other union constituents that must still be inspected. With `type R = () => R | any; declare const x: R; declare function f(x: object): void; f(x as object);`, upstream 8.68.0 finds `any` and emits no diagnostic, but Rslint skips the whole `R | any` return, reports `contextuallyUnnecessary`, and removes the assertion. Cut only the recursive edge while retaining sibling return constituents.

internal/plugins/typescript/rules/no_unnecessary_type_assertion/no_unnecessary_type_assertion.go
```diff
@@ -224,110 +692,238 @@ var NoUnnecessaryTypeAssertionRule = rule.CreateRule(rule.Rule{
...
 			}
 
-			if !uncastHasUndefined {
+			if (utils.IsTypeFlagSetWithUnion(uncast, checker.TypeFlagsNonPrimitive) && !utils.IsTypeFlagSetWithUnion(cast, checker.TypeFlagsNonPrimitive)) ||
+				(assertionHasIndexSignature(ctx.TypeChecker, uncast) && !assertionHasIndexSignature(ctx.TypeChecker, cast)) ||
+				assertionTypeContainsAny(ctx.TypeChecker, uncast) ||
+				assertionTypeContainsAny(ctx.TypeChecker, cast) ||
+				(assertionTypeContainsTypeVariable(ctx.TypeChecker, cast) && !assertionTypeContainsTypeVariable(ctx.TypeChecker, uncast)) {
 				return false
 			}
```
swwind, line 742: Fixed in 831528ea7. This now uses the shared union-aware flag helper, matching upstream's `isTypeFlagSet` behavior and preserving assertions that would otherwise produce an unsafe fix. Added a regression case.

internal/plugins/typescript/rules/no_unnecessary_type_assertion/no_unnecessary_type_assertion.go
```diff
@@ -224,110 +692,238 @@ var NoUnnecessaryTypeAssertionRule = rule.CreateRule(rule.Rule{
...
+				reportAssertion(buildUnnecessaryAssertionMessage())
 				return
 			}
 
-			msg := buildUnnecessaryAssertionMessage()
-
-			if node.Kind == ast.KindAsExpression {
-				asKeywordRange := getTokenRange(expression.End())
-
-				startPos := asKeywordRange.Pos()
```
swwind, line 314: Fixed in 831528ea7. I removed the branch-local optional/logical undefined-union suppression; unchanged assertions in optional property/element/call access and `||`, `&&`, and `??` now report as upstream does. Added regression coverage.

internal/plugins/typescript/rules/no_unnecessary_type_assertion/no_unnecessary_type_assertion.go
```diff
@@ -68,6 +69,473 @@ func parseOptions(options []any) NoUnnecessaryTypeAssertionOptions {
...
+	}
+	return false
+}
+
+func assertionTypeArguments(typeChecker *checker.Checker, t *checker.Type) []*checker.Type {
+	if alias := checker.Type_alias(t); alias != nil && len(alias.TypeArguments()) > 0 {
+		return alias.TypeArguments()
+	}
+	if utils.IsTypeReference(t) {
+		return checker.Checker_getTypeArguments(typeChecker, t)
```
swwind, line 113: Fixed in 831528ea7. An empty ts-go alias type-argument slice is now treated like upstream's absent `aliasTypeArguments`, falling back to the underlying type reference arguments. Added a regression case.

internal/plugins/typescript/rules/no_unnecessary_type_assertion/no_unnecessary_type_assertion.go
```diff
@@ -68,6 +69,473 @@ func parseOptions(options []any) NoUnnecessaryTypeAssertionOptions {
...
+			return true
+		}
+		if t.IsUnion() || t.IsIntersection() {
+			stack = append(stack, t.Types()...)
+			continue
+		}
+		stack = append(stack, assertionTypeArguments(typeChecker, t)...)
+		for _, signature := range utils.GetCallSignatures(typeChecker, t) {
+			if !skipRecursiveCallReturns || !assertionSignatureReturnsContainingAlias(signature) {
+				stack = append(stack, checker.Checker_getReturnTypeOfSignature(typeChecker, signature))
```
swwind, line 176: Deliberately diverged. A direct port of upstream's `typeContains` recursion does not terminate in ts-go for this recursive return shape because it produces fresh type instances. I retained the existing recursion guard rather than add a rule-local alias-expansion heuristic.

elecmonkey (approved):
LBTM

PR #1929, opened 2026-08-28
feat(rstest): add hoisted-apis-on-top rule

elecmonkey:
## Summary

Report the module-mock APIs Rstest lifts to the top of a module — `rs.mock`, `rs.mockRequire`, `rs.unmock`, `rs.unmockRequire` and `rs.hoisted` — when they are written anywhere else. All five run before the rest of the file, ahead of its own imports, so a call written inside an `if`, a loop or a test body reads as conditional or scoped when it is neither. The four `do*` counterparts run where they are written and are never reported.

The rule matches exactly the shapes the Rstest build rewrites, each confirmed by running it under Rstest with `@rspack/core@2.1.10` rather than inferred from the transform, which lives in the Rust `RstestPlugin`. Two things have to hold. The call has to be written as the whole of a statement, or as the whole initializer of a variable declaration — the latter optionally behind an `await`; anywhere else (an argument, an object property, an arrow's expression body, an operand of `&&`, the right-hand side of an assignment, an `await` in statement position) it is left alone and throws where it is written. And it has to name the API as a plain dotted member off `rs` or `rstest`, with no import or scope analysis behind the receiver, since the transform matches the name as written. Parentheses and TypeScript wrappers are transparent throughout — around the receiver, around the callee, and around the complete call.

Two suggestions: move the statement to the top of the file, or switch to the non-hoisted counterpart. Registered in the rstest plugin and left out of the `recommended` preset, matching upstream.

## Credits

Port from [`eslint-plugin-vitest`](https://github.com/vitest-dev/eslint-plugin-vitest)'s `hoisted-apis-on-top`.

- **The set of hoisted APIs is Rstest's, not Vitest's.** Upstream watches `mock`, `unmock` and `hoisted`. Rstest also lifts `mockRequire` and `unmockRequire`, the CommonJS `require()` counterparts of the first two, so both are reported here.

- **`unmock` and `unmockRequire` get a non-hoisted-counterpart suggestion too.** Upstream offers `vi.doMock` only for `vi.mock`; Rstest exposes `doUnmock` and `doUnmockRequire` as well, and the suggestion is the same one in every case. The message id is `suggestUseNonHoistedApi` rather than upstream's `suggestReplaceMockWithDoMock`, and it names the receiver as written: `Replace 'rstest.unmock()' with 'rstest.doUnmock()', which is not hoisted.`

- **Only the two positions Rstest actually rewrites are reported.** Upstream reports the API wherever it appears and replaces the call with `undefined` when its value is used. Rstest rewrites only a call that is the whole expression of a statement or the whole initializer of a variable declaration — `c && rs.mock('./dep')`, `() => rs.mock('./dep')`, `register(rs.hoisted(f))`, `{ v: rs.hoisted(f) }`, `v = rs.hoisted(f)` and `await rs.mock('./dep');` in statement position all throw `mock() was not transformed by Rstest` instead. Reporting them would describe a lift that does not happen, so this port stays silent. Depth is not part of the test: a statement inside a function body or a declaration inside a block is rewritten like any other.

- **The receiver is matched by the name written at the call site, with no import or scope analysis.** Upstream tracks Vitest's import aliases, so `import { vi as v }` makes `v.mock()` reportable and an unrelated `import { vi } from 'some-other-module'` makes `vi.mock()` valid. Rstest's transform does neither: a renamed binding is never rewritten and throws where it is written (Rstest has its own e2e test for this), while a `rs` from another package, a namespace import, or a local `const rs = { mock() {} }` all still have their calls lifted. Upstream's resolution would report calls Rstest leaves alone and miss calls it lifts.

- **Parentheses and TypeScript wrappers are transparent on all three sides of the call.** Upstream requires the receiver to be a bare identifier and the callee to be a direct member expression, so `(vi).mock()`, `vi!.mock()` and `(vi.mock)()` go unreported. TypeScript erases those wrappers before Rstest's mock transform runs, so `(rs).mock()`, `rs!.mock()`, `(rs as any).mock()`, `(rs.mock as any)()` are all lifted and all reported — and, on the other side, `rs.mock('./dep') as unknown;` and `const v = (await rs.hoisted(f)) as number` are recognized as the top-level shapes they are rather than reported.

- **An optional chain is not reported.** `vi?.mock()` is reportable upstream. Rstest does not rewrite one, so the call runs and throws where it is written — a different problem from the one this rule describes.

- **A statement becomes an empty statement rather than being deleted.** Upstream's move suggestion removes the call node and leaves whatever surrounds it, producing the syntax error `if (foo) ` when the `if` was written without braces and without a semicolon. Replacing the whole statement with `;` gives upstream's output byte for byte when a semicolon was present, and stays valid when it was not.

- **A declaration moves with its binding, or not at all.** Rstest lifts only the call and leaves the declaration in place, so moving just the call would assign the binding when its block runs while the factory had already executed. This port moves the whole declaration when it is a single-declarator variable statement, and withholds the move when the declaration cannot travel alone — a second declarator, or a `for` header — or when a name it binds already exists at the top level, where re-declaring it would not parse.

- **The insertion point respects a shebang and the directive prologue.** Upstream inserts at offset 0 when the file has no imports, which pushes a `#!` line off line one and takes `'use strict'` out of the prologue. This port inserts after the last of the shebang, the leading directives and the final import declaration.

## Related Links

Related #935

## Checklist


internal/plugins/rstest/rules/hoisted_apis_on_top/hoisted_apis_on_top.go
```diff
@@ -0,0 +1,312 @@
...
+			if !insertResolved {
+				insertPos = lastImportEnd(ctx.SourceFile)
+				insertResolved = true
+			}
+			return insertPos
+		}
+
+		return rule.RuleListeners{
+			ast.KindCallExpression: func(node *ast.Node) {
+				accessor, namespace, api, ok := hoistedAPICall(node)
```
elecmonkey, line 81: [P2] Restrict reports to call shapes Rstest actually hoists

Rstest’s parser hook only handles calls that are the direct expression of an expression statement or the direct initializer of a variable declaration (optionally under `await`). With `@rspack/core@2.1.10`, inputs such as `if (c) { c && rs.mock('./a') }`, `() => rs.unmock('./a')`, and `register(rs.hoisted(() => 1))` remain untransformed, while this listener reports all of them and offers behavior-changing suggestions. Please mirror the transform’s parent-shape restrictions before reporting a call.

internal/plugins/rstest/rules/hoisted_apis_on_top/hoisted_apis_on_top.go
```diff
@@ -0,0 +1,312 @@
...
+// carry parentheses and TypeScript wrappers — `(rs).mock()`, `rs!.mock()`,
+// `(rs as any).mock()`, `(rs satisfies RstestUtilities).mock()` are all lifted,
+// because those are erased before the module-mock rewrite runs. The API name
+// must be written as a plain dotted member: `rs['mock']()` and a call reached
+// through `import.meta.rstest` are left as ordinary runtime calls and throw,
+// and so is anything on an optional chain, so none of them are reported here.
+func hoistedAPICall(node *ast.Node) (*ast.Node, string, string, bool) {
+	call := node.AsCallExpression()
+	if call == nil || call.Expression == nil ||
+		call.Expression.Kind != ast.KindPropertyAccessExpression ||
```
elecmonkey: [P2] Unwrap transparent wrappers around the callee

Requiring `call.Expression` to be a direct `PropertyAccessExpression` misses `(rs.mock)('./dep')` and `(rs.mock as any)('./dep')`. The actual Rstest transform hoists both forms, and `@vitest/eslint-plugin@1.6.27` also reports the parenthesized form. Please unwrap parentheses and TypeScript assertion wrappers around the callee before checking the member access, while retaining the optional-chain exclusion.

internal/plugins/rstest/rules/hoisted_apis_on_top/hoisted_apis_on_top.go
```diff
@@ -0,0 +1,312 @@
...
+// isWrittenAtTopLevel reports whether the call is already written where it
+// runs: as a statement of the file itself.
+//
+// `hoisted` is also accepted as the initializer of a top-level variable, with
+// or without an `await`, because that is the shape the API is designed for —
+// its return value is what the rest of the file uses. The other hoisted APIs
+// evaluate to undefined, so a top-level statement is the only shape that reads
+// as intended.
+func isWrittenAtTopLevel(call *ast.Node, api string) bool {
+	outer := utils.OutermostParenthesizedExpression(call)
```
elecmonkey: [P2] Treat wrappers around the complete call as transparent

`OutermostParenthesizedExpression` does not cross `as`, `satisfies`, or non-null assertion wrappers, so valid top-level forms such as `const value = rs.hoisted(() => 1) as number` and `const value = rs.hoisted(() => 1)!` are reported even though Rstest hoists their complete declarations. `rs.mock('./dep') as unknown;` is similarly misclassified as non-top-level. Please unwrap these transparent wrappers when determining the call’s containing statement or initializer.

internal/plugins/rstest/rules/hoisted_apis_on_top/hoisted_apis_on_top.go
```diff
@@ -0,0 +1,312 @@
...
+		rule.RuleFixReplace(ctx.SourceFile, call, "undefined"),
+	)
+}
+
+// insertAtTop places one complete statement after the file's last import, or at
+// the very start of a file that has none.
+func insertAtTop(statement string, insertPos int) []rule.RuleFix {
+	if insertPos < 0 {
+		return []rule.RuleFix{
+			rule.RuleFixReplaceRange(core.NewTextRange(0, 0), statement+"\n"),
```
elecmonkey: [P2] Preserve shebangs and directive prologues when inserting

In a file without imports, inserting at offset 0 can corrupt leading syntax or semantics. Applying this suggestion to `#!/usr/bin/env node
if (c) rs.mock('./dep')` moves the shebang to line 2 and produces a parse error; placing it before `'use strict'` also removes that string from the directive prologue. Please insert after any shebang and leading directives, or omit the move suggestion when a safe insertion point cannot be established.

internal/plugins/rstest/rules/hoisted_apis_on_top/hoisted_apis_on_top.go
```diff
@@ -0,0 +1,312 @@
...
+		statement := movableVariableStatement(outer)
+		if statement == nil {
+			return nil
+		}
+		text := utils.TrimmedNodeText(ctx.SourceFile, statement)
+		if !strings.HasSuffix(text, ";") {
+			text += ";"
+		}
+		return append(
+			insertAtTop(text, insertPos),
```
elecmonkey: [P2] Avoid introducing top-level binding collisions

Moving a block-local declaration verbatim can make the suggested output invalid. For `const shared = 2; if (c) { const shared = rs.hoisted(() => 1); console.log(shared) }`, the suggestion creates two top-level `const shared` declarations and fails with `Identifier 'shared' has already been declared`. Rstest’s transform hygienically renames the lifted binding and updates its references; this suggestion should either do equivalent binding-aware renaming or be withheld when the destination scope conflicts.

internal/plugins/rstest/rules/hoisted_apis_on_top/hoisted_apis_on_top.go
elecmonkey: Confirmed and fixed in d3d148f9. Checked each shape against Rstest with `@rspack/core@2.1.10`: `c && rs.mock('./a')`, `() => rs.mock('./a')`, `identity(rs.hoisted(f))`, `{ v: rs.hoisted(f) }` and `v = rs.hoisted(f)` all throw `... was not transformed by Rstest`, while a statement inside a function body and a declaration initializer inside a block are both rewritten — so nesting depth is irrelevant and only the parent shape matters. `liftedCallPosition` now accepts exactly the two positions you describe. One extra finding while pinning this down: `await rs.mock('./a');` in statement position is not rewritten either, so the `await` is followed only on the way to a declaration. The `undefined` replacement went away with the branch that produced it.

internal/plugins/rstest/rules/hoisted_apis_on_top/hoisted_apis_on_top.go
elecmonkey: Confirmed and fixed in d3d148f9. `(rs.mock)('./dep')` and `(rs.mock as any)('./dep')` both hoist under `@rspack/core@2.1.10`. The callee now goes through `utils.SkipAssertionsAndParens` before the member-access check, and the optional-chain exclusion is kept on both the call and the unwrapped callee — `rs?.mock()`, `rs.mock?.()` and `(rs?.mock)()` all stay unreported, since Rstest does not rewrite them.

internal/plugins/rstest/rules/hoisted_apis_on_top/hoisted_apis_on_top.go
elecmonkey: Confirmed and fixed in d3d148f9. `const value = rs.hoisted(() => 100) as number`, the `satisfies` form, `const value = (await rs.hoisted(async () => 100)) as number`, `rs.mock('./dep') as unknown;` and `rs.mock('./dep')!;` all hoist and all run correctly under `@rspack/core@2.1.10`. `skipTransparentAncestors` now walks out of parentheses, `as`, `satisfies`, `!` and `<T>` when locating the containing statement or initializer.

internal/plugins/rstest/rules/hoisted_apis_on_top/hoisted_apis_on_top.go
elecmonkey: Confirmed and fixed in d3d148f9. `topLevelInsertionPoint` now returns the last of: the end of the shebang (`scanner.GetShebang`), the end of the leading directive prologue, and the end of the file's last import declaration. Whether the newline goes before or after the inserted statement depends on whether the insertion point is already at a line start, so the shebang case comes out as `#!/usr/bin/env node` / `rs.mock('./dep');` / `if (c) ;`. Both the shebang and the `'use strict'` case are locked in as tests.

internal/plugins/rstest/rules/hoisted_apis_on_top/hoisted_apis_on_top.go
elecmonkey: The defect is real and the move is now withheld: `movableVariableStatement` collects the names the declaration binds and drops the suggestion when the file's top level already declares one of them (top-level variable statements, imports, and named declarations). Fixed in d3d148f9, with the exact input from this comment plus an import-bound variant as tests.

One correction to the rationale, though. Rstest does not rename hygienically — running `const shared = 2; if (c) { const shared = rs.hoisted(() => 1); console.log(shared) }` under Rstest fails with that same `Identifier 'shared' has already been declared`, so the input is already broken before the rule sees it. Withholding is the right call, but equivalent binding-aware renaming would not match the transform.

swwind (approved):
LGTM

PR #1930, opened 2026-08-28
feat(rstest): add prefer-import-in-mock rule

elecmonkey:
## Summary

Report the module path of `rs.mock('./module')` and `rs.doMock('./module')` and rewrite it to `rs.mock(import('./module'))`, which ties the call to the mocked module's real type: a factory returning the wrong shape, a stale export name and a path that no longer resolves become type errors instead of a mock that silently does nothing. The rule is autofixable, and the `fixable` option turns the fix off. Registered in the rstest plugin but left out of the `recommended` preset, matching upstream.

Only `mock` and `doMock` are covered — they are the two module mock APIs whose first parameter is typed `string | Promise<T>`. `mockRequire`, `doMockRequire`, `unmock`, `importActual`, `importMock` and the rest of the registry APIs take a module name only, so their paths stay as written.

Rstest rewrites these two APIs with a syntactic transform rather than resolving them at run time, so the rule matches what the transform matches. The receiver is read as the name written at the call site, `rs` or `rstest`, with no attempt to resolve where the binding came from; the call has to stand alone as a statement, since the transform lifts it above the module's imports; and its argument list has to be one the transform reads positionally. Parentheses and TypeScript's type-only syntax are transparent throughout. Every one of those conditions was checked against rstest 0.11.8 by mocking a real module and observing whether the mock took effect, whether the call threw `mock() was not transformed by Rstest`, or whether the build failed.

A path that is not a quoted string has no type to recover, and a call with an explicit type argument has already stated the mocked shape; both are left alone. The fix wraps the path as written and touches nothing else.

## Credits

Port from [`eslint-plugin-vitest`](https://github.com/vitest-dev/eslint-plugin-vitest)'s `prefer-import-in-mock`.

- **The receiver is matched by its spelling, not resolved to an import.** Four of upstream's eight invalid cases are `import { vi as renamedVi } from 'vitest'; renamedVi.mock('./foo.js')`. Rstest never rewrites that form, so it throws `[Rstest] mock() was not transformed by Rstest … called through an import alias` (`packages/core/src/runtime/api/utilities.ts`, pinned by Rstest's `e2e/mock/tests/aliasRuntimeError.test.ts`); a fix there would leave the call failing, so those cases became valid ones. The same spelling rule runs the other way: a local `const rs = getMocker()` and even a parameter named `rs` are rewritten by the transform and are reported here, which resolving the identifier through the type checker would have missed.

- **A call whose value is consumed is not reported.** Upstream only declines when the parent is a `CallExpression` or `MemberExpression`. Rstest lifts the call above the module's imports, so a variable initializer, a comma operand and an awaited call all fail too — `const x = rs.mock('./dep')` is lifted out and leaves `const x = ;`, a `SyntaxError`. The rule requires an expression statement. That is deliberately narrower than the transform for the non-hoisted `doMock`, which is rewritten in place and also accepts an initializer; nothing is gained by mocking into a discarded `void`, so this errs toward the false negative.

- **A spread or third argument is not reported.** Upstream does not inspect the argument list. A spread makes Rstest's transform give up, and a third argument fails the build, so wrapping the path repairs neither.

- **Parentheses and type-only syntax are transparent.** ESTree has no parenthesis node, so upstream gets parentheses for free but stops at a type assertion. Rstest's transform sees neither: `as`, `satisfies` and `!` are erased before it runs. `(rs as any).mock('./dep')`, `rs!.mock('./dep')`, `(rs.mock as any)('./dep')` and `rs.mock(('./dep'))` are all rewritten by Rstest and all reported here.

- **A call with an explicit type argument is not reported.** `rs.mock<{ d1: number }>('../src/d', factory)`, a shape Rstest uses in `e2e/mock/tests/mockTypedArg.test.ts`, already supplies the type this rule exists to recover; wrapping the path would check that annotation against the real module instead of against the factory. Upstream has no type arguments to consider.

- **Only a string literal is reported.** Upstream tests `pathArg.type === 'Literal'`, which in ESTree also covers numeric, boolean, `null` and regex literals, so `vi.mock(123)` is rewritten to `import('123')`. A string is the only literal that is a module path, and a template literal path fails Rstest's build whether or not this rule rewrites it.

- **The fix replaces the whole argument, reusing the literal's source text.** Upstream writes `` `import('${pathArg.value}')` `` over the literal node, so `vi.mock("./it's")` becomes `import('./it's')`, which does not parse; an escape sequence is re-emitted unescaped, and every path is normalized to single quotes. Replacing the whole argument also drops parentheses and an assertion such as `'./dep' as string` that no longer describes what the argument holds. A comment in the space that goes with them withholds the fix.

- **The diagnostic covers the argument, not the whole call.** Upstream reports the entire `CallExpression`, spanning the mock factory too.

## Related Links

Related #935

## Checklist


internal/plugins/rstest/rules/prefer_import_in_mock/prefer_import_in_mock.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+	localName := receiver.Text()
+	if !utilityNames[localName] {
+		return false
+	}
+
+	// A local `rs` or `rstest` that is neither imported from `@rstest/core`
+	// nor the framework global resolves to the empty name, which keeps a
+	// same-named binding from another library out of the rule.
+	resolvedName, _, _ := testFramework.ResolveFunctionIdentifierReference(
```
elecmonkey: [P2] Match the transform's syntactic receiver semantics

Rstest identifies these APIs by the call-site spelling, not by the binding's origin: Rstest 0.11.8 rewrites even `const rs = localObject; rs.mock('./dep')`. Resolving the identifier through the TypeChecker therefore makes project-backed files miss local, cross-module, and dynamic-import bindings that Rstest actually transforms, while source-only files report the same calls. Please match the transform's syntactic `rs`/`rstest` check instead of requiring an `@rstest/core` symbol.

internal/plugins/rstest/rules/prefer_import_in_mock/prefer_import_in_mock.go
```diff
@@ -0,0 +1,171 @@
...
+
+// mockPathArgument returns the module path of a `rs.mock` / `rs.doMock` call
+// that is written as a plain string, or nil when the call is anything else.
+func mockPathArgument(ctx rule.RuleContext, node *ast.Node) *ast.Node {
+	call := node.AsCallExpression()
+	if call == nil || call.QuestionDotToken != nil {
+		return nil
+	}
+
+	if !isMockCallee(ctx, call.Expression) {
```
elecmonkey: [P2] Restrict reports to transformable call positions

With `consume(rs.mock('./dep'))`, Rstest 0.11.8 throws `mock() was not transformed by Rstest`, and `@vitest/eslint-plugin@1.6.27` also skips the nested call; Rslint currently reports it and offers a fix that leaves the runtime failure intact. Since this port explicitly excludes calls outside Rstest's transform, please validate the call's parent/container before returning its path.

internal/plugins/rstest/rules/prefer_import_in_mock/prefer_import_in_mock.go
```diff
@@ -0,0 +1,171 @@
...
+		return nil
+	}
+	return pathArgument
+}
+
+// isMockCallee reports whether expression is `rs.mock`, `rs.doMock`, or the
+// same two under the `rstest` name, read off the utilities object exported by
+// `@rstest/core` or off the Rstest global.
+func isMockCallee(ctx rule.RuleContext, expression *ast.Node) bool {
+	if expression == nil || expression.Kind != ast.KindPropertyAccessExpression {
```
elecmonkey: [P2] Unwrap transparent callee and receiver parentheses

Both `(rs.mock)('./dep')` and `(rs).mock('./dep')` are successfully transformed by Rstest 0.11.8 and reported by `@vitest/eslint-plugin@1.6.27`, but these strict kind checks reject them. Please unwrap transparent parentheses around the callee and receiver while continuing to exclude optional chains.

internal/plugins/rstest/rules/prefer_import_in_mock/prefer_import_in_mock.go
```diff
@@ -0,0 +1,171 @@
...
+	// that argument against the real module's type instead, so the call is
+	// left alone.
+	if call.TypeArguments != nil && len(call.TypeArguments.Nodes) > 0 {
+		return nil
+	}
+
+	if call.Arguments == nil || len(call.Arguments.Nodes) == 0 {
+		return nil
+	}
+	pathArgument := call.Arguments.Nodes[0]
```
elecmonkey: [P2] Unwrap parentheses around the path literal

`rs.mock(('./dep'))` is transformed by Rstest 0.11.8 and reported by `@vitest/eslint-plugin@1.6.27`, but ts-go represents its first argument as a `ParenthesizedExpression`, so this check silently skips it. Please unwrap parentheses before testing for the string literal and constructing the replacement.

internal/plugins/rstest/rules/prefer_import_in_mock/prefer_import_in_mock.go
```diff
@@ -0,0 +1,171 @@
...
+
+	// An explicit type argument, `rs.mock<{ value: number }>('./module')`,
+	// already states the mocked module's shape. Wrapping the path would pin
+	// that argument against the real module's type instead, so the call is
+	// left alone.
+	if call.TypeArguments != nil && len(call.TypeArguments.Nodes) > 0 {
+		return nil
+	}
+
+	if call.Arguments == nil || len(call.Arguments.Nodes) == 0 {
```
elecmonkey: [P2] Reject argument lists Rstest cannot transform

`rs.mock('./dep', ...args)` currently receives a diagnostic and autofix, but Rstest 0.11.8 bypasses its transform when an argument is spread, leaving the call to throw `mock() was not transformed by Rstest`; calls with more than two arguments fail during the build. Wrapping the path resolves neither failure, so please require one or two non-spread arguments before returning the path.

internal/plugins/rstest/rules/prefer_import_in_mock/prefer_import_in_mock.go
elecmonkey: Fixed in `f18d26d9f1`. Verified against rstest 0.11.8 by mocking a real module and checking the mock took effect: both `const rs = { mock: () => {} }; rs.mock(…)` and an `rs` that arrives as a function parameter are rewritten and the mock applies, so the binding's origin is indeed irrelevant. The `ResolveFunctionIdentifierReference` call is gone and the receiver is now matched on the call-site spelling alone; two former valid cases covering a local binding and a parameter are now invalid ones.

internal/plugins/rstest/rules/prefer_import_in_mock/prefer_import_in_mock.go
elecmonkey: Fixed in `f18d26d9f1`. Confirmed on 0.11.8 that `consume(rs.mock(…))`, a comma operand and `await rs.mock(…)` all throw `not transformed`, and that `const x = rs.mock(…)` is worse than untransformed — the call is lifted out and leaves `const x = ;`, a `SyntaxError`. The rule now requires the call to stand alone as an expression statement, climbing through wrappers first so `(rs.mock('./dep'));` still counts. One note: this is deliberately narrower than the transform for the non-hoisted `doMock`, which is rewritten in place and so also accepts a variable initializer. Since nothing is gained by mocking into a discarded `void`, the guard errs toward the false negative rather than modelling a second position class.

internal/plugins/rstest/rules/prefer_import_in_mock/prefer_import_in_mock.go
elecmonkey: Fixed in `f18d26d9f1`, and widened. Parentheses are not the only thing the transform cannot see: TypeScript's type-only syntax is erased before it runs, so `(rs as any).mock(…)`, `rs!.mock(…)`, `(rs satisfies object).mock(…)` and `(rs.mock as any)(…)` are all rewritten too — each checked on 0.11.8. Both the callee and the receiver now go through `utils.SkipAssertionsAndParens` (`138a63139a` replaced the rule's own byte-identical mask with that shared helper). Optional chains stay excluded, with `rs?.mock(…)` and `rs.mock?.(…)` confirmed to throw.

internal/plugins/rstest/rules/prefer_import_in_mock/prefer_import_in_mock.go
elecmonkey: Fixed in `f18d26d9f1`. The first argument is unwrapped the same way before the string-literal test. The replacement covers the whole argument rather than the inner literal, so `rs.mock(('./dep'))` becomes `rs.mock(import('./dep'))` without a leftover paren, and `rs.mock('./dep' as string)` drops an assertion that no longer describes what the argument holds. When a comment sits in the space that replacement removes, the call is reported without a fix instead.

internal/plugins/rstest/rules/prefer_import_in_mock/prefer_import_in_mock.go
elecmonkey: Fixed in `f18d26d9f1`. Both failure modes reproduced on 0.11.8: a spread throws `not transformed`, and a third argument fails the build. The rule now requires one or two non-spread arguments. The same probing turned up one more build-time failure — a template-literal path fails the build even without a substitution — so that shape stays unreported as well.

swwind (approved):
LGTM

PR #1941, opened 2026-08-28
feat(react): port rule jsx-no-useless-fragment

### LINKED ISSUE #1937: [Feature]: implement react/jsx-no-useless-fragment
### What problem does this feature solve?

We are migrating the [storybook-rsbuild](https://github.com/rstackjs/storybook-rsbuild) monorepo from Biome to Rslint. Almost all of our previous lint policy mapped 1:1 onto Rslint's existing rules — the one gap left on the React side is `react/jsx-no-useless-fragment`.

The rule is already listed (unchecked) in the eslint-plugin-react tracking issue #225, so this is a request to prioritize it. Reasons it pulls ahead of the rest of the backlog:

- It is widely deployed: eslint-config-airbnb enables it at `error` level ([rules/react.js](https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb/rules/react.js)), and it has been ported by Biome (as [`noUselessFragments`](https://biomejs.dev/linter/rules/no-useless-fragments/)) and by Oxlint ([react/jsx-no-useless-fragment](https://oxc.rs/docs/guide/usage/linter/rules/react/jsx-no-useless-fragment)). Migrations from either tool hit this gap immediately.
- Original rule for reference: [eslint-plugin-react docs](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-no-useless-fragment.md).

### What does the proposed API of configuration look like?

Same as upstream eslint-plugin-react, no new API surface:

```js
rules: {
  'react/jsx-no-useless-fragment': ['warn', { allowExpressions: true }],
}
```

Options to match upstream: `allowExpressions` (default `false`).

elecmonkey:
## Summary

Port the `react/jsx-no-useless-fragment` rule from eslint-plugin-react to rslint.

The rule reports fragments that are redundant — a fragment with fewer than two children, or a fragment passed directly to an HTML element — while leaving keyed fragments and single-text fragments used as prop values alone. Both the `<>…</>` shorthand and the long `<Fragment>` / `<React.Fragment>` forms are checked, resolved through the `react` / `fragment` shared settings. The `allowExpressions` option (default `false`) is supported.

## Credits

Port from [`eslint-plugin-react`](https://github.com/jsx-eslint/eslint-plugin-react)'s `jsx-no-useless-fragment`.

## Related Links

- ESLint rule: https://github.com/jsx-eslint/eslint-plugin-react/blob/v7.37.5/docs/rules/jsx-no-useless-fragment.md
- Source code: https://github.com/jsx-eslint/eslint-plugin-react/blob/v7.37.5/lib/rules/jsx-no-useless-fragment.js
- Closes #1937

## Checklist

### DIFF
--- a/internal/plugins/react/rules/jsx_no_useless_fragment/jsx_no_useless_fragment.go
+++ b/internal/plugins/react/rules/jsx_no_useless_fragment/jsx_no_useless_fragment.go
@@ -23,7 +23,7 @@ var JsxNoUselessFragmentRule = rule.Rule{
 	Schema: rule.NewSchema(schemaJSON),
 	Run: func(ctx rule.RuleContext, options []any) rule.RuleListeners {
 		allowExpressions := parseAllowExpressions(options)
-		reactPragma := reactutil.GetReactPragma(ctx.Settings)
+		reactPragma := reactutil.GetReactPragmaFromContext(ctx)
 		fragmentPragma := reactutil.GetReactFragmentPragma(ctx.Settings)
 		sourceText := ctx.SourceFile.Text()
 
@@ -75,7 +75,7 @@ var JsxNoUselessFragmentRule = rule.Rule{
 
 		isFragmentWithSingleExpression := func(node *ast.Node) bool {
 			kept := nonPaddingChildren(node)
-			return len(kept) == 1 && kept[0].Kind == ast.KindJsxExpression
+			return len(kept) == 1 && isExpressionContainer(kept[0])
 		}
 
 		isChildOfHtmlElement := func(node *ast.Node) bool {
@@ -104,7 +104,7 @@ var JsxNoUselessFragmentRule = rule.Rule{
 			if node.Kind == ast.KindJsxText {
 				return !ecmascript.IsBlank(rawText(node))
 			}
-			return node.Kind == ast.KindJsxExpression
+			return isExpressionContainer(node)
 		}
 
 		canFix := func(node *ast.Node) bool {
@@ -231,13 +231,29 @@ func parseAllowExpressions(options []any) bool {
 	return allow
 }
 
+// isExpressionContainer reports whether a JSX child is an ESTree
+// JSXExpressionContainer — `{expr}`. tsgo parses a JSX spread child
+// (`{...expr}`) into the same KindJsxExpression, distinguished only by
+// DotDotDotToken, whereas ESTree gives it a separate JSXSpreadChild type that
+// upstream's `type === 'JSXExpressionContainer'` tests all reject. Every
+// container-specific check therefore has to exclude the spread form.
+func isExpressionContainer(node *ast.Node) bool {
+	return node != nil &&
+		node.Kind == ast.KindJsxExpression &&
+		node.AsJsxExpression().DotDotDotToken == nil
+}
+
 // containsCallExpression mirrors upstream's helper: a JSX expression container
 // whose expression is a plain CallExpression. Parentheses are transparent
-// because ESTree has no ParenthesizedExpression node, while an optional call
-// (`{foo?.()}`) is NOT a match — ESTree wraps it in a ChainExpression, which
-// upstream's `type === 'CallExpression'` check rejects.
+// because ESTree has no ParenthesizedExpression node, while three call-shaped
+// forms are NOT matches, because ESTree gives each of them its own node type
+// that upstream's `type === 'CallExpression'` check rejects:
+// an optional call (`{foo?.()}`, a ChainExpression), a dynamic import
+// (`{import("x")}`, an ImportExpression), and a spread child (`{...make()}`,
+// a JSXSpreadChild). tsgo collapses the first two into KindCallExpression and
+// the third into KindJsxExpression, so each needs an explicit exclusion.
 func containsCallExpression(node *ast.Node) bool {
-	if node == nil || node.Kind != ast.KindJsxExpression {
+	if !isExpressionContainer(node) {
 		return false
 	}
 	// `{}` — an empty JSX expression container — has no expression at all.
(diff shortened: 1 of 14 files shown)


internal/plugins/react/rules/jsx_no_useless_fragment/jsx_no_useless_fragment.go
```diff
@@ -0,0 +1,324 @@
...
+	}
+	// `{}` — an empty JSX expression container — has no expression at all.
+	if node.AsJsxExpression().Expression == nil {
+		return false
+	}
+	expression := ast.SkipParentheses(node.AsJsxExpression().Expression)
+	if expression == nil {
+		return false
+	}
+	return expression.Kind == ast.KindCallExpression && !ast.IsOptionalChain(expression)
```
elecmonkey: [P2] Exclude dynamic imports from the call-expression exemption

With `const x = <>{import('x')}</>;`, eslint-plugin-react@7.37.5 reports `NeedsMoreChildren`, while Rslint reports nothing. ts-go represents dynamic import as a `KindCallExpression`, whereas @typescript-eslint/parser exposes an `ImportExpression`, so this condition incorrectly treats it like the ordinary calls that upstream exempts. Please exclude `ast.IsImportCall(expression)` here.

internal/plugins/react/rules/jsx_no_useless_fragment/jsx_no_useless_fragment.go
```diff
@@ -0,0 +1,324 @@
...
+			var first *ast.Node
+			if len(kept) == 1 {
+				first = kept[0]
+			}
+			return !containsCallExpression(first)
+		}
+
+		isFragmentWithSingleExpression := func(node *ast.Node) bool {
+			kept := nonPaddingChildren(node)
+			return len(kept) == 1 && kept[0].Kind == ast.KindJsxExpression
```
elecmonkey: [P2] Distinguish JSX spread children from expression containers

For TSX such as `<>{...value}</>`, @typescript-eslint/parser exposes a `JSXSpreadChild`, but ts-go stores it as `KindJsxExpression` with a non-nil `DotDotDotToken`. Treating it as a normal expression container makes `allowExpressions: true` suppress the required diagnostic; the same issue also lets `<>{...make()}</>` enter the call-expression exemption and prevents upstream-compatible autofixing. Please require `DotDotDotToken == nil` in the expression-container-specific checks.

internal/plugins/react/rules/jsx_no_useless_fragment/jsx_no_useless_fragment.go
```diff
@@ -0,0 +1,324 @@
...
+	needsMoreChildrenDescription  = "Fragments should contain more than one child - otherwise, there’s no need for a Fragment at all."
+	childOfHtmlElementDescription = "Passing a fragment to an HTML element is useless."
+)
+
+var JsxNoUselessFragmentRule = rule.Rule{
+	Name:   "react/jsx-no-useless-fragment",
+	Schema: rule.NewSchema(schemaJSON),
+	Run: func(ctx rule.RuleContext, options []any) rule.RuleListeners {
+		allowExpressions := parseAllowExpressions(options)
+		reactPragma := reactutil.GetReactPragma(ctx.Settings)
```
elecmonkey: [P2] Honor source-level @jsx pragmas when matching fragments

With `/** @jsx Preact.h */` and `<Preact.Fragment>{value}</Preact.Fragment>`, eslint-plugin-react@7.37.5 derives `Preact` from the comment and reports `NeedsMoreChildren`, while Rslint falls back to `React` and emits nothing; conversely, `<React.Fragment>` becomes a false positive. Please resolve the first source-level `@jsx` pragma before falling back to `settings.react.pragma`.

internal/plugins/react/rules/jsx_no_useless_fragment/jsx_no_useless_fragment.go
elecmonkey: Fixed in 296a17d2ed. `containsCallExpression` now excludes `ast.IsImportCall(expression)`, so `<>{import("x")}</>` reports `NeedsMoreChildren` like upstream. Locked in with a dedicated extras case.

internal/plugins/react/rules/jsx_no_useless_fragment/jsx_no_useless_fragment.go
elecmonkey: Fixed in 296a17d2ed. Added an `isExpressionContainer` helper requiring `DotDotDotToken == nil`, and routed all three container-specific checks (`isFragmentWithSingleExpression`, `containsCallExpression`, `isNonSpaceJsxTextOrJsxCurly`) through it. So `<>{...value}</>` is reported even under `allowExpressions`, `<>{...make()}</>` no longer hits the call exemption, and the fix is now emitted. Extras cases wrap the fragment in `const a = ...` because the bare fixed output `{...value}` is not a valid statement — upstream produces the same text there.

internal/plugins/react/rules/jsx_no_useless_fragment/jsx_no_useless_fragment.go
elecmonkey: Fixed in 296a17d2ed. Ported upstream `pragmaUtil.getFromContext` as `reactutil.GetReactPragmaFromContext`: the first comment matching `@jsx <name>` wins over `settings.react.pragma`, only the head of a dotted name counts (`@jsx Preact.h` -> `Preact`), and a non-identifier name falls back to `React`. `getFragmentFromContext` is settings-only upstream, so the fragment half is unchanged.

Since `react/no-deprecated` already carried a near-equivalent `detectJsxPragma`, it now shares the new helper rather than keeping a second copy. That also fixes two divergences there: its regex scanned raw source text (so `@jsx` inside a string literal counted) and it kept the dotted tail.

swwind (approved):
No false cases found here, LGTM

PR #1942, opened 2026-08-28
feat(rstest): add consistent-rstest-namespace rule

elecmonkey:
## Summary

Adds `rstest/consistent-rstest-namespace`, which keeps a file on one spelling of the Rstest utility namespace. `@rstest/core` exposes it as both `rs` and `rstest`; the rule requires one and reports the other.

It reports two things: a plain named import of the reported spelling from `@rstest/core`, and every call made through that namespace — `rstest.mock('./service')` under the default preference. The namespace is recognised whether it is imported or read as a global. Aliased imports, namespace imports, same-named imports from other modules, `require` of the whole module, locals, parameters, property names and `import.meta.rstest` are all left alone.

The fix replaces the namespace object alone, so the method name, optional chaining, arguments and comments survive. An import is renamed to the preferred spelling, or — when the file already imports it — dropped from its specifier list along with the separating comma.

The option is `{ "fn": "rs" | "rstest" }`, defaulting to `"rs"`.

## Credits

Port from [`eslint-plugin-vitest`](https://github.com/vitest-dev/eslint-plugin-vitest)'s `consistent-vitest-vi`.

- **A call is reported only when the imported name is the reported spelling, not just the name at the call site.** Upstream matches the call-site accessor, so under a preference for `rs` it reports and rewrites `rstest.mock()` in a file that wrote `import { rs as rstest } from '@rstest/core'` — turning a working call into one on an undefined `rs`. Requiring the imported name to match removes that false positive.
- **Removing a specifier preserves the aliases of the ones that stay.** Upstream rebuilds the specifier list from local names, so `import { expect as check, rs, rstest }` loses the alias and becomes `import { expect, rs }`. This port deletes the reported specifier and its comma in place and leaves the rest of the list byte for byte.
- **"The preferred spelling is already imported" is answered across the file, not within one declaration.** Two `@rstest/core` imports in one file would otherwise be renamed into a duplicate binding.
- **A report carries no fix when the rewrite cannot reach every use of the binding.** That covers a namespace bound by a destructured `require`, a `export { rstest }` re-export, a reference that is not a call such as `const { fn } = rstest`, and a specifier whose removal would empty its declaration and leave a bare side-effect import. Upstream renames or removes unconditionally, which can leave identifiers naming nothing.
- **An aliased specifier is never reported**, `rstest as rstest` included. Upstream reports the specifier whenever local and imported names are equal; here an alias means the author chose the name, and there is no plain binding to rewrite.

## Related Links

Related #935

## Checklist


swwind (commented):
I added focused regression probes in internal/plugins/rstest/rules/consistent_rstest_namespace/consistent_rstest_namespace_extras_test.go.

  Findings:

  - High: autofix can produce invalid TypeScript for a trailing comment on the first removed specifier:
    `import { rstest /* remove */, rs } ...`
    becomes `import { /* remove */, rs } ...`, which fails parsing. The issue is in commentAwareEnd, which stops the removal before the comment but retains the
    comma.

  - Medium: false negative for non-null assertions:
    `rstest!.mock()` receives no report/fix. `isNamespaceCallRoot` explicitly understands `NonNullExpression`, but `namespaceCallRoot` / `ResolveFirstIdentifier` do
    not.

  - Medium: false negative for TypeScript assertions:
    `(rstest as typeof rstest).mock()` reports only the import, not the runtime call. The test correctly expects no fix, since the typeof rstest type query
    keeps the import binding live.

  - Low: tagged-template invocation ``rstest.fn`service` ``is not reported, because the rule only listens toCallExpression. Computed normal calls such as
   `rstest'mock' ()` do work.

elecmonkey:
@swwind Addressed all four findings in f6787d74:

- Fixed first-specifier removal so comments before the separator comma are removed with the specifier, while comments after the comma remain attached to the next specifier.
- Unified invocation-root handling for parentheses, non-null assertions, `as`/angle-bracket assertions, and `satisfies`; imported `rstest!.mock()` now rewrites both the import and the call.
- Type-asserted calls are reported; cases such as `(rstest as typeof rstest).mock()` remain intentionally unfixed when a type query would survive.
- Added tagged-template detection and fixes.
- Added Go and JS regression coverage and updated the rule documentation.

Validation:

- `go test ./internal/plugins/rstest/... -count=1`
- `pnpm --filter @rslint/test-tools exec rstest run tests/rstest/rules/consistent-rstest-namespace.test.ts`

internal/plugins/rstest/rules/consistent_rstest_namespace/consistent_rstest_namespace.go
```diff
@@ -0,0 +1,421 @@
...
+	return file.disallowedImport != nil && file.rewritable
+}
+
+func (file *fileNamespaces) importFixes(elements []*ast.Node, index int) []rule.RuleFix {
+	file.scan()
+	if !file.rewritable {
+		return nil
+	}
+	element := elements[index]
+	if !file.preferredImported {
```
swwind, line 304: The autofix does not account for an existing binding named `rs`, so it can produce invalid code or silently retarget the rewritten call.

For example:

```ts
import { rstest } from '@rstest/core';
const rs = { mock() {} };
rstest.mock();
```

is fixed to two top-level `rs` declarations. A nested binding is more subtle:

```ts
import { rstest } from '@rstest/core';
function run() {
  const rs = { mock() {} };
  rstest.mock();
}
```

becomes `rs.mock()` resolving to the local object instead of the imported Rstest namespace. The same capture occurs when rewriting a global `rstest` reference inside a scope that declares `rs`.

`preferredImported` only checks plain named imports from `@rstest/core`; it does not establish that every replacement can safely resolve to the preferred binding. Please suppress the fix when the preferred name would collide with or be captured by an existing binding, and add top-level, nested-scope, and global-reference regression cases.

internal/plugins/rstest/rules/consistent_rstest_namespace/consistent_rstest_namespace.go
```diff
@@ -0,0 +1,421 @@
...
+
+// specifierRemovalRange spans the specifier at index together with the comma
+// that separates it from its neighbour. A comment written for the specifier
+// that stays is kept; one written for the specifier that goes is removed with
+// it.
+func specifierRemovalRange(sourceFile *ast.SourceFile, elements []*ast.Node, index int) core.TextRange {
+	elementRange := utils.TrimNodeTextRange(sourceFile, elements[index])
+	if index > 0 {
+		previousEnd := utils.TrimNodeTextRange(sourceFile, elements[index-1]).End()
+		return core.NewTextRange(previousEnd, elementRange.End())
```
swwind: Removing a non-first specifier leaves its trailing comment attached to the preceding import:

```ts
import { rs, rstest /* remove with namespace */ } from '@rstest/core';
```

currently becomes:

```ts
import { rs /* remove with namespace */ } from '@rstest/core';
```

The same issue occurs in the middle of a list, where the comment is left between the preceding and following specifiers. This contradicts the helper's comment that a comment belonging to the removed specifier is removed with it.

The `index > 0` branch ends the range at `elementRange.End()`, so it never considers trailing comments. Please extend that branch through comments belonging to the removed specifier while preserving comments after the separating comma, and add both middle- and last-specifier regression cases.

internal/plugins/rstest/rules/consistent_rstest_namespace/consistent_rstest_namespace.go
elecmonkey: Fixed in 8d09784a. `fileNamespaces` now also answers whether the preferred spelling is free to write: the file walk flags any declaration of that name that is not an `@rstest/core` import of `rs`/`rstest` — a variable, parameter, binding element, function/class/enum/module declaration, or an import of another name — and every fix is withheld when one exists. The guard runs before the import-mode check, so a global namespace reference is covered as well, which it was not before. Scopes are deliberately not distinguished: a declaration anywhere in the file withholds the fix, which costs a fix in a file that would have accepted it and never points a rewritten call at the binding next to it. Regression cases cover the top-level, nested-scope and global-reference forms, plus an import of `rs` from another module. The rule doc records the condition.

internal/plugins/rstest/rules/consistent_rstest_namespace/consistent_rstest_namespace.go
elecmonkey: Fixed in 8d09784a. The `index > 0` branch now extends the removal end through the whitespace and comments that follow the specifier, stopping at the next comma for a specifier in the middle of the list and at the closing brace for the last one. A comment written for the removed specifier goes with it; one written past the separating comma stays with the specifier that follows. Regression cases cover the middle and last positions and the keep-side comment after the comma.

internal/plugins/rstest/rules/consistent_rstest_namespace/consistent_rstest_namespace.go
```diff
@@ -0,0 +1,562 @@
...
+		return false
+	}
+	name := imported.AsIdentifier().Text
+	return name == namespaceRs || name == namespaceRstest
+}
+
+// isSurvivingUse reports whether node names the imported disallowed spelling
+// somewhere the rule's fixes do not reach.
+func (file *fileNamespaces) isSurvivingUse(node *ast.Node, boundName *ast.Node) bool {
+	if node.Kind == ast.KindExportSpecifier {
```
swwind, line 446: **[P1] Keep aliased re-exports from losing their binding**

`isSurvivingUse` only recognizes `export { rstest }`, because it requires `PropertyName == nil`. An aliased re-export is still a live use of the imported binding:

```ts
import { rstest } from '@rstest/core';
export { rstest as helper };
rstest.mock('./service');
```

The fix currently produces:

```ts
import { rs } from '@rstest/core';
export { rstest as helper };
rs.mock('./service');
```

so the export references an identifier that no longer exists. `export type { rstest as RstestNamespace }` breaks in the same way. Treat export specifiers that reference the disallowed imported binding—including aliases and type-only exports—as surviving uses, and withhold all related fixes.

elecmonkey: Fixed in 58698f69. An export specifier's reference to the enclosing module is
its property name when the export is aliased and its exported name otherwise,
so `isSurvivingUse` now reads that name instead of requiring `PropertyName ==
nil`. `export { rstest as helper }` and `export type { rstest as
RstestNamespace }` are therefore surviving uses and withhold every related fix,
the import rename included. Regression cases cover both forms alongside the
existing plain `export { rstest }`. The rule doc names aliased re-exports in
the list of uses the rewrite does not reach.

swwind (approved):
LGTM

PR #1944, opened 2026-08-28
feat(rstest): add require-mock-type-parameters rule and extract the shared utility-call parser

elecmonkey:
## Summary

Report `rs.fn()` written without the type parameter that gives the mock a call signature, and, under `checkImportFunctions`, the four module loaders — `importActual`, `importMock`, `requireActual`, `requireMock` — that return `Record<string, unknown>` unless told what the module exports. Registered in the rstest plugin but left out of the `recommended` preset, matching upstream.

## Two kinds of API, and the shared piece this PR extracts

`rs` and `rstest` carry two kinds of member, and this one rule has to handle one of each.

`fn` is an ordinary function reached through an ordinary binding, so the rule resolves the receiver: a renamed import, a namespace as in `core.rs.fn()`, a `require` and the installed globals are all reported, and a local object or a parameter named `rs` is not. The four module loaders are among the fourteen members Rstest's build rewrites instead of calling — at run time they are stubs that throw `[Rstest] importActual() was not transformed by Rstest` — and the build matches them on the name written at the call site, so the rule reads the receiver instead of resolving it.

Two merged rules already carry their own copy of that syntactic matcher, and this one would have made three. Both move onto a shared `ParseRstestUtilityCall` in `internal/plugins/rstest/utils/`, alongside the fourteen-name table behind `IsPluginManagedAPI` that lets a rule pick a strategy per member. What differs per API stays with each rule, because the two families really do differ: `rs.mock` is hoisted even when `rs` is a local object or a parameter, while `rs.importActual` on that same object is left alone; `rs['mock']` is never rewritten while `rs['importActual']` is; `rs.mock` takes one or two arguments while `rs.importActual` fails the build on any count but one. Every condition was checked against rstest 0.11.8 by mocking a real module and observing whether the mock took effect, whether the call threw, or whether the build failed. Both migrated rules pass their existing suites unchanged.

## Credits

Port from [`eslint-plugin-vitest`](https://github.com/vitest-dev/eslint-plugin-vitest)'s `require-mock-type-parameters`.

- **`requireActual` and `requireMock` are covered by `checkImportFunctions` too.** Rstest has two CommonJS loaders Vitest lacks, typed `<T = Record<string, unknown>>(path: string)` like the other two. Skipping them would leave the same untyped result unreported for half the API.

- **The receiver of a module loader is matched by its spelling, not resolved.** `import { rs as mocker }` followed by `mocker.importActual('./m')` is never rewritten and throws where it stands, so a type parameter would not save it; a `rs` that is not from `@rstest/core` at all is still rewritten and still reported. A local `const rs = { … }` or a parameter named `rs` is the one case the build defers to the binding, and is excluded.

- **A module loader call the build cannot rewrite is not reported.** Upstream does not inspect the argument list. Any argument count but one fails Rstest's build, and a path that is not a plain quoted string is left as the stub that throws; a type parameter repairs neither.

- **An uncalled `fn` is not reported.** Upstream reads the enclosing call as `member.parent.parent` without checking it is one, so `vi.fn.mockReturnValue(1)` is reported although `fn` is never called and there is no call to write a type parameter on.

- **A mock created in an argument position is reported.** Upstream's call parser bails out when the parent is a `CallExpression`, so `spy.mockImplementation(vi.fn())` goes unreported. The mock is just as untyped when its value is consumed on the spot.

## Related Links

Related #935

## Checklist


swwind (commented):
Confirmed findings:

- Medium: false negative for tagged-template factory calls: `` rs.fn`service` `` is untyped but gets no diagnostic. The rule only listens for `CallExpression`; tagged templates use a distinct AST node.
- Medium: false negative for TypeScript `import = require` namespace bindings:
  `import core = require('@rstest/core'); core.rs.fn()`
  gets no diagnostic. `isShadowed` recognizes `ImportEqualsDeclaration` as an import, but `IsModuleNamespaceSymbol` does not recognize it as an Rstest module namespace.

internal/plugins/rstest/rules/require_mock_type_parameters/require_mock_type_parameters.go
```diff
@@ -0,0 +1,229 @@
...
+				}
+				// The type parameter is written on the call, so a call that
+				// already carries one is what this rule asks for. An empty
+				// list does not parse, so any list at all is an explicit one.
+				if call.TypeArguments != nil && len(call.TypeArguments.Nodes) > 0 {
+					return
+				}
+
+				if opts.CheckImportFunctions {
+					if utility := rstestUtils.ParseRstestUtilityCall(node); utility != nil &&
```
elecmonkey: [P2] Handle the actual-module loader shapes separately

When `checkImportFunctions` is enabled, routing every loader through `ParseRstestUtilityCall` misses `rs['importActual']('./subject.js')`, `rs[`importActual`]('./subject.js')`, and `rs.importActual?.('./subject.js')` (and the corresponding `requireActual` forms). I ran each form under `@rstest/core@0.11.8`; Rstest transformed them successfully, while the PR-head binary emitted no diagnostic even though the inferred result remained `Record<string, unknown>`. The mock-family APIs do have the narrower syntax, so please make the actual-module-loader matcher accept its broader transform surface instead of relying solely on the common parser.

internal/plugins/rstest/rules/require_mock_type_parameters/require_mock_type_parameters.go
```diff
@@ -0,0 +1,229 @@
...
+		// `rs` and `rstest` name the same object
+		// (packages/core/src/runtime/api/public.ts), either may be imported
+		// under a further name, and under `globals: true` both are on
+		// `globalThis` with no import at all. A local declaration of the name
+		// shadows both, and resolution reports that by returning no name.
+		//
+		// ctx.Refs.Resolve places an import and a local declaration from the
+		// binder alone, so a file that declares the name is recognized whether
+		// or not a TypeChecker is configured.
+		name, _, _ := testFramework.ResolveFunctionIdentifierReferenceFromSymbol(
```
elecmonkey, line 234: [P2] Resolve awaited dynamic-import destructuring

`const { rs } = await import('@rstest/core'); rs.fn()` creates and invokes the real utilities object, but this resolver only recognizes static imports and `require` bindings. `@vitest/eslint-plugin@1.6.27` reports the equivalent `const { vi } = await import('vitest'); vi.fn()`, while the PR-head binary emits nothing with or without a tsconfig. Please extend receiver resolution to cover awaited dynamic-import destructuring, including renamed destructures.

internal/plugins/rstest/rules/require_mock_type_parameters/require_mock_type_parameters.go
```diff
@@ -0,0 +1,229 @@
...
+	"importActual":  true,
+	"importMock":    true,
+	"requireActual": true,
+	"requireMock":   true,
+}
+
+func missingTypeParameterMessage(member string) rule.RuleMessage {
+	return rule.RuleMessage{
+		Id:          "missingTypeParameter",
+		Description: "'" + member + "' is called without a type parameter, so its result is untyped.",
```
elecmonkey: [P2] Preserve the upstream diagnostic contract

The pinned upstream rule emits messageId `noTypeParameter` and message `Missing type parameters`; this port instead exposes `missingTypeParameter` plus member-specific text. That breaks consumers and tests keyed to the rule's public diagnostic contract and is not listed as an intentional difference. The new message is also inaccurate for `rs.fn((value: number) => value.toString())`, where TypeScript infers the mock signature despite there being no explicit type argument. Please keep the upstream ID and text unless the contract change is explicitly accepted and documented.

internal/plugins/rstest/rules/require_mock_type_parameters/require_mock_type_parameters.go
elecmonkey: Confirmed and fixed in 483298f689.

I re-measured on `@rstest/core@0.11.8`, with `rs` imported from `@rstest/core` (an earlier measurement of mine was confounded by aliasing `rs` into a local `const` first, which by itself blocks the loader rewrite):

| call | result |
| --- | --- |
| `rs.importActual('../src/b')` | rewritten |
| `rs['importActual']('../src/b')` | rewritten |
| ``rs[`importActual`]('../src/b')`` | rewritten |
| `rs.importActual?.('../src/b')` | rewritten |
| `rs?.importActual('../src/b')` | throws `not transformed by Rstest` |
| `rs['requireActual']('../src/b')` | rewritten |

So the computed key and the optional call are reported now; the optional receiver stays out, because that one really is left as the throwing stub.

The shared parser is unchanged on purpose. The hoisted family is narrower, measured the same way: `rs['mock']('../src/increment')` and `rs.mock?.(...)` both throw, while `(rs as any).mock(...)` is rewritten. Widening `ParseRstestUtilityCall` would turn those into false positives in `prefer-import-in-mock` and `hoisted-apis-on-top`, so the wider surface lives in `parseModuleLoaderCall` next to the rule that needs it.

internal/plugins/rstest/rules/require_mock_type_parameters/require_mock_type_parameters.go
elecmonkey: Confirmed as a real false negative, and not fixed in this PR — it is not specific to this rule.

`const { rs } = await import('@rstest/core')` is not recognized because the shared resolver in `internal/utils/test_framework/module_reference.go` only understands static imports and `require(...)` initializers; there is no `await import(...)` branch. Every rstest rule that resolves a receiver has the same gap, so patching it inside `require-mock-type-parameters` would split the behaviour across rules.

I would rather add the branch to the shared resolver in its own change, so all of them pick it up at once. Recording it here so it is not lost.

internal/plugins/rstest/rules/require_mock_type_parameters/require_mock_type_parameters.go
elecmonkey: Half taken, in 483298f689.

The message no longer asserts anything about the resulting type — it now reads `'fn' is called without a type parameter.` You are right that the old wording was wrong for `rs.fn((value: number) => value.toString())`, where the signature is inferred from the implementation and the mock is not untyped.

The message id stays `missingTypeParameter`. This is a new rule in this repo rather than a drop-in for an existing one, so there are no consumers keyed to another id, and the rstest rules are written to stand on their own rather than to mirror another plugin's identifiers.

elecmonkey:
On the two findings in the earlier review — both are real, and both are left as they are.

**Tagged templates.** ``rs.fn`service` `` is a `TaggedTemplateExpression`, and the rule only listens for call expressions, so it is not reported. Using `rs.fn` as a template tag passes the `TemplateStringsArray` as the mock implementation, which is not a shape worth carrying a branch for.

**`import core = require('@rstest/core'); core.rs.fn()`.** Not reported: `IsModuleNamespaceSymbol` in `internal/utils/test_framework/module_reference.go` recognizes a namespace import and a whole-module `require`, but not an import-equals declaration. That is a gap in the shared resolver rather than in this rule — the same receiver would go unrecognized in every rstest rule that resolves one — so if it is worth closing, it belongs in `module_reference.go` together with the `await import(...)` case raised on the other thread.

Both are false negatives, which this rule prefers over reporting a call Rstest does rewrite.

internal/plugins/rstest/rules/require_mock_type_parameters/require_mock_type_parameters.md
```diff
@@ -0,0 +1,58 @@
+# require-mock-type-parameters
+
+## Rule Details
+
+Requires `rs.fn()` to be given a type parameter. Without one the mock is typed as a function that accepts any arguments and returns `any`, so nothing about how the test drives it is checked: a call with the wrong arguments, a `mockResolvedValue` of the wrong shape, and an assertion against a property the real function never returns all pass the compiler. Writing the signature — `rs.fn<(id: number) => Promise<User>>()` — makes the mock stand in for the function it replaces, in the editor as well as at build time.
```
swwind: [P3] Correct the explanation for inferred mock signatures

This paragraph still says that every `rs.fn()` call without a type argument produces a function accepting and returning `any`. As the previous review established, `rs.fn((value: number) => value.toString())` infers `(value: number) => string`; commit 483298f corrected the diagnostic wording but left the same inaccurate claim here. Please qualify this explanation for calls whose signature cannot be inferred, and explain why the rule still requires an explicit type argument when an implementation is supplied.

internal/plugins/rstest/rules/require_mock_type_parameters/require_mock_type_parameters_extras_test.go
```diff
@@ -0,0 +1,240 @@
...
+			// Unlike the hoisted module-mock APIs, a local binding of the name
+			// keeps the rewrite from happening at all.
+			{Code: `const rs = { importActual: async (p) => ({}) }; rs.importActual('./example.js');`, Options: checkImports},
+			{Code: `function load(rs) { return rs.requireActual('./example.js'); }`, Options: checkImports},
+			{Code: `const rs = { importActual: async (p) => ({}) }; rs['importActual']('./example.js');`, Options: checkImports},
+			// An optional chain on the receiver is left as the runtime stub.
+			{Code: `rs?.importActual('./example.js')`, Options: checkImports},
+
+			// The build resolves the path, so anything it cannot read at build
+			// time is either a build failure or the runtime stub.
```
swwind, line 123: [P3] Add regression cases for the accepted false-negative boundaries

The previous review identified ``rs.fn`service` `` and `import core = require('@rstest/core'); core.rs.fn()`, and the follow-up discussion also records awaited dynamic-import destructuring. These are intentionally left unreported for different reasons, but none appears in the test suite. Please add commented valid cases for these accepted boundaries so their current resolution remains explicit and future shared-resolver changes force the intended expectations to be revisited.

internal/plugins/rstest/utils/utility_call.go
```diff
@@ -0,0 +1,128 @@
...
+// call that does the work is emitted by Rstest's build, which rewrites the call
+// site itself.
+//
+// The distinction decides how a rule may recognize a call. Everything else on
+// the object — `fn`, `spyOn`, `mocked`, the timer and stub helpers — is an
+// ordinary function reached through an ordinary binding, so a renamed import
+// still works and a rule should resolve the receiver. These fourteen are
+// matched by the build on the name written at the call site, so a rule must
+// read the receiver instead of resolving it: see ParseRstestUtilityCall.
+var pluginManagedAPIs = map[string]bool{
```
swwind: [P3] Avoid adding unused classifier and result fields

`pluginManagedAPIs`/`IsPluginManagedAPI` currently have no production caller; their test only verifies the table itself. Likewise, `RstestUtilityCall.NamespaceNode` and `Call` are populated but never consumed. The new rule separately maintains `moduleLoaders`, while the migrated rules only use `Namespace`, `Member`, and `MemberNode`. Please either connect this API surface to real callers or remove it until it is needed, avoiding duplicate sources of truth that can drift.

internal/plugins/rstest/rules/require_mock_type_parameters/require_mock_type_parameters.md
elecmonkey: Confirmed and fixed in 7f531d3f.

The first paragraph now says the `(...args: any[]) => any` fallback happens only
when nothing is passed for the compiler to infer from, and a new paragraph gives
the reason a call that does pass an implementation is still reported: the
inferred signature describes the throwaway body the test happened to write, not
the function the mock stands in for, so it moves whenever the body does —
relaxing a parameter to `unknown`, or writing the `() => {}` a stub usually is,
quietly relaxes what every later argument list, `mockReturnValue` and assertion
is checked against, and the second checks nothing at all.

internal/plugins/rstest/rules/require_mock_type_parameters/require_mock_type_parameters_extras_test.go
elecmonkey: Added in 7f531d3f, as a labelled "Accepted false negatives" block in the valid
cases: the tagged template, `import core = require('@rstest/core')`, and a
destructured `await import('@rstest/core')` both renamed and not. Each carries
the reason it is unreported — a tagged template is not a `CallExpression`, and
the other two are gaps in the shared resolver in
`internal/utils/test_framework` — and the block says that a change which starts
reporting one has to move it to the invalid list deliberately.

On the tagged template specifically: it stays a valid case rather than becoming
a fix, since `fn` takes the implementation to infer from and a tag is handed a
`TemplateStringsArray` that no `FunctionLike` accepts. The case records the AST
boundary, not a shape worth reporting.

internal/plugins/rstest/utils/utility_call.go
elecmonkey: Confirmed and removed in 7f531d3f.

`IsPluginManagedAPI` and its fourteen-name table had no production caller, and
`NamespaceNode` and `Call` had no reader left once the module loaders moved to
`parseModuleLoaderCall` in 483298f6 — `NamespaceNode` was what the first version
fed to `isShadowed`. All four are gone, so the loader names are named once, in
the rule that acts on them.

What the table documented is the reason the parser exists at all, so it stays as
prose on `ParseRstestUtilityCall`: these members never run as written, the build
rewrites the call site, and everything else on the object is an ordinary
function whose receiver a rule should resolve instead.

swwind (approved):
LGTM

PR #1946, opened 2026-08-29
fix: align no-magic-array-flat-depth with upstream

fansenze:
## Summary

- Align `unicorn/no-magic-array-flat-depth` receiver classification and comment boundaries with Unicorn v73.
- Add source-only static evaluation for upstream-supported calls, variables, compound expressions, and conversions while preserving typed consumers.
- Bound recursive/static evaluation work and cache classification results; targeted benchmarks show lower allocation and runtime on common and deep inputs.
- Add upstream-parity, comment-boundary, resource, and JavaScript integration coverage.

## Related Links

N/A

## Checklist

### DIFF
--- a/internal/plugins/unicorn/rules/no_magic_array_flat_depth/comments_test.go
+++ b/internal/plugins/unicorn/rules/no_magic_array_flat_depth/comments_test.go
@@ -1,0 +1,36 @@
+package no_magic_array_flat_depth_test
+
+import (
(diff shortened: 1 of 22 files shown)


fansenze:
Superseded by #1949, which provides the same intended Unicorn v73 compatibility fixes with a smaller implementation built on existing tsgo/rslint utilities.

PR #1962, opened 2026-08-31
fix(prefer-regex-literals): respect ecma version in suggestions

swwind:
## Summary

- make regex literal validation respect the configured ECMAScript version
- avoid suggestions that use unavailable regex flags or inline modifiers
- enable the previously skipped `prefer-regex-literals` upstream test cases

## Related Links

None.

## Checklist


internal/utils/ecmascript/regexp_literal.go
```diff
@@ -33,6 +42,75 @@ func IsValidRegexLiteral(literal string) bool {
...
+	if ecmaVersion == 0 {
+		return true
+	}
+	if ecmaVersion < 2015 && strings.ContainsAny(flags, "uy") ||
+		ecmaVersion < 2018 && strings.Contains(flags, "s") ||
+		ecmaVersion < 2022 && strings.Contains(flags, "d") ||
+		ecmaVersion < 2024 && strings.Contains(flags, "v") {
+		return false
+	}
+	return ecmaVersion >= 2025 || !hasRegExpPatternModifiers(literal)
```
elecmonkey, line 68: [P1] Gate all versioned RegExp pattern syntax

The fallback returns true for non-modifier patterns even when their syntax was introduced after the configured edition, because ts-go's `ScriptTarget` does not enforce every regexpp version gate. With `ecmaVersion: 2017`, `new RegExp('(?<=a)b')` gets a `/(?<=a)b/` suggestion in Rslint, while ESLint 10.9.0 reports the diagnostic with zero suggestions; the same mismatch occurs for Unicode property escapes before ES2018 and duplicate named groups before ES2025. Applying these suggestions produces syntax unavailable under the selected target, so this helper needs regexpp-equivalent pattern feature validation in addition to the flag and modifier checks.

swwind: Fixed in 8cacfe103. canFixTo now validates the generated literal with both the existing literal/flag checker and the shared ECMAScript pattern validator, using the configured ecmaVersion. This rejects suggestions for pre-2018 lookbehind/Unicode property escapes and pre-2025 duplicate named groups while preserving valid suggestions. Added regression coverage for these cases.

elecmonkey (approved):
LGTM

fansenze (approved):

PR #1964, opened 2026-08-31
fix: align strict rule sourceType behavior

swwind:
## Summary

Align the `strict` rule's `sourceType` behavior with ESLint v10.9.1.

- Use `languageOptions.sourceType` instead of inferring module mode from import/export syntax.
- Apply global mode for the default and `"safe"` options in CommonJS files.
- Restore upstream module and CommonJS test cases in the Go and JS suites.

## Related Links

- [ESLint: strict](https://eslint.org/docs/latest/rules/strict)
- [ESLint v10.9.1 source](https://github.com/eslint/eslint/blob/v10.9.1/lib/rules/strict.js)

## Checklist


packages/rslint-test-tools/tests/eslint/rule-tester.ts
```diff
@@ -123,10 +127,17 @@ export class RuleTester {
...
           const options =
             typeof validCase === 'object' ? validCase.options : undefined;
-          const languageOptions =
+          const caseLanguageOptions =
             typeof validCase === 'object'
               ? validCase.languageOptions
               : undefined;
+          const languageOptions =
+            this.defaultLanguageOptions || caseLanguageOptions
+              ? {
```
elecmonkey: [P2] Please deep-merge default and per-case languageOptions

`languageOptions` in ESLint's RuleTester is deep-merged, but this top-level spread drops default nested fields whenever a case provides the same nested object. For example, with constructor defaults `{ globals: { defaultGlobal: 'readonly' } }` and a case-level `{ globals: { caseGlobal: 'readonly' } }`, a valid `no-undef` case using both globals reports `defaultGlobal` as undefined here, while the same case passes in ESLint 10.9.1. `parserOptions` has the same overwrite behavior, and the invalid-case path duplicates this merge logic. Please deep-merge nested language options, or pass the defaults and case options as separate config entries so the existing config merger handles them.

swwind: Fixed in `2a96317df`. RuleTester now passes default and case language options as separate flat-config entries, preserving nested `globals` and `parserOptions` in both valid and invalid cases. Added regression coverage for this merge behavior. Typecheck, formatting, and focused tests pass.

elecmonkey (approved):

PR #1966, opened 2026-08-31
feat: port rule prefer-then-catch

hebeiqing:
## Summary

Port the `unicorn/prefer-then-catch` rule from eslint-plugin-unicorn (v74.0.0) to rslint.

The rule flags `.then(onFulfilled, onRejected)` calls and offers a suggestion to rewrite them as `.then(onFulfilled).catch(onRejected)` when the rejection handler can safely be moved.

It is type-aware: it only reports when the receiver type is the built-in `Promise`, so non-Promise `.then()` calls on user-defined thenables stay untouched. It also requires the result's `.catch` to be callable.

The suggestion is suppressed when removing the rejection handler would drop a comment, when the handler has trailing-comment baggage, or when the handler is not a safe identifier/function expression (for example, call expressions or member access).

## Files added

- `internal/plugins/unicorn/rules/prefer_then_catch/prefer_then_catch.go`
- `internal/plugins/unicorn/rules/prefer_then_catch/prefer_then_catch.md`
- `internal/plugins/unicorn/rules/prefer_then_catch/prefer_then_catch_upstream_test.go`
- `internal/plugins/unicorn/rules/prefer_then_catch/prefer_then_catch_extras_test.go`
- `packages/rslint-test-tools/tests/eslint-plugin-unicorn/rules/prefer-then-catch.test.ts`

## Files modified

- `internal/plugins/unicorn/all.go` — register the rule in `GetAllRules()`
- `packages/rslint-test-tools/rstest.config.mts` — include the JS test file

## Related links

- [ESLint rule documentation](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v74.0.0/docs/rules/prefer-then-catch.md)
- [Upstream source code](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v74.0.0/rules/prefer-then-catch.js)

## Checklist

- [x] Tests updated.
- [x] Documentation updated.


internal/plugins/unicorn/rules/prefer_then_catch/prefer_then_catch.go
```diff
@@ -0,0 +1,308 @@
...
+		return nil
+	}
+
+	argsCloseParen := callExpression.End() - 1
+	if sourceFile.Text()[argsCloseParen] != ')' {
+		argsCloseParen = callExpression.End()
+	}
+
+	comments := ctx.Comments.All()
+	if utils.HasCommentInSpan(comments, removalRange.Pos(), removalRange.End()) {
```
swwind: `removalRange` includes the rejection handler, but its source is reinserted in `.catch(...)`. This therefore treats comments inside a movable function as removed and suppresses a suggestion that upstream preserves.

For example, eslint-plugin-unicorn v74.0.0 suggests a rewrite for each of these:

```js
promise.then(a, x => { /* keep */ return x; });
promise.then(a, x => /* keep */ x);
promise.then(a, function (/* keep */ x) { return x; });
```

Rslint reports without a suggestion. Please exclude comments inside `rejectionHandler` from the removal check, while retaining checks for comments in the removed separator/trivia.

internal/plugins/unicorn/rules/prefer_then_catch/prefer_then_catch.go
```diff
@@ -0,0 +1,308 @@
...
+		// upstream's try/catch by reporting on failure.
+		_ = recover()
+	}()
+
+	tc := ctx.TypeChecker
+	receiverType := tc.GetNonNullableType(tc.GetTypeAtLocation(call.Object))
+	if utils.IsTypeAnyType(receiverType) || utils.IsTypeUnknownType(receiverType) {
+		return true
+	}
+	if !utils.IsPromiseLike(ctx.Program(), tc, receiverType) {
```
swwind: Upstream requires the receiver type’s own symbol to be the default-library `Promise`; it does not treat inheritance, intersections, or a type-parameter constraint as native Promise identity. `utils.IsPromiseLike` recursively accepts all of these, so this reports cases upstream keeps valid:

```ts
class DerivedPromise extends Promise<string> {}
declare const derived: DerivedPromise;
derived.then(a, b);

declare const intersection: Promise<string> & { readonly tag: 'x' };
intersection.then(a, b);

function f<T extends Promise<string>>(value: T) {
  value.then(a, b);
}
```

Please mirror `isNativePromiseType` here: inspect only the receiver type’s direct symbol and require a default-library symbol named `Promise`, rather than using the broader `IsPromiseLike` helper.

swwind (commented):
Found two upstream-parity issues in `unicorn/prefer-then-catch`.

swwind:
Please fix the failing `golangci-lint` errors. No other issues found, thanks for your contribution.

swwind (approved):

PR #1969, opened 2026-08-31
fix: align no-restricted-syntax with ESLint 10.9.1

swwind:
## Summary

- align no-restricted-syntax selector parsing and matching with ESLint 10.9.1 and esquery 1.7.0
- add missing ESTree mappings for TSEnumDeclaration, BigInt literals, decorators, MethodDefinition.value, ClassBody, and JSX wrapper nodes
- update no-restricted-syntax documentation and Go parity coverage

## Validation

- go test ./internal/rules/no_restricted_syntax


internal/rules/no_restricted_syntax/mapping.go
```diff
@@ -53,10 +53,15 @@ var estreeKindMap = map[string][]ast.Kind{
...
 
 	// Classes
-	"ClassDeclaration":   {ast.KindClassDeclaration},
-	"ClassExpression":    {ast.KindClassExpression},
+	"ClassDeclaration": {ast.KindClassDeclaration},
+	"ClassExpression":  {ast.KindClassExpression},
+	// tsgo stores the class body directly on the class node. The matcher uses
+	// that same node as a virtual ClassBody facade when the selector asks for
+	// the ESTree wrapper.
+	"ClassBody":          {ast.KindClassDeclaration, ast.KindClassExpression},
```
elecmonkey, line 66: [P2] Preserve virtual ESTree node identity for ClassBody

Mapping `ClassBody` onto the class declaration/expression node lets simple `ClassBody` selectors report a class-body range, but attribute/combinator matching still evaluates against the physical class node. For example, ESLint 10.9.1 reports one diagnostic for `class C { m() {} }` with `ClassBody[parent.type='ClassDeclaration']` at the body range, while a temporary Rslint test on this PR reports 0 diagnostics because `parent.type` resolves to `Program`. It also collapses `:is(ClassDeclaration, ClassBody)` to one diagnostic where ESLint reports both the class and its body. Any config that restricts class bodies through parent/type/union selectors will silently under-report.

internal/rules/no_restricted_syntax/no_restricted_syntax.go
```diff
@@ -387,16 +399,54 @@ func (bucket *ruleBucket) matchAndReport(index int, node *ast.Node, mc *matchCon
...
+	message := rule.RuleMessage{
 		Id:          "restrictedSyntax",
 		Description: entry.formatMessage(),
-	})
+	}
+	if selectorTargetsClassBody(entry.compiled) && isClassLikeNode(node) {
+		ctx.ReportRange(classBodyTextRange(mc.sf, node), message)
+		return
+	}
+	if selectorTargetsJSXEmptyExpression(entry.compiled) && isEmptyJSXExpression(node) {
```
elecmonkey: [P2] Preserve JSXEmptyExpression as a separate selector target

`JSXEmptyExpression` is mapped to the same `JsxExpression` node as `JSXExpressionContainer`, but the matcher only special-cases reporting range and direct child matching. ESLint 10.9.1 reports one diagnostic for `const x = <div>{/* hi */}</div>;` with `JSXEmptyExpression[parent.type='JSXExpressionContainer']`, and two diagnostics for `:is(JSXExpressionContainer, JSXEmptyExpression)`. A temporary Rslint test on this PR reports 0 for the parent selector and collapses the union to the physical container node, so selectors involving the virtual empty expression are not 1:1 with esquery.

internal/rules/no_restricted_syntax/mapping.go
swwind: Fixed in `2bffb2f8b`. ClassBody is now matched through a separate virtual target, including its ESTree parent/type fields. Selectors such as `ClassBody[parent.type='ClassDeclaration']` report the class-body range, and `:is(ClassDeclaration, ClassBody)` produces one diagnostic for each ESTree node. Added regression coverage.

internal/rules/no_restricted_syntax/no_restricted_syntax.go
swwind: Fixed in `2bffb2f8b`. JSXEmptyExpression is now matched as a separate virtual target with JSXExpressionContainer as its ESTree parent. Parent selectors and unions now preserve ESLint's diagnostic count and range. Added regression coverage.

internal/rules/no_restricted_syntax/no_restricted_syntax.go
```diff
@@ -384,19 +396,61 @@ func (bucket *ruleBucket) matchAndReport(index int, node *ast.Node, mc *matchCon
...
-	}
-	ctx.ReportNode(node, rule.RuleMessage{
+	message := rule.RuleMessage{
 		Id:          "restrictedSyntax",
 		Description: entry.formatMessage(),
-	})
+	}
+
+	physicalMatch := matchesInScopeTarget(compiled, node, mc, nil, "physical")
+	classBodyMatch := isClassLikeNode(node) && selectorTargetsClassBody(compiled) &&
```
elecmonkey: [P1] Detect virtual nodes in attribute-only selectors

`classBodyMatch` and `jsxEmptyMatch` are attempted only when `selectorTargetsEstreeType` finds an explicit node-name selector. With `class C {}` and `[type='ClassBody']`, ESLint 10.9.1 reports the body at 1:9–1:11 while Rslint reports nothing; `[type='JSXEmptyExpression']` similarly misses an empty JSX expression. Attribute-only and nested pseudo selectors must also be evaluated against applicable virtual identities.

internal/rules/no_restricted_syntax/matcher.go
```diff
@@ -467,9 +559,21 @@ func lookupStringAttrPath(node *ast.Node, path []string, mc *matchContext) (text
 // matchesCombinator handles `>` (parent), descendant, `+` (prev sibling),
 // `~` (any prior sibling). The right-hand side has already been verified
 // against the current node before this function is called.
-func matchesCombinator(c combinatorSelector, node *ast.Node, mc *matchContext, scopeRoot *ast.Node) bool {
+func matchesCombinatorTarget(c combinatorSelector, node *ast.Node, mc *matchContext, scopeRoot *ast.Node, target string) bool {
 	switch c.Kind {
 	case combChild:
+		if target == "ClassBody" && isClassLikeNode(node) && selectorTargetsClassBody(c.Right) {
```
elecmonkey: [P1] Model virtual wrappers during structural traversal

The special cases cover only direct-child matching and do not establish the virtual nodes in descendant or `:has()` traversal. ESLint reports one diagnostic for `class C { m() {} }` with either `ClassDeclaration ClassBody` or `ClassDeclaration:has(> ClassBody)`, whereas Rslint reports zero; `JSXExpressionContainer JSXEmptyExpression` fails the same way. Virtual wrappers need to participate in the complete structural traversal rather than only these child shortcuts.

internal/rules/no_restricted_syntax/matcher.go
```diff
@@ -724,6 +898,45 @@ func stepAttrPath(current interface{}, segment string, mc *matchContext) (interf
...
+				return virtualNodeFacade{node: vn.node, typeName: "JSXExpressionContainer"}, true
+			default:
+				return estreeParent(vn.node), true
+			}
+		case "expression":
+			if vn.typeName == "JSXExpressionContainer" {
+				return unwrapExpression(vn.node.AsJsxExpression().Expression), true
+			}
+		}
+		return readNodeAttr(vn.node, segment, mc)
```
elecmonkey: [P2] Do not expose class fields on ClassBody

Falling through to `readNodeAttr` makes the virtual `ClassBody` inherit fields from its physical class node. ESLint reports nothing for `ClassBody[id]` or `ClassBody[superClass]`, but Rslint reports the class body; conversely, `ClassBody:not([id])` reports upstream and is suppressed here. The facade should expose only fields belonging to the ESTree `ClassBody`, such as `body`, `type`, and `parent`.

internal/rules/no_restricted_syntax/matcher.go
```diff
@@ -185,6 +209,20 @@ func matchesIdentifier(sel identifierSelector, node *ast.Node) bool {
...
 func refineEstreeMatch(name string, node *ast.Node) bool {
 	switch name {
+	case "Identifier":
+		return !isJSXIdentifier(node)
+	case "JSXIdentifier":
+		return isJSXIdentifier(node)
+	case "MemberExpression":
+		return !isJSXMemberExpression(node)
+	case "JSXMemberExpression":
+		return isJSXMemberExpression(node)
```
elecmonkey, line 219: [P2] Preserve JSX identities in attribute paths

The new bare-node refinements identify JSX names correctly, but attribute resolution still sees their physical ts-go identities. With `<Foo.Bar />`, ESLint reports for `JSXMemberExpression[type='JSXMemberExpression']` and `JSXMemberExpression[object.type='JSXIdentifier']`, while Rslint reports zero; `JSXIdentifier[type='JSXIdentifier']` similarly misses both identifiers. These remapped JSX nodes need facade-aware `type`, `parent`, `object`, and `property` traversal.

internal/rules/no_restricted_syntax/matcher.go
```diff
@@ -724,6 +898,45 @@ func stepAttrPath(current interface{}, segment string, mc *matchContext) (interf
...
+				return estreeParent(vn.node), true
+			}
+		case "expression":
+			if vn.typeName == "JSXExpressionContainer" {
+				return unwrapExpression(vn.node.AsJsxExpression().Expression), true
+			}
+		}
+		return readNodeAttr(vn.node, segment, mc)
+	}
+	if fv, ok := current.(functionValueFacade); ok {
```
elecmonkey, line 1254: [P2] Expose MethodDefinition.value as a matchable node

`functionValueFacade` supports a few dotted attributes but is not represented as the synthetic `FunctionExpression` child that ESTree exposes, and it has no `parent` field. ESLint reports for `MethodDefinition[value.parent.type='MethodDefinition']`, `MethodDefinition > FunctionExpression`, and `FunctionExpression > BlockStatement`; Rslint reports zero for all three. Model the synthetic node and its parent/children rather than limiting it to scalar attribute lookup.

internal/rules/no_restricted_syntax/mapping.go
```diff
@@ -53,10 +53,15 @@ var estreeKindMap = map[string][]ast.Kind{
 	"FunctionDeclaration":     {ast.KindFunctionDeclaration},
 	"FunctionExpression":      {ast.KindFunctionExpression},
 	"ArrowFunctionExpression": {ast.KindArrowFunction},
+	"TSEnumDeclaration":       {ast.KindEnumDeclaration},
```
elecmonkey, line 56: [P2] Add the TSEnumDeclaration ESTree fields

Mapping `KindEnumDeclaration` to `TSEnumDeclaration` makes only the bare type selector work. Using the repository's `@typescript-eslint/parser@8.65.0`, ESLint reports `enum E { A, B }` for `TSEnumDeclaration[id.name='E']`, while Rslint reports zero because `readIdAttr` and the field map do not handle enums; `body`/`TSEnumBody` and members are likewise absent. Add the TS-ESTree shape alongside the new type mapping.

internal/rules/no_restricted_syntax/matcher.go
```diff
@@ -1673,7 +1890,7 @@ func readValueAttr(node *ast.Node, mc *matchContext) (interface{}, bool) {
 		}
 		return text, true
 	case ast.KindBigIntLiteral:
-		return node.AsBigIntLiteral().Text, true
+		return utils.NormalizeBigIntLiteral(node.AsBigIntLiteral().Text), true
```
elecmonkey, line 2284: [P2] Normalize Literal.bigint as well as Literal.value

This normalizes the BigInt `value`, but the sibling `bigint` attribute still strips only the trailing `n` from source text. ESLint reports `Literal[bigint='1']` for `0x1n` and `Literal[bigint='1000']` for `1_000n`; Rslint reports zero for both. Reuse `NormalizeBigIntLiteral` for `readBigintAttr` so both ESTree properties have normalized decimal values.

internal/rules/no_restricted_syntax/no_restricted_syntax.go
```diff
@@ -384,19 +396,61 @@ func (bucket *ruleBucket) matchAndReport(index int, node *ast.Node, mc *matchCon
...
+	text := sf.Text()
+	leftLimit := max(node.Pos(), 0)
+	rightLimit := min(max(members.Pos()+1, leftLimit), len(text))
+	openOffset := strings.LastIndex(text[leftLimit:rightLimit], "{")
+	if openOffset < 0 {
+		return core.NewTextRange(node.Pos(), node.End())
+	}
+	open := leftLimit + openOffset
+	closeStart := min(max(members.End(), open+1), len(text))
+	closeOffset := strings.Index(text[closeStart:min(node.End(), len(text))], "}")
```
elecmonkey: [P2] Skip comment braces when locating ClassBody's end

`strings.Index` accepts the first `}` after the last member even when it is inside a comment. For `class C { m() {} /* } */ }`, ESLint's `ClassBody` diagnostic ends at column 27, while Rslint ends at column 22 inside the comment. Derive the brace range from tokens/parser positions rather than raw substring searches.

internal/rules/no_restricted_syntax/no_restricted_syntax.md
```diff
@@ -65,40 +65,45 @@ The rule accepts an array of restriction entries. Each entry is either:
 
 ### Supported selector forms
 
-The implementation covers the subset of [esquery] used in real-world ESLint
-configurations and in the upstream `no-restricted-syntax` test suite:
-
-- ESTree node names (e.g. `Identifier`, `FunctionDeclaration`, `BinaryExpression`).
+The implementation follows the esquery 1.7 selector forms used by ESLint
```
elecmonkey, line 68: [P2] Preserve the documented malformed-selector divergence

The updated documentation says the implementation follows esquery 1.7, but `parseRuleOptions` still silently drops parse errors. ESLint 10.9.1 throws a selector syntax error for `[`, whereas Rslint runs with no restriction, so a typo can silently disable a policy. Either reject malformed selectors before linting or retain the explicit compatibility warning removed by this patch.

internal/rules/no_restricted_syntax/probe_test.go
```diff
@@ -467,11 +464,10 @@ func TestNoRestrictedSyntax_ProbeDeepEdges(t *testing.T) {
 			},
 			// Probe: ban specific decorator name
 			{
-				Code:    `@sealed class A {}`,
+				Code:    `@sealed() class A {}`,
 				Options: []interface{}{`Decorator > CallExpression[callee.name='sealed']`},
-				Skip:    true, // tsgo wraps decorator over Identifier directly here, no Call. Skip; cover via :has below.
 				Errors: []rule_tester.InvalidTestCaseError{
-					{MessageId: "restrictedSyntax"},
+					{MessageId: "restrictedSyntax", Message: `Using 'Decorator > CallExpression[callee.name='sealed']' is not allowed.`},
```
elecmonkey, line 470: [P2] Cover Decorator.expression rather than only physical ancestry

The newly enabled test verifies `Decorator > CallExpression`, which happens to work through the physical ts-go parent, but the actual TS-ESTree `Decorator.expression` field remains absent. With `@sealed() class A {}`, ESLint reports `Decorator[expression.type='CallExpression']` while Rslint reports zero. Add `expression` attribute/field mapping for `KindDecorator` before documenting decorator selector parity.

internal/rules/no_restricted_syntax/no_restricted_syntax.go
```diff
@@ -535,6 +591,13 @@ func buildEntryFromObject(m map[string]interface{}) (ruleEntry, bool) {
 	return ruleEntry{selector: rawSel, compiled: compiled, message: msg}, true
 }
 
+// ESLint handles the event-phase suffix outside esquery. This rule emits the
+// same diagnostic on enter or exit, so matching can discard the suffix while
+// retaining it in the diagnostic text.
+func parseRuleSelector(sel string) (selector, error) {
+	return parseSelector(strings.TrimSuffix(sel, ":exit"))
```
elecmonkey: [P3] Dispatch :exit selectors on exit

Removing `:exit` and registering the selector as an enter listener changes observable diagnostic ordering. For `foo();` with `CallExpression:exit` and `Identifier`, ESLint reports the identifier first and the call on exit, while Rslint reports the call first; both ranges begin at the same offset, so stable final sorting does not correct this. Register the parsed selector on the corresponding exit listener instead of discarding its phase.

internal/rules/no_restricted_syntax/matcher.go
swwind: Fixed: ClassBody now exposes only the fields defined by ESTree.

internal/rules/no_restricted_syntax/matcher.go
swwind: Fixed: JSX member and identifier facade attribute paths now match ESTree.

internal/rules/no_restricted_syntax/matcher.go
swwind: Fixed: Literal.bigint now uses the same decimal BigInt normalization.

internal/rules/no_restricted_syntax/mapping.go
swwind: Fixed: Added TSEnumDeclaration, TSEnumBody, and TSEnumMember field mappings.

internal/rules/no_restricted_syntax/probe_test.go
swwind: Fixed: Decorator.expression is now mapped.

internal/rules/no_restricted_syntax/matcher.go
swwind: Fixed: MethodDefinition.value is now modeled as a traversable FunctionExpression facade.

internal/rules/no_restricted_syntax/no_restricted_syntax.md
swwind: Intentional compatibility divergence retained: rslint drops malformed selectors; the documentation now explicitly notes that ESLint rejects the whole configuration.

internal/rules/no_restricted_syntax/no_restricted_syntax.go
swwind: Fixed: `:exit` selectors are now registered on exit listeners and preserve diagnostic ordering.

internal/rules/no_restricted_syntax/no_restricted_syntax.go
swwind: Fixed: virtual nodes now support attribute-only selectors.

internal/rules/no_restricted_syntax/no_restricted_syntax.go
swwind: Fixed: ClassBody ranges now use parser tokens and are not truncated by `}` inside comments.

internal/rules/no_restricted_syntax/matcher.go
swwind: Fixed: virtual wrappers now participate in complete structural traversal and `:has()` matching.

internal/rules/no_restricted_syntax/mapping.go
elecmonkey, line 66: [P2] Preserve virtual ESTree node identity for ClassBody

Mapping `ClassBody` onto the class declaration/expression node lets simple `ClassBody` selectors report a class-body range, but attribute/combinator matching still evaluates against the physical class node. For example, ESLint 10.9.1 reports one diagnostic for `class C { m() {} }` with `ClassBody[parent.type='ClassDeclaration']` at the body range, while a temporary Rslint test on this PR reports 0 diagnostics because `parent.type` resolves to `Program`. It also collapses `:is(ClassDeclaration, ClassBody)` to one diagnostic where ESLint reports both the class and its body. Any config that restricts class bodies through parent/type/union selectors will silently under-report.

internal/rules/no_restricted_syntax/no_restricted_syntax.go
elecmonkey: [P2] Preserve JSXEmptyExpression as a separate selector target

`JSXEmptyExpression` is mapped to the same `JsxExpression` node as `JSXExpressionContainer`, but the matcher only special-cases reporting range and direct child matching. ESLint 10.9.1 reports one diagnostic for `const x = <div>{/* hi */}</div>;` with `JSXEmptyExpression[parent.type='JSXExpressionContainer']`, and two diagnostics for `:is(JSXExpressionContainer, JSXEmptyExpression)`. A temporary Rslint test on this PR reports 0 for the parent selector and collapses the union to the physical container node, so selectors involving the virtual empty expression are not 1:1 with esquery.

internal/rules/no_restricted_syntax/matcher.go
```diff
@@ -467,13 +565,36 @@ func lookupStringAttrPath(node *ast.Node, path []string, mc *matchContext) (text
 // matchesCombinator handles `>` (parent), descendant, `+` (prev sibling),
 // `~` (any prior sibling). The right-hand side has already been verified
 // against the current node before this function is called.
-func matchesCombinator(c combinatorSelector, node *ast.Node, mc *matchContext, scopeRoot *ast.Node) bool {
+func matchesCombinatorTarget(c combinatorSelector, node *ast.Node, mc *matchContext, scopeRoot *ast.Node, target string) bool {
+	if target != "" {
```
elecmonkey: [P1] Keep physical combinators on the physical path

Every rule match starts with `target == "physical"`, but this branch treats every non-empty target as virtual and calls `virtualParent`, which has no `physical` case. Consequently all child, descendant, adjacent, and sibling selectors return false: ESLint 10.9.1 reports `ArrowFunctionExpression > BlockStatement` for `() => {}`, while this head reports nothing and fails its own Go and npm tests. Only enter this branch for actual virtual targets.

internal/rules/no_restricted_syntax/no_restricted_syntax.go
```diff
@@ -384,19 +406,79 @@ func (bucket *ruleBucket) matchAndReport(index int, node *ast.Node, mc *matchCon
...
-	}
-	ctx.ReportNode(node, rule.RuleMessage{
+	message := rule.RuleMessage{
 		Id:          "restrictedSyntax",
 		Description: entry.formatMessage(),
-	})
+	}
+	physicalMatch := matchesInScopeTarget(compiled, node, mc, nil, "physical")
+
+	for _, target := range virtualTargets(node) {
```
elecmonkey, line 445: [P1] Recheck virtual targets with the unstripped selector

When string-attribute dispatch selects `[type='MethodDefinition']`, `compiled` has the type condition stripped before this loop, so the resulting wildcard also matches the method's virtual `FunctionExpression`; Rslint emits two identical diagnostics while ESLint emits one. `[type='ClassDeclaration']` similarly reports both the class and its `ClassBody`. Virtual targets must be checked with the original selector, or skipped for an indexed physical-only match.

internal/rules/no_restricted_syntax/matcher.go
```diff
@@ -467,13 +565,36 @@ func lookupStringAttrPath(node *ast.Node, path []string, mc *matchContext) (text
...
+			return false
+		}
+	}
 	switch c.Kind {
 	case combChild:
 		parent := matchingParent(node, scopeRoot)
 		if parent == nil {
 			return false
 		}
+		if matchesNodeVariants(c.Left, parent, mc, scopeRoot) {
```
elecmonkey: [P2] Preserve wrapper nodes in structural ancestry

Trying every virtual identity for the same physical parent does not reproduce the ESTree graph. After correcting the physical-target guard, Rslint reports `ClassDeclaration > MethodDefinition`, `MethodDefinition > BlockStatement`, and `TSEnumDeclaration > TSEnumMember`, whereas ESLint reports none because `ClassBody`, the synthetic `FunctionExpression`, and `TSEnumBody` respectively intervene. Structural matching needs explicit virtual parent-child edges rather than treating wrappers as alternate names for a physical node.

internal/rules/no_restricted_syntax/matcher.go
```diff
@@ -533,34 +654,182 @@ func matchesCombinator(c combinatorSelector, node *ast.Node, mc *matchContext, s
...
 	}
 	return false
 }
 
-func matchesPseudo(p pseudoSelector, node *ast.Node, mc *matchContext, scopeRoot *ast.Node) bool {
+func isEmptyJSXExpression(node *ast.Node) bool {
+	return node != nil && node.Kind == ast.KindJsxExpression && node.AsJsxExpression().DotDotDotToken == nil && node.AsJsxExpression().Expression == nil
+}
+
+func matchesPseudoTarget(p pseudoSelector, node *ast.Node, mc *matchContext, scopeRoot *ast.Node, target string) bool {
```
elecmonkey, line 974: [P2] Evaluate pseudos against the selected virtual target

`matchesPseudoTarget` propagates `target` only through `:is()` and `:not()`; `:has()`, positional pseudos, and semantic classes still inspect the physical node. Rslint therefore reports `ClassBody:has(Identifier[name='C'])` and `ClassBody:first-child` although ESLint reports neither, while missing `FunctionExpression:function` for class methods. These pseudos need the virtual target's children, list position, and type.

internal/rules/no_restricted_syntax/matcher.go
```diff
@@ -1196,6 +1523,12 @@ func readNodeAttr(node *ast.Node, name string, mc *matchContext) (interface{}, b
 	}
 	switch name {
 	case "type":
+		if isJSXMemberExpression(node) {
```
elecmonkey, line 1787: [P2] Use refined JSX identities in type dispatch

The new JSX refinements are bypassed by `lookupStringAttrPath`, which resolves a final `.type` through `estreeNameForKind`; that function still returns `MemberExpression` or `Identifier`, and has no JSX container/element cases. On the exact head, ESLint reports `JSXMemberExpression[type='JSXMemberExpression']`, `JSXIdentifier[type='JSXIdentifier']`, and `[type='JSXExpressionContainer']`, while Rslint reports zero for each. Type dispatch must use the same refined JSX identity as normal matching.

internal/rules/no_restricted_syntax/no_restricted_syntax.go
```diff
@@ -384,19 +406,79 @@ func (bucket *ruleBucket) matchAndReport(index int, node *ast.Node, mc *matchCon
...
+		switch target {
+		case "ClassBody":
+			ctx.ReportRange(classBodyTextRange(mc.sf, node), message)
+		case "TSEnumBody":
+			ctx.ReportRange(enumBodyTextRange(mc.sf, node), message)
+		case "JSXEmptyExpression":
+			nodeRange := utils.TrimNodeTextRange(mc.sf, node)
+			ctx.ReportRange(core.NewTextRange(nodeRange.Pos()+1, max(nodeRange.Pos()+1, nodeRange.End()-1)), message)
+		default:
+			ctx.ReportNode(node, message)
```
elecmonkey, line 469: [P2] Report the synthetic FunctionExpression range

The default virtual-target branch reports the entire physical method node, but TS-ESTree's synthetic `MethodDefinition.value` starts after the method key and modifiers. For `class C { m() {} }`, ESLint reports the `FunctionExpression` at columns 12–17 while Rslint reports 11–17; for a constructor the mismatch is 22–27 versus 11–27. Compute and report the synthetic function-value range instead of calling `ReportNode` on the method.

internal/rules/no_restricted_syntax/matcher.go
```diff
@@ -2278,10 +2614,117 @@ func readExpressionAttr(node *ast.Node) (interface{}, bool) {
...
+	switch facade.typeName {
+	case "ClassBody":
+		if name == "body" {
+			return readBodyAttr(node)
+		}
+		if name == "type" || name == "parent" {
+			return nil, false
+		}
+	case "TSEnumBody":
+		if name == "body" || name == "members" {
```
elecmonkey: [P2] Expose only TSEnumBody.members

TS-ESTree 8.65.0 defines `TSEnumBody` with a `members` field, not a `body` alias. This branch makes `TSEnumBody[body]` match `enum E { A }` in Rslint even though ESLint reports nothing, creating a false positive for a nonexistent property. Remove the `body` alias from this facade.

internal/rules/no_restricted_syntax/matcher.go
```diff
@@ -533,34 +654,182 @@ func matchesCombinator(c combinatorSelector, node *ast.Node, mc *matchContext, s
...
+		return true
+	}
+	return false
+}
+
+func isEnumDeclarationNode(node *ast.Node) bool {
+	return node != nil && node.Kind == ast.KindEnumDeclaration
+}
+
+func virtualTargets(node *ast.Node) []string {
```
elecmonkey, line 747: [P2] Split self-closing JSX into both ESTree nodes

A ts-go `JsxSelfClosingElement` represents both a TS-ESTree `JSXElement` and its `JSXOpeningElement`, but `virtualTargets` creates neither identity separately. ESLint reports two diagnostics for `:is(JSXElement, JSXOpeningElement)` on `<Foo />` and matches `JSXElement > JSXOpeningElement`; Rslint reports one and zero respectively. Add distinct virtual targets and their parent relationship for this collapsed shape.

elecmonkey (commented):
Reviewed against ESLint 10.9.1 and esquery 1.7.0 with the complete upstream no-restricted-syntax suite and targeted differential cases. The current head also fails its own Go and npm tests.

internal/rules/no_restricted_syntax/matcher.go
```diff
@@ -467,13 +565,36 @@ func lookupStringAttrPath(node *ast.Node, path []string, mc *matchContext) (text
 // matchesCombinator handles `>` (parent), descendant, `+` (prev sibling),
 // `~` (any prior sibling). The right-hand side has already been verified
 // against the current node before this function is called.
-func matchesCombinator(c combinatorSelector, node *ast.Node, mc *matchContext, scopeRoot *ast.Node) bool {
+func matchesCombinatorTarget(c combinatorSelector, node *ast.Node, mc *matchContext, scopeRoot *ast.Node, target string) bool {
```
elecmonkey, line 606: [P1] Keep physical targets on the normal combinator path

`matchAndReport` calls this function with `target == "physical"`, but this branch treats every non-empty target as virtual and returns false when `virtualParent` rejects that sentinel. As a result, ordinary child, descendant, adjacent, and sibling selectors stop matching. For example, ESLint 10.9.1 reports `() => {}` for `ArrowFunctionExpression > BlockStatement`, while this head reports nothing; both the Go package test and the JS integration test fail on this upstream case.

internal/rules/no_restricted_syntax/no_restricted_syntax.go
elecmonkey, line 445: [P2] Reapply dispatch predicates to virtual targets

When the dispatch index matches, `compiled` has the indexed attribute removed before this loop. For `class C { foo() {} }` with `[type='MethodDefinition']`, the stripped selector is `*`, so both the physical MethodDefinition and its virtual FunctionExpression are reported. ESLint 10.9.1 emits one diagnostic, while this head emits two; virtual variants need to be checked against the original selector rather than the dispatch-stripped form.

internal/rules/no_restricted_syntax/no_restricted_syntax.go
```diff
@@ -292,7 +303,13 @@ func supportsSingleDispatch(path []string) bool {
 }
 
 func collectDispatchAttrs(sel selector, attrs []dispatchAttr) []dispatchAttr {
```
elecmonkey, line 326: [P2] Keep JSX type dispatch consistent with the new mappings

Selectors ending in `[type='…']` are indexed through `lookupStringAttrPath`, whose fast path still derives the physical ts-go type. With `const x = <Foo.Bar />` and `JSXMemberExpression[type='JSXMemberExpression']`, ESLint 10.9.1 reports one diagnostic but this head reports none; the PR's own `invalid-59` test fails. Please either exclude remapped JSX identities from this dispatch or make the dispatch lookup facade-aware.

internal/rules/no_restricted_syntax/matcher.go
```diff
@@ -533,34 +654,182 @@ func matchesCombinator(c combinatorSelector, node *ast.Node, mc *matchContext, s
...
+		}
+	}
+	return false
+}
+
+// virtualParent returns the physical node and wrapper identity that is the
+// parent of a virtual ESTree node. The wrapper itself is represented by the
+// same tsgo node, so this is enough to evaluate structural selectors without
+// allocating synthetic AST nodes.
+func virtualParent(node *ast.Node, target string) (*ast.Node, string, bool) {
```
elecmonkey, line 786: [P2] Insert virtual wrappers into child parent chains

This models how a virtual node reaches its parent, but physical children still resolve `parent` through the physical ts-go tree. Thus `MethodDefinition[parent.type='ClassBody']` on `class C { m() {} }` reports once in ESLint 10.9.1 and zero times here; `TSEnumMember[parent.type='TSEnumBody']`, deeper FunctionExpression ancestry, and virtual field selectors have the same gap. The virtual wrapper also needs to become the logical parent of its ESTree children.

internal/rules/no_restricted_syntax/matcher.go
elecmonkey, line 974: [P2] Evaluate semantic classes against the virtual type

`matchesPseudoTarget` receives the virtual target but delegates semantic classes to `matchesNodeClass`, which derives the physical node type. Consequently `ClassBody:declaration` incorrectly reports here although ESLint reports nothing, while `FunctionExpression:function` on a class method is missed although ESLint reports it. Please evaluate these classes using the target identity.

internal/rules/no_restricted_syntax/no_restricted_syntax.go
```diff
@@ -384,19 +406,79 @@ func (bucket *ruleBucket) matchAndReport(index int, node *ast.Node, mc *matchCon
...
+		}
+		switch target {
+		case "ClassBody":
+			ctx.ReportRange(classBodyTextRange(mc.sf, node), message)
+		case "TSEnumBody":
+			ctx.ReportRange(enumBodyTextRange(mc.sf, node), message)
+		case "JSXEmptyExpression":
+			nodeRange := utils.TrimNodeTextRange(mc.sf, node)
+			ctx.ReportRange(core.NewTextRange(nodeRange.Pos()+1, max(nodeRange.Pos()+1, nodeRange.End()-1)), message)
+		default:
```
elecmonkey, line 468: [P2] Report the synthetic FunctionExpression range

The default virtual-target branch reports the underlying method node rather than the synthetic FunctionExpression span. For `class C { m() {} }` with `FunctionExpression`, ESLint 10.9.1 reports columns 12–17, while this head reports columns 11–17 and includes the method key. The MethodDefinition value facade needs its own ESTree-compatible range.

internal/rules/no_restricted_syntax/matcher.go
swwind: Fixed in e35b340c6. Physical matching now stays on the physical combinator path; virtual-parent handling is entered only for actual virtual targets. Added regression coverage for child, descendant, adjacent, and sibling selectors.

internal/rules/no_restricted_syntax/matcher.go
swwind: Fixed in e35b340c6. Physical matching now stays on the physical combinator path; virtual-parent handling is entered only for actual virtual targets. Added regression coverage for child, descendant, adjacent, and sibling selectors.

internal/rules/no_restricted_syntax/no_restricted_syntax.go
swwind: Fixed in e35b340c6. Virtual targets are now rechecked with the original selector after dispatch predicates are stripped, preventing duplicate diagnostics.

swwind: Fixed in e35b340c6. Virtual targets are now rechecked with the original selector after dispatch predicates are stripped, preventing duplicate diagnostics.

internal/rules/no_restricted_syntax/matcher.go
swwind: Fixed in e35b340c6. Structural ancestry now models ClassBody, FunctionExpression, and TSEnumBody explicitly, so direct selectors no longer skip ESTree wrapper nodes.

internal/rules/no_restricted_syntax/matcher.go
swwind: Fixed in e35b340c6. :has, positional pseudos, and semantic classes now evaluate the selected virtual identity and its logical children/list position.

internal/rules/no_restricted_syntax/matcher.go
swwind: Fixed in e35b340c6. Type dispatch now uses the refined ESTree/facade identity for JSX nodes.

internal/rules/no_restricted_syntax/no_restricted_syntax.go
swwind: Fixed in e35b340c6. Type dispatch now uses the refined ESTree/facade identity for JSX nodes.

internal/rules/no_restricted_syntax/no_restricted_syntax.go
swwind: Fixed in e35b340c6. Synthetic FunctionExpression matches now report the ESTree MethodDefinition.value range instead of the entire physical method node.

internal/rules/no_restricted_syntax/no_restricted_syntax.go
swwind: Fixed in e35b340c6. Synthetic FunctionExpression matches now report the ESTree MethodDefinition.value range instead of the entire physical method node.

internal/rules/no_restricted_syntax/matcher.go
swwind: Fixed in e35b340c6. TSEnumBody now exposes only the ESTree members field; the body alias was removed.

internal/rules/no_restricted_syntax/matcher.go
swwind: Fixed in e35b340c6. Self-closing JSX nodes now expose distinct virtual JSXElement and JSXOpeningElement identities with the correct parent relationship.

internal/rules/no_restricted_syntax/matcher.go
swwind: Fixed in e35b340c6. Semantic classes now evaluate the selected virtual ESTree identity, so virtual ClassBody and FunctionExpression nodes follow the correct semantics.

internal/rules/no_restricted_syntax/no_restricted_syntax.go
```diff
@@ -384,19 +407,127 @@ func (bucket *ruleBucket) matchAndReport(index int, node *ast.Node, mc *matchCon
...
-	ctx.ReportNode(node, rule.RuleMessage{
+	message := rule.RuleMessage{
 		Id:          "restrictedSyntax",
 		Description: entry.formatMessage(),
-	})
+	}
+	physicalMatch := matchesInScopeTarget(compiled, node, mc, nil, "physical")
+
+	for _, target := range virtualTargets(node) {
+		if !selectorTargetsEstreeType(entry.compiled, target) {
```
elecmonkey: [P2] Try virtual targets for semantic class selectors

This gate still skips virtual ESTree identities unless the selector explicitly names that virtual type. ESLint 10.9.1 reports the synthetic method value for `class C { m() {} }` with `:function`, `MethodDefinition > :function`, and `:function[body.type='BlockStatement']`, but a temporary Rslint overlay test on this head reports 0 diagnostics because `selectorTargetsEstreeType` never recognizes `:function` as targeting the virtual `FunctionExpression`. Please also evaluate applicable virtual targets for semantic class selectors, not only explicit `FunctionExpression` selectors.

swwind: Fixed in dc17fd2fa. Semantic pseudo-class selectors now identify applicable virtual ESTree targets, so `:function`, `MethodDefinition > :function`, and `:function[body.type='BlockStatement']` match class method `FunctionExpression` values and report the synthetic `MethodDefinition.value` range. Added regression coverage.

internal/rules/no_restricted_syntax/matcher.go
```diff
@@ -533,62 +689,320 @@ func matchesCombinator(c combinatorSelector, node *ast.Node, mc *matchContext, s
...
+	}
+	var targets []string
+	if isClassLikeNode(node) {
+		targets = append(targets, "ClassBody")
+	}
+	if isEnumDeclarationNode(node) {
+		targets = append(targets, "TSEnumBody")
+	}
+	if node.Kind == ast.KindJsxSelfClosingElement {
+		targets = append(targets, "JSXElement", "JSXOpeningElement")
```
elecmonkey, line 759: [P2] Avoid double-reporting self-closing JSXElement type selectors

Self-closing JSX nodes already expose `type: "JSXElement"` on the physical path via `estreeNameForKind`, but this virtual target makes attribute-only selectors match the same tsgo node again. With `const x = <Foo />;` and `[type='JSXElement']`, ESLint 10.9.1 reports one diagnostic at the JSX node, while this PR reports two identical diagnostics for the same range. Please avoid adding the virtual `JSXElement` report when the physical path already represents that ESTree identity.

internal/rules/no_restricted_syntax/no_restricted_syntax.go
```diff
@@ -384,19 +407,127 @@ func (bucket *ruleBucket) matchAndReport(index int, node *ast.Node, mc *matchCon
...
+		}
+		if token.Kind == ast.KindOpenParenToken {
+			openIndex = index
+		}
+	}
+	if openIndex < 0 {
+		return core.NewTextRange(node.Pos(), node.End())
+	}
+	start := tokens[openIndex].Start
+	if openIndex > 0 && tokens[openIndex-1].Kind == ast.KindGreaterThanToken {
```
elecmonkey, line 527: [P2] Include generic method type parameters in FunctionExpression ranges

The synthetic `MethodDefinition.value` range should match TS-ESTree's `FunctionExpression` span, but this backtracking starts from the `>` before `(` with `depth := 1`, so `m<T>() {}` never reaches depth zero and the reported start remains at the opening paren. ESLint 10.9.1 reports `class C { m<T>() {} }` for `FunctionExpression` starting at the `<` (column 12), while this PR reports from `(` (column 15); async and constrained generic methods have the same range drift.

internal/rules/no_restricted_syntax/matcher.go
swwind: Fixed in 00cb4cc71. Self-closing JSXElement is already represented by the physical tsgo node for type-attribute matching, so the duplicate virtual JSXElement report is now suppressed while JSXOpeningElement and union selectors remain distinct. Added regression coverage for `<Foo />` with `[type='JSXElement']`; focused Go and JS tests pass.

internal/rules/no_restricted_syntax/no_restricted_syntax.go
swwind: Fixed in 00cb4cc71. Generic method type-parameter angle brackets are now included when computing synthetic FunctionExpression ranges, covering regular, async constrained, and constrained generic methods. Added exact-range regressions; focused Go/JS and ESLint corpus differential tests pass.

internal/rules/no_restricted_syntax/matcher.go
```diff
@@ -2278,10 +2874,137 @@ func readExpressionAttr(node *ast.Node) (interface{}, bool) {
...
+		}
+		if name == "closingElement" || name == "type" || name == "parent" {
+			return nil, true
+		}
+	case "JSXOpeningElement":
+		if name == "name" || name == "tagName" {
+			return jsxFacade(node.AsJsxSelfClosingElement().TagName), true
+		}
+		if name == "attributes" {
+			return node.AsJsxSelfClosingElement().Attributes, true
```
elecmonkey: [P2] Return JSX attribute node slices from the virtual facade

For self-closing JSX, the new virtual `JSXOpeningElement` path returns the raw ts-go `NodeList`, but `stepAttrPath` only implements `.length` and numeric segments for `[]*ast.Node`. As a result, `const x = <Foo a b={c} />;` with `JSXOpeningElement[attributes.length=2]` or `JSXOpeningElement[attributes.0.name.name='a']` reports in ESLint 10.9.1 but Rslint reports nothing, so policies that inspect opening-element attributes silently under-report. Return the underlying `Attributes.Nodes` slice here, matching other list-valued ESTree fields.

internal/rules/no_restricted_syntax/no_restricted_syntax.go
```diff
@@ -384,19 +407,133 @@ func (bucket *ruleBucket) matchAndReport(index int, node *ast.Node, mc *matchCon
...
-	}
-	ctx.ReportNode(node, rule.RuleMessage{
+	message := rule.RuleMessage{
 		Id:          "restrictedSyntax",
 		Description: entry.formatMessage(),
-	})
+	}
+	physicalMatch := matchesInScopeTarget(compiled, node, mc, nil, "physical")
+
+	for _, target := range virtualTargets(node) {
```
elecmonkey, line 445: [P3] Preserve physical JSXElement report order

This loop reports virtual targets before the physical node is reported below, which changes observable diagnostic order for self-closing JSX when multiple selectors match the same range. For `const x = <Foo />;` with `:is(JSXElement, JSXOpeningElement)`, `[type='JSXElement']`, and `JSXElement > JSXOpeningElement`, ESLint reports the physical `JSXElement` matches first and then the synthetic opening element, while Rslint emits the virtual opening-element matches before `[type='JSXElement']`. Since the rule snapshots and prior `:exit` fix treat ordering as parity-sensitive, the physical report should be emitted before walking the synthetic `JSXOpeningElement` target.

elecmonkey (commented):
Reviewed against ESLint 10.9.1 and esquery 1.7.0 with the upstream no-restricted-syntax suite, focused Go/JS parity tests, and targeted virtual-node differential cases.

internal/rules/no_restricted_syntax/matcher.go
swwind: Fixed in `a51677d2e`. The virtual `JSXOpeningElement` facade now returns the underlying `[]*ast.Node` attribute slice, so `.length` and numeric attribute paths are evaluated correctly. Added regressions for both selectors; focused Go, JS, and ESLint 10.9.1 differential tests pass.

internal/rules/no_restricted_syntax/no_restricted_syntax.go
swwind: Fixed in `a51677d2e`. Self-closing JSX matching now reports the physical `JSXElement` before the synthetic `JSXOpeningElement`, preserving ESLint's observable ordering. Added an exact ordering regression; focused Go, JS, and ESLint 10.9.1 differential tests pass.

internal/rules/no_restricted_syntax/no_restricted_syntax.go
```diff
@@ -384,7 +425,66 @@ func (bucket *ruleBucket) matchAndReport(index int, node *ast.Node, mc *matchCon
...
 		compiled = bucket.dispatch.matched[index]
 	}
-	if !matches(compiled, node, mc) {
+	message := rule.RuleMessage{
+		Id:          "restrictedSyntax",
+		Description: entry.formatMessage(),
+	}
+	physicalMatch := matchesInScopeTarget(compiled, node, mc, nil, "physical")
+
+	for _, target := range virtualTargets(node) {
```
elecmonkey, line 445: [P3] Preserve physical reports before virtual facades

This loop emits every virtual target before the physical node is reported below, which still changes ESLint's observable diagnostic order for non-JSX facades. For `class C { m() {} }` with `:is(ClassDeclaration, ClassBody)` followed by `[type='ClassDeclaration']`, ESLint 10.9.1 reports `union`, then `physical`, then the ClassBody `union`; current Rslint reports the ClassBody `union` first, then the physical class reports. The same happens for `:is(MethodDefinition, FunctionExpression)` plus `[type='MethodDefinition']`. Since this PR already treats selector/report order as parity-sensitive for self-closing JSX, the same physical-before-virtual ordering needs to apply here too.

swwind: Fixed in 8678718fe. Physical matches are now reported in selector order before virtual facade matches, preserving ESLint's ordering for ClassBody and synthetic FunctionExpression targets. Added regressions for both cases. The focused Go and JS tests, ESLint 10.9.1 corpus differential test, spell check, and Go lint all pass.

internal/rules/no_restricted_syntax/no_restricted_syntax.go
```diff
@@ -378,13 +422,71 @@ func (bucket *ruleBucket) candidates(node *ast.Node, mc *matchContext) (first, s
...
+}
+
+func (bucket *ruleBucket) matchVirtual(index int, node *ast.Node, mc *matchContext, ctx *rule.RuleContext) {
+	entry := &bucket.entries[index]
+	message := rule.RuleMessage{
+		Id:          "restrictedSyntax",
+		Description: entry.formatMessage(),
+	}
+	for _, target := range virtualTargets(node) {
+		if !selectorTargetsEstreeType(entry.compiled, target) {
```
elecmonkey: [P2] Try virtual facades for non-explicit type selectors

This gate skips virtual targets unless the selector positively names the facade, but esquery also matches virtual ESTree nodes through broader selectors such as `[type=/^(ClassBody|FunctionExpression)$/]` and negative selectors like `:not(Identifier):not(ClassDeclaration):not(MethodDefinition):not(BlockStatement):not(Program)`. With `class C { m() {} }`, ESLint 10.9.1 reports both the `ClassBody` and synthetic `FunctionExpression` ranges for those selectors, while current Rslint reports none for the virtual nodes. Since the PR documents these wrappers as modeled except for bare `*`, this still silently under-reports real no-restricted-syntax configs.

swwind: Fixed in d88cb2114. Broad selectors now evaluate applicable virtual ESTree facades, including regex type selectors such as `[type=/^(ClassBody|FunctionExpression)$/]` and chained negative selectors. Bare `*` retains physical-tree-only behavior. Added permanent Go regressions; focused Go/JS tests, ESLint 10.9.1 corpus differential tests, spell check, and Go lint pass.

internal/rules/no_restricted_syntax/no_restricted_syntax.go
```diff
@@ -378,13 +422,74 @@ func (bucket *ruleBucket) candidates(node *ast.Node, mc *matchContext) (first, s
...
+
+	physicalMatch := matchesInScopeTarget(compiled, node, mc, nil, "physical")
+	if target == "JSXElement" && physicalMatch {
+		ctx.ReportNode(node, rule.RuleMessage{
+			Id:          "restrictedSyntax",
+			Description: entry.formatMessage(),
+		})
+		return
+	}
+	if !selectorTargetsEstreeType(entry.compiled, target) {
```
elecmonkey, line 489: [P2] Match broad selectors against self-closing JSX openings

This guard only reports the synthetic `JSXOpeningElement` facade when `selectorTargetsEstreeType` can prove the selector explicitly names that type, so broader selectors never get evaluated against the virtual opening node. I reproduced this with ESLint 10.9.1 on `const x = <Foo a b={c} />;`: `[type=/^(JSXElement|JSXOpeningElement)$/]` reports two diagnostics, but current Rslint reports only the physical `JSXElement`; `JSXElement > [type=/^JSXOpeningElement$/]` reports one in ESLint and zero in Rslint. This leaves real `no-restricted-syntax` configs using regex/type predicates or broad structural selectors silently under-reporting self-closing JSX.

fansenze (approved):

PR #1970, opened 2026-08-31
feat: port rule react/prop-types

swwind:
## Summary

Port the `react/prop-types` rule from eslint-plugin-react to rslint.

The rule reports React component props that are used without corresponding
`propTypes` validation, supporting class components, `createReactClass`,
stateless function components, nested prop paths, and the `ignore` and
`skipUndeclared` options.

## Related Links

- ESLint rule: https://github.com/jsx-eslint/eslint-plugin-react/blob/v7.37.5/docs/rules/prop-types.md
- Source code: https://github.com/jsx-eslint/eslint-plugin-react/blob/v7.37.5/lib/rules/prop-types.js

## Checklist

- [x] Tests updated.
- [x] Documentation updated.


internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,435 @@
...
+			case ast.KindObjectLiteralExpression:
+				if reactutil.IsCreateReactClassObjectArg(n, pragma, createClass) {
+					c := &component{node: n, declared: map[string]propType{}, destructured: map[string]string{}}
+					comps = append(comps, c)
+				}
+			case ast.KindFunctionDeclaration, ast.KindFunctionExpression, ast.KindArrowFunction, ast.KindMethodDeclaration:
+				if reactutil.IsStatelessReactComponentWithWrappers(n, pragma, ctx.TypeChecker, wrappers) {
+					c := &component{node: n, declared: map[string]propType{}, destructured: map[string]string{}}
+					comps = append(comps, c)
+					if name := className(n); name != "" {
```
elecmonkey: [P1] Associate variable-assigned function components with propTypes

For `FunctionExpression` and `ArrowFunction` components this uses `BindingIdentifierName(n)`, which is empty because the binding lives on the parent `VariableDeclaration`, so `const Hello = props => <div>{props.name}</div>; Hello.propTypes = { name: PropTypes.string };` never populates `byName["Hello"]`. The later assignment handler therefore leaves the component undeclared and Rslint reports `'name' is missing`, while `eslint-plugin-react@7.37.5` reports no diagnostics for the same source. This makes the common `const Component = (...) => ...; Component.propTypes = ...` pattern produce false positives.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,435 @@
...
+
+func validatorType(n *ast.Node) propType {
+	n = unwrap(n)
+	if n == nil {
+		return propType{any: true}
+	}
+	if n.Kind == ast.KindPropertyAccessExpression && keyName(n.AsPropertyAccessExpression().Name()) == "isRequired" {
+		n = unwrap(n.AsPropertyAccessExpression().Expression)
+	}
+	if n.Kind != ast.KindCallExpression {
```
elecmonkey, line 177: [P1] Treat broad PropTypes validators as declaring nested reads

Non-call validators fall through to an empty `propType`, so `PropTypes.object.isRequired`, `PropTypes.array`, `PropTypes.string`, and similar declarations only validate the top-level prop and then reject nested reads like `this.props.name.firstname`. Upstream marks `name: PropTypes.object.isRequired` with `this.props.name.firstname` as valid because broad validators accept child access; the local probe reports `'name.firstname' is missing`. This will create false positives for ordinary object, array, string, function, and number props when their members are read.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,435 @@
...
+							c.declaredBlock = true
+						} else {
+							c.declaredBlock = true
+							c.declared = map[string]propType{"__ANY_KEY__": {any: true}}
+						}
+					}
+				}
+			case ast.KindBinaryExpression:
+				b := n.AsBinaryExpression()
+				if b.OperatorToken != nil && b.OperatorToken.Kind == ast.KindEqualsToken && b.Left != nil && b.Left.Kind == ast.KindPropertyAccessExpression {
```
elecmonkey, line 1541: [P2] Merge nested propTypes assignments

The assignment handler only accepts an LHS whose final property is exactly `propTypes`, so it ignores declarations added as `Hello.propTypes.a = PropTypes.shape({ b: PropTypes.string })`. Upstream treats that incremental assignment as declaring `a.b`; the local probe reports both `'a.b'` and `'a'` as missing for `this.props.a.b`. Codebases often build propTypes incrementally, so this creates false positives whenever declarations are split across `Component.propTypes = {}` plus member assignments.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,435 @@
...
+						if m, ok := declared(pa.Initializer); ok {
+							c.declared = m
+							c.declaredBlock = true
+						} else {
+							c.declaredBlock = true
+							c.declared = map[string]propType{"__ANY_KEY__": {any: true}}
+						}
+					}
+				}
+			case ast.KindPropertyDeclaration:
```
elecmonkey, line 1493: [P2] Read propTypes declared by static getters

Class members are limited to `PropertyDeclaration`, so `static get propTypes() { return { name: PropTypes.string }; }` is skipped and the component is treated as missing validation. `eslint-plugin-react@7.37.5` accepts that declaration form and reports no diagnostics, while the local Rslint probe reports `'name' is missing`. Static getter propTypes are part of the upstream rule surface and need to populate `declaredPropTypes`.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,435 @@
...
+					comps = append(comps, c)
+					if name := className(n); name != "" {
+						byName[name] = c
+					}
+					ps := n.Parameters()
+					if len(ps) > 0 && ps[0].Kind == ast.KindParameter {
+						name := ps[0].AsParameterDeclaration().Name()
+						if name != nil && name.Kind == ast.KindIdentifier {
+							c.props = name.AsIdentifier().Text
+						} else if name != nil && name.Kind == ast.KindObjectBindingPattern {
```
elecmonkey, line 1444: [P1] Track destructuring from the props binding inside component bodies

Only destructuring in the first parameter is recorded, so `function Hello(props) { const { name } = props; return <div>{name}</div>; }` produces no used prop in Rslint. Upstream reports `missingPropType` for `name` on the same source. This misses a very common props access pattern, including class-side `const { name } = this.props`, and lets components use undeclared props without diagnostics.

internal/plugins/react/rules/prop_types/prop_types.go
swwind: Fixed in 832bd9978. Variable-assigned function components now resolve their binding through expression and component-wrapper chains, including assignment-form components; regression coverage was added.

internal/plugins/react/rules/prop_types/prop_types.go
swwind: Fixed in 832bd9978. Broad non-call validators now declare nested reads, while arrayOf/objectOf retain their element-shape validation; regression coverage was added.

internal/plugins/react/rules/prop_types/prop_types.go
swwind: Fixed in 832bd9978. Nested and incremental propTypes assignments now update the declaration tree, including namespaced component paths; regression coverage was added.

internal/plugins/react/rules/prop_types/prop_types.go
swwind: Fixed in 832bd9978. Only static propTypes getters are recognized, using the last top-level return as upstream does; regression coverage was added.

internal/plugins/react/rules/prop_types/prop_types.go
swwind: Fixed in 832bd9978. Props tracking now covers body destructuring, nested paths, aliases, class props destructuring, lifecycle parameters, and setState updater props; regression coverage was added.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,933 @@
...
+	name := ""
+	if callee != nil && callee.Kind == ast.KindPropertyAccessExpression {
+		name = keyName(callee.AsPropertyAccessExpression().Name())
+		if object := unwrap(callee.AsPropertyAccessExpression().Expression); object != nil && object.Kind == ast.KindIdentifier && slices.Contains(customValidators, object.AsIdentifier().Text) {
+			return propType{open: true}
+		}
+	}
+	if name == "shape" || name == "exact" {
+		if call.Arguments != nil && len(call.Arguments.Nodes) > 0 {
+			if m, ok := propMap(call.Arguments.Nodes[0], customValidators); ok {
```
elecmonkey: [P1] Preserve the recursion guard when descending into shapes

A recursive validator such as `const nodeType = PropTypes.shape({ child: nodeType }); Tree.propTypes = { node: nodeType };` makes Rslint terminate with a Go stack overflow, while eslint-plugin-react@7.37.5 completes without diagnostics. The `seen` map reaches `validatorTypeSeen`, but the nested `propMap` call invokes `validatorType`, which creates a fresh map and re-enters the cycle indefinitely.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,933 @@
...
+
+func declared(n *ast.Node, customValidators []string, propWrappers []reactutil.PropWrapperEntry) (map[string]propType, bool) {
+	n = unwrap(n)
+	if n != nil && n.Kind == ast.KindCallExpression && reactutil.IsPropWrapperCall(n, propWrappers) {
+		call := n.AsCallExpression()
+		if call.Arguments != nil && len(call.Arguments.Nodes) > 0 {
+			return declared(call.Arguments.Nodes[0], customValidators, propWrappers)
+		}
+	}
+	m, ok := propMap(n, customValidators)
```
elecmonkey, line 252: [P1] Resolve local propTypes identifiers before accepting all props

For `const types = { name: PropTypes.string }; function Hello(props) { return <div>{props.name}{props.missing}</div>; } Hello.propTypes = types;`, upstream reports only `missing`, whereas Rslint reports nothing. `declared` only unwraps configured calls and object literals, so the caller converts every identifier—including locally resolvable object literals—into `__ANY_KEY__`; resolve local bindings first and reserve the opaque fallback for genuinely external declarations.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,933 @@
...
+				if candidate != nil && candidate.Kind != ast.KindOmittedExpression {
+					union = append(union, validatorTypeSeen(candidate, customValidators, seen))
+				}
+			}
+			if len(union) > 0 {
+				return propType{union: union}
+			}
+		}
+	}
+	return propType{}
```
elecmonkey: [P1] Keep unmodeled call validators open

Common validators implemented as calls fall through to an empty `propType`, so nested reads under `PropTypes.instanceOf(User)`, `PropTypes.oneOf(...)`, or `PropTypes.shape(dynamicValue)` are reported by Rslint even though upstream treats those validators as accepting nested access. For example, `user: PropTypes.instanceOf(User)` with `props.user.name` produces `user.name` only in Rslint.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,933 @@
...
+	case ast.KindTypeLiteral:
+		members = n.AsTypeLiteralNode().Members
+	case ast.KindInterfaceDeclaration:
+		members = n.AsInterfaceDeclaration().Members
+	case ast.KindTypeAliasDeclaration:
+		return declaredType(n.AsTypeAliasDeclaration().Type, aliases, depth-1)
+	case ast.KindTypeReference:
+		name := reactutil.EntityNameRightmost(n.AsTypeReferenceNode().TypeName)
+		if name != nil {
+			return declaredType(aliases[name.AsIdentifier().Text], aliases, depth-1)
```
elecmonkey: [P1] Treat unresolved TypeScript annotations as opaque

With `import type { Props } from './types'; function Hello(props: Props) { return <div>{props.name}</div>; }`, eslint-plugin-react@7.37.5 emits no diagnostic because an unresolved or imported type disables validation, but Rslint returns `false` here and later reports `name`. Imported prop interfaces are routine, so unresolved references must preserve upstream's unknown/ignore state rather than behave like an absent declaration.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,933 @@
...
+
+func declaredType(n *ast.Node, aliases map[string]*ast.Node, depth int) (map[string]propType, bool) {
+	if n == nil || depth == 0 {
+		return nil, false
+	}
+	var members *ast.TypeElementList
+	switch n.Kind {
+	case ast.KindTypeLiteral:
+		members = n.AsTypeLiteralNode().Members
+	case ast.KindInterfaceDeclaration:
```
elecmonkey, line 351: [P1] Traverse complete TypeScript interface declarations

The interface branch reads only direct members, and the member loop accepts only property signatures. Consequently both `interface Props extends Base {}` and direct method declarations such as `interface Props { onClick(): void }` produce false positives for inherited or method props, while upstream accepts them. Interface heritage and `MethodSignature` members need to participate in declaration construction.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,933 @@
...
+							c.props = name.AsIdentifier().Text
+						} else if name != nil && name.Kind == ast.KindObjectBindingPattern {
+							addDestructured(c, name, nil)
+						}
+						if parameter.Type != nil {
+							if declared, ok := declaredType(parameter.Type.AsNode(), typeAliases, 8); ok {
+								c.declared = declared
+								c.declaredBlock = true
+							}
+						} else if declared, ok := declaredType(firstTypeArgument(componentVariableType(n)), typeAliases, 8); ok {
```
elecmonkey: [P1] Infer props only from supported React generic positions

The fallback unconditionally takes the first type argument of any variable annotation. This suppresses an upstream diagnostic for unrelated annotations such as `const Hello: Wrapper<Props> = props => <div>{props.name}</div>`, while failing to validate `React.forwardRef<HTMLDivElement, Props>(...)` because its props are in the second type argument. Match upstream's import-aware React generic whitelist and per-generic argument index.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,933 @@
...
+		typeAliases := map[string]*ast.Node{}
+		var collectTypeAliases ast.Visitor
+		collectTypeAliases = func(n *ast.Node) bool {
+			if n == nil {
+				return false
+			}
+			switch n.Kind {
+			case ast.KindTypeAliasDeclaration:
+				if name := n.AsTypeAliasDeclaration().Name(); name != nil && name.Kind == ast.KindIdentifier {
+					typeAliases[name.AsIdentifier().Text] = n
```
elecmonkey: [P2] Resolve TypeScript aliases in lexical scope

All aliases and interfaces are stored in one file-wide name map, so an inner `interface Props` overwrites an outer declaration with the same name. A component outside that block then resolves against the wrong type; the differential case with outer `{ outer: string }` and inner `{ inner: string }` reports `outer` only in Rslint. Alias lookup needs binding or lexical-scope identity rather than the last declaration textually visited.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,933 @@
...
+								}
+							}
+						}
+					}
+				}
+			case ast.KindBinaryExpression:
+				b := n.AsBinaryExpression()
+				if b.OperatorToken != nil && b.OperatorToken.Kind == ast.KindEqualsToken && b.Left != nil && b.Left.Kind == ast.KindIdentifier {
+					if c := componentFor(n, comps); c != nil && (b.Left.AsIdentifier().Text == c.props || isClassPropsParameter(b.Left, c)) {
+						c.ignore = true
```
elecmonkey: [P1] Do not disable validation after assigning to props

For `function Hello(props) { props = {}; return <div>{props.name}</div>; }`, upstream still reports `name`, but this assignment sets `c.ignore` and suppresses every diagnostic for the component. The same happens when an inner function assigns to its own shadowing `props` parameter, hiding unrelated outer-component violations; assignment-target filtering should exclude the write itself without disabling component-wide validation.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,933 @@
...
+				}
+				if b.OperatorToken != nil && b.OperatorToken.Kind == ast.KindEqualsToken && b.Left != nil && b.Left.Kind == ast.KindPropertyAccessExpression {
+					root, names, ok := memberNames(b.Left)
+					if ok && root != nil && root.Kind == ast.KindIdentifier {
+						propTypesIndex := slices.Index(names, "propTypes")
+						if propTypesIndex < 0 {
+							break
+						}
+						componentParts := append([]string{root.AsIdentifier().Text}, names[:propTypesIndex]...)
+						if c := byName[strings.Join(componentParts, ".")]; c != nil {
```
elecmonkey: [P2] Resolve propTypes assignments by binding and declaration order

The file-wide `byName` lookup is populated only after a component is visited and is overwritten by same-named components in nested scopes. As a result, a hoisted declaration such as `Hello.propTypes = {...}; function Hello(props) {...}` is ignored, and two scoped `Hello` components can both receive the wrong declarations. Upstream resolves these assignments to the corresponding component binding.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,933 @@
...
+				}
+			case ast.KindPropertyAccessExpression, ast.KindElementAccessExpression:
+				if isAssignmentTarget(n) {
+					break
+				}
+				root, names, ok := memberNames(n)
+				if !ok || root == nil {
+					break
+				}
+				c := componentFor(n, comps)
```
elecmonkey: [P2] Attribute captured props through nested components

Always returning the nearest component loses outer props captured by a nested component. In `function Outer(props) { function Inner(innerProps) { return <span>{props.outer}</span>; } Inner.propTypes = {}; return <Inner />; }`, upstream reports `outer` against `Outer`, while Rslint emits nothing because it tries only `Inner` and `props` does not match that component's parameter.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,933 @@
...
+						propTypesIndex := slices.Index(names, "propTypes")
+						if propTypesIndex < 0 {
+							break
+						}
+						componentParts := append([]string{root.AsIdentifier().Text}, names[:propTypesIndex]...)
+						if c := byName[strings.Join(componentParts, ".")]; c != nil {
+							propNames := names[propTypesIndex+1:]
+							if len(propNames) == 0 {
+								if m, ok := declared(b.Right, o.customValidators, propWrappers); ok {
+									c.declared = m
```
elecmonkey: [P2] Merge repeated complete propTypes declarations

Each complete assignment replaces `c.declared` instead of merging it. With separate `Hello.propTypes = { first: ... }` and `Hello.propTypes = { last: ... }` assignments, upstream considers both props declared, while Rslint reports `first` after the second assignment discards it.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,933 @@
...
+		return nil, false
+	}
+	if n.Kind == ast.KindObjectLiteralExpression {
+		out := map[string]propType{}
+		for _, p := range n.AsObjectLiteralExpression().Properties.Nodes {
+			if p == nil {
+				continue
+			}
+			if p.Kind == ast.KindSpreadAssignment {
+				out["__ANY_KEY__"] = propType{any: true}
```
elecmonkey: [P2] Do not make nested shape spreads accept every child

Every spread assignment becomes `__ANY_KEY__`, including spreads inside `PropTypes.shape`. For `user: PropTypes.shape({ ...fields, name: PropTypes.string })` followed by `props.user.missing`, upstream reports `user.missing`, but Rslint treats the spread as a wildcard and emits nothing. The upstream whole-component escape for top-level composed propTypes must not be reused as nested shape semantics.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,933 @@
...
+				continue
+			}
+			var name, value *ast.Node
+			switch p.Kind {
+			case ast.KindPropertyAssignment:
+				pa := p.AsPropertyAssignment()
+				name, value = pa.Name(), pa.Initializer
+			case ast.KindShorthandPropertyAssignment:
+				name, value = p.AsShorthandPropertyAssignment().Name(), p.AsShorthandPropertyAssignment().Name()
+			default:
```
elecmonkey, line 133: [P2] Recognize method-style prop validators

`Hello.propTypes = { name() {} }` is accepted by upstream as a declaration of `name`, but `propMap` skips every object member except property and shorthand assignments, so Rslint reports `name` as missing. Method and accessor property forms need to register their static keys as validators.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,933 @@
...
+		}
+		ctx.SourceFile.Node.ForEachChild(walk)
+		for _, c := range comps {
+			if c.ignore || len(c.used) == 0 || !c.declaredBlock && o.skipUndeclared {
+				continue
+			}
+			seen := map[string]bool{}
+			for _, u := range c.used {
+				name := u.names[0]
+				if isObjectPrototypeProperty(name) || slices.Contains(o.ignore, name) || seen[reportName(u.names)] {
```
elecmonkey: [P2] Emit one diagnostic for every undeclared use

The `seen` map deduplicates by property path across the entire component. For two separate `props.name` reads, eslint-plugin-react@7.37.5 emits two diagnostics at the two occurrences, while Rslint emits only one. This changes diagnostic counts and leaves later occurrences unannotated; deduplication should follow upstream's node-level used-prop entries.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,933 @@
...
+					continue
+				}
+				seen[reportName(u.names)] = true
+				if !isDeclared(c.declared, u.names) {
+					ctx.ReportNode(u.node, rule.RuleMessage{Id: "missingPropType", Description: fmt.Sprintf("'%s' is missing in props validation", reportName(u.names)), Data: map[string]string{"name": reportName(u.names)}})
```
elecmonkey, line 1625: [P2] Report the terminal property instead of the full chain

Member uses store and report the complete access expression. For `props.name`, upstream reports columns covering only `name`, whereas Rslint covers `props.name`; for `props.user.name`, Rslint covers the whole chain instead of the terminal property. This affects editor highlighting and produces 138 location mismatches in the differential upstream corpus.

internal/plugins/react/rules/prop_types/prop_types.go
elecmonkey: [P2] Exempt Object prototype members at nested path segments

The exemption checks only `u.names[0]`. With `user: PropTypes.shape({})` and `props.user.toString()`, upstream emits no diagnostic because `toString` is an Object prototype member, while Rslint reports `user.toString`. Apply the prototype-member check to the accessed segment that generated each used-prop entry.

internal/plugins/react/all.go
```diff
@@ -143,6 +144,7 @@ func GetAllRules() []rule.Rule {
 		no_will_update_set_state.NoWillUpdateSetStateRule,
 		prefer_es6_class.PreferEs6ClassRule,
 		prefer_stateless_function.PreferStatelessFunctionRule,
+		prop_types.PropTypesRule,
```
elecmonkey, line 149: [P2] Enable the ported rule in the React recommended preset

The rule is added to the native React catalog here, but `packages/rslint/src/config/presets/react.ts` still comments out `react/prop-types` as ‘not implemented’. eslint-plugin-react@7.37.5 enables this rule in its recommended preset, so users of `reactPlugin.configs.recommended` will not receive the newly ported behavior.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,933 @@
...
+
+func elementName(n *ast.Node) string {
+	n = unwrap(n)
+	if n == nil {
+		return ""
+	}
+	switch n.Kind {
+	case ast.KindStringLiteral,
+		ast.KindNumericLiteral,
+		ast.KindBigIntLiteral,
```
elecmonkey: [P3] Keep BigInt element accesses computed

Rslint treats a BigInt literal as a static property name, so `props[1n]` reports `'1' is missing`; eslint-plugin-react@7.37.5 treats this as an unanalyzable computed access and emits no diagnostic. Restrict statically recognized element keys to the string and number forms accepted upstream.

internal/plugins/react/rules/prop_types/prop_types.go
swwind: Fixed in d6cba77af with regression coverage.

internal/plugins/react/rules/prop_types/prop_types.go
swwind: Fixed in d6cba77af with regression coverage.

internal/plugins/react/rules/prop_types/prop_types.go
swwind: Fixed in d6cba77af with regression coverage.

internal/plugins/react/rules/prop_types/prop_types.go
swwind: Fixed in d6cba77af with regression coverage.

internal/plugins/react/rules/prop_types/prop_types.go
swwind: Fixed in d6cba77af with regression coverage.

internal/plugins/react/rules/prop_types/prop_types.go
swwind: Fixed in d6cba77af with regression coverage.

internal/plugins/react/rules/prop_types/prop_types.go
swwind: Fixed in d6cba77af with regression coverage.

internal/plugins/react/rules/prop_types/prop_types.go
swwind: Fixed in d6cba77af with regression coverage.

internal/plugins/react/rules/prop_types/prop_types.go
swwind: Fixed in d6cba77af with regression coverage.

internal/plugins/react/rules/prop_types/prop_types.go
swwind: Fixed in d6cba77af with regression coverage.

internal/plugins/react/rules/prop_types/prop_types.go
swwind: Fixed in d6cba77af with regression coverage.

internal/plugins/react/rules/prop_types/prop_types.go
swwind: Fixed in d6cba77af with regression coverage.

internal/plugins/react/rules/prop_types/prop_types.go
swwind: Fixed in d6cba77af with regression coverage.

internal/plugins/react/rules/prop_types/prop_types.go
swwind: Fixed in d6cba77af with regression coverage.

internal/plugins/react/rules/prop_types/prop_types.go
swwind: Fixed in d6cba77af with regression coverage.

internal/plugins/react/rules/prop_types/prop_types.go
swwind: Fixed in d6cba77af with regression coverage.

internal/plugins/react/all.go
swwind: Fixed in d6cba77af with regression coverage.

internal/plugins/react/rules/prop_types/prop_types.go
swwind: Fixed in d6cba77af with regression coverage.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,1193 @@
...
+		}
+	}
+	// Calls whose structure this port does not model (instanceOf, oneOf, or a
+	// dynamic shape argument) are broad validators upstream.
+	return propType{open: true}
+}
+
+func declared(n *ast.Node, customValidators []string, propWrappers []reactutil.PropWrapperEntry) (map[string]propType, bool) {
+	n = unwrap(n)
+	if n != nil && n.Kind == ast.KindIdentifier {
```
elecmonkey, line 231: [P1] Guard identifier declaration resolution against cycles

Resolving a top-level `propTypes` identifier recursively has no visited-set or depth guard. With `var first = second; var second = first; function Hello(props) { return <div>{props.name}</div>; } Hello.propTypes = first;`, eslint-plugin-react@7.37.5 returns one normal `missingPropType` diagnostic, while Rslint terminates with a Go stack overflow in `declared`.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,1193 @@
...
+			continue
+		}
+		key := keyName(be.PropertyName)
+		if key == "" {
+			key = keyName(be.Name())
+		}
+		if key == "" || be.Name() == nil {
+			continue
+		}
+		path := append(append([]string{}, prefix...), key)
```
elecmonkey, line 933: [P1] Record destructuring patterns as prop uses

`addDestructured` only records aliases and waits for a later identifier read, but upstream considers the destructuring itself a prop use. Consequently `function Hello({ name }) { return <div>Hello</div>; }` and body destructuring such as `const { name } = props` produce one upstream diagnostic and none in Rslint; nested patterns also lose the parent diagnostic (`user`) that upstream emits.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,1193 @@
...
+					if c.binding != nil {
+						byBinding[c.binding] = c
+					}
+					if name := componentName(n); name != "" {
+						byName[name] = c
+					}
+					ps := n.Parameters()
+					if len(ps) > 0 && ps[0].Kind == ast.KindParameter {
+						parameter := ps[0].AsParameterDeclaration()
+						name := parameter.Name()
```
elecmonkey, line 1441: [P2] Restrict function member tracking to upstream prop bindings

Saving every first parameter name as `c.props` broadens the rule beyond upstream, whose ordinary function-component member tracking recognizes `props` specifically. For `function Hello(input) { return <div>{input.name}</div>; }`, eslint-plugin-react@7.37.5 reports nothing while Rslint reports `'name' is missing`, creating false positives for components that use another parameter name.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,1193 @@
...
+		case ast.KindPropertyAccessExpression:
+			pa := cur.AsPropertyAccessExpression()
+			k := keyName(pa.Name())
+			if k == "" {
+				return nil, nil, false
+			}
+			names = append([]string{k}, names...)
+			report = pa.Name()
+			cur = unwrap(pa.Expression)
+		case ast.KindElementAccessExpression:
```
elecmonkey, line 1159: [P1] Preserve dynamic computed segments in member paths

Returning `false` for a nonliteral element access drops the entire outer chain. For `props.users[i].name` with `users: PropTypes.arrayOf(PropTypes.shape({}))`, upstream represents the index as `__COMPUTED_PROP__` and reports `'users[].name'`, while Rslint emits nothing; nested numeric indices are also rendered as `.0` instead of `[]` because they are always treated as static.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,1193 @@
...
+								symbol = ctx.TypeChecker.GetSymbolAtLocation(root)
+							}
+							if bySymbol := byBinding[symbol]; bySymbol != nil {
+								c = bySymbol
+							}
+						}
+						if c != nil {
+							propNames := names[propTypesIndex+1:]
+							if len(propNames) == 0 {
+								if m, ok := declared(b.Right, o.customValidators, propWrappers); ok {
```
elecmonkey: [P1] Do not accept every opaque propTypes expression

When `declared` cannot parse the right-hand side, this fallback marks every property as valid. For `Hello.propTypes = makePropTypes()` followed by `props.name`, upstream reports `'name' is missing` because an unconfigured call is not an opaque external declaration, while Rslint reports nothing. The wildcard should be limited to upstream's actual ignore cases, such as unresolved external identifiers.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,1193 @@
...
+		return nil
+	}
+	return nil
+}
+
+func reactComponentTypeArgument(typeNode *ast.Node) *ast.Node {
+	if typeNode == nil || typeNode.Kind != ast.KindTypeReference {
+		return nil
+	}
+	name := reactutil.EntityNameRightmost(typeNode.AsTypeReferenceNode().TypeName)
```
elecmonkey, line 641: [P1] Resolve React generic annotations through imports

The annotation check uses only the rightmost spelling, so unrelated types are accepted while imported aliases and supported names are missed. `const Hello: Fake.FC<Props> = ...` suppresses an upstream diagnostic in Rslint, whereas `import { FC as Component } from 'react'; const Hello: Component<Props> = ...` produces a Rslint false positive despite being valid upstream; `VFC` and `VoidFunctionComponent` are omitted as well.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,1193 @@
...
+
+func forwardRefPropsType(n *ast.Node) *ast.Node {
+	for current := n; current != nil && current.Parent != nil; current = current.Parent {
+		parent := current.Parent
+		if parent.Kind == ast.KindCallExpression {
+			call := parent.AsCallExpression()
+			if call.Arguments == nil || len(call.Arguments.Nodes) == 0 || unwrap(call.Arguments.Nodes[0]) != unwrap(current) {
+				return nil
+			}
+			callee := unwrap(call.Expression)
```
elecmonkey, line 763: [P1] Recognize bare imported forwardRef calls

`forwardRefPropsType` accepts only property-access callees, even though component detection accepts a bare named import. For `import { forwardRef } from 'react'; type Props = { name: string }; const Hello = forwardRef<HTMLDivElement, Props>((props, ref) => <div ref={ref}>{props.name}</div>);`, upstream reports nothing but Rslint reports `'name' is missing` because the second type argument is discarded.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,1193 @@
...
+						mergeDeclared(declared, base)
+					}
+				}
+			}
+		}
+	}
+	return declared, true
+}
+
+func propTypeFromType(n *ast.Node, aliases map[string][]*ast.Node, aliasesBySymbol map[*ast.Symbol]*ast.Node, depth int) propType {
```
elecmonkey: [P1] Match upstream's shallow TypeScript property validation

Recursively converting each TypeScript property type makes this port stricter than eslint-plugin-react@7.37.5. For `type Props = { user: { name: string } }; function Hello(props: Props) { return <div>{props.user.missing}</div>; }`, upstream treats the declared `user` property as accepting deeper reads and emits nothing, while Rslint reports `'user.missing'`; this affects ordinary nested object, array, and union annotations.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,1193 @@
...
+				return declaredType(declaration, aliases, aliasesBySymbol, depth-1)
+			}
+		}
+		if declaration := visibleTypeAlias(n, aliases[n.AsIdentifier().Text]); declaration != nil {
+			return declaredType(declaration, aliases, aliasesBySymbol, depth-1)
+		}
+		return map[string]propType{"__ANY_KEY__": {any: true}}, true
+	case ast.KindParenthesizedType:
+		return declaredType(n.AsParenthesizedTypeNode().Type, aliases, aliasesBySymbol, depth-1)
+	case ast.KindIntersectionType:
```
elecmonkey, line 428: [P2] Treat unsupported top-level TS annotations as opaque

Unsupported annotation kinds return `(nil, false)`, which is interpreted as no declaration. A component parameter typed `A | B` or a mapped type is ignored by upstream when it cannot construct prop declarations, but Rslint reports its property accesses as missing. These branches should preserve upstream's ignore/opaque state rather than fall through to undeclared props.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,1193 @@
...
+				if name := n.AsTypeAliasDeclaration().Name(); name != nil && name.Kind == ast.KindIdentifier {
+					typeAliases[name.AsIdentifier().Text] = append(typeAliases[name.AsIdentifier().Text], n)
+					if symbol := name.Symbol(); symbol != nil {
+						typeAliasesBySymbol[symbol] = n
+					}
+					if symbol := n.Symbol(); symbol != nil {
+						typeAliasesBySymbol[symbol] = n
+					}
+				}
+			case ast.KindInterfaceDeclaration:
```
elecmonkey, line 1316: [P2] Merge declarations for the same TypeScript interface

Declaration merging is lost because every interface symbol is mapped to a single node and later declarations overwrite earlier ones. With `interface Props { first: string } interface Props { last: string }`, upstream accepts reads of both properties, while Rslint resolves only the final declaration and reports `last` or `first` depending on which declaration was retained.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,1193 @@
...
+				}
+			case ast.KindVariableDeclaration:
+				vd := n.AsVariableDeclaration()
+				if vd.Name() != nil {
+					if root, names, ok := memberNames(vd.Initializer); ok && root != nil {
+						if c := componentFor(n, comps); c != nil {
+							if root.Kind == ast.KindThisKeyword && len(names) == 0 && vd.Name().Kind == ast.KindObjectBindingPattern {
+								addThisPropsDestructured(c, vd.Name())
+							}
+							if path, ok := propsPath(root, names, c); ok {
```
elecmonkey, line 1526: [P2] Keep plain props aliases out of used-prop tracking

Recording `const copy = props` as a tracked alias changes an explicit upstream behavior: eslint-plugin-react treats accesses through such reassigned props objects as unanalyzable and skips validation. For `function Hello(props) { const copy = props; return <div>{copy.name}</div>; }`, upstream emits nothing while Rslint reports `'name' is missing`.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,1193 @@
...
+	var report *ast.Node
+	for cur != nil {
+		switch cur.Kind {
+		case ast.KindPropertyAccessExpression:
+			pa := cur.AsPropertyAccessExpression()
+			k := keyName(pa.Name())
+			if k == "" {
+				return nil, nil, false
+			}
+			names = append([]string{k}, names...)
```
elecmonkey, line 1154: [P2] Report the terminal property range

Although `memberNames` now stores a property node, each iteration overwrites `report`, leaving the property closest to the root. On `props.name`, upstream reports columns 42–46 for `name`, while Rslint reports from column 36 and highlights `props.name`; nested chains similarly anchor to the first segment instead of the terminal property.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,1193 @@
...
+							break
+						}
+					}
+					if contains && rootMatches(root, c) {
+						if path, ok := propsPath(root, names, c); ok {
+							appendUse(c, n, path)
+						}
+					}
+				}
+			case ast.KindIdentifier:
```
elecmonkey: [P2] Attribute captured destructured props to outer components

Identifier uses are assigned only to `componentFor`, the nearest component, unlike member-expression uses which search all enclosing components. Thus `function Outer({ outer }) { function Inner() { return <span>{outer}</span>; } return <Inner />; }` produces one upstream diagnostic for `outer` but none in Rslint because the identifier is looked up only in `Inner`'s destructuring map.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,1193 @@
...
+				name, value = pa.Name(), pa.Initializer
+			case ast.KindShorthandPropertyAssignment:
+				name, value = p.AsShorthandPropertyAssignment().Name(), p.AsShorthandPropertyAssignment().Name()
+			case ast.KindMethodDeclaration, ast.KindGetAccessor, ast.KindSetAccessor:
+				name = p.Name()
+				value = p
+			default:
+				continue
+			}
+			k := keyName(name)
```
elecmonkey, line 136: [P2] Do not wildcard computed propTypes keys

A property whose key cannot be statically resolved inserts `__ANY_KEY__`, causing every used prop to pass validation. For `const key = 'other'; Hello.propTypes = { [key]: PropTypes.string }` with `props.name`, eslint-plugin-react@7.37.5 reports `name`, while Rslint emits nothing; upstream treats the key as a declaration named `key` rather than as permission for all keys.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,1193 @@
...
+					}
+					break
+				}
+			}
+		}
+	}
+	return best
+}
+
+func declaredType(n *ast.Node, aliases map[string][]*ast.Node, aliasesBySymbol map[*ast.Symbol]*ast.Node, depth int) (map[string]propType, bool) {
```
elecmonkey: [P2] Replace the fixed type-alias recursion limit

The fixed depth guard turns otherwise resolvable annotations into absent declarations. A six-link alias chain ending in `{ name: string }` is accepted by eslint-plugin-react@7.37.5, but Rslint exhausts the depth budget and reports `name`; cycle detection should use visited declarations so valid acyclic aliases are not rejected based on an arbitrary nesting count.

internal/plugins/react/rules/prop_types/prop_types.go
swwind: Fixed in ec433704e with regression coverage.

internal/plugins/react/rules/prop_types/prop_types.go
swwind: Fixed in ec433704e with regression coverage.

internal/plugins/react/rules/prop_types/prop_types.go
swwind: Fixed in ec433704e with regression coverage.

internal/plugins/react/rules/prop_types/prop_types.go
swwind: Fixed in ec433704e with regression coverage.

internal/plugins/react/rules/prop_types/prop_types.go
swwind: Fixed in ec433704e with regression coverage.

internal/plugins/react/rules/prop_types/prop_types.go
swwind: Fixed in ec433704e with regression coverage.

internal/plugins/react/rules/prop_types/prop_types.go
swwind: Fixed in ec433704e with regression coverage.

internal/plugins/react/rules/prop_types/prop_types.go
swwind: Fixed in ec433704e with regression coverage.

internal/plugins/react/rules/prop_types/prop_types.go
swwind: Fixed in ec433704e with regression coverage.

internal/plugins/react/rules/prop_types/prop_types.go
swwind: Fixed in ec433704e with regression coverage.

internal/plugins/react/rules/prop_types/prop_types.go
swwind: Fixed in ec433704e with regression coverage.

internal/plugins/react/rules/prop_types/prop_types.go
swwind: Fixed in ec433704e with regression coverage.

internal/plugins/react/rules/prop_types/prop_types.go
swwind: Fixed in ec433704e with regression coverage.

internal/plugins/react/rules/prop_types/prop_types.go
swwind: Fixed in ec433704e with regression coverage.

internal/plugins/react/rules/prop_types/prop_types.go
swwind: Fixed in ec433704e with regression coverage.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,1490 @@
...
+		if name != nil {
+			// TypeScript declaration merging exposes one symbol for every interface
+			// declaration. Preserve all same-name interface members, as upstream does.
+			var interfaces []*ast.Node
+			for _, candidate := range aliases[name.AsIdentifier().Text] {
+				if candidate != nil && candidate.Kind == ast.KindInterfaceDeclaration {
+					interfaces = append(interfaces, candidate)
+				}
+			}
+			if len(interfaces) > 1 {
```
elecmonkey, line 392: [P1] Do not merge same-name interfaces across lexical scopes

This merges every interface named `Props` in the file before checking symbol identity or lexical visibility. With `interface Props { outer: string } { interface Props { inner: string } function Inner(props: Props) { return <div>{props.inner}</div>; } } function Outer(props: Props) { return <div>{props.inner}</div>; }`, upstream reports missing `inner` for `Outer`, but Rslint treats the inner block interface as merged into the outer `Props` and reports nothing.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,1490 @@
...
+	}
+	if typeName == "" {
+		qualified := typeNode.AsTypeReferenceNode().TypeName
+		if qualified == nil || qualified.Kind != ast.KindQualifiedName || qualified.AsQualifiedName().Left == nil || qualified.AsQualifiedName().Left.Kind != ast.KindIdentifier {
+			return nil
+		}
+		qualifier := qualified.AsQualifiedName().Left.AsIdentifier().Text
+		if qualifier != "React" {
+			root := ast.GetSourceFileOfNode(name)
+			if root == nil || !strings.Contains(root.Text(), "* as "+qualifier) || (!strings.Contains(root.Text(), `"react"`) && !strings.Contains(root.Text(), `'react'`)) {
```
elecmonkey: [P1] Resolve React namespace imports by binding instead of source text

This source-text check can combine unrelated imports: `import * as R from 'not-react'; import React from 'react'; type Props = { name: string }; const Hello: R.FC<Props> = props => <div>{props.name}</div>;` is treated as a React generic because the file contains both `* as R` and a separate React import. `eslint-plugin-react@7.37.5` reports `'name' is missing in props validation`, while Rslint suppresses it, so non-React namespace types can create false negatives.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,1490 @@
...
+	}
+	body := unwrap(initializer.Body())
+	if body == nil || body.Kind != ast.KindObjectLiteralExpression {
+		return nil, false
+	}
+	return propMap(body, nil)
+}
+
+func declaredType(n *ast.Node, aliases map[string][]*ast.Node, aliasesBySymbol map[*ast.Symbol]*ast.Node, depth int) (map[string]propType, bool) {
+	if n == nil || depth == 0 {
```
elecmonkey: [P2] Replace the fixed alias-depth budget with cycle detection

This still rejects valid acyclic aliases once the chain is longer than the fixed depth. A long `type A00 = A01; ...; type A65 = { name: string }; function Hello(props: A00) { return <div>{props.name}</div>; }` chain is accepted by `eslint-plugin-react@7.37.5`, but Rslint exhausts the budget and reports `name` as missing; a visited-declaration set would stop real cycles without making valid code depend on an arbitrary nesting limit.

swwind: Fixed in `033577298`. Type alias resolution now uses path-local cycle detection instead of the fixed depth-64 budget. Added regression coverage for a 65-link acyclic alias chain.

internal/plugins/react/rules/prop_types/prop_types.go
swwind: Fixed in `033577298`. Type declaration resolution now selects the nearest lexical scope visible at the reference and merges same-name interfaces only within that scope. Added regression coverage for same-name interfaces across nested scopes.

internal/plugins/react/rules/prop_types/prop_types.go
swwind: Fixed in `033577298`. Qualified React component types now require the qualifier to be an actual namespace import from `react`, so unrelated namespace imports are no longer accepted. Added regression coverage.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,1553 @@
...
+	// named import.  A project-local `FC<Props>` must not suppress this rule.
+	if typeName == "" && typeNode.AsTypeReferenceNode().TypeName.Kind == ast.KindIdentifier {
+		return nil
+	}
+	if typeName == "" {
+		qualified := typeNode.AsTypeReferenceNode().TypeName
+		if qualified == nil || qualified.Kind != ast.KindQualifiedName || qualified.AsQualifiedName().Left == nil || qualified.AsQualifiedName().Left.Kind != ast.KindIdentifier {
+			return nil
+		}
+		qualifier := qualified.AsQualifiedName().Left.AsIdentifier().Text
```
elecmonkey, line 660: [P1] Resolve qualified React types through import bindings

When the annotation is qualified, this accepts the literal `React` spelling without checking an import and only recognizes aliases from namespace imports. For `type Props = { name: string }; const Hello: React.FC<Props> = props => <div>{props.name}</div>`, `eslint-plugin-react@7.37.5` reports `'name' is missing in props validation`, while Rslint reports nothing; conversely, `import R from 'react'; const Hello: R.FC<Props> = ...` is valid upstream but Rslint reports `name`. Please resolve both default and namespace qualifiers through the actual `react` import binding.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,1553 @@
...
+			if symbol := name.Symbol(); symbol != nil {
+				if declaration := aliasesBySymbol[symbol]; declaration != nil {
+					return declaredType(declaration, aliases, aliasesBySymbol, seen)
+				}
+			}
+			if declaration := visibleTypeAlias(n, aliases[name.AsIdentifier().Text]); declaration != nil {
+				return declaredType(declaration, aliases, aliasesBySymbol, seen)
+			}
+		}
+		// Imported or otherwise unresolved annotations are opaque upstream and
```
elecmonkey, line 413: [P1] Unwrap supported React utility types

Known React wrappers used directly on component parameters fall through to the opaque-reference path. For `function Hello(props: React.PropsWithChildren<{ name: string }>) { return <div>{props.missing}</div>; }`, `eslint-plugin-react@7.37.5` reports `missing`, while Rslint treats the annotation as accepting every property. Please traverse the props type argument of upstream-supported wrappers such as `PropsWithChildren` before using the unresolved-type fallback.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,1553 @@
...
+		qualifier := qualified.AsQualifiedName().Left.AsIdentifier().Text
+		if qualifier != "React" && !importedReactNamespace(name, qualifier) {
+			return nil
+		}
+		typeName = name.AsIdentifier().Text
+	}
+	if typeName == "ForwardRefRenderFunction" && len(arguments.Nodes) >= 2 {
+		return arguments.Nodes[1]
+	}
+	if !slices.Contains([]string{"FC", "FunctionComponent", "SFC", "StatelessComponent", "ComponentType", "VFC", "VoidFunctionComponent"}, typeName) {
```
elecmonkey: [P2] Remove unsupported component annotation wrappers

The sibling-annotation whitelist includes `ComponentType`, which upstream does not use to declare function props. `const Hello: React.ComponentType<{ name: string }> = props => <div>{props.name}</div>` therefore reports `name` in `eslint-plugin-react@7.37.5` but nothing in Rslint. Please match the upstream wrapper set rather than accepting additional generic names.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,1553 @@
...
+		clauses = class.AsClassDeclaration().HeritageClauses
+	case ast.KindClassExpression:
+		clauses = class.AsClassExpression().HeritageClauses
+	default:
+		return nil
+	}
+	if clauses == nil {
+		return nil
+	}
+	for _, clause := range clauses.Nodes {
```
elecmonkey, line 539: [P1] Read class props only from the React extends clause

`classPropsType` returns the first generic heritage entry without checking whether it belongs to `extends` or `implements`. With `class Hello extends React.Component implements Other<Props>`, `eslint-plugin-react@7.37.5` reports the undeclared `this.props.name`, but Rslint adopts `Props` from `implements Other<Props>` and suppresses it. Please restrict extraction to the React component superclass and its actual props argument.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,1553 @@
...
+
+func forwardRefPropsType(n *ast.Node) *ast.Node {
+	for current := n; current != nil && current.Parent != nil; current = current.Parent {
+		parent := current.Parent
+		if parent.Kind == ast.KindCallExpression {
+			call := parent.AsCallExpression()
+			if call.Arguments == nil || len(call.Arguments.Nodes) == 0 || unwrap(call.Arguments.Nodes[0]) != unwrap(current) {
+				return nil
+			}
+			callee := unwrap(call.Expression)
```
elecmonkey, line 763: [P2] Restrict forwardRef generic extraction to React

Every property-access callee ending in `forwardRef` is treated as React. Under `settings.react.pragma = "Foo"`, `Foo.forwardRef<HTMLDivElement, Props>(...)` is reported by `eslint-plugin-react@7.37.5` when `props.name` lacks validation, but Rslint consumes `Props` and emits nothing. Please verify that the callee is an upstream-supported React form rather than matching the terminal property name alone.

internal/plugins/react/rules/prop_types/prop_types.go
```diff
@@ -0,0 +1,1553 @@
...
+			binary := assignment.AsBinaryExpression()
+			root, names, ok := memberNames(binary.Left)
+			if !ok || root == nil || root.Kind != ast.KindIdentifier {
+				return
+			}
+			propTypesIndex := slices.Index(names, "propTypes")
+			if propTypesIndex < 0 {
+				return
+			}
+			componentParts := append([]string{root.AsIdentifier().Text}, names[:propTypesIndex]...)
```
elecmonkey, line 1365: [P2] Key namespaced components by lexical binding

Namespaced components use a file-wide textual key such as `A.X`, so a block-scoped component with the same path overwrites the outer component. With an outer `A.X` validated by `{ outer: ... }` and an inner shadowed `A.X` validated only by `{ inner: ... }`, upstream reports the inner `props.outer`, while Rslint reports nothing. Please include the root binding or lexical scope in this lookup.

internal/plugins/react/rules/prop_types/prop_types_upstream_test.go
```diff
@@ -0,0 +1,546 @@
...
+			{Code: "\n        type Props = { foo: string }\n        function higherOrderComponent<Props>() {\n          return class extends React.Component<Props> {\n            render() {\n              return <div>{this.props.foo}</div>\n            }\n          }\n        }\n      ", Tsx: true, Skip: true},
+			{Code: "\n        function higherOrderComponent<P: { foo: string }>() {\n          return class extends React.Component<P> {\n            render() {\n              return <div>{this.props.foo}</div>\n            }\n          }\n        }\n      ", Tsx: true, Skip: true},
+			{Code: "\n        const withOverlayState = <P: { foo: string }>(WrappedComponent: ComponentType<P>): ComponentType<P> => (\n          class extends React.Component<P> {\n            constructor(props) {\n              super(props);\n              this.state = { foo: props.foo };\n            }\n            render() {\n              return <div>Hello World</div>;\n            }\n          }\n        )\n      ", Tsx: true, Skip: true},
+			{Tsx: true},
```
elecmonkey: [P2] Restore the omitted upstream valid cases

The claimed exhaustive migration replaces three upstream sources with empty `{Tsx: true}` cases: the two issue-1288 `props.bar = ...; <div {...props} />` inputs and the lowercase JSX-returning `noAComponent` input. Empty source files cannot verify those behaviors, and no adjacent explanation maps them to another case. Please preserve the actual sources or document an explicit equivalent mapping.

internal/plugins/react/rules/prop_types/prop_types_upstream_test.go
```diff
@@ -0,0 +1,546 @@
...
+			{Code: "\n        import React from \"react\"\n\n        export function Heading({ className, ...props }: React.HTMLAttributes<HTMLDivElement>) {\n          return <div className={cn(\"font-semibold text-lg\", className)} {...props} />\n        }\n      ", Tsx: true},
+			{Code: "\n        import React from 'react';\n        type TDelIconProps = React.HTMLProps<HTMLDivElement>;\n\n        const DelIcon: React.FC<TDelIconProps> = ({ className, ...rest }) => (\n            <div className={classNames('del flex-center', className)} {...rest}>\n                <i className=\"iconfont icon-del f12\" />\n            </div>\n        );\n      ", Tsx: true},
+		}, []rule_tester.InvalidTestCase{
+			{Code: "\n        type Props = {\n          name: string,\n        };\n        class Hello extends React.Component {\n          foo(props: Props) {}\n          render() {\n            return this.props.name;\n          }\n        }\n      ", Tsx: true, Errors: []rule_tester.InvalidTestCaseError{{MessageId: "missingPropType", Line: 8, Column: 31}}},
```
elecmonkey, line 341: [P2] Preserve upstream diagnostic message assertions

Upstream supplies a concrete `data.name` for 204 diagnostics, but this migration asserts `Message` only 23 times; an empty `Message` makes `RunRuleTester` skip text validation. The first invalid cases already discard the expected `'name' is missing in props validation` value. Since the dynamic property path is core rule behavior, please migrate these message assertions instead of checking only `messageId` and count.

internal/plugins/react/rules/prop_types/prop_types.md
```diff
@@ -0,0 +1,28 @@
...
+```
+
+The `ignore` option excludes named props. `skipUndeclared` limits checking to
+components with a `propTypes` declaration. `customValidators` is accepted for
+configuration compatibility; custom validator calls are treated as declared
+prop types.
+
+## Differences from ESLint
+
+TypeScript and Flow type declarations are not yet used as prop declarations by
```
elecmonkey: [P3] Correct the documented TypeScript behavior

This page states that TypeScript declarations are not used, but the implementation resolves parameter annotations, class generics, interfaces, aliases, intersections, `ReturnType`, and React component generics. This gives users the opposite expectation from the shipped behavior. Please describe the supported TypeScript forms and any actual remaining limits.

elecmonkey:
i beg you don't use Terra ()

swwind:
> i beg you don't use Terra ()

You are absolutely right! --- reply from luna.

fansenze (approved):

PR #1972, opened 2026-08-31
feat(rstest): add prefer-to-be-truthy, prefer-to-be-falsy and prefer-strict-boolean-matchers rules

elecmonkey:
## Summary

Three rules about how a boolean assertion is spelled. `prefer-to-be-truthy` and `prefer-to-be-falsy` report an equality assertion made against a boolean literal — `expect(value).toBe(true)` and its `toEqual` / `toStrictEqual` spellings — and rewrite it to `toBeTruthy()`. `prefer-strict-boolean-matchers` is their exact inverse: it reports `expect(value).toBeTruthy()` and rewrites it to `expect(value).toBe(true)`. All three are registered in the rstest plugin and left out of the `recommended` preset, matching upstream.

The three are mutually exclusive by construction — enabling `prefer-strict-boolean-matchers` next to either of the other two makes the two fixes undo each other — so they are opt-in and a project picks one direction. That is upstream's arrangement too, and it is why they ship together here: the same expect-chain analysis, the same accessor rewrite, and the same argument-list edit serve both directions, and doing them separately would mean building that twice.

## One rule with the boolean flipped, and the helpers this PR extracts

`prefer-to-be-truthy` and `prefer-to-be-falsy` differ only in which literal they match, which matcher they write, and what they say, so `MakePreferBooleanMatcherRule` in `internal/plugins/rstest/utils/` takes those three as configuration and each rule package is a fifteen-line wrapper around it — the same shape `reactutil.MakeNoMethodSetStateRule` already uses for the three `no-*-set-state` rules. `prefer-strict-boolean-matchers` carries two messages and inserts rather than removes, so it stays its own rule, but it is built from the same pieces.

Those pieces — three primitives every matcher-rewriting rule in this plugin needs — move into `internal/plugins/rstest/utils/matcher_edit.go`. `MatcherCall` walks up from the matcher accessor to the call it is the callee of, because `MemberEntry.Call` is overwritten by each enclosing call and points at the wrong parentheses for `expect(fn).toBe(true)()`. `FollowTypeAssertionChain` reads through the `as` and `<T>` wrappers, and the parentheses ESTree does not represent at all, that upstream's `getFirstMatcherArg` sees through. `MatcherArgumentListRange` returns the span between the matcher's parentheses, scanning for the opening one rather than assuming it follows the name, so `x['toBe'](true)` and a call split across lines are handled alike; a rule that empties an argument list replaces that span, and one that fills an empty list writes at its start. `prefer-called-once` and `prefer-called-times` each carried their own copy of the first two and now call the shared versions; `prefer-called-times` keeps inserting its count at the start of the argument list rather than writing over the whole span, so a comment already written between the parentheses survives its fix. Both migrated rules pass their existing suites unchanged.

## Credits

Port from [`eslint-plugin-vitest`](https://github.com/vitest-dev/eslint-plugin-vitest)'s `prefer-to-be-truthy`, `prefer-to-be-falsy` and `prefer-strict-boolean-matchers`.

- **`expectTypeOf` assertions are not covered.** All three upstream rules accept a parse of type `expect` or `expectTypeOf`. Rstest exposes no `expectTypeOf` at all, so those chains are ordinary calls rather than assertions; the upstream cases written on it are kept in the test files as skipped cases with the reason.

- **A chain where no assertion factory ran is not reported.** Upstream reports `expect.toBe(true)` and `expect.toBeTruthy()`, whose author never passed a value to assert on, and rewrites them into assertions that still assert nothing. Matching the neighbouring `prefer-called-once`, they are left alone rather than made to look right.

- **A matcher named by a computed key is not reported.** `expect(value)[matcherName](true)` names its matcher at run time, so `toBe` in that position is a variable reference; reporting it would flag whatever assertion the variable holds and the fix would rename the reference.

- **`prefer-to-be-truthy` and `prefer-to-be-falsy` empty the span between the matcher's parentheses instead of removing the argument node.** Upstream's `fixer.remove(args[0])` leaves everything written around the literal behind, so `expect(value).toBe(true,)` becomes `expect(value).toBeTruthy(,)`, a syntax error. Taking the whole span removes the trailing comma and the whitespace of a multi-line argument list with it. When a comment sits inside those parentheses the fix is withheld entirely and the diagnostic stands alone, since emptying the list would delete it.

- **`prefer-strict-boolean-matchers` does not report a truthiness matcher that was given arguments.** `toBeTruthy` and `toBeFalsy` take none, and upstream inserts the literal without looking at what is already between the parentheses, welding the two into a single run-together identifier — code that still compiles and asserts something else. The insertion point is also scanned for rather than computed as `matcher.range[1] + 1`, so a quoted accessor or a comment between the name and the parentheses does not displace the literal.

## Related Links

Related #935

## Checklist


internal/plugins/rstest/utils/prefer_boolean_matcher_rule.go
```diff
@@ -0,0 +1,149 @@
...
+
+	return rule.Rule{
+		Name:   cfg.RuleName,
+		Schema: rule.EmptyArraySchema,
+		Run: func(ctx rule.RuleContext, options []any) rule.RuleListeners {
+			analysis := GetRstestCallAnalysis(ctx)
+			return rule.RuleListeners{
+				ast.KindCallExpression: func(node *ast.Node) {
+					matcher, matcherCall, ok := booleanEqualityMatcherEntry(
+						analysis.ParseExpectCall(node),
```
swwind, line 55: Could we follow `const` aliases of a recognized Rstest `expect` here? For example:

```ts
import { expect } from '@rstest/core';
const check = expect;
check(value).toBe(true);
```

This does not trigger `rstest/prefer-to-be-truthy` (and similarly misses the other two new rules), even though `check` is the same assertion API. Please add a regression test and resolve the alias before parsing the expect chain.

swwind (commented):
One false-negative case.

swwind (dismissed):

swwind (approved):

PR #1978, opened 2026-09-01
feat(rstest): add consistent-test-filename rule

elecmonkey:
## Summary

Report a test file whose name does not follow the project's convention, which Rstest running both `*.test.*` and `*.spec.*` files by default otherwise lets drift.

## Credits

Port from [`eslint-plugin-vitest`](https://github.com/vitest-dev/eslint-plugin-vitest)'s `consistent-test-filename`.

- **Both default patterns accept `.cts`, `.mts`, `.cjs`, and `.mjs`.** Upstream's defaults end in `[tj]sx?`, which does not cover every extension Rstest's default `include` glob (`**/*.{test,spec}.?(c|m)[jt]s?(x)`) does. `user-service.spec.mts` is a test Rstest actually runs, and under upstream's defaults it would not even be recognized as a test file.

## Related Links

Related #935

## Checklist


internal/plugins/rstest/rules/consistent_test_filename/consistent_test_filename.go
```diff
@@ -0,0 +1,105 @@
...
+
+		if !opts.allTestPattern.TestOrTimeout(fileName) {
+			return nil
+		}
+		if opts.pattern.TestOrTimeout(fileName) {
+			return nil
+		}
+
+		ctx.ReportRange(
+			core.NewTextRange(0, 0),
```
swwind: Upstream reports this on the `Program` node, so the diagnostic spans the source file. Reporting `NewTextRange(0, 0)` instead makes every diagnostic end at column 1. Please report the full `SourceFile` range and update the position assertions to preserve the upstream range.

swwind (commented):
Review comment.

internal/plugins/rstest/rules/consistent_test_filename/consistent_test_filename.go
elecmonkey: Fixed in a1b41abc2. The diagnostic now spans the full source file range (`core.NewTextRange(0, ctx.SourceFile.End())`) rather than a zero-width range at offset 0. `ReportNode` would have trimmed leading trivia off that range, so the range is built directly. Position assertions in both Go suites and in the JS rule test were updated to the whole-file range.

swwind (dismissed):
LGTM

swwind (approved):

Issue #1309, opened 2026-07-20
[Tracking]: eslint-plugin-unicorn rules


fansenze:
This is a tracking issue for porting the existing rules from [`eslint-plugin-unicorn`](https://github.com/sindresorhus/eslint-plugin-unicorn#rules).

There are **338** rules (+ **8 deprecated**).

| Category | ✅ Done | 🚫 Won't | Total |
| :--- | :---: | :---: | :---: |
| Recommended | 47 | 0 | 306 |
| Not recommended | 3 | 0 | 32 |
| Deprecated | 0 | 8 | 8 |
| **Total** | **50** | **8** | **346** |

> ✅ = Implemented &nbsp;|&nbsp; 🚫 = Not intending to implement

---

## Recommended rules

<details open>
<summary>✅ 47 &nbsp;|&nbsp; 🚫 0 &nbsp;/&nbsp; Total: 306</summary>

| Status | Name |
| :----: | :--- |
|  | [unicorn/better-dom-traversing](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/better-dom-traversing.md) |
| ✅ | [unicorn/catch-error-name](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/catch-error-name.md) |
|  | [unicorn/class-reference-in-static-methods](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/class-reference-in-static-methods.md) |
|  | [unicorn/consistent-assert](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/consistent-assert.md) |
|  | [unicorn/consistent-boolean-name](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/consistent-boolean-name.md) |
|  | [unicorn/consistent-class-member-order](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/consistent-class-member-order.md) |
|  | [unicorn/consistent-compound-words](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/consistent-compound-words.md) |
|  | [unicorn/consistent-conditional-object-spread](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/consistent-conditional-object-spread.md) |
[...]

PR #1980, opened 2026-09-01
feat: port rule unicorn/consistent-date-clone

shiyudesu:
## Summary

Port `unicorn/consistent-date-clone` from
`eslint-plugin-unicorn@v74.0.0` to rslint.

The rule reports redundant `Date#getTime()` calls when cloning an existing
`Date`:

```js
// Before
const cloned = new Date(date.getTime());

// After
const cloned = new Date(date);
```

## Implementation

- Matches `new Date(date.getTime())` with exactly one constructor argument.
- Preserves parentheses and TypeScript expression wrappers during autofix.
- Excludes optional calls, computed properties, spread arguments, and
  additional constructor arguments.
- Preserves upstream syntax-based behavior when `Date` is locally shadowed.
- Extracts the existing `removeMethodCall` behavior into a shared Unicorn
  fixer.
- Registers the rule and enables it in the Unicorn recommended preset.

## Tests

- Migrated the complete upstream v74.0.0 test suite.
- Added Go coverage for TypeScript syntax, parentheses, optional chains,
  comments, nested calls, diagnostic ranges, and edit-demand behavior.
- Added the JS end-to-end RuleTester suite.
- Differentially verified diagnostic messages, positions, and autofixes
  against `eslint-plugin-unicorn@v74.0.0`.

Verification completed:

- `go test -count=1 ./internal/plugins/unicorn/rules/consistent_date_clone`
- `go test -count=1 ./internal/plugins/unicorn/rules/prefer_array_flat_map`
- `go test -count=1 ./internal/plugins/unicorn/unicornutil`
- `go test -count=1 ./internal/plugins/unicorn/...`
- `pnpm build`
- `pnpm exec rstest run --testTimeout=10000 consistent-date-clone`
- `pnpm exec rstest run --testTimeout=10000 presets`
- `pnpm typecheck`
- `pnpm lint`
- `pnpm -w run check-spell`
- `pnpm format:check`
- `golangci-lint`: 0 issues

## Related Links

- [Documentation](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v74.0.0/docs/rules/consistent-date-clone.md)
- [Source](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v74.0.0/rules/consistent-date-clone.js)
- [Upstream tests](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v74.0.0/test/consistent-date-clone.js)

Tracks #1309.

## Checklist

- [x] Tests updated.
- [x] Documentation updated.
- [x] Rule registered in the Unicorn catalog.
- [x] Unicorn recommended preset updated.

### DIFF
--- a/internal/plugins/unicorn/rules/consistent_date_clone/consistent_date_clone.go
+++ b/internal/plugins/unicorn/rules/consistent_date_clone/consistent_date_clone.go
@@ -43,6 +43,9 @@ var ConsistentDateCloneRule = rule.Rule{
 				}
 
 				argument := ast.SkipParentheses(node.Arguments()[0])
+				if argument == nil || ast.IsOptionalChain(argument) {
+					return
+				}
 				getTimeCall, ok := unicornutil.MatchDotMethodCall(argument, unicornutil.DotMethodCallOptions{
 					Method:          "getTime",
 					ArgumentsLength: &zeroArguments,
--- a/internal/plugins/unicorn/rules/consistent_date_clone/consistent_date_clone.go
+++ b/internal/plugins/unicorn/rules/consistent_date_clone/consistent_date_clone.go
@@ -1,0 +1,62 @@
+package consistent_date_clone
+
+import (
+	"github.com/microsoft/typescript-go/shim/ast"
+	"github.com/web-infra-dev/rslint/internal/plugins/unicorn/unicornutil"
+	"github.com/web-infra-dev/rslint/internal/rule"
+	"github.com/web-infra-dev/rslint/internal/utils"
+)
+
+const messageID = "consistent-date-clone/error"
+
+var unnecessaryGetTimeMessage = rule.RuleMessage{
+	Id:          messageID,
+	Description: "Unnecessary `.getTime()` call.",
+}
+
+func isSingleArgumentDateConstruction(node *ast.Node) bool {
+	if node == nil || !ast.IsNewExpression(node) {
+		return false
+	}
+
+	arguments := node.Arguments()
+	if len(arguments) != 1 || ast.IsSpreadElement(arguments[0]) {
+		return false
+	}
+
+	callee := ast.SkipParentheses(node.Expression())
+	return callee != nil && ast.IsIdentifier(callee) && callee.AsIdentifier().Text == "Date"
+}
+
+// https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v74.0.0/docs/rules/consistent-date-clone.md
+var ConsistentDateCloneRule = rule.Rule{
+	Name:   "unicorn/consistent-date-clone",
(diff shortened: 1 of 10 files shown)


internal/plugins/unicorn/rules/consistent_date_clone/consistent_date_clone.go
```diff
@@ -0,0 +1,62 @@
...
+		zeroArguments := 0
+		return rule.RuleListeners{
+			// Exit order reports a nested clone before its enclosing clone, matching
+			// the source-ordered diagnostics exposed by ESLint.
+			rule.ListenerOnExit(ast.KindNewExpression): func(node *ast.Node) {
+				if !isSingleArgumentDateConstruction(node) {
+					return
+				}
+
+				argument := ast.SkipParentheses(node.Arguments()[0])
```
elecmonkey, line 45: [P2] Reject calls within an unbroken optional chain

When the receiver starts an unbroken optional chain, such as `new Date(date?.value.getTime())`, ts-go marks the outer argument call with `IsOptionalChain` but not `IsOptionalChainRoot`, so this call passes `MatchDotMethodCall`. `eslint-plugin-unicorn@74.0.0` reports nothing because the argument is a `ChainExpression`, while Rslint reports at 1:22–31 and `--fix` rewrites it to `new Date(date?.value)`. Reject an optional-chain argument before matching, while continuing to accept the parenthesized `(date?.value).getTime()` form, which upstream does report.

shiyudesu: Thanks, good catch. I now reject the argument before method matching when `ast.IsOptionalChain(argument)` is true, which maps tsgo's chain flags to the `ChainExpression` gate that upstream gets from ESTree.

I added regression coverage for both sides of the boundary:

- `new Date(date?.value.getTime())` and `new Date((date?.value.getTime)())` are ignored because the `getTime()` call remains part of an unbroken optional chain.
- `new Date((date?.value).getTime())` is still reported and fixed to `new Date((date?.value))` because the receiver parentheses end the optional chain.

I also verified these cases against eslint-plugin-unicorn v74.0.0 and reran the consistent-date-clone, shared Unicorn utility, and prefer-array-flat-map Go tests.

elecmonkey (dismissed):
LGTM

fansenze:
@shiyudesu Thanks for the contribution! This PR currently has conflicts with the main branch. Please resolve them and ping me afterward.

shiyudesu:
@fansenze Conflicts have been resolved. I rebased the branch onto the latest `main` and updated the test registration to the new `packages/rslint-test-tools/rstack.config.mts` layout introduced by #1981. GitHub now reports the PR as mergeable.

elecmonkey (approved):

PR #1982, opened 2026-09-01
feat: port rule react/hook-use-state

swwind:
## Summary

Port the `react/hook-use-state` rule from ESLint to rslint.

Ensures React `useState` calls use a symmetric `[value, setValue]` destructure, with editor suggestions for setter names and eligible `useMemo` conversions.

## Related Links

- ESLint rule: https://github.com/jsx-eslint/eslint-plugin-react/blob/v7.37.5/docs/rules/hook-use-state.md
- Source code: https://github.com/jsx-eslint/eslint-plugin-react/blob/v7.37.5/lib/rules/hook-use-state.js

## Checklist


internal/plugins/react/rules/hook_use_state/hook_use_state.go
```diff
@@ -0,0 +1,350 @@
...
+}
+
+// isReactUseStateCall ports Components#isReactHookCall(node, ["useState"]).
+// Parentheses are transparent because ESTree drops them; TS assertion wrappers
+// deliberately are not, matching the upstream parser's explicit TS nodes.
+func isReactUseStateCall(ctx rule.RuleContext, imports importInfo, node *ast.Node) bool {
+	if node == nil || node.Kind != ast.KindCallExpression {
+		return false
+	}
+	callee := ast.SkipParentheses(node.AsCallExpression().Expression)
```
elecmonkey: This unwrap makes `(React?.useState)()` look like a plain `React.useState()` call. Upstream sees the outer call's callee as a ChainExpression and `isReactHookCall` returns false, so `import React from 'react'; const result = (React?.useState)()` produces no report in eslint-plugin-react@7.37.5. Rslint currently reports it as non-destructured, so this is a false positive; avoid unwrapping optional-chain callees before deciding the call is a React hook.

internal/plugins/react/rules/hook_use_state/hook_use_state.go
```diff
@@ -0,0 +1,350 @@
...
+	}
+	clause := decl.ImportClause.AsImportClause()
+	if clause.Name() != nil && clause.Name().Kind == ast.KindIdentifier {
+		info.defaultName = clause.Name().AsIdentifier().Text
+		info.defaultSpecifier = decl.ImportClause
+	}
+	if clause.NamedBindings == nil || clause.NamedBindings.Kind != ast.KindNamedImports {
+		return info
+	}
+	for _, spec := range clause.NamedBindings.AsNamedImports().Elements.Nodes {
```
elecmonkey, line 93: Within a single import declaration this loop keeps overwriting matching specifiers, while upstream uses `namedReactImports.find(...)` and therefore keeps the first match. For example, `import { useState as useFirst, useState as useSecond } from 'react'; const [value] = useSecond(initial)` should insert `useMemo` after `useFirst`, but Rslint inserts after `useSecond`; similarly duplicate `useMemo` aliases should use `memo1`, not `memo2`. Please preserve the first matching specifier inside `reactImportInfo`.

internal/plugins/react/rules/hook_use_state/hook_use_state.go
```diff
@@ -0,0 +1,350 @@
...
+							fix := rule.RuleFixInsertAfter(imports.useStateSpecifier, ", useMemo")
+							importFix = &fix
+						}
+						fixes := make([]rule.RuleFix, 0, 3)
+						if importFix != nil {
+							fixes = append(fixes, *importFix)
+						}
+						fixes = append(fixes,
+							rule.RuleFixReplace(ctx.SourceFile, pattern, valueName),
+							rule.RuleFixReplace(ctx.SourceFile, node, memoName+"(() => "+utils.TrimmedNodeText(ctx.SourceFile, argument)+", [])"),
```
elecmonkey: `TrimmedNodeText` includes ts-go's explicit ParenthesizedExpression text, but upstream `getText(context, node.arguments[0])` reads the ESTree argument range with redundant parentheses elided. As a result, `useState((initial))` is suggested as `useMemo(() => (initial), [])` in Rslint, while eslint-plugin-react@7.37.5 suggests `useMemo(() => initial, [])`; `useState((first, second))` similarly differs. Use the text range of the de-parenthesized argument for this suggestion to match upstream.

internal/plugins/react/rules/hook_use_state/hook_use_state.go
```diff
@@ -0,0 +1,358 @@
...
+// deliberately are not, matching the upstream parser's explicit TS nodes.
+func isReactUseStateCall(ctx rule.RuleContext, imports importInfo, node *ast.Node) bool {
+	if node == nil || node.Kind != ast.KindCallExpression {
+		return false
+	}
+	callee := ast.SkipParentheses(node.AsCallExpression().Expression)
+	if callee == nil {
+		return false
+	}
+	if ast.IsOptionalChain(callee) {
```
swwind: Fixed in `8c7786470`. Optional-chain callees are now excluded from React hook detection, while `useState?.()` remains recognized. Added regression coverage for both forms.

internal/plugins/react/rules/hook_use_state/hook_use_state.go
```diff
@@ -0,0 +1,358 @@
...
+		info.defaultName = clause.Name().AsIdentifier().Text
+		info.defaultSpecifier = decl.ImportClause
+	}
+	if clause.NamedBindings == nil || clause.NamedBindings.Kind != ast.KindNamedImports {
+		return info
+	}
+	for _, spec := range clause.NamedBindings.AsNamedImports().Elements.Nodes {
+		switch importedName(spec) {
+		case "useState":
+			if info.useStateSpecifier == nil {
```
swwind, line 101: Fixed in `8c7786470`. `reactImportInfo` now preserves the first matching `useState` and `useMemo` import, matching upstream behavior.

internal/plugins/react/rules/hook_use_state/hook_use_state.go
```diff
@@ -0,0 +1,358 @@
...
+							fix := rule.RuleFixInsertAfter(imports.useStateSpecifier, ", useMemo")
+							importFix = &fix
+						}
+						fixes := make([]rule.RuleFix, 0, 3)
+						if importFix != nil {
+							fixes = append(fixes, *importFix)
+						}
+						fixes = append(fixes,
+							rule.RuleFixReplace(ctx.SourceFile, pattern, valueName),
+							rule.RuleFixReplace(ctx.SourceFile, node, memoName+"(() => "+utils.TrimmedNodeText(ctx.SourceFile, ast.SkipParentheses(argument))+
```
swwind, line 421: Fixed in `8c7786470`. Memo suggestions now use the de-parenthesized argument text, matching ESTree’s range behavior.

internal/plugins/react/rules/hook_use_state/hook_use_state.go
elecmonkey: Upstream `eslint-plugin-react@7.37.5` still reports direct optional-member calls like `import React from 'react'; const result = React?.useState()` as `useStateErrorMessage`. This guard rejects every optional-chain callee before checking whether it is the direct `React?.useState()` shape, so Rslint emits no diagnostic for that case. `(React?.useState)()` is intentionally ignored upstream, but `React?.useState()` should still be reported; please narrow this optional-chain handling so only the upstream-ignored shape is skipped.

swwind: Fixed in `32ebb2e77`. Optional-member calls are now handled like `eslint-plugin-react@7.37.5`: `(React?.useState)()` is ignored, while direct `React?.useState()` and `useState?.()` are reported. Added Go regression coverage and all focused checks pass.

internal/plugins/react/rules/hook_use_state/hook_use_state.go
```diff
@@ -0,0 +1,362 @@
...
+	if ctx.Refs == nil || ident == nil || ident.Kind != ast.KindIdentifier {
+		return false
+	}
+	// Components#isReactHookCall requires the local hook spelling to match
+	// /^use[A-Z]/, even when it is imported under another name.
+	name := ident.AsIdentifier().Text
+	if len(name) < 4 || name[:3] != "use" || name[3] < 'A' || name[3] > 'Z' {
+		return false
+	}
+	symbol := ctx.Refs.Resolve(ident)
```
elecmonkey: [P2] Match upstream's reference-selection semantics

When another imported React hook is referenced before a locally shadowed `useState`, resolving this callee directly suppresses a diagnostic that upstream emits. For example, `import { useFoo, useState } from 'react'; function f() { useFoo(); function useState() {} const result = useState(); }` produces `useStateErrorMessage` in `eslint-plugin-react@7.37.5`, while Rslint produces no diagnostic. Upstream uses the first matching React-hook reference from `scope.references`, so this lookup needs to preserve that selection behavior for 1:1 compatibility.

internal/plugins/react/rules/hook_use_state/hook_use_state.go
```diff
@@ -0,0 +1,362 @@
...
+					return
+				}
+				if onlyDestructuredValue {
+					ctx.ReportNode(pattern, destructuredStateErrorMessage())
+					return
+				}
+
+				ctx.ReportNodeWithDeferredSuggestions(pattern, useStateErrorMessage(), func() []rule.RuleSuggestion {
+					suggestions := make([]rule.RuleSuggestion, 0, 2)
+					if value != nil && len(elements) == 1 && valueName != "" && node.AsCallExpression().Arguments != nil && len(node.AsCallExpression().Arguments.Nodes) == 1 {
```
elecmonkey: [P2] Preserve the memo suggestion for non-Identifier single getters

Requiring `valueName != ""` drops a suggestion emitted upstream for single-element AssignmentPattern and RestElement bindings. For both `const [value = fallback] = useState(initial)` and `const [...value] = useState(initial)`, `eslint-plugin-react@7.37.5` emits `suggestMemo` (with its existing `undefined` replacement), while Rslint emits no suggestion. Since suggestion output is part of compatibility, determine single-getter eligibility from element presence/count as upstream does rather than identifier extraction.

internal/plugins/react/rules/hook_use_state/hook_use_state.go
swwind: Fixed in 7604a0244. Named React hook detection now mirrors Components#isReactHookCall by selecting the first hook-shaped reference in the current scope before checking shadowing. Added a regression test for the useFoo-before-shadowed-useState case; the hook-use-state package test passes.

internal/plugins/react/rules/hook_use_state/hook_use_state.go
swwind: Fixed in 7604a0244. Single-getter memo suggestion eligibility now follows the upstream element/count check, so AssignmentPattern and RestElement bindings retain suggestMemo behavior, including the existing empty replacement value. Added regressions for both forms; the hook-use-state package test passes.

internal/plugins/react/rules/hook_use_state/hook_use_state.go
```diff
@@ -0,0 +1,390 @@
...
+	reference := firstReactHookReference(referenceScopes, imports, ident)
+	if reference == nil {
+		return false
+	}
+	for _, declaration := range reference.Declarations {
+		if declaration != nil && declaration.Name == name && declaration.Kind != scope.DefImport {
+			return false
+		}
+	}
+	return imports.namedHookNames[name] == "useState" || name == "useState"
```
elecmonkey: [P2] Respect the imported hook behind a local `useState` alias

For `import { useFoo as useState } from 'react'; const result = useState();`, eslint-plugin-react@7.37.5 reports nothing, but Rslint emits `useStateErrorMessage`. Upstream uses the mapped imported name (`useFoo`) when one exists; the unconditional `name == "useState"` fallback instead classifies this call as React's `useState`. Only fall back to the local name when no named-import mapping exists.

internal/plugins/react/rules/hook_use_state/hook_use_state.go
```diff
@@ -0,0 +1,390 @@
...
+}
+
+func resolvesToDefaultReactImport(ctx rule.RuleContext, defaultSpecifier, ident *ast.Node) bool {
+	if ctx.Refs == nil || defaultSpecifier == nil || ident == nil || ident.Kind != ast.KindIdentifier ||
+		defaultSpecifier.Pos() >= ident.Pos() {
+		return false
+	}
+	symbol := ctx.Refs.Resolve(ident)
```
elecmonkey: [P2] Mirror first-reference semantics for the default React import

For `import React from 'react'; function f(x = React) { const React = {}; const result = React.useState(); }`, eslint-plugin-react@7.37.5 reports the call, while Rslint reports nothing. Upstream checks the first `React` reference in the current function scope, which is the parameter-default reference resolving to the import; resolving only the current callee instead sees the body-local declaration. The named-hook path now models this first-reference behavior, but the default-import path still needs the equivalent handling.

internal/plugins/react/rules/hook_use_state/hook_use_state.go
```diff
@@ -0,0 +1,390 @@
...
+					return
+				}
+				if onlyDestructuredValue {
+					ctx.ReportNode(pattern, destructuredStateErrorMessage())
+					return
+				}
+
+				ctx.ReportNodeWithDeferredSuggestions(pattern, useStateErrorMessage(), func() []rule.RuleSuggestion {
+					suggestions := make([]rule.RuleSuggestion, 0, 2)
+					if value != nil && len(elements) == 1 && node.AsCallExpression().Arguments != nil && len(node.AsCallExpression().Arguments.Nodes) == 1 {
```
elecmonkey: [P2] Exclude omitted binding elements from memo suggestions

For `import { useState } from 'react'; const [,] = useState(initial)`, upstream emits the diagnostic without suggestions, while Rslint offers `suggestMemo` and rewrites the declaration to invalid `const  = useMemo(() => initial, [])`. ts-go represents the omitted slot with a non-nil node, so `value != nil` does not implement upstream's `valueVariable` truthiness check; distinguish an omitted element from actual AssignmentPattern/RestElement bindings before enabling this suggestion.

internal/plugins/react/rules/hook_use_state/hook_use_state.go
```diff
@@ -0,0 +1,390 @@
...
+						if imports.useStateSpecifier != nil && (imports.useMemoSpecifier == nil || imports.defaultName != "") {
+							fix := rule.RuleFixInsertAfter(imports.useStateSpecifier, ", useMemo")
+							importFix = &fix
+						}
+						fixes := make([]rule.RuleFix, 0, 3)
+						if importFix != nil {
+							fixes = append(fixes, *importFix)
+						}
+						fixes = append(fixes,
+							rule.RuleFixReplace(ctx.SourceFile, pattern, valueName),
```
elecmonkey: [P2] Preserve the upstream replacement for non-Identifier getters

For `const [value = fallback] = useState(initial)` and `const [...value] = useState(initial)`, eslint-plugin-react@7.37.5 under ESLint 8.57.1 and 9.39.5 produces a `suggestMemo` result containing `const undefined = useMemo(...)`; Rslint currently emits `const  = useMemo(...)`. The restored suggestion therefore still differs from the upstream output because `identifierBindingName` returns an empty string here rather than preserving JavaScript's `undefined` coercion.

internal/plugins/react/rules/hook_use_state/hook_use_state.go
swwind: Fixed in `f1aa559a4`. Named React hook detection now uses the imported hook name rather than falling back to the local spelling, so `useFoo as useState` is not reported. Added regression coverage.

internal/plugins/react/rules/hook_use_state/hook_use_state.go
swwind: Fixed in `f1aa559a4`. Default React hook detection now mirrors upstream's first-reference selection within the current scope, including the parameter-default case. Added regression coverage for this upstream-compatible behavior.

internal/plugins/react/rules/hook_use_state/hook_use_state.go
swwind: Fixed in `f1aa559a4`. Omitted binding elements are now excluded from single-getter memo suggestions. Added regression coverage.

internal/plugins/react/rules/hook_use_state/hook_use_state.go
swwind: Fixed in `f1aa559a4`. AssignmentPattern and RestElement suggestions now preserve upstream's `undefined` replacement output. Added regression coverage.

internal/plugins/react/rules/hook_use_state/hook_use_state.go
```diff
@@ -0,0 +1,426 @@
...
+	reference := firstReactHookReference(referenceScopes, imports, ident)
+	if reference == nil {
+		return false
+	}
+	for _, declaration := range reference.Declarations {
+		if declaration != nil && declaration.Name == name && declaration.Kind != scope.DefImport {
+			return false
+		}
+	}
+	return imports.namedHookNames[name] == "useState"
```
elecmonkey: [P2] Preserve upstream fallback for bare useState calls

This final check rejects a bare `useState()` unless the callee itself is a named React import mapped to `useState`, but upstream falls back to the callee spelling after selecting the first hook-shaped reference in the scope. Repro: `import { useEffect } from 'react'; useEffect(); const result = useState();`. `eslint-plugin-react@7.37.5` reports `useStateErrorMessage`, while Rslint reports nothing, so this is a false negative against the port’s 1:1 compatibility target.

internal/plugins/react/rules/hook_use_state/hook_use_state.go
```diff
@@ -0,0 +1,426 @@
...
+		}
+	}
+	return nil
+}
+
+func resolvesToDefaultReactImport(referenceScopes map[*ast.Node]*scope.Scope, imports importInfo, ident *ast.Node) bool {
+	if imports.defaultName == "" || imports.defaultSpecifier == nil {
+		return false
+	}
+	reference := firstReference(referenceScopes, ident, imports.defaultName)
```
elecmonkey, line 174: [P2] Require the callee object to be the first React import

This lookup can classify `B.useState()` as a React hook call just because the same scope has an earlier reference to the first default React import. Repro: `import A from 'react'; import B from 'react'; A; const result = B.useState();`. `eslint-plugin-react@7.37.5` reports no diagnostics because the member object is not the first default import name, but Rslint reports `useStateErrorMessage`, creating false positives when a file has multiple React default aliases.

internal/plugins/react/rules/hook_use_state/hook_use_state.go
swwind: Fixed in `b03c4886d`. Bare `useState` calls now preserve upstream’s local-name fallback after selecting the first hook-shaped reference. Added regression coverage.

internal/plugins/react/rules/hook_use_state/hook_use_state.go
swwind: Fixed in `b03c4886d`. Default React hook detection now requires the member object to match the first default React import. Existing regression coverage verifies that `B.useState()` is ignored when `A` is the first import.

elecmonkey (approved):
Good

PR #1983, opened 2026-09-01
feat: port rule react/jsx-sort-props

swwind:
## Summary

Port the `react/jsx-sort-props` rule from ESLint to rslint.

Enforce configurable JSX prop ordering, including alphabetical sorting, reserved props, callbacks, shorthand props, multiline props, spread boundaries, and autofixes.

## Related Links

- ESLint rule: https://github.com/jsx-eslint/eslint-plugin-react/blob/v7.37.5/docs/rules/jsx-sort-props.md
- Source code: https://github.com/jsx-eslint/eslint-plugin-react/blob/v7.37.5/lib/rules/jsx-sort-props.js

## Checklist


internal/plugins/react/rules/jsx_sort_props/jsx_sort_props.go
```diff
@@ -0,0 +1,440 @@
...
+				left := utils.TrimNodeTextRange(ctx.SourceFile, pending[i].attr)
+				right := utils.TrimNodeTextRange(ctx.SourceFile, pending[j].attr)
+				return left.Pos() < right.Pos()
+		})
+		for _, report := range pending {
+			reportNode := report.attr.AsJsxAttribute().Name()
+				if report.wholeAttribute {
+					reportNode = report.attr
+				}
+				ctx.ReportNodeWithDeferredFixes(reportNode, rule.RuleMessage{Id: report.id, Description: report.description, Data: report.data}, func() []rule.RuleFix {
```
elecmonkey: [P2] Do not attach sorting fixes to invalid reservedFirst diagnostics

With `reservedFirst: []`, `<App b a />` produces two `listIsEmpty` diagnostics. eslint-plugin-react@7.37.5 attaches no fix and leaves the source unchanged, but Rslint routes these validation reports through `ReportNodeWithDeferredFixes` and rewrites it to `<App a b />`. Invalid-option diagnostics should not invoke the sorting fixer.

internal/plugins/react/rules/jsx_sort_props/jsx_sort_props.go
```diff
@@ -0,0 +1,440 @@
...
+				if ids[id] {
+					return
+				}
+				ids[id] = true
+				pending = append(pending, pendingReport{attr, id, description, data, wholeAttribute})
+			}
+
+			memo := attrs[0]
+			for index, current := range attrs {
+				if index == 0 {
```
elecmonkey: [P2] Process a leading spread before skipping the first attribute

For `<App {...p} key={1} />` with `reservedFirst: []`, eslint-plugin-react@7.37.5 reports `listIsEmpty` on `key`, while Rslint reports nothing. The unconditional `index == 0` branch skips the first spread before it can advance `memo` to the following attribute, so the only ordinary prop never enters reserved-list validation.

internal/plugins/react/rules/jsx_sort_props/jsx_sort_props.go
```diff
@@ -0,0 +1,440 @@
...
+	}
+	opts.callbacksLast, _ = config["callbacksLast"].(bool)
+	opts.shorthandFirst, _ = config["shorthandFirst"].(bool)
+	opts.shorthandLast, _ = config["shorthandLast"].(bool)
+	opts.ignoreCase, _ = config["ignoreCase"].(bool)
+	opts.noSortAlphabetically, _ = config["noSortAlphabetically"].(bool)
+	if value, ok := config["multiline"].(string); ok {
+		opts.multiline = value
+	}
+	if value, ok := config["locale"].(string); ok {
```
elecmonkey: [P2] Treat an empty locale as auto

The schema permits `locale: ""`, and upstream uses `configuration.locale || 'auto'`. Consequently `<App Z a />` is valid there, but Rslint preserves the empty string, enters locale collation, and reports `a`; normalize a falsey empty locale back to `auto` when parsing the option.

internal/plugins/react/rules/jsx_sort_props/jsx_sort_props.go
```diff
@@ -0,0 +1,440 @@
...
+}
+
+func isMultiline(ctx rule.RuleContext, node *ast.Node) bool {
+	r := utils.TrimNodeTextRange(ctx.SourceFile, node)
+	lines := ctx.SourceFile.ECMALineMap()
+	return scanner.ComputeLineOfPosition(lines, r.Pos()) != scanner.ComputeLineOfPosition(lines, r.End())
+}
+
+func compareNames(left, right string, opts options) int {
+	if opts.ignoreCase || opts.locale != "auto" {
```
elecmonkey, line 271: [P2] Match localeCompare for explicit locales

The Go collation used here is not fully equivalent to JavaScript `localeCompare`. For example, with `locale: "da"`, eslint-plugin-react@7.37.5 accepts `<App cH ch />`, whereas Rslint reports `ch` and would reorder the props because `x/text/collate` gives the opposite case ordering. The only documented divergence covers `locale: "auto"`, so explicit locales still need compatible ordering.

elecmonkey (commented):
Port-rule review findings.

internal/plugins/react/rules/jsx_sort_props/jsx_sort_props.go
```diff
@@ -0,0 +1,440 @@
...
+	}
+	switch value := config["reservedFirst"].(type) {
+	case bool:
+		opts.reservedFirst = value
+	case []any:
+		opts.reservedFirst = true
+		opts.reservedList = make([]string, 0, len(value))
+		invalid := make([]string, 0)
+		for _, item := range value {
+			name, _ := item.(string)
```
elecmonkey, line 237: [P3] Preserve non-string reservedFirst values in the diagnostic

The copied schema allows arbitrary array items. For `reservedFirst: [1]`, upstream reports `Remove: 1`, but this failed string assertion converts the item to `""`, so Rslint reports `Remove: ` instead. Preserve JavaScript-style stringification or constrain the schema to strings if that incompatibility is intentional.

internal/plugins/react/rules/jsx_sort_props/jsx_sort_props.go
```diff
@@ -0,0 +1,440 @@
...
+						continue
+					}
+				}
+				if !opts.noSortAlphabetically && compareNames(previousName, currentName, opts) > 0 {
+					reportAttr(current, "sortPropsByAlpha", "Props should be sorted alphabetically", nil, false)
+					continue
+				}
+				memo = current
+			}
+			sort.SliceStable(pending, func(i, j int) bool {
```
elecmonkey, line 176: [P3] Run gofmt on the new rule

`gofmt -d` changes the indentation of the `sort.SliceStable` closure and following loop at this block. The repository guidelines require Go sources to use gofmt, so the new file should be formatted before merging.

internal/plugins/react/rules/jsx_sort_props/jsx_sort_props.go
```diff
@@ -0,0 +1,451 @@
...
+				right := utils.TrimNodeTextRange(ctx.SourceFile, pending[j].attr)
+				return left.Pos() < right.Pos()
+			})
+			for _, report := range pending {
+				reportNode := report.attr.AsJsxAttribute().Name()
+				if report.wholeAttribute {
+					reportNode = report.attr
+				}
+				message := rule.RuleMessage{Id: report.id, Description: report.description, Data: report.data}
+				if !report.fixable {
```
swwind, line 187: Fixed. Invalid reservedFirst configurations now report without attaching a sorting fix. Added a <App b a /> regression case to verify that the output remains unchanged.

internal/plugins/react/rules/jsx_sort_props/jsx_sort_props.go
```diff
@@ -0,0 +1,451 @@
...
+				if ids[id] {
+					return
+				}
+				ids[id] = true
+				pending = append(pending, pendingReport{attr, id, description, data, wholeAttribute, opts.reservedError == ""})
+			}
+
+			memo := attrs[0]
+			for index, current := range attrs {
+				if current.Kind == ast.KindJsxSpreadAttribute {
```
swwind, line 73: Fixed. A leading spread is now processed before the following regular prop is validated against reservedFirst. Added a <App {...p} key={1} /> regression case.

internal/plugins/react/rules/jsx_sort_props/jsx_sort_props.go
```diff
@@ -0,0 +1,451 @@
...
+	}
+	opts.callbacksLast, _ = config["callbacksLast"].(bool)
+	opts.shorthandFirst, _ = config["shorthandFirst"].(bool)
+	opts.shorthandLast, _ = config["shorthandLast"].(bool)
+	opts.ignoreCase, _ = config["ignoreCase"].(bool)
+	opts.noSortAlphabetically, _ = config["noSortAlphabetically"].(bool)
+	if value, ok := config["multiline"].(string); ok {
+		opts.multiline = value
+	}
+	if value, ok := config["locale"].(string); ok && value != "" {
```
swwind, line 226: Fixed. parseOptions now keeps an empty locale at the default value, "auto". Added a <App Z a /> regression case.

internal/plugins/react/rules/jsx_sort_props/jsx_sort_props.go
```diff
@@ -0,0 +1,451 @@
...
+	r := utils.TrimNodeTextRange(ctx.SourceFile, node)
+	lines := ctx.SourceFile.ECMALineMap()
+	return scanner.ComputeLineOfPosition(lines, r.Pos()) != scanner.ComputeLineOfPosition(lines, r.End())
+}
+
+func compareNames(left, right string, opts options) int {
+	if opts.ignoreCase || opts.locale != "auto" {
+		locale := language.Make(opts.locale)
+		base, _ := locale.Base()
+		if base.String() == "da" && ecmascript.StringToLowerCase(left) == ecmascript.StringToLowerCase(right) {
```
swwind, line 274: Fixed the reported Danish case-ordering behavior and cover da-* locales. Added a <App cH ch /> regression case with locale: "da".

internal/plugins/react/rules/jsx_sort_props/jsx_sort_props.schema.json
```diff
@@ -0,0 +1,24 @@
...
+        "callbacksLast": { "type": "boolean" },
+        "shorthandFirst": { "type": "boolean" },
+        "shorthandLast": { "type": "boolean" },
+        "multiline": {
+          "enum": ["ignore", "first", "last"],
+          "default": "ignore"
+        },
+        "ignoreCase": { "type": "boolean" },
+        "noSortAlphabetically": { "type": "boolean" },
+        "reservedFirst": { "type": ["array", "boolean"], "items": { "type": "string" } },
```
swwind: This is an intentional divergence: reservedFirst arrays now accept only string items in the schema, so non-string values fail configuration validation. The difference is documented in the rule documentation.

internal/plugins/react/rules/jsx_sort_props/jsx_sort_props.go
```diff
@@ -0,0 +1,451 @@
...
+						continue
+					}
+				}
+				if !opts.noSortAlphabetically && compareNames(previousName, currentName, opts) > 0 {
+					reportAttr(current, "sortPropsByAlpha", "Props should be sorted alphabetically", nil, false)
+					continue
+				}
+				memo = current
+			}
+			sort.SliceStable(pending, func(i, j int) bool {
```
swwind, line 176: Ran gofmt.

swwind (commented):
Addressed all review findings in f2c152f03. Validation: go test ./internal/plugins/react/rules/jsx_sort_props; pnpm run check-spell; golangci-lint run --timeout=1m ./internal/plugins/react/rules/jsx_sort_props.

internal/plugins/react/rules/jsx_sort_props/jsx_sort_props.schema.json
elecmonkey: [P2] Preserve upstream sortFirst option support

eslint-plugin-react@7.37.5 includes `sortFirst` in `jsx-sort-props` and the upstream suite covers valid and invalid cases where `className`/`id` must be ordered before all other constraints. This schema omits `sortFirst` while using `additionalProperties: false`, so configs such as `{ sortFirst: ["className"] }` are rejected by Rslint instead of reporting/fixing `<App name="John" className="test" />` to `<App className="test" name="John" />`. I verified the upstream RuleTester case passes and a temporary Rslint schema probe fails with `additional properties "sortFirst" not allowed`.

elecmonkey (commented):
Port-rule review finding.

internal/plugins/react/rules/jsx_sort_props/jsx_sort_props.schema.json
swwind: I can't find `sortFirst` option in eslint-plugin-react@7.37.5.

elecmonkey (approved):

PR #1984, opened 2026-09-01
feat: port rule react/sort-prop-types

swwind:
## Summary

Port the `react/sort-prop-types` rule from ESLint to rslint.

Enforces ordering for React propTypes declarations, including required-first, callbacks-last, shape props, and TypeScript component prop types.

## Related Links

- ESLint rule: https://github.com/jsx-eslint/eslint-plugin-react/blob/v7.37.5/docs/rules/sort-prop-types.md
- Source code: https://github.com/jsx-eslint/eslint-plugin-react/blob/v7.37.5/lib/rules/sort-prop-types.js

## Checklist


internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
```diff
@@ -0,0 +1,360 @@
...
+				if current == nil {
+					continue
+				}
+				if current.Kind == ast.KindSpreadAssignment {
+					// A spread starts an independent ordering group.
+					previous = nil
+					continue
+				}
+				name, ok := declarationName(ctx.SourceFile, current)
+				if !ok {
```
elecmonkey, line 93: Upstream's reducer still advances when it sees members without a sortable key, so a nameless member breaks the comparison chain. Here `continue` leaves `previous` pointing at `z`, so `type P = { z: string; (): void; a: string }; function C(p: P) {}` (and the same shape with an index signature) reports `a` in Rslint while eslint-plugin-react@7.37.5 is silent. Please advance the previous node for these non-spread, unkeyed declarations instead of comparing across them.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
```diff
@@ -0,0 +1,360 @@
...
+			case ast.KindObjectLiteralExpression:
+				checkSorted(value.AsObjectLiteralExpression().Properties.Nodes)
+			case ast.KindIdentifier:
+				// The upstream rule resolves an identifier initialized with an
+				// object. The binder lookup avoids a whole-file text scan here.
+				if decl := declarationObject(ctx.Refs.Resolve(value)); decl != nil {
+					checkSorted(decl.AsObjectLiteralExpression().Properties.Nodes)
+				}
+			case ast.KindCallExpression:
+				if !reactutil.IsPropWrapperCall(value, wrappers) {
```
elecmonkey: Passing the whole call into `IsPropWrapperCall` changes this rule's wrapper semantics. Upstream passes only `node.callee.name`, so `{ object: "Object", property: "assign" }` matches bare `assign(...)` and does not match `Object.assign(...)`; Rslint currently misses the former and reports the latter. The rule needs the same callee-name lookup rather than the shared full-callee matcher.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
```diff
@@ -0,0 +1,360 @@
...
+		}
+
+		listeners := rule.RuleListeners{
+			ast.KindObjectLiteralExpression: func(node *ast.Node) {
+				for _, prop := range node.AsObjectLiteralExpression().Properties.Nodes {
+					if prop.Kind != ast.KindPropertyAssignment {
+						continue
+					}
+					assignment := prop.AsPropertyAssignment()
+					name, ok := utils.GetStaticPropertyName(assignment.Name())
```
elecmonkey: Using `GetStaticPropertyName` does not match upstream's propTypes declaration test. Upstream reads `key.name`, so it ignores `"propTypes"` / `["propTypes"]` keys but reports computed identifiers like `[propTypes]`; this branch does the reverse for object literals. For example, `const C = { "propTypes": { z: P.string, a: P.string } }` is a Rslint-only false positive, while `const C = { [propTypes]: { z: P.string, a: P.string } }` is a false negative.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
```diff
@@ -0,0 +1,360 @@
...
+						value := ast.SkipParentheses(assignment.Initializer)
+						if value != nil && value.Kind == ast.KindObjectLiteralExpression {
+							checkSorted(value.AsObjectLiteralExpression().Properties.Nodes)
+						}
+					}
+				}
+			},
+			ast.KindPropertyDeclaration: func(node *ast.Node) {
+				property := node.AsPropertyDeclaration()
+				name, ok := utils.GetStaticPropertyName(property.Name())
```
elecmonkey: Upstream treats a typed class field named `props` as a prop-types declaration (`node.typeAnnotation && node.key.name === "props"`). This branch only accepts `propTypes`, so `class C { static props: unknown = { z: P.string, a: P.string }; }` is reported by eslint-plugin-react@7.37.5 but missed here.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
```diff
@@ -0,0 +1,360 @@
...
+			},
+		}
+		if opts.sortShapeProp {
+			listeners[ast.KindCallExpression] = func(node *ast.Node) {
+				call := node.AsCallExpression()
+				callee := reactutil.SkipExpressionWrappers(call.Expression)
+				if !isShapeCall(callee) || call.Arguments == nil || len(call.Arguments.Nodes) == 0 {
+					return
+				}
+				checkValue(call.Arguments.Nodes[0])
```
elecmonkey: Upstream's shape listener only sorts the first argument when it is an object expression or identifier. Routing it through `checkValue` also unwraps configured wrapper calls, so `PropTypes.shape(forbidExtraProps({ z: P.string, a: P.string }))` reports in Rslint with `sortShapeProp` plus that wrapper setting, while eslint-plugin-react@7.37.5 is silent.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
```diff
@@ -0,0 +1,360 @@
...
+	}
+	if current.isNumber {
+		previousNumber, ok := ecmascript.StringToNumber(previous.text)
+		return ok && current.number < previousNumber
+	}
+	if previous.isNumber {
+		currentNumber, ok := ecmascript.StringToNumber(current.text)
+		return ok && currentNumber < previous.number
+	}
+	return current.text < previous.text
```
elecmonkey: String comparison here is still Go UTF-8 byte ordering for non-numeric keys, but upstream uses JavaScript string `<`, which compares UTF-16 code units. `Component.propTypes = { "\u{10000}": P.string, "\uE000": P.string };` is valid upstream and reported by Rslint because the byte order reverses those two keys. Please compare using ECMAScript code-unit ordering for this branch too.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
```diff
@@ -0,0 +1,360 @@
...
+		return nil
+	}
+	return param.Type
+}
+
+func isRequired(node *ast.Node) bool {
+	if node == nil || node.Kind != ast.KindPropertyAssignment {
+		return false
+	}
+	value := reactutil.SkipExpressionWrappers(node.AsPropertyAssignment().Initializer)
```
elecmonkey: Unwrapping TypeScript expressions here changes required-first behavior. Upstream checks `node.value.property.name` directly, so `(P.string.isRequired as any)` is not considered required; with `requiredFirst: true`, `Component.propTypes = { a: P.string, b: (P.string.isRequired as any) }` is accepted upstream but reported by Rslint.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
```diff
@@ -0,0 +1,360 @@
...
+
+func declarationObject(symbol *ast.Symbol) *ast.Node {
+	if symbol == nil || symbol.ValueDeclaration == nil {
+		return nil
+	}
+	decl := symbol.ValueDeclaration
+	if decl.Kind != ast.KindVariableDeclaration {
+		return nil
+	}
+	value := reactutil.SkipExpressionWrappers(decl.AsVariableDeclaration().Initializer)
```
elecmonkey: Identifier resolution should keep TypeScript assertions opaque in the same way `checkValue` now does. Upstream receives the variable initializer as a `TSAsExpression` and ignores it, but this unwrap makes `const p = ({ z: P.string, a: P.string } as const); Component.propTypes = p;` report in Rslint only.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
```diff
@@ -0,0 +1,410 @@
...
+					// A spread starts an independent ordering group.
+					previous = nil
+					continue
+				}
+				name, ok := declarationName(ctx.SourceFile, current)
+				if !ok {
+					// An unkeyed declaration breaks the comparison chain. Keep it
+					// as the previous node so the next keyed declaration is not
+					// compared with the declaration before the unkeyed one.
+					previous = current
```
swwind, line 97: Fixed. Unkeyed, non-spread declarations now advance previous, and comparisons restart after them. Added a TypeScript regression case.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
```diff
@@ -0,0 +1,410 @@
...
+			case ast.KindObjectLiteralExpression:
+				checkSorted(value.AsObjectLiteralExpression().Properties.Nodes)
+			case ast.KindIdentifier:
+				// The upstream rule resolves an identifier initialized with an
+				// object. The binder lookup avoids a whole-file text scan here.
+				if decl := declarationObject(ctx.Refs.Resolve(value)); decl != nil {
+					checkSorted(decl.AsObjectLiteralExpression().Properties.Nodes)
+				}
+			case ast.KindCallExpression:
+				if !isPropWrapperCallByCalleeName(value, wrappers) {
```
swwind, line 175: Fixed. Wrapper matching now follows upstream and uses only the direct callee name: {object: "Object", property: "assign"} matches assign(...) but not Object.assign(...). Added regression coverage.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
```diff
@@ -0,0 +1,410 @@
...
+		}
+
+		listeners := rule.RuleListeners{
+			ast.KindObjectLiteralExpression: func(node *ast.Node) {
+				for _, prop := range node.AsObjectLiteralExpression().Properties.Nodes {
+					if prop.Kind != ast.KindPropertyAssignment {
+						continue
+					}
+					assignment := prop.AsPropertyAssignment()
+					if isAuthoredPropertyName(assignment.Name(), "propTypes") {
```
swwind, line 192: Fixed. Object-literal propTypes detection now accepts only authored identifier keys; quoted keys are ignored while computed identifiers are checked. Added regression coverage.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
```diff
@@ -0,0 +1,410 @@
...
+						value := ast.SkipParentheses(assignment.Initializer)
+						if value != nil && value.Kind == ast.KindObjectLiteralExpression {
+							checkSorted(value.AsObjectLiteralExpression().Properties.Nodes)
+						}
+					}
+				}
+			},
+			ast.KindPropertyDeclaration: func(node *ast.Node) {
+				property := node.AsPropertyDeclaration()
+				if property.Type != nil && isAuthoredPropertyName(property.Name(), "props") {
```
swwind, line 207: Fixed. Typed class fields named props are now checked, matching eslint-plugin-react behavior. Added a regression case.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
```diff
@@ -0,0 +1,410 @@
...
+			},
+		}
+		if opts.sortShapeProp {
+			listeners[ast.KindCallExpression] = func(node *ast.Node) {
+				call := node.AsCallExpression()
+				callee := reactutil.SkipExpressionWrappers(call.Expression)
+				if !isShapeCall(callee) || call.Arguments == nil || len(call.Arguments.Nodes) == 0 {
+					return
+				}
+				firstArg := ast.SkipParentheses(call.Arguments.Nodes[0])
```
swwind, line 249: Fixed. sortShapeProp now checks only direct object or identifier arguments and does not recurse through wrapper calls. Added a regression case.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
```diff
@@ -0,0 +1,410 @@
...
+	}
+	if current.isNumber {
+		previousNumber, ok := ecmascript.StringToNumber(previous.text)
+		return ok && current.number < previousNumber
+	}
+	if previous.isNumber {
+		currentNumber, ok := ecmascript.StringToNumber(current.text)
+		return ok && currentNumber < previous.number
+	}
+	return ecmascript.CompareStrings(current.text, previous.text) < 0
```
swwind: Fixed. Property-name ordering now uses ECMAScript UTF-16 code-unit comparison. The corresponding reusable compatibility fact was recorded as JS-14.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
```diff
@@ -0,0 +1,410 @@
...
+		return nil
+	}
+	return param.Type
+}
+
+func isRequired(node *ast.Node) bool {
+	if node == nil || node.Kind != ast.KindPropertyAssignment {
+		return false
+	}
+	value := ast.SkipParentheses(node.AsPropertyAssignment().Initializer)
```
swwind, line 437: Fixed. Required detection skips parentheses but does not cross TypeScript assertion wrappers. Added a regression case.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
```diff
@@ -0,0 +1,410 @@
...
+
+func declarationObject(symbol *ast.Symbol) *ast.Node {
+	if symbol == nil || symbol.ValueDeclaration == nil {
+		return nil
+	}
+	decl := symbol.ValueDeclaration
+	if decl.Kind != ast.KindVariableDeclaration {
+		return nil
+	}
+	value := ast.SkipParentheses(decl.AsVariableDeclaration().Initializer)
```
swwind, line 560: Fixed. Identifier initializer resolution skips parentheses but keeps TypeScript assertion wrappers opaque. Added a regression case.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
```diff
@@ -0,0 +1,410 @@
...
+					}
+				}
+			},
+			ast.KindPropertyDeclaration: func(node *ast.Node) {
+				property := node.AsPropertyDeclaration()
+				if property.Type != nil && isAuthoredPropertyName(property.Name(), "props") {
+					checkValue(property.Initializer)
+					return
+				}
+				name, ok := utils.GetStaticPropertyName(property.Name())
```
elecmonkey: Upstream's class-property check uses propsUtil.isPropTypesDeclaration(), whose class-field branch reads node.key.name rather than the static value. That means eslint-plugin-react@7.37.5 ignores static ["propTypes"] = { z: P.string, a: P.string } but this branch reports it, and upstream reports static [propTypes] = { z: P.string, a: P.string } while this branch misses it. Please use the authored identifier/computed-identifier name, matching the object-literal path, instead of GetStaticPropertyName here.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
```diff
@@ -0,0 +1,410 @@
...
+				if ok && name == "propTypes" {
+					checkValue(property.Initializer)
+				}
+			},
+			ast.KindBinaryExpression: func(node *ast.Node) {
+				binary := node.AsBinaryExpression()
+				left := reactutil.SkipExpressionWrappers(binary.Left)
```
elecmonkey: Upstream listens to ESTree MemberExpression and getPropertyName() returns the identifier name for computed members, so C[propTypes] = { z: P.string, a: P.string } is diagnosed. This listener only accepts ts-go PropertyAccessExpression, so the same computed-identifier propTypes assignment is skipped entirely. Please include ElementAccessExpression with an identifier argument for parity.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
```diff
@@ -0,0 +1,410 @@
...
+
+func isPropWrapperCallByCalleeName(call *ast.Node, wrappers []reactutil.PropWrapperEntry) bool {
+	if call == nil || call.Kind != ast.KindCallExpression {
+		return false
+	}
+	callee := ast.SkipParentheses(call.AsCallExpression().Expression)
+	if callee == nil || callee.Kind != ast.KindIdentifier {
+		return false
+	}
+	name := callee.AsIdentifier().Text
```
elecmonkey, line 516: For this rule upstream calls isPropWrapperFunction(context, node.callee.name), so a setting like propWrapperFunctions: ["Object.assign"] does not match assign(...) because the raw string is compared with "assign". GetPropWrapperFunctions has already split that string into {Object, assign}, and this check compares only wrapper.Property, so Component.propTypes = assign({ z: P.string, a: P.string }) becomes a Rslint-only report. This path needs to preserve the original string-vs-object distinction for sort-prop-types.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
```diff
@@ -0,0 +1,410 @@
...
+				name := left.AsPropertyAccessExpression().Name()
+				if name != nil && name.Kind == ast.KindIdentifier && name.AsIdentifier().Text == "propTypes" {
+					checkValue(binary.Right)
+				}
+			},
+		}
+		if opts.sortShapeProp {
+			listeners[ast.KindCallExpression] = func(node *ast.Node) {
+				call := node.AsCallExpression()
+				callee := reactutil.SkipExpressionWrappers(call.Expression)
```
elecmonkey: Upstream checks node.callee.property.name directly in isShapeProp(), so TypeScript wrappers around the callee remain opaque. By unwrapping before isShapeCall, Rslint reports inside (PropTypes.shape as any)({ z: P.string, a: P.string }) with sortShapeProp enabled, while eslint-plugin-react@7.37.5 leaves it alone. Please use the direct callee shape here, as the rule already does for value-side TS wrappers.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
swwind: Fixed in ae184da3f. Class-field propTypes detection now matches only authored identifiers and computed identifiers, so static ["propTypes"] is ignored while static [propTypes] is checked. Added regression coverage.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
swwind: Fixed in ae184da3f. The assignment listener now handles ElementAccessExpression with an identifier argument, so C[propTypes] is checked. Added regression coverage.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
swwind: Fixed in ae184da3f. Raw string and object/property wrapper configurations are now distinguished, so ["Object.assign"] no longer matches bare assign(...), while object/property entries retain their upstream behavior. Added regression coverage.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
swwind: Fixed in ae184da3f. Shape matching skips parentheses only and keeps TypeScript assertion wrappers opaque. Added regression coverage.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
```diff
@@ -0,0 +1,417 @@
...
+					checkValue(property.Initializer)
+					return
+				}
+				if isAuthoredPropertyName(property.Name(), "propTypes") {
+					checkValue(property.Initializer)
+				}
+			},
+			ast.KindBinaryExpression: func(node *ast.Node) {
+				binary := node.AsBinaryExpression()
+				left := reactutil.SkipExpressionWrappers(binary.Left)
```
elecmonkey: The assignment listener strips TypeScript expression wrappers from the left-hand side before checking for `propTypes`, which makes Rslint report cases that upstream ignores. For example, eslint-plugin-react@7.37.5 emits no `sort-prop-types` diagnostic for `(Component.propTypes as any) = { z: P.string, a: P.string }` or `(Component.propTypes!) = { z: P.string, a: P.string }`, because the `MemberExpression`'s parent is the TS wrapper rather than the `AssignmentExpression`. Rslint currently unwraps the LHS here and reports `propsNotSorted`. Please keep TS wrappers opaque on this assignment-target path, matching the direct ESTree member behavior upstream uses.

swwind: Fixed in 8a7ba370f. The assignment listener now skips only parentheses and keeps TypeScript assertion and non-null wrappers opaque, matching upstream's direct MemberExpression parent check. Added regression coverage for `(Component.propTypes as any) = ...` and `(Component.propTypes!) = ...`; the focused Go package test passes.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
```diff
@@ -0,0 +1,419 @@
...
+			if value == nil {
+				return
+			}
+			switch value.Kind {
+			case ast.KindObjectLiteralExpression:
+				checkSorted(value.AsObjectLiteralExpression().Properties.Nodes)
+			case ast.KindIdentifier:
+				// The upstream rule resolves an identifier initialized with an
+				// object. The binder lookup avoids a whole-file text scan here.
+				if decl := declarationObject(ctx.Refs.Resolve(value)); decl != nil {
```
elecmonkey: [P1] Keep resolved initializer nodes in the current source file

With a TypeChecker, `ctx.Refs.Resolve` can return a declaration from another source file. For `global.ts` containing `const sharedProps = { z: P.string, a: P.string }` and `cross-file.ts` containing `C.propTypes = sharedProps`, eslint-plugin-react@7.37.5 reports nothing, while the built Rslint CLI aborts with `invalid range [34,46) for source length 27` because `checkSorted` reports the foreign AST node through the current file's context. Both identifier-resolution paths need a binder-only or same-source guard.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
```diff
@@ -0,0 +1,419 @@
...
+					checkSorted(decl.AsObjectLiteralExpression().Properties.Nodes)
+				}
+			case ast.KindCallExpression:
+				if !isPropWrapperCallByCalleeName(value, wrappers) {
```
elecmonkey, line 175: [P2] Treat optional wrapper calls as opaque

With `propWrapperFunctions: ["wrap"]`, eslint-plugin-react@7.37.5 ignores `C.propTypes = wrap?.({ z: P.string, a: P.string })` because the optional call is wrapped in an ESTree `ChainExpression`, but Rslint reports `propsNotSorted`. ts-go retains it as an optional `CallExpression`, so this path must reject optional-chain calls before matching configured wrappers.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
```diff
@@ -0,0 +1,419 @@
...
+				// wrappers around the member, so only skip parentheses here.
+				left := ast.SkipParentheses(binary.Left)
+				if left == nil {
+					return
+				}
+				isPropTypes := false
+				switch left.Kind {
+				case ast.KindPropertyAccessExpression:
+					name := left.AsPropertyAccessExpression().Name()
+					isPropTypes = name != nil && name.Kind == ast.KindIdentifier && name.AsIdentifier().Text == "propTypes"
```
elecmonkey: [P2] Match private propTypes names

Upstream exposes `PrivateIdentifier.name` without the `#`, so it diagnoses both `class C { static #propTypes = { z: P.string, a: P.string } }` and `this.#propTypes = { z: P.string, a: P.string }`. Rslint emits nothing because the assignment check here and `isAuthoredPropertyName` accept only ordinary identifiers; handle private identifiers, stripping the leading `#`, in both declaration paths.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
```diff
@@ -0,0 +1,419 @@
...
+						// Upstream only examines an inline object in this path; identifier
+						// resolution belongs to assignment and class-property declarations.
+						value := ast.SkipParentheses(assignment.Initializer)
+						if value != nil && value.Kind == ast.KindObjectLiteralExpression {
+							checkSorted(value.AsObjectLiteralExpression().Properties.Nodes)
+						}
+					}
+				}
+			},
+			ast.KindPropertyDeclaration: func(node *ast.Node) {
```
elecmonkey, line 202: [P2] Exclude auto-accessor fields

For `class C { static accessor propTypes = { z: P.string, a: P.string } }`, eslint-plugin-react@7.37.5 is silent because `AccessorProperty` is not one of its `ClassProperty, PropertyDefinition` listeners, while Rslint reports because ts-go represents the declaration with `KindPropertyDeclaration`. Guard out accessor declarations before applying the class-field logic.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
```diff
@@ -0,0 +1,419 @@
...
+
+func isCallback(name string) bool {
+	return len(name) >= 3 && strings.HasPrefix(name, "on") && name[2] >= 'A' && name[2] <= 'Z'
+}
+
+func isShapeCall(callee *ast.Node) bool {
+	if callee == nil {
+		return false
+	}
+	if callee.Kind != ast.KindPropertyAccessExpression {
```
elecmonkey: [P2] Recognize computed shape callees

With `sortShapeProp: true`, upstream reports the inner ordering error in `C.propTypes = { x: P[shape]({ z: P.string, a: P.string }) }` because it reads `callee.property.name` even for a computed member. Rslint requires `PropertyAccessExpression`, so the equivalent `ElementAccessExpression` is missed.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
```diff
@@ -0,0 +1,419 @@
...
+		return propName{}, false
+	}
+	if numeric := numericPropertyName(name); numeric != nil {
+		text := utils.TrimmedNodeText(sourceFile, numeric)
+		if value, ok := ecmascript.StringToNumber(utils.NormalizeNumericLiteral(numeric.AsNumericLiteral().Text)); ok && value != 0 {
+			return propName{text: utils.NormalizeNumericLiteral(numeric.AsNumericLiteral().Text), number: value, isNumber: true}, true
+		}
+		return propName{text: text}, true
+	}
+	if value, ok := utils.GetStaticPropertyName(name); ok {
```
elecmonkey: [P2] Preserve upstream key values before comparing

Converting every static key to a string does not reproduce upstream's `getKey` plus JavaScript relational comparison. Differential runs show Rslint misses upstream diagnostics for `{ 2: P, [true]: P }`, `{ a: P, [`z`]: P }`, and `{ 10n: P, 2n: P }`, while it adds a false positive for `{ 1: P, "": P }`. Preserve the original value type and the authored-token fallback, then apply JavaScript `<` coercion semantics.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
```diff
@@ -0,0 +1,419 @@
...
+	}
+	return param.Type
+}
+
+func isRequired(node *ast.Node) bool {
+	if node == nil || node.Kind != ast.KindPropertyAssignment {
+		return false
+	}
+	value := ast.SkipParentheses(node.AsPropertyAssignment().Initializer)
+	return value != nil && value.Kind == ast.KindPropertyAccessExpression &&
```
elecmonkey: [P2] Match optional and computed required access exactly

The ESTree check reads `node.value.property.name`: it treats `P[isRequired]` as required but leaves `P?.isRequired` opaque behind `ChainExpression`. Rslint does the reverse—missing the computed case and reporting the optional-chain case—because this helper accepts any ts-go `PropertyAccessExpression` and rejects element access. Account for computed identifier access and exclude optional chains.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
```diff
@@ -0,0 +1,419 @@
...
+		}
+	}
+	return false
+}
+
+func checkTypeNode(typeNode *ast.Node, aliases map[string]*ast.Node, check func([]*ast.Node)) {
+	if typeNode == nil {
+		return
+	}
+	switch typeNode.Kind {
```
elecmonkey, line 530: [P2] Unwrap parenthesized type annotations

With `checkTypes: true`, eslint-plugin-react@7.37.5 reports `function C(p: ({ z: string; a: string })) {}` because typescript-eslint exposes the annotation directly as `TSTypeLiteral`; Rslint emits nothing because ts-go preserves a parenthesized type that this switch does not unwrap. Skip parenthesized type nodes before selecting the type-literal/reference branch.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
```diff
@@ -0,0 +1,419 @@
...
+				property := node.AsPropertyDeclaration()
+				if property.Type != nil && isAuthoredPropertyName(property.Name(), "props") {
+					checkValue(property.Initializer)
+					return
+				}
+				if isAuthoredPropertyName(property.Name(), "propTypes") {
+					checkValue(property.Initializer)
+				}
+			},
+			ast.KindBinaryExpression: func(node *ast.Node) {
```
elecmonkey, line 215: [P2] Exclude comma expressions from assignment handling

Ts-go represents a comma expression as `BinaryExpression`, so `C.propTypes, ({ z: P.string, a: P.string })` reaches this listener and Rslint reports the right-hand object. ESTree represents it as `SequenceExpression`, and eslint-plugin-react@7.37.5 reports nothing. Exclude the comma operator before interpreting the binary node as the member-expression parent.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
```diff
@@ -0,0 +1,419 @@
...
+			}
+		}
+	}
+}
+
+func declarationObject(symbol *ast.Symbol) *ast.Node {
+	if symbol == nil || symbol.ValueDeclaration == nil {
+		return nil
+	}
+	decl := symbol.ValueDeclaration
```
elecmonkey, line 547: [P2] Resolve destructured bindings to the declarator initializer

For `const { p } = { z: P.string, a: P.string }; C.propTypes = p`, eslint-plugin-react@7.37.5 obtains the containing `VariableDeclarator` initializer and reports the unsorted object, while Rslint emits nothing because the resolved binding's value declaration is not itself `KindVariableDeclaration`. Walk from a destructured binding declaration to its containing variable declarator before inspecting the initializer.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
swwind: Fixed in b9e1b5d4e. Resolved initializer objects are now used only when their declaration belongs to the current source file, so cross-file bindings remain opaque and cannot produce out-of-range diagnostics. Added a multi-file regression test.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
swwind: Fixed in b9e1b5d4e. Optional-chain calls are now opaque to configured wrapper matching, so wrap?.(...) is ignored as in eslint-plugin-react@7.37.5. Added regression coverage.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
swwind: Fixed in b9e1b5d4e. Class and member propTypes detection now accepts private identifiers with the leading # stripped, covering static #propTypes and this.#propTypes. Added regression coverage.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
swwind: Fixed in b9e1b5d4e. Auto-accessor fields are excluded from the class-property path, matching eslint-plugin-react@7.37.5. Added regression coverage.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
swwind: Fixed in b9e1b5d4e. Shape detection now also handles computed identifier callees such as P[shape](...). Added regression coverage.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
swwind: Fixed in b9e1b5d4e. Key extraction now preserves numeric, boolean, BigInt, and authored-text fallback values before JavaScript relational comparison, including empty-string keys and template literals. Added regression coverage.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
swwind: Fixed in b9e1b5d4e. Required detection now recognizes computed identifier access such as P[isRequired] while keeping optional-chain access such as P?.isRequired opaque. Added regression coverage.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
swwind: Fixed in b9e1b5d4e. Parenthesized type annotations are now unwrapped before selecting the type-literal path. Added regression coverage.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
swwind: Fixed in b9e1b5d4e. Assignment handling now requires the equals operator, so comma expressions are no longer interpreted as propTypes assignments. Added regression coverage.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
swwind: Fixed in b9e1b5d4e. Destructured bindings now resolve through their containing variable declaration before inspecting the initializer. Added regression coverage.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
```diff
@@ -0,0 +1,545 @@
...
+					checkValue(property.Initializer)
+					return
+				}
+				if isAuthoredPropertyName(property.Name(), "propTypes") {
+					checkValue(property.Initializer)
+				}
+			},
+			ast.KindBinaryExpression: func(node *ast.Node) {
+				binary := node.AsBinaryExpression()
+				if binary.OperatorToken == nil || binary.OperatorToken.Kind != ast.KindEqualsToken {
```
elecmonkey: [P2] Upstream still inspects the right-hand side for non-comma binary parents, not only plain assignment. For example, eslint-plugin-react@7.37.5 reports propsNotSorted for C.propTypes += { z: P, a: P } and C.propTypes + ({ z: P, a: P }), while this EqualsToken guard makes Rslint miss both. The guard should exclude the comma operator case that ts-go models as BinaryExpression, rather than every non-= operator.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
```diff
@@ -0,0 +1,545 @@
...
+					if decl := declarationObject(ctx.SourceFile, ctx.Refs.Resolve(firstArg)); decl != nil {
+						checkSorted(decl.AsObjectLiteralExpression().Properties.Nodes)
+					}
+				}
+			}
+		}
+		if opts.checkTypes {
+			listeners[ast.KindTypeAliasDeclaration] = func(node *ast.Node) {
+				decl := node.AsTypeAliasDeclaration()
+				if decl.Name() == nil || decl.Name().Kind != ast.KindIdentifier || decl.Type == nil || decl.Type.Kind != ast.KindTypeLiteral {
```
elecmonkey: [P2] Parenthesized type aliases are still skipped here. With checkTypes enabled, eslint-plugin-react@7.37.5 reports type Props = ({ z: string; a: string }); function C(p: Props) {}, but Rslint never registers the alias because decl.Type is ParenthesizedType instead of TypeLiteral. Please unwrap parenthesized type nodes before deciding whether to store the alias.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
```diff
@@ -0,0 +1,545 @@
...
+		value := new(big.Int)
+		if normalized := utils.NormalizeBigIntLiteral(node.AsBigIntLiteral().Text); normalized != "" {
+			if _, ok := value.SetString(normalized, 10); ok && value.Sign() != 0 {
+				return propName{bigInt: value, kind: propNameBigInt}
+			}
+		}
+		return propName{text: sourceText, kind: propNameString}
+	case ast.KindTrueKeyword:
+		return propName{number: 1, kind: propNameBoolean}
+	default:
```
elecmonkey, line 337: [P2] Computed regular-expression keys need the same canonical value as upstream. In eslint-plugin-react@7.37.5, C.propTypes = { [/a/mi]: P, "/a/j": P } is accepted because the key compares as RegExp.prototype.toString() with canonical flags, but Rslint falls back to the authored /a/mi text and reports propsNotSorted. Please normalize RegularExpressionLiteral keys before comparing.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
```diff
@@ -0,0 +1,545 @@
...
+	previousNumber, previousOK := propNameNumberValue(previous)
+	return currentOK && previousOK && currentNumber < previousNumber
+}
+
+func isBigIntPropNameBefore(current, previous propName) bool {
+	if current.kind == propNameBigInt && previous.kind == propNameBigInt {
+		return current.bigInt.Cmp(previous.bigInt) < 0
+	}
+	if current.kind == propNameBigInt && previous.kind == propNameString {
+		value, ok := new(big.Int).SetString(ecmascript.StringTrim(previous.text), 0)
```
elecmonkey: [P2] BigInt string comparison is using Go base-0 parsing rather than JavaScript's StringToBigInt rules. JavaScript parses "010" as decimal 10, so C.propTypes = { 10n: P, "010": P } is accepted upstream, but Rslint parses it as octal 8 and reports; conversely { 10n: P, "08": P } should report upstream but is missed here. This path needs a JS-compatible string-to-BigInt conversion.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
```diff
@@ -0,0 +1,545 @@
...
+		return false
+	}
+}
+
+func isCallback(name string) bool {
+	return len(name) >= 3 && strings.HasPrefix(name, "on") && name[2] >= 'A' && name[2] <= 'Z'
+}
+
+func isShapeCall(callee *ast.Node) bool {
+	if callee == nil || ast.IsOptionalChain(callee) {
```
elecmonkey: [P2] Optional receiver shape calls are still visible to upstream's shape check. eslint-plugin-react@7.37.5 reports propsNotSorted for P?.shape({ z: P, a: P }) and P?.[shape]({ z: P, a: P }), because it can still read the callee property name, but this optional-chain rejection makes Rslint skip them. Please keep optional-call opacity where upstream has it, but allow optional receivers whose property name is shape.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
```diff
@@ -0,0 +1,545 @@
...
+			}
+		}
+	}
+}
+
+func declarationObject(sourceFile *ast.SourceFile, symbol *ast.Symbol) *ast.Node {
+	if sourceFile == nil || symbol == nil || symbol.ValueDeclaration == nil {
+		return nil
+	}
+	decl := symbol.ValueDeclaration
```
elecmonkey, line 547: [P2] Walking from any resolved declaration up to a VariableDeclaration makes non-binding declarations look like the outer initializer. For const holder = { z: P, method(p) { C.propTypes = p; } }, upstream treats p as the method parameter and reports nothing, while Rslint walks from that parameter to the holder variable and reports the object literal. This parent walk should be limited to destructured binding declarations; other declaration kinds should stay opaque.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
swwind: Fixed in 86163b45c. The assignment listener now ignores only comma binary expressions and preserves upstream RHS inspection for compound and ordinary binary parents. Added regression coverage for += and +.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
swwind: Fixed in 86163b45c. Computed RegExp keys now use the canonical RegExp.prototype.toString() value, including canonical flag ordering. Added regression coverage.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
swwind: Fixed in 86163b45c. StringToBigInt now parses non-prefixed strings as decimal, including leading-zero values, while retaining prefixed-radix and invalid-string behavior. Added regression coverage for "010" and "08".

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
swwind: Fixed in 86163b45c. Optional calls remain opaque, while optional receiver property/element accesses expose the shape member name. Added regression coverage for P?.shape, P?.[shape], and P.shape?.().

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
swwind: Fixed in 86163b45c. Object initializer resolution now walks upward only for BindingElement declarations; parameters and other declaration kinds remain opaque. Added regression coverage.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
swwind: Fixed in 86163b45c. Parenthesized type aliases are now unwrapped before registration, matching typescript-eslint's direct object-type view. Added regression coverage.

internal/plugins/react/rules/sort_prop_types/sort_prop_types.go
```diff
@@ -0,0 +1,569 @@
...
+				if current == nil {
+					continue
+				}
+				if current.Kind == ast.KindSpreadAssignment {
+					// A spread starts an independent ordering group.
+					previous = nil
+					continue
+				}
+				name, ok := declarationName(ctx.SourceFile, current)
+				if !ok {
```
elecmonkey, line 93: [P2] Preserve checkTypes ordering for nameless signatures

When `checkTypes` is enabled, upstream does not treat nameless type members as an ordering break: its `getKey()` fallback still lets these members participate in the reducer. For example, eslint-plugin-react@7.37.5 reports `propsNotSorted` for `type Props = { z: string; (): void; a: string }; function C(p: Props) {}`, and reports `callbackPropsLast` for `type Props = { onChange: () => void; (): void; name: string }; ...` with `callbacksLast`. This early `continue` makes Rslint report zero diagnostics for both cases, so type literals containing call/index/construct signatures can miss ordering violations that upstream flags.

swwind: Fixed in 0b3b946b3. TypeScript type-member sorting now preserves typescript-eslint's source-text fallback for nameless call, index, and construct signatures, while ordinary object-member lists retain the existing boundary behavior. Added regression coverage for sorting and callbacksLast. Focused package tests pass.

elecmonkey (approved):

PR #1986, opened 2026-09-01
feat: rstestPlugin supports rstack-cli

elecmonkey:
## Summary

The Rstest API is also reachable as `rstack/test`. The rules matched `@rstest/core` alone, so those files went unchecked.

### rstest/no-import-node-test

The fix writes a specifier, and an Rstack project depends on `rstack` alone. It now follows the file: only a file naming `rstack/test` and never `@rstest/core` gets the Rstack spelling, read from its imports, `require` calls and `/// <reference types="..." />` directives.

### rstest/consistent-rstest-namespace

The fix rewrites the import specifier in place and leaves the module as authored, so it never repoints an import at the other specifier.

## Related Links

- #1981


internal/plugins/rstest/rules/no_import_node_test/rule.go
```diff
@@ -4,24 +4,152 @@ import (
...
+		case directive.FileName == rstestUtils.RstestImportModule ||
+			strings.HasPrefix(directive.FileName, rstestUtils.RstestImportModule+"/"):
+			return rstestUtils.RstestImportModule
+		case directive.FileName == rstestUtils.RstackTestImportModule ||
+			strings.HasPrefix(directive.FileName, rstestUtils.RstackTestImportModule+"/"):
+			sawRstack = true
+		}
+	}
+
+	for _, statement := range sourceFile.Statements.Nodes {
```
swwind: [P1] Cover all static `require()` calls when selecting the replacement module

`replacementModuleForFile` only inspects `sourceFile.Statements.Nodes`, and `requiredModuleSpecifier` is reached only for a variable declaration whose entire initializer is `require(...)`. It therefore misses valid uses such as `function setup() { require('rstack/test'); }` and `const expect = require('rstack/test').expect`. In an Rstack-only project, a file that also imports `node:test` is autofixed to `@rstest/core`, which may not resolve. Please traverse the file for static `require()` calls (or otherwise include these forms) and add a regression test.

elecmonkey: Fixed in 8f3133b2e. `replacementModuleForFile` now walks the whole file instead of only `sourceFile.Statements.Nodes`, and a new `referencedModuleSpecifier` recognizes the specifier in an import declaration, an `import x = require('m')`, and any `require('m')` call wherever it appears — so `function setup() { require('rstack/test'); }` and `const expect = require('rstack/test').expect` now select `rstack/test`. Added regression tests for both forms, plus one where a nested `require('@rstest/core')` still wins over a top-level `rstack/test` import.

swwind (approved):

PR #1987, opened 2026-09-01
fix: handle async imports from missing project outputs

fansenze:
## Summary

- Align `@typescript-eslint/require-await` with upstream when an imported project-reference source is redirected to a missing declaration output and tsgo consequently exposes the call as `any`.
- Recover only direct, unannotated async named exports from unloaded TypeScript sources, while preserving diagnostics for explicit `any`, overloaded or redeclared bindings, synchronous exports, async generators, local values, and shadowed imports.
- Add a missing-output project-reference fixture covering concise returns, block returns, async-generator yields, renamed imports, async variable exports, async function declarations, and adversarial controls including an `any` overload and both variable-redeclaration orders.
- Verify the reported TSX file against the final binary with no diagnostics; targeted Go and JS rule tests pass with no snapshot changes. Benchmarks show unchanged allocations and no listener-time regression, so no benchmark file or performance-only change is included.

## Related Links

- [typescript-eslint require-await rule](https://typescript-eslint.io/rules/require-await/)
- [Latest upstream implementation](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/rules/require-await.ts)

## Checklist

### DIFF
--- a/internal/plugins/typescript/rules/require_await/require_await.go
+++ b/internal/plugins/typescript/rules/require_await/require_await.go
@@ -6,6 +6,8 @@ import (
 	"github.com/microsoft/typescript-go/shim/ast"
 	"github.com/microsoft/typescript-go/shim/checker"
 	"github.com/microsoft/typescript-go/shim/core"
+	"github.com/microsoft/typescript-go/shim/parser"
+	"github.com/microsoft/typescript-go/shim/tspath"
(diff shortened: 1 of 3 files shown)


fansenze:
Closing because the reported difference came from mismatched typescript-eslint parser/project settings in the comparison. With equivalent type-aware project configuration, current rslint already matches upstream behavior, so no rule change is needed.

PR #1991, opened 2026-09-01
feat: port rule react/no-invalid-html-attribute

swwind:
## Summary

- port `react/no-invalid-html-attribute` from eslint-plugin-react v7.37.5
- register the rule and add its schema and documentation
- migrate the complete upstream fixture suite as explicit Go test cases
- add rslint-specific coverage for supported syntax and suggestions

## Tests

- go test -count=1 ./internal/plugins/react/rules/no_invalid_html_attribute


elecmonkey:
cspell fail

internal/plugins/react/rules/no_invalid_html_attribute/no_invalid_html_attribute.go
```diff
@@ -0,0 +1,373 @@
...
+			}
+			if len(call.Arguments.Nodes) < 2 {
+				return
+			}
+			props := ast.SkipParentheses(call.Arguments.Nodes[1])
+			if props.Kind != ast.KindObjectLiteralExpression {
+				return
+			}
+			for _, property := range props.AsObjectLiteralExpression().Properties.Nodes {
+				if property.Kind != ast.KindPropertyAssignment && property.Kind != ast.KindMethodDeclaration {
```
elecmonkey: [P2] Preserve upstream reports before skipping shorthand props

In `React.createElement` props, upstream `checkCreateProps` checks the `rel` key and reports `onlyMeaningfulFor` for invalid target elements before it skips shorthand properties. This port drops shorthand object members at this kind filter, so `React.createElement("html", { rel })` reports 0 diagnostics in Rslint while eslint-plugin-react@7.37.5 reports `onlyMeaningfulFor` at `rel`. That is an under-reporting parity bug for shorthand `rel` props on unsupported tags.

internal/plugins/react/rules/no_invalid_html_attribute/no_invalid_html_attribute_upstream_test.go
```diff
@@ -0,0 +1,294 @@
...
+		{Code: `React.createElement("a", { rel: ["alternate alternatex"] })`, Tsx: true, Errors: []rule_tester.InvalidTestCaseError{{MessageId: "neverValid", Suggestions: []rule_tester.InvalidTestCaseSuggestion{{MessageId: "suggestRemoveInvalid", Output: `React.createElement("a", { rel: [""] })`}}}}},
+		{Code: `<html rel></html>`, Tsx: true, Errors: []rule_tester.InvalidTestCaseError{{MessageId: "onlyMeaningfulFor", Suggestions: []rule_tester.InvalidTestCaseSuggestion{{MessageId: "suggestRemoveDefault", Output: `<html ></html>`}}}}},
+		{Code: `React.createElement("html", { rel: 1 })`, Tsx: true, Errors: []rule_tester.InvalidTestCaseError{{MessageId: "onlyMeaningfulFor"}}},
+		{Code: `<a rel></a>`, Tsx: true, Errors: []rule_tester.InvalidTestCaseError{{MessageId: "emptyIsMeaningless", Suggestions: []rule_tester.InvalidTestCaseSuggestion{{MessageId: "suggestRemoveEmpty", Output: `<a ></a>`}}}}},
+		// ESLint 8 accepts this upstream suggestion although its output is not
```
elecmonkey: [P3] Re-enable the migrated numeric suggestion assertion

This upstream invalid fixture is skipped with a comment saying Rslint's test runner reparses suggestions, but `RunRuleTester` applies suggestion edits as text and compares the output without reparsing. I temporarily unskipped the equivalent case and `go test -run TestReviewNumberSuggestionParity` passed with the expected `React.createElement("a", { rel:  })` output, so the current `Skip: true` removes one upstream fixture from the claimed complete migration and can let this behavior regress unnoticed.

internal/plugins/react/rules/no_invalid_html_attribute/no_invalid_html_attribute.go
```diff
@@ -0,0 +1,373 @@
...
+
+func sourceText(ctx rule.RuleContext, node *ast.Node) string {
+	if node == nil || ctx.SourceFile == nil {
+		return ""
+	}
+	r := utils.TrimNodeTextRange(ctx.SourceFile, node)
+	return ctx.SourceFile.Text()[r.Pos():r.End()]
+}
+
+func removeStaticValue(ctx rule.RuleContext, node *ast.Node, value string) rule.RuleFix {
```
elecmonkey, line 257: [P3] Mirror regexp literal suggestion replacement

For `React.createElement("a", { rel: /invalid/ })`, upstream reports `neverValid` with a `suggestRemoveInvalid` replacement of `//` because `value.raw.replace(value.value, '')` treats the RegExp object as a regular expression. Here `text` comes from `GetStaticExpressionValue`, but `strings.Replace` treats `/invalid/` as a literal string and removes the whole token, yielding `rel:  ` instead of upstream's `rel: //`. Since the port promises fix/suggestion parity, regexp-literal suggestions should preserve this source-text behavior.

internal/plugins/react/rules/no_invalid_html_attribute/no_invalid_html_attribute.go
```diff
@@ -0,0 +1,373 @@
...
+				}
+				var key, value *ast.Node
+				method := property.Kind == ast.KindMethodDeclaration
+				if method {
+					key = property.AsMethodDeclaration().Name()
+				} else {
+					p := property.AsPropertyAssignment()
+					key, value = p.Name(), p.Initializer
+				}
+				if key == nil || key.Kind != ast.KindIdentifier || key.AsIdentifier().Text != "rel" {
```
elecmonkey, line 445: [P2] Do not discard computed rel keys before upstream checks

Upstream still observes computed property keys like `[rel]` as an Identifier key and only skips computed props after reporting unsupported target elements and method props. This gate requires the ts-go key node itself to be `Identifier`, so `React.createElement("html", { [rel]: 1 })` and `React.createElement("a", { [rel](){ return 1; } })` produce no Rslint diagnostics, while eslint-plugin-react@7.37.5 reports `onlyMeaningfulFor` and `noMethod` respectively. Computed `rel` keys need to reach those earlier checks before value validation is skipped.

internal/plugins/react/rules/no_invalid_html_attribute/no_invalid_html_attribute.go
swwind: Fixed. Before skipping shorthand values, rslint now checks unsupported elements first and reports onlyMeaningfulFor, matching upstream. Added a regression test for React.createElement("html", { rel }).

internal/plugins/react/rules/no_invalid_html_attribute/no_invalid_html_attribute.go
swwind: Fixed. Computed keys are now unwrapped to their inner identifier before the onlyMeaningfulFor/noMethod checks, while value validation is still skipped afterward. Added regression tests for [rel] and [rel]().

internal/plugins/react/rules/no_invalid_html_attribute/no_invalid_html_attribute_upstream_test.go
swwind: We deliberately diverge from upstream here. Numeric createElement rel values are now skipped before string lookup and suggestion generation, and the upstream fixture remains explicitly skipped. Added regression coverage to ensure no malformed rel:  suggestion is produced.

internal/plugins/react/rules/no_invalid_html_attribute/no_invalid_html_attribute.go
swwind: We deliberately diverge from upstream here. RegExp literals are not treated as strings, so rslint reports no neverValid diagnostic and offers no suggestion for this case. Added regression coverage to ensure no malformed rel: // suggestion is produced.

swwind:
Fixed the cspell failure: added `reparses` and `batgo` to `scripts/dictionary.txt`. Local `pnpm run check-spell` passes.

internal/plugins/react/rules/no_invalid_html_attribute/no_invalid_html_attribute.go
```diff
@@ -0,0 +1,384 @@
...
+				return
+			}
+			init := attr.Initializer
+			if init == nil {
+				ctx.ReportNodeWithDeferredSuggestions(name, message("emptyIsMeaningless", "An empty “rel” attribute is meaningless.", map[string]string{"attributeName": "rel"}), func() []rule.RuleSuggestion {
+					return suggestion("suggestRemoveEmpty", "remove empty attribute rel", rule.RuleFixRemove(ctx.SourceFile, node))
+				})
+				return
+			}
+			if init.Kind == ast.KindStringLiteral {
```
elecmonkey, line 376: [P2] Decode entities in direct JSX attribute values

ts-go preserves entities in direct JSX string attributes, whereas ESTree exposes their decoded value. Consequently, `<a rel="no&#x6f;pener" />` reports `neverValid` in Rslint even though eslint-plugin-react@7.37.5 decodes it to `noopener` and reports nothing. Decode direct JSX initializers before validation while leaving expression-container strings unchanged.

internal/plugins/react/rules/no_invalid_html_attribute/no_invalid_html_attribute.go
```diff
@@ -0,0 +1,384 @@
...
+				case ast.KindMethodDeclaration:
+					method = true
+					key = property.AsMethodDeclaration().Name()
+				case ast.KindPropertyAssignment:
+					p := property.AsPropertyAssignment()
+					key, value = p.Name(), p.Initializer
+				case ast.KindShorthandPropertyAssignment:
+					shorthand = true
+					key = property.AsShorthandPropertyAssignment().Name()
+				default:
```
elecmonkey, line 438: [P2] Handle accessor properties before skipping object members

`get rel()` and `set rel()` are represented by ts-go as `KindGetAccessor` and `KindSetAccessor`, so they fall through this switch. For `React.createElement("html", { get rel() {} })`, upstream reports `onlyMeaningfulFor` on `rel`, while Rslint emits no diagnostic. Include accessor declarations so unsupported-element checks run before their non-method values are skipped.

internal/plugins/react/rules/no_invalid_html_attribute/no_invalid_html_attribute.go
```diff
@@ -0,0 +1,384 @@
...
+			}
+			if expr.Kind == ast.KindObjectLiteralExpression || (expr.Kind == ast.KindIdentifier && expr.AsIdentifier().Text == "undefined") {
+				ctx.ReportNodeWithDeferredSuggestions(init, message("onlyStrings", "“rel” attribute only supports strings.", map[string]string{"attributeName": "rel"}), func() []rule.RuleSuggestion {
+					return suggestion("suggestRemoveDefault", "remove rel", rule.RuleFixRemove(ctx.SourceFile, node))
+				})
+			}
+		}
+		checkCall := func(node *ast.Node) {
+			call := node.AsCallExpression()
+			if !reactutil.IsCreateElementCall(call.Expression, reactutil.DefaultReactPragma) || call.Arguments == nil || len(call.Arguments.Nodes) == 0 {
```
elecmonkey: [P2] Recognize computed identifier createElement calls

The upstream predicate accepts a computed member whose property is an identifier because `node.callee.property.name` is still `createElement`. With `const createElement = "createElement"; React[createElement]("a", { rel: "bad" })`, eslint-plugin-react@7.37.5 reports `neverValid`, but this helper rejects the element-access callee and Rslint reports nothing. Preserve that computed-identifier form without also accepting `React["createElement"]`, which upstream skips.

internal/plugins/react/rules/no_invalid_html_attribute/no_invalid_html_attribute.go
```diff
@@ -0,0 +1,384 @@
...
+					data["elementName"] = element
+				}
+				if id != "" {
+					partRange := literalRange(ctx, valueNode, part)
+					ctx.ReportNodeWithDeferredSuggestions(valueNode, message(id, desc, data), func() []rule.RuleSuggestion {
+						return suggestion("suggestRemoveInvalid", "“remove invalid attribute "+word+"”", rule.RuleFixRemoveRange(partRange))
+					})
+				}
+			}
+			parts := stringParts(value)
```
elecmonkey: [P2] Preserve the upstream pair matcher at internal word boundaries

The upstream lookahead `/(?=(\b\S+\s*\S+))/g` can begin at a word boundary inside a non-whitespace token, but this loop only examines complete tokens equal to `shortcut`. Thus `<link rel=".shortcut" />` produces both `neverValid` and `notAlone` upstream but only `neverValid` in Rslint; `<link rel="-shortcut foo" />` similarly loses `notPaired`.

internal/plugins/react/rules/no_invalid_html_attribute/no_invalid_html_attribute.go
```diff
@@ -0,0 +1,384 @@
...
+		}
+		reportLiteral := func(valueNode, removeNode *ast.Node, element string) {
+			value, isString := literalString(valueNode)
+			if !isString {
+				ctx.ReportNodeWithDeferredSuggestions(valueNode, message("onlyStrings", "“rel” attribute only supports strings.", map[string]string{"attributeName": "rel"}), func() []rule.RuleSuggestion {
+					return suggestion("suggestRemoveNonString", "remove non-string value in “rel”", rule.RuleFixRemove(ctx.SourceFile, removeNode))
+				})
+				return
+			}
+			if ecmascript.IsBlank(value) {
```
elecmonkey, line 310: [P2] Preserve the empty expression-string suggestion range

For `<a rel={""} />`, upstream removes the literal's parent expression container and suggests `<a rel= />`; passing the outer attribute as `removeNode` makes Rslint remove the entire attribute and suggest `<a  />`. Direct `rel=""` should still remove the attribute, but expression-container literals need the upstream parent range unless this additional divergence is explicitly accepted.

internal/plugins/react/rules/no_invalid_html_attribute/no_invalid_html_attribute.go
```diff
@@ -0,0 +1,384 @@
...
+		}
+	}
+	return false
+}
+
+func message(id, description string, data map[string]string) rule.RuleMessage {
+	return rule.RuleMessage{Id: id, Description: description, Data: data}
+}
+
+func suggestion(id, description string, fix rule.RuleFix) []rule.RuleSuggestion {
```
elecmonkey, line 60: [P2] Populate structured data on suggestion messages

Every suggestion created through this helper has `Message.Data == nil`, although the public API exposes suggestion data and upstream supplies it. For example, the suggestion for `<a rel="bad" />` contains `{ attributeName: "rel", reportingValue: "bad" }` upstream but no data in Rslint; all 78 suggestion-bearing migrated cases exhibit this metadata loss.

internal/plugins/react/rules/no_invalid_html_attribute/no_invalid_html_attribute.go
```diff
@@ -0,0 +1,384 @@
...
+			tag := reactutil.GetJsxTagName(parent)
+			if tag == nil || tag.Kind != ast.KindIdentifier {
+				return
+			}
+			element := tag.AsIdentifier().Text
+			if _, ok := htmlElements[element]; !ok {
+				return
+			}
+			if _, ok := relElements[element]; !ok {
+				ctx.ReportNodeWithDeferredSuggestions(name, message("onlyMeaningfulFor", "The ”rel“ attribute only has meaning on the tags: \"<link>\", \"<a>\", \"<area>\", \"<form>\"", map[string]string{"attributeName": "rel", "tagNames": "\"<link>\", \"<a>\", \"<area>\", \"<form>\""}), func() []rule.RuleSuggestion {
```
elecmonkey: [P3] Keep the quoted default and empty suggestion descriptions

The upstream message templates for `suggestRemoveDefault` and `suggestRemoveEmpty` include literal double quotes, so their displayed descriptions are `"remove rel"` and `"remove empty attribute rel"`. These call sites omit those quotes, producing observably different suggestion text even though the IDs and edits match.

internal/plugins/react/rules/no_invalid_html_attribute/no_invalid_html_attribute.go
swwind: Fixed in 002e1aaa5. Direct JSX string attribute values are now decoded with JSX entity semantics before validation, while expression-container strings remain unchanged. Added regression coverage for <a rel="no&#x6f;pener" />.

internal/plugins/react/rules/no_invalid_html_attribute/no_invalid_html_attribute.go
swwind: Fixed in 002e1aaa5. GetAccessorDeclaration and SetAccessorDeclaration are now included in the pre-validation key checks without being classified as ordinary methods. Added getter and setter regression coverage.

internal/plugins/react/rules/no_invalid_html_attribute/no_invalid_html_attribute.go
swwind: Fixed in 002e1aaa5. The rule now accepts React[createElement] while continuing to reject React["createElement"], matching the upstream predicate. Added regression coverage.

internal/plugins/react/rules/no_invalid_html_attribute/no_invalid_html_attribute.go
swwind: Fixed in 002e1aaa5. The shortcut pair matcher now preserves upstream word-boundary behavior, including .shortcut and -shortcut foo cases.

internal/plugins/react/rules/no_invalid_html_attribute/no_invalid_html_attribute.go
swwind: Fixed in 002e1aaa5. Empty string literals in JSX expression containers now remove the expression container, while direct JSX attributes keep the direct-attribute removal behavior.

internal/plugins/react/rules/no_invalid_html_attribute/no_invalid_html_attribute.go
swwind: Fixed in 002e1aaa5. Suggestion messages now carry the same structured data as their diagnostics, including attributeName and reportingValue where applicable.

internal/plugins/react/rules/no_invalid_html_attribute/no_invalid_html_attribute.go
swwind: Fixed in 002e1aaa5. The default and empty suggestion descriptions now preserve the literal double quotes from upstream.

internal/plugins/react/rules/no_invalid_html_attribute/no_invalid_html_attribute.go
```diff
@@ -0,0 +1,464 @@
...
+
+func removeStaticValue(ctx rule.RuleContext, node *ast.Node, value string) rule.RuleFix {
+	raw := sourceText(ctx, node)
+	raw = strings.Replace(raw, value, "", 1)
+	return rule.RuleFixReplace(ctx.SourceFile, node, raw)
+}
+
+func literalRange(ctx rule.RuleContext, node *ast.Node, offset [2]int) core.TextRange {
+	r := utils.TrimNodeTextRange(ctx.SourceFile, node)
+	return core.NewTextRange(r.Pos()+1+offset[0], r.Pos()+1+offset[1])
```
elecmonkey: [P2] Keep decoded JSX offsets in raw-source coordinates

After `jsxLiteralString` decodes entities, `part` contains byte offsets into the decoded value, but `literalRange` adds them to raw token offsets. For `<a rel="noopener&#x20;"/>`, upstream reports nothing while Rslint emits `spaceDelimited` and suggests `<a rel="noopener#x20;"/>`; `<a rel="&copy; bad"/>` similarly suggests `<a rel="opy; bad"/>` instead of upstream's `<a rel="copy; bad"/>`. Please keep a decoded-to-raw/UTF-16 range map before building suggestions.

internal/utils/ecmascript/jsx_text.go
```diff
@@ -1,11 +1,41 @@
...
+func DecodeJSXEntities(raw string) string {
+	var decoded strings.Builder
+	for pos := 0; pos < len(raw); {
+		if raw[pos] == '&' {
+			first, second, count, next, ok := decodeJSXTextEntity(raw, pos)
+			if ok {
+				if count == 2 {
+					decoded.WriteRune(utf16.DecodeRune(rune(first), rune(second)))
+				} else {
+					decoded.WriteRune(rune(first))
```
elecmonkey: [P2] Preserve adjacent surrogate entity code units

`DecodeJSXEntities` writes each numeric entity through `WriteRune`, so a lone surrogate is converted to U+FFFD before the following surrogate can combine with it. As a result `<a rel="&#xD83D;&#xDE00;"/>` is reported as `��`, while Acorn/ESLint decodes the adjacent UTF-16 units as `😀`. Accumulate decoded UTF-16 units and convert them together.

internal/plugins/react/rules/no_invalid_html_attribute/no_invalid_html_attribute.go
```diff
@@ -0,0 +1,464 @@
...
+		if !configured(options) {
+			return rule.RuleListeners{}
+		}
+		reportLiteral := func(valueNode, nonStringRemoveNode, emptyRemoveNode *ast.Node, element string, decodeJSX bool) {
+			value, isString := literalString(valueNode)
+			if decodeJSX {
+				value, isString = jsxLiteralString(ctx, valueNode)
+			}
+			if !isString {
+				data := map[string]string{"attributeName": "rel"}
```
elecmonkey: [P3] Preserve reportingValue for non-string JSX values

This branch drops upstream's `reportingValue` from diagnostic and suggestion data. For `<a rel={true}/>`, eslint-plugin-react@7.37.5 exposes `{ attributeName: "rel", reportingValue: true }` on `suggestRemoveNonString`, while Rslint exposes only `attributeName`; numeric and null literals have the same omission. Please preserve the literal value in the public suggestion data, extending the data representation if needed.

internal/plugins/react/rules/no_invalid_html_attribute/no_invalid_html_attribute_upstream_test.go
```diff
@@ -0,0 +1,293 @@
...
+		{Code: `<link rel="apple-touch-icon" sizes="60x60" href="apple-touch-icon-60x60.png" />`, Tsx: true},
+		{Code: `<link rel="apple-touch-icon" sizes="76x76" href="apple-touch-icon-76x76.png" />`, Tsx: true},
+		{Code: `<link rel="apple-touch-icon" sizes="120x120" href="apple-touch-icon-120x120.png" />`, Tsx: true},
+		{Code: `<link rel="apple-touch-icon" sizes="152x152" href="apple-touch-icon-152x152.png" />`, Tsx: true},
+		{Code: `<link rel="apple-touch-icon" sizes="180x180" href="apple-touch-icon-180x180.png" />`, Tsx: true},
+		{Code: `<link rel="apple-touch-startup-image" href="launch.png" />`, Tsx: true},
+		{Code: `<link rel="apple-touch-startup-image" href="iphone5.png" media="(device-width: 320px) and (device-height: 568px) and (-webkit-device-pixel-ratio: 2)" />`, Tsx: true},
+		{Code: `<link rel="mask-icon" href="/safari-pinned-tab.svg" color="#fff" />`, Tsx: true},
+	}, []rule_tester.InvalidTestCase{
+		{Code: `<a rel="alternatex"></a>`, Tsx: true, Errors: []rule_tester.InvalidTestCaseError{{MessageId: "neverValid", Suggestions: []rule_tester.InvalidTestCaseSuggestion{{MessageId: "suggestRemoveInvalid", Output: `<a rel=""></a>`}}}}},
```
elecmonkey: [P2] Preserve upstream assertions in migrated cases

The upstream inputs are present, but most migrated invalid cases assert only `messageId` and suggestion output, dropping upstream diagnostic data, suggestion data, node type, and explicit positions. That lets `<a rel={true}>` pass despite the missing `reportingValue` above. Please keep each applicable upstream assertion or add a specific per-case explanation where ts-go cannot represent it.

elecmonkey (commented):
Inline review findings for react/no-invalid-html-attribute parity.

internal/plugins/react/rules/no_invalid_html_attribute/no_invalid_html_attribute.go
swwind: Fixed in a8116ad96. Direct JSX values now keep UTF-16 offsets separate from decoded Go byte spans and map report/fix ranges back to the raw literal. Added coverage for entity-expanded trailing whitespace and entity offsets.

internal/utils/ecmascript/jsx_text.go
swwind: Fixed in a8116ad96. JSX entity decoding now accumulates UTF-16 code units before converting to UTF-8, so adjacent surrogate entities decode to the same astral character as Acorn/ESLint. Added decoder regression coverage.

internal/plugins/react/rules/no_invalid_html_attribute/no_invalid_html_attribute.go
swwind: Fixed in a8116ad96. Non-string JSX literal diagnostics and suggestions now carry the upstream-compatible reportingValue metadata for boolean true, while preserving the upstream null/numeric cases. Added data assertions and coverage.

internal/plugins/react/rules/no_invalid_html_attribute/no_invalid_html_attribute_upstream_test.go
swwind: Fixed in a8116ad96. The rule tester now supports diagnostic and suggestion data assertions, and the migrated cases now assert the affected metadata and ranges. ESTree node type assertions remain documented as intentionally omitted because ts-go node kinds do not map one-to-one to ESTree types.

internal/plugins/react/rules/no_invalid_html_attribute/no_invalid_html_attribute.go
```diff
@@ -0,0 +1,516 @@
...
```
elecmonkey: [P3] Preserve data for every non-string JSX literal

This branch only adds `reportingValue` for `true`, but upstream `eslint-plugin-react@7.37.5` passes `reportingValue: node.value` for every non-string JSX literal. I verified with upstream RuleTester that `<a rel={false}>`, `<a rel={5}>`, and `<a rel={null}>` all accept diagnostic and `suggestRemoveNonString` data containing their literal values. Rslint currently exposes only `attributeName` for those cases, and the migrated tests now lock in that mismatch. Please carry non-string JSON values in the diagnostic/suggestion data instead of special-casing `true` as a string.

swwind: This is intentional. Native Go rules use pre-rendered descriptions and do not expose ESLint `context.report()` data metadata. I reverted the rule-specific `Data` plumbing and related RuleTester assertions. The existing createElement non-string divergence remains unchanged.

elecmonkey (approved):

PR #1994, opened 2026-09-01
feat(website): share playground config in the URL

elecmonkey:
## Summary

The playground's code, `rslint.config.js`, `tsconfig.json` and the selected Rslint version are encoded into the URL, so a shared link reproduces the whole setup.

## Encoding

A URL looks like `/playground#1AbCd…`. The leading `1` is the format version — reserved, in case the dictionary ever changes, see below — and the rest is base64url. Decoding it yields a deflate-raw stream, which inflates to a frame:

```
[flags, 1 byte]   bit0=code bit1=config bit2=tsconfig bit3=version
[version]         major/minor/patch as three LEB128 numbers, present only when bit3 is set
[text sections]*  in bit0→bit2 order; every section but the last is prefixed with its LEB128 byte length
```

Whatever an editor holds that equals its default stays out of the frame; the version is always written. Anything the frame does not carry is shown as its default, the version as the newest release.

## Dictionary

The defaults of `rslint.config.js` and `tsconfig.json` serve as the preset dictionary, which keeps a URL short when the user has made a small edit to a default. Since the defaults are unlikely to change much, the dictionary can be frozen from here on, so that URLs already produced never stop working.

## Fallback

A URL that does not parse as this format is tried against the older format, so links produced before this change keep working.

## Dependency

`fflate` is added for compression and decompression. It contributes about 8.2 KB raw, 4.5 KB gzipped, to the site build.

## Running a config safely

A config is arbitrary JavaScript and this proposal puts one in a link, so the playground no longer evaluates it in the page origin. It runs inside an iframe sandboxed without `allow-same-origin`: the frame's origin is opaque, which puts the page's DOM, storage, cookies and credentialed same-origin requests out of reach, and the source reaches the frame by `postMessage` rather than being inlined into its markup, so a `</script` sequence in a config cannot break out either. The default export comes back JSON-serialized and is validated in the page.

The frame also carries a Content-Security-Policy that denies everything except script and connections to the package CDNs a config is meant to import from. Without it a config could still fetch, beacon or dynamically import from a host of its choosing, and a link that looked like it pointed at the docs would quietly hand the reader's address to whoever crafted it. Allowing the CDNs costs nothing, because an attacker cannot read their logs.

Evaluation is given twenty seconds, after which the frame is removed, so a config that never finishes reports an error instead of hanging the run; the frame has its own thread, so it cannot block the page while the deadline runs.

Decompression is bounded at both ends. DEFLATE cannot expand its input by more than about 1032:1, so the compressed payload is capped at 64 KiB, which caps the work an inflate can be made to perform; the decompressed size is capped at 1 MiB separately, because fflate keeps inflating past a full output buffer and discards the overflow rather than stopping.

## Example

An Rstest plugin config: https://feat-playground-config-url.rslint.pages.dev/playground/?tab=lint#142bgZFCAB2xJanEJLCCV8vJTUq1AIkp4A15HoagYpCogpzQ9M2-gogHZDbgMGuJRBQA


swwind:
I would suggest let's using some compress algorithm to compress the [code, config, tsconfig] triple to shrink the size, and maybe just leave empty when the value is untouched.

FYI, the default config is ~391 bytes, after gzip ~236 bytes, after gzip + base64 ~312 bytes.

elecmonkey:
> I would suggest let's using some compress algorithm to compress the [code, config, tsconfig] triple to shrink the size, and maybe just leave empty when the value is untouched.
> 
> FYI, the default config is ~391 bytes, after gzip ~236 bytes, after gzip + base64 ~312 bytes.

The default is already left empty. I think compression is definitely something i should do.

website/theme/components/Playground/share-url.ts
```diff
@@ -0,0 +1,273 @@
...
+  ).filter(([flag]) => flags & flag);
+  keys.forEach(([, key], index) => {
+    if (index === keys.length - 1) {
+      state[key] = decoder.decode(frame.subarray(cursor.offset));
+      cursor.offset = frame.length;
+      return;
+    }
+    const length = readVarint(frame, cursor);
+    if (cursor.offset + length > frame.length) throw new RangeError();
+    state[key] = decoder.decode(
```
swwind, line 193: This lets an attacker-controlled `rslintConfig` from the URL flow into `Client.runLint`, where `evaluateConfig` dynamically imports it from a Blob and executes it in the page origin. Opening a crafted share link therefore runs arbitrary JavaScript automatically. Please do not execute configuration source loaded from the URL, or isolate it in a sandbox with no page privileges.

website/theme/components/Playground/share-url.ts
```diff
@@ -0,0 +1,273 @@
...
+
+/**
+ * Decodes a share fragment. A truncated, corrupt or future-version link decodes
+ * to `null` so the caller can fall back to the defaults instead of failing.
+ */
+export function decodeShareState(payload: string): ShareState | null {
+  if (payload[0] !== SHARE_VERSION) return null;
+  try {
+    return decodeFrame(
+      inflateSync(fromBase64Url(payload.slice(1)), {
```
swwind: The compressed payload is fully controlled by the URL, but `inflateSync` has no output limit here. A small DEFLATE bomb can therefore block the main thread or allocate excessive memory before `decodeFrame` gets a chance to validate anything. Please enforce a reasonable maximum decompressed size during inflation.

website/theme/components/Playground/EditorTabs.tsx
```diff
@@ -148,45 +132,32 @@ export const EditorTabs = ({
...
+        tsconfig:
+          tsconfigEditorRef.current?.getValue() ?? initialState.tsconfig,
+        wasmVersion: wasmVersionRef.current,
+      });
     }, 300);
   }
 
+  useEffect(() => {
+    wasmVersionRef.current = wasmVersion;
+    scheduleSerializeToUrl();
```
swwind, line 165: This only schedules the URL update for 300 ms later. If the user changes the wasm version or config and immediately clicks Share, `ResultPanel.copyShareUrl` reads `window.location.href` before this callback runs, so the copied link contains stale state. Please flush the pending serialization before copying, or generate the share URL directly from the current editor values and selected version.

swwind (commented):
Please address these security and correctness issues before merging.

website/theme/components/Playground/share-url.ts
elecmonkey: Fixed in 397fc1a9c. `evaluateConfig` no longer imports the config in the page origin. It runs inside an `<iframe sandbox="allow-scripts">` — no `allow-same-origin`, so the frame has an opaque origin — and the source reaches it by `postMessage` rather than being inlined into the markup, so a `</script` sequence in a config cannot break out either. The default export comes back JSON-serialized and is validated in the page.

Verified in Chrome against crafted configs: reading `localStorage`, `document.cookie`, a parent global and `parent.document` all fail with `SecurityError`, and a credentialed same-origin `fetch` fails outright.

The frame also gets a five-second deadline and is removed when it expires, so a config that never finishes reports an error instead of hanging the run. It has its own thread, so it cannot block the page while the deadline runs.

website/theme/components/Playground/share-url.ts
elecmonkey: Fixed in 18cb083c1. Both ends of the decompression are bounded now.

DEFLATE cannot expand its input by more than about 1032:1, so the compressed payload is capped at 64 KiB, which caps the work an inflate can be made to perform. That alone does not bound memory — fflate keeps inflating past a full output buffer and discards the overflow, so a bounded `out` costs more time, not less — so the decompressed size is capped at 1 MiB separately. The output buffer is sized to whatever the payload could actually expand into and one byte past the cap, which is what makes hitting the cap distinguishable from landing exactly on it, since fflate reports the real length when the output fits and the buffer's length when it had to truncate.

Measured: a 200 MB bomb is refused in 1ms without inflating at all, and the worst payload that still gets as far as inflating costs 197ms and a megabyte. Neither cap is reachable by hand — source compresses about 4:1, so the compressed cap sits around 270 KiB of editor text, an 87,000-character fragment — and `encodeShareState` refuses oversized state rather than minting a link it would then reject.

website/theme/components/Playground/EditorTabs.tsx
elecmonkey: Fixed in 361460da1. `EditorTabsRef` gained a `flushShareUrl` that cancels the pending write and performs it immediately, and `ResultPanel` calls it before reading `window.location.href`. `writeShareState` replaces the history entry synchronously, so by the time the click is handled the address bar and the clipboard agree.

Reading the URL rather than composing a second one from the editor values keeps a single source of truth: what gets copied is exactly what a reader would get by copying out of the address bar.

fansenze (approved):

PR #1998, opened 2026-09-01
feat: port rule unicorn/consistent-tuple-labels

shiyudesu:
## Summary

Port `unicorn/consistent-tuple-labels` from
`eslint-plugin-unicorn@v74.0.0` to rslint.

The rule reports unlabeled tuple elements when any element in the same tuple
has a label:

```ts
// Incorrect
type Point = [x: number, number];

// Correct
type Point = [x: number, y: number];
type Coordinates = [number, number];
```

There is no autofix because meaningful tuple labels cannot be inferred.

## Implementation

- Handles ordinary, optional, rest, readonly, and nested tuple forms.
- Reports every unlabeled element in a partially labeled tuple.
- Preserves independent diagnostics for nested tuples.
- Registers the rule in the Unicorn catalog and enables it in the Unicorn
  recommended preset.

## Tests

- Migrated the complete upstream v74.0.0 suite: 17 valid and 13 invalid cases.
- Added Go coverage for tsgo tuple node shapes, nested tuples, parenthesized and
  union element types, declaration containers, and exact diagnostic ranges.
- Added the JS end-to-end RuleTester suite.
- Compared message IDs, messages, and report ranges with the upstream v74.0.0
  snapshots.

Verification completed:

- `go test -count=1 ./internal/plugins/unicorn ./internal/plugins/unicorn/rules/consistent_tuple_labels`
- `pnpm build`
- `pnpm --dir packages/rslint-test-tools exec rs test --testTimeout=10000 consistent-tuple-labels`
- `pnpm typecheck`
- `pnpm lint`
- `pnpm -w run check-spell`
- `pnpm format:check`
- `golangci-lint`: 0 issues

## Related Links

- [Documentation](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v74.0.0/docs/rules/consistent-tuple-labels.md)
- [Source](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v74.0.0/rules/consistent-tuple-labels.js)
- [Upstream tests](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v74.0.0/test/consistent-tuple-labels.js)

Tracks #1309.

## Checklist

- [x] Tests updated.
- [x] Documentation updated.
- [x] Rule registered in the Unicorn catalog.
- [x] Unicorn recommended preset updated.

### DIFF
--- a/internal/plugins/unicorn/rules/consistent_tuple_labels/consistent_tuple_labels.go
+++ b/internal/plugins/unicorn/rules/consistent_tuple_labels/consistent_tuple_labels.go
@@ -1,0 +1,56 @@
+package consistent_tuple_labels
+
+import (
+	"github.com/microsoft/typescript-go/shim/ast"
+	"github.com/web-infra-dev/rslint/internal/rule"
+)
+
+const messageID = "consistent-tuple-labels"
+
+var message = rule.RuleMessage{
+	Id:          messageID,
+	Description: "This tuple element should have a label, just like the other elements.",
+}
+
+func isLabeledElement(element *ast.Node) bool {
+	// tsgo represents every labeled tuple element as NamedTupleMember,
+	// including optional and rest elements. Unlabeled optional and rest
+	// elements instead use OptionalType and RestType nodes.
+	return element != nil && element.Kind == ast.KindNamedTupleMember
+}
+
+// ConsistentTupleLabelsRule requires every element of a tuple type to have a
+// label when any element is labeled.
+//
+// https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v74.0.0/rules/consistent-tuple-labels.js
+var ConsistentTupleLabelsRule = rule.Rule{
+	Name:   "unicorn/consistent-tuple-labels",
+	Schema: rule.EmptyArraySchema,
+	Run: func(ctx rule.RuleContext, options []any) rule.RuleListeners {
+		return rule.RuleListeners{
+			ast.KindTupleType: func(node *ast.Node) {
+				tuple := node.AsTupleTypeNode()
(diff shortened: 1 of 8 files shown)


internal/plugins/unicorn/rules/consistent_tuple_labels/consistent_tuple_labels.go
```diff
@@ -0,0 +1,56 @@
...
+					}
+				}
+
+				if labeledCount == 0 || labeledCount == len(tuple.Elements.Nodes) {
+					return
+				}
+
+				for _, element := range tuple.Elements.Nodes {
+					if !isLabeledElement(element) {
+						ctx.ReportNode(element, message)
```
swwind: [P2] Report the inner type range for parenthesized tuple elements

For `type T = [a: string, (number)]`, upstream reports `number` (columns 23–29), while this implementation reports `(number)` (columns 22–30). Nested parentheses widen the range further. Please unwrap `KindParenthesizedType` only when determining the report range, while retaining the wrapper for unlabeled-element classification, and add a regression test.

shiyudesu: Fixed in 4cf013f1. The rule now keeps the original tuple element for label classification and applies `ast.SkipTypeParentheses` only when selecting the report node, so single and nested type parentheses are excluded from the diagnostic range. I updated the existing parenthesized assertion and added regressions for `(((number)))` and `((number | undefined))`. I also compared all three cases directly against `eslint-plugin-unicorn@74.0.0`, then reran the Go tests, full build, JS integration test, typecheck, lint, spell check, format check, and Go lint.

elecmonkey (approved):
no more issues found.

swwind (approved):
LGTM

PR #2003, opened 2026-09-02
feat: port rule unicorn/no-unnecessary-array-flat-depth

shiyudesu:
## Summary

Port the `unicorn/no-unnecessary-array-flat-depth` rule from eslint-plugin-unicorn v74.0.0.

The rule reports an explicit depth of `1` passed to `Array#flat()`, since `1` is already the default.

- Recognize equivalent numeric literal forms such as `1.0` and `0b01`.
- Preserve optional member access and TypeScript expression wrappers.
- Skip optional calls, computed property access, non-literal arguments, and receivers known not to be arrays.
- Remove a trailing comma together with the redundant argument while preserving surrounding comments.
- Defer autofix construction until fixes are requested.
- Enable the rule in the native Unicorn recommended preset.

Validation:

- `go test -count=1 ./internal/plugins/unicorn/rules/no_unnecessary_array_flat_depth`
- `pnpm build`
- `pnpm typecheck`
- `pnpm lint`
- `pnpm -w run check-spell`
- `pnpm format:check`
- `pnpm exec rs test --testTimeout=10000 no-unnecessary-array-flat-depth presets`
- `golangci-lint v2.12.2` on the changed Go packages

## Related Links

- Rule documentation: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v74.0.0/docs/rules/no-unnecessary-array-flat-depth.md
- Upstream source: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v74.0.0/rules/no-unnecessary-array-flat-depth.js
- Upstream tests: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v74.0.0/test/no-unnecessary-array-flat-depth.js

## Checklist


internal/plugins/unicorn/rules/no_unnecessary_array_flat_depth/no_unnecessary_array_flat_depth.go
```diff
@@ -0,0 +1,63 @@
...
+					Method:              "flat",
+					ArgumentsLength:     &oneArgument,
+					AllowOptionalMember: true,
+				})
+				if !ok {
+					return
+				}
+
+				rawArgument := call.Call.Arguments()[0]
+				depth := ast.SkipParentheses(rawArgument)
```
swwind: **[P1] Unwrap JSDoc-only argument wrappers before literal check**

`tsgo` materializes `array.flat(/** @type {number} */ (1))` as a JSDoc wrapper, while ESLint sees the argument as the numeric literal `1`. Since `ast.SkipParentheses` leaves that wrapper intact, this JavaScript case is silently missed. Please use `utils.ESTreeRuntimeExpression(rawArgument)` here and add a regression test.

internal/plugins/unicorn/rules/no_unnecessary_array_flat_depth/no_unnecessary_array_flat_depth.go
```diff
@@ -0,0 +1,63 @@
...
+				if depth == nil || depth.Kind != ast.KindNumericLiteral ||
+					utils.NormalizeNumericLiteral(depth.AsNumericLiteral().Text) != "1" {
+					return
+				}
+
+				if unicornutil.ShouldSkipKnownNonArrayReceiver(ctx, call.Object) {
```
elecmonkey: [P2] Keep source-only call receivers reportable

This skip suppresses cases that upstream still reports. In `eslint-plugin-unicorn@74.0.0`, `Number(1).flat(1)`, `String("x").flat(1)`, `Boolean(0).flat(1)`, `parseInt("1", 10).flat(1)`, and `Math.abs(-1).flat(1)` all report and autofix to remove the redundant depth, because `CallExpression` receivers remain unknown to `isKnownNonIndexedCollection`. Rslint's source-only receiver classifier folds those built-in calls to known non-arrays, so this rule returns here and reports nothing. Please keep those receivers reportable for this rule, or add a rule-specific receiver check that matches upstream's behavior.

internal/plugins/unicorn/rules/no_unnecessary_array_flat_depth/no_unnecessary_array_flat_depth.go
shiyudesu: Thank you for catching this. I replaced ast.SkipParentheses with utils.ESTreeRuntimeExpression for the argument and added regression coverage for both JSDoc @type and @satisfies casts, including autofix outputs that preserve the comments.

internal/plugins/unicorn/rules/no_unnecessary_array_flat_depth/no_unnecessary_array_flat_depth.go
shiyudesu: Thank you for pointing this out. I added a rule-specific receiver check that keeps source-only JavaScript CallExpression receivers reportable while preserving the shared classifier for other shapes. I also added regression cases for Number, String, Boolean, parseInt, Math.abs, and Object calls, plus a TypeScript Map-returning call to verify that typed keyed collections are still skipped.

internal/plugins/unicorn/rules/no_unnecessary_array_flat_depth/no_unnecessary_array_flat_depth.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+	return []rule.RuleFix{rule.RuleFixRemoveRange(removalRange)}
+}
+
+func shouldSkipFlatReceiver(ctx rule.RuleContext, receiver *ast.Node) bool {
+	// For source-only JavaScript, Unicorn leaves CallExpression receivers unknown.
+	// The shared classifier can fold built-in calls such as Number(1), so keep
+	// JavaScript call receivers reportable for this rule.
+	runtimeReceiver := utils.ESTreeRuntimeExpression(receiver)
+	if runtimeReceiver != nil && ast.IsInJSFile(runtimeReceiver) && ast.IsCallExpression(runtimeReceiver) {
```
elecmonkey, line 40: [P2] Preserve upstream pass-through receiver skips

This blanket exemption makes every JavaScript `CallExpression` receiver reportable, but upstream `eslint-plugin-unicorn@74.0.0` still runs its static-value path for safe pass-through calls and skips known non-arrays such as `Object.freeze({}).flat(1)`, `Object.seal({}).flat(1)`, `Object.preventExtensions({}).flat(1)`, and `Object.freeze("x").flat(1)`. Rslint now reports and autofixes those cases, so the previous fix for `Number(1).flat(1)` needs to be narrowed rather than bypassing `ShouldSkipKnownNonArrayReceiver` for all JS call receivers.

swwind, line 40: [P2] Restrict the source-only call exception

The unconditional `CallExpression` escape makes this rule report cases that upstream intentionally skips because the receiver is statically known to be non-array. For example, `Object.freeze({}).flat(1)`, `Object.freeze(1).flat(1)`, `Object.seal({}).flat(1)`, and `Object.preventExtensions({}).flat(1)` are valid under eslint-plugin-unicorn 74.0.0, but rslint reports them and removes `1`. Please preserve the source-only exception for calls that upstream leaves unknown while still consulting the static classifier for calls with a known non-array result, and add a regression test for the smallest case.

shiyudesu: Thank you, agreed. The source-only exception is now restricted so safe Object pass-through calls still consult ShouldSkipKnownNonArrayReceiver. I added Object.freeze({}).flat(1) as the minimal skipped regression and also covered seal, preventExtensions, primitive and array results, extra arguments, shadowing, and computed access.

shiyudesu: Thank you for catching the over-broad exception. I narrowed it so exact one-argument Object.freeze, Object.seal, and Object.preventExtensions dot calls continue through the shared static classifier, while calls that upstream leaves unknown remain reportable. I added regression coverage for the skipped non-array pass-through cases, plus array results, extra arguments, a shadowed Object binding, and computed access.

swwind (dismissed):
LGTM

internal/plugins/unicorn/rules/no_unnecessary_array_flat_depth/no_unnecessary_array_flat_depth.go
```diff
@@ -0,0 +1,96 @@
...
+func shouldSkipFlatReceiver(ctx rule.RuleContext, receiver *ast.Node) bool {
+	// For source-only JavaScript, Unicorn leaves CallExpression receivers unknown.
+	// The exception does not cover safe Object pass-through calls, whose static
+	// result Unicorn still classifies.
+	runtimeReceiver := utils.ESTreeRuntimeExpression(receiver)
+	if runtimeReceiver != nil && ast.IsInJSFile(runtimeReceiver) && ast.IsCallExpression(runtimeReceiver) &&
+		!isPotentialStaticObjectPassThroughCall(runtimeReceiver) {
+		return false
+	}
+	return unicornutil.ShouldSkipKnownNonArrayReceiver(ctx, receiver)
```
elecmonkey, line 48: [P2] Preserve static pass-through skips through const aliases

Upstream `eslint-plugin-unicorn@74.0.0` skips `const obj = {}; Object.freeze(obj).flat(1)` because `Object.freeze(obj)` is a safe pass-through call and `getStaticValue` resolves the stable const `obj` to a non-array object. This branch routes pass-through-shaped calls back into `ShouldSkipKnownNonArrayReceiver`, but the source-only classifier rejects local identifier arguments in `staticCallArgumentsKnown`, so the receiver stays unknown and this rule reports/autofixes a case upstream treats as valid. The same applies to `Object.seal(obj)` and `Object.preventExtensions(obj)` when the argument is a stable non-array const.

shiyudesu: Thank you for identifying this. Exact Object.freeze, Object.seal, and Object.preventExtensions pass-through calls now resolve stable const aliases through a cached static evaluator, with a conservative safety walk that rejects non-const dependencies, calls, constructions, assignments, mutations, and cycles. I added coverage for direct and chained const object aliases, nested pass-through calls, and static expressions, along with negative cases for array aliases, let bindings, call initializers, destructuring, functions, and mutated objects.

elecmonkey (dismissed):
LGTM

fansenze:
@shiyudesu Thanks for the contribution! This PR currently has a few conflicts that need to be resolved first. Once those are fixed, it should be ready to merge.

elecmonkey (approved):

PR #2006, opened 2026-09-02
feat: port rule react/no-namespace

swwind:
## Summary

Port the `react/no-namespace` rule from ESLint to rslint.

Enforce that namespaces are not used in React elements, including namespace-qualified JSX tags and `React.createElement` string names.

## Related Links

- ESLint rule: https://github.com/jsx-eslint/eslint-plugin-react/blob/v7.37.5/docs/rules/no-namespace.md
- Source code: https://github.com/jsx-eslint/eslint-plugin-react/blob/v7.37.5/lib/rules/no-namespace.js

## Checklist


internal/plugins/react/rules/no_namespace/no_namespace.go
```diff
@@ -0,0 +1,57 @@
...
+			// tsgo separates paired and self-closing JSX, so listen to both.
+			reportIfNamespaced(node, reactutil.GetJsxElementTypeString(node))
+		}
+
+		return rule.RuleListeners{
+			ast.KindJsxOpeningElement:     checkJSX,
+			ast.KindJsxSelfClosingElement: checkJSX,
+			ast.KindCallExpression: func(node *ast.Node) {
+				call := node.AsCallExpression()
+				if !reactutil.IsCreateElementCallWithChecker(call.Expression, pragma, ctx.TypeChecker) {
```
elecmonkey: [P2] Avoid treating shadowed createElement as React

This call now uses `IsCreateElementCallWithChecker`, but when `ctx.TypeChecker` is nil the helper falls back to a source-file-wide syntax scan that ignores lexical shadowing. In a source-only run, `import { createElement } from "react"; function f() { const createElement = () => null; createElement("ns:Panel"); }` reports `react/no-namespace`, while `eslint-plugin-react@7.37.5` reports no diagnostics because scope resolution picks the local function. That gives false positives for gap/source-only files whenever an imported React `createElement` name is shadowed; the bare-callee branch needs scope-aware resolution or should not be enabled without it.

swwind: Fixed in `ce7fddcdd`. When TypeChecker is unavailable, `react/no-namespace` now uses the file-local RefStore to resolve bare `createElement` bindings, so a lexical shadow is no longer treated as the React import. The existing import case remains covered by a regression test.

internal/plugins/react/rules/no_namespace/no_namespace.go
```diff
@@ -0,0 +1,57 @@
...
+			// tsgo separates paired and self-closing JSX, so listen to both.
+			reportIfNamespaced(node, reactutil.GetJsxElementTypeString(node))
+		}
+
+		return rule.RuleListeners{
+			ast.KindJsxOpeningElement:     checkJSX,
+			ast.KindJsxSelfClosingElement: checkJSX,
+			ast.KindCallExpression: func(node *ast.Node) {
+				call := node.AsCallExpression()
+				if !reactutil.IsCreateElementCallWithRefs(call.Expression, pragma, ctx.TypeChecker, ctx.Refs) {
```
elecmonkey, line 41: [P2] Match upstream's latest createElement binding

The bare-callee path here delegates to `IsDestructuredFromPragmaImportWithRefs`, which picks the ts-go symbol's primary declaration rather than ESLint's latest variable definition. That makes redeclarations diverge from `eslint-plugin-react@7.37.5`: `var createElement = React.createElement; var createElement = other; createElement("ns:Panel")` is reported by Rslint but upstream reports nothing, while reversing the two `var` initializers reports upstream but Rslint misses it. Any file that redeclares `createElement` can therefore get false positives or false negatives depending on declaration order; this helper needs to model the latest value definition for the resolved binding.

swwind: Fixed in `e9b0680a9`. `isDestructuredFromPragmaSymbol` now selects the latest source declaration instead of ts-go's primary `ValueDeclaration`, matching eslint-plugin-react@7.37.5's latest variable definition. Added source-only regressions for both redeclaration orders. Focused tests pass: `go test ./internal/plugins/react/rules/no_namespace` and `go test ./internal/plugins/react/reactutil`.

elecmonkey, line 41: [P2] Recognize default and namespace React imports

When the callee binding comes from `import createElement from "react"` or `import * as createElement from "react"`, `eslint-plugin-react@7.37.5` reports `createElement("ns:Panel")`, but Rslint reports nothing in both source-only and checker-backed runs. `IsDestructuredFromPragmaImportWithRefs` only accepts an `ImportSpecifier`, while upstream accepts any definition whose parent is an `ImportDeclaration`; please handle `ImportClause` and `NamespaceImport` declarations as React bindings too.

elecmonkey, line 41: [P2] Preserve upstream initializer AST-shape checks

For `const createElement = React["anything"]`, upstream reports but Rslint does not; for `const createElement = React?.anything` and `const createElement = React.anything as any`, upstream is silent but Rslint reports. The helper rejects computed member expressions and uses `SkipExpressionWrappers`, so this bare-callee path diverges in both directions; please match upstream's exact `VariableDeclarator.init` and `init.object` node-shape checks.

internal/plugins/react/reactutil/pragma_binding.go
```diff
@@ -55,16 +81,27 @@ func IsDestructuredFromPragmaImport(ident *ast.Node, pragma string, tc *checker.
...
+
+	// Pick the latest definition. Upstream walks `defs[defs.length - 1]`,
+	// whereas ts-go's ValueDeclaration is the symbol's primary declaration
+	// and can point at an earlier redeclaration. Declarations are source
+	// nodes, so selecting the one with the greatest position mirrors the
+	// latest definition regardless of the binder's primary declaration.
 	var decl *ast.Node
-	if symbol.ValueDeclaration != nil {
+	for _, candidate := range symbol.Declarations {
+		if candidate != nil && (decl == nil || candidate.Pos() > decl.Pos()) {
```
elecmonkey: [P2] Avoid ordering merged declarations by file-local offsets

`candidate.Pos()` is only meaningful within one source file, but checker symbols for global scripts can contain declarations from multiple files. With another included script declaring `var createElement = React.createElement` at a larger offset and the linted file declaring `var createElement = other; createElement("ns:Panel")`, Rslint reports while upstream is silent. This also misses `function createElement() {}; var createElement = React.x; createElement("ns:Panel")`, because ts-go exposes only the function in `symbol.Declarations` while ESLint includes the later variable definition. Please model ESLint’s value-definition order instead of choosing the largest AST offset.

internal/plugins/react/rules/no_namespace/no_namespace.go
swwind: Fixed in 3a32457a0. Bare createElement bindings from default and namespace imports are now recognized, with source-only and checker-backed regressions added.

internal/plugins/react/reactutil/pragma_binding.go
swwind: Fixed in 3a32457a0. The rule now uses file-local ESLint-like value-definition order instead of comparing AST offsets across merged checker declarations. Cross-file and function/variable regressions were added.

internal/plugins/react/rules/no_namespace/no_namespace.go
swwind: Fixed in 3a32457a0. Initializer matching now follows the upstream AST shapes: computed members are accepted, while optional chains and TypeScript wrappers are not. Regression coverage was added.

internal/plugins/react/reactutil/pragma_binding.go
```diff
@@ -55,59 +87,67 @@ func IsDestructuredFromPragmaImport(ident *ast.Node, pragma string, tc *checker.
...
-	// ImportSpecifier (which has no Initializer of its own), upstream
-	// walks `latestDef.parent.type === 'ImportDeclaration'`. We mirror
-	// by trying ValueDeclaration first then Declarations[0].
-	var decl *ast.Node
-	if symbol.ValueDeclaration != nil {
-		decl = symbol.ValueDeclaration
-	} else if len(symbol.Declarations) > 0 {
-		decl = symbol.Declarations[0]
-	}
+	return isDestructuredFromPragmaSymbol(symbol, pragma, pragmaLower)
```
elecmonkey, line 80: [P2] Avoid falling back to cross-file checker bindings here

When `LatestValueDefinitionInFile` finds no declaration in the current file, this falls through to `tc.GetSymbolAtLocation`, so a global script declaration from another project file can make an otherwise unbound `createElement("ns:Panel")` look like React’s factory. Repro: with `declaration.ts` containing `var createElement = React.createElement;` and `usage.ts` containing only `createElement("ns:Panel");` in the same tsconfig, Rslint reports `react/no-namespace` in `usage.ts`, while `eslint-plugin-react@7.37.5` reports no diagnostics because ESLint scope lookup is file-local. The RefStore-aware path should not use checker fallback when the current file has no value definition for the bare callee.

swwind: Fixed in `62a7fb323`. When a RefStore is available, file-local resolution is now authoritative: if the current file has no value definition for the bare `createElement` callee, the rule returns false instead of falling back to TypeChecker symbols from another global script. Added `TestNoNamespaceDoesNotUseCrossFileCheckerBinding` and appended the corresponding unbound case to the persistent corpus. Focused tests and the ESLint corpus bridge pass.

internal/plugins/react/rules/no_namespace/no_namespace.go
elecmonkey, line 41: [P2] Recognize computed createElement member calls

For `React[createElement]("ns:Panel")`, `eslint-plugin-react@7.37.5` reports `react/no-namespace`, because upstream only checks the computed member identifier property name, but this path delegates to `IsCreateElementCallWithRefs` / `isPragmaFactoryCallCore`, which returns false unless the callee is a `PropertyAccessExpression`. I confirmed the differential with upstream and a temporary local Rslint test, which reported 0 diagnostics, so this form silently misses namespaces; the member-access branch needs to handle `ElementAccessExpression` when the property expression is an identifier named `createElement`.

swwind: Fixed in 3525b0876. The React factory matcher now recognizes computed member calls whose property is an identifier, such as `React[createElement](...)`, while preserving the upstream behavior for string-literal properties. Added a permanent Go regression and the corresponding persistent corpus case; the focused Go tests and ESLint corpus bridge pass.

internal/rule/ref_store.go
```diff
@@ -264,6 +265,36 @@ func (s *RefStore) ResolveInFile(node *ast.Node) *ast.Symbol {
...
+			ReferenceNames:    map[string]struct{}{node.Text(): {}},
+		})
+	}
+	for _, reference := range s.scope.References {
+		if reference.Identifier != node {
+			continue
+		}
+		for i := len(reference.Declarations) - 1; i >= 0; i-- {
+			declaration := reference.Declarations[i]
+			if declaration != nil && declaration.IsValueBinding {
```
elecmonkey: [P2] Preserve the actual latest scope definition

Filtering out definitions where `IsValueBinding` is false does not match upstream's `variable.defs[variable.defs.length - 1]`. With `import type { createElement } from "react"; createElement("ns:Panel")`, eslint-plugin-react@7.37.5 reports one diagnostic while Rslint reports none. Conversely, `const createElement = React.createElement; type createElement = {}; createElement("ns:Panel")` is silent upstream but reported by Rslint because this loop skips the trailing type definition and falls back to the earlier value definition; a trailing interface produces the same mismatch.

internal/plugins/react/rules/no_namespace/no_namespace.go
elecmonkey, line 41: [P2] Keep authored TypeScript wrappers around require calls

`SkipExpressionWrappers` removes authored TypeScript assertions that `@typescript-eslint/parser` preserves. Consequently, both `const { createElement } = require("react" as string); createElement("ns:Panel")` and `const { createElement } = (require as any)("react"); createElement("ns:Panel")` produce one Rslint diagnostic while eslint-plugin-react@7.37.5 produces none. The require matcher should preserve authored TypeScript wrappers around both the callee and its argument.

internal/plugins/react/rules/no_namespace/no_namespace.go
```diff
@@ -0,0 +1,57 @@
...
+					return
+				}
+				if call.Arguments == nil || len(call.Arguments.Nodes) == 0 {
+					return
+				}
+
+				// Upstream checks Literal, not TemplateLiteral or a TypeScript
+				// wrapper. Parentheses are skipped because ESTree flattens them.
+				argument := ast.SkipParentheses(call.Arguments.Nodes[0])
```
elecmonkey: [P2] Unwrap synthesized JSDoc casts from the first argument

ts-go synthesizes a wrapper for JavaScript JSDoc casts that Espree does not expose. For `React.createElement(/** @type {any} */ ("ns:Panel"))`, eslint-plugin-react@7.37.5 sees a `Literal` and reports one diagnostic, while Rslint leaves the synthesized wrapper after `SkipParentheses` and reports none. This path should remove parentheses and synthesized JSDoc wrappers while continuing to preserve authored TypeScript assertions.

internal/plugins/react/reactutil/pragma_binding.go
```diff
@@ -319,22 +371,31 @@ func initializerMatchesPragma(init *ast.Node, pragma, pragmaLower string) bool {
 	if init == nil {
 		return false
 	}
-	init = SkipExpressionWrappers(init)
+	init = ast.SkipParentheses(init)
```
elecmonkey: [P2] Unwrap synthesized JSDoc casts from pragma initializers

The same synthesized-wrapper mismatch affects bare bindings: `const createElement = /** @type {any} */ (React.anything); createElement("ns:Panel")` produces one diagnostic with eslint-plugin-react@7.37.5 but none with Rslint because `SkipParentheses` cannot reach `React.anything`. Please use the ESTree runtime view that removes parentheses and synthesized JSDoc wrappers without removing authored TypeScript assertions.

internal/rule/ref_store.go
swwind: Fixed in d57c32f8. LatestValueDefinitionInFile now returns the actual latest scope definition, including type-only definitions, matching variable.defs[defs.length - 1]. Added regressions for import type and trailing type/interface definitions; focused tests and the ESLint corpus bridge pass.

internal/plugins/react/rules/no_namespace/no_namespace.go
swwind: Fixed in d57c32f8. require matching now uses the ESTree runtime view: authored TypeScript assertions around the callee or module argument remain visible and prevent a match, while parentheses and JSDoc wrappers are removed. Added both reported regressions; focused tests and the corpus bridge pass.

internal/plugins/react/rules/no_namespace/no_namespace.go
swwind: Fixed in d57c32f8. The first createElement argument now uses the ESTree runtime view, so synthesized JSDoc casts are unwrapped while authored TypeScript assertions remain visible. Added the reported regression; focused tests and the corpus bridge pass.

internal/plugins/react/reactutil/pragma_binding.go
swwind: Fixed in d57c32f8. Pragma initializer matching now uses the ESTree runtime view, removing synthesized JSDoc wrappers while preserving authored TypeScript wrappers. Added the reported regression; focused tests and the corpus bridge pass.

elecmonkey (dismissed):

elecmonkey:
why conflict

swwind:
> why conflict

I think you give me the same shared utils issue for all of my PRs, and I fixed the same issue in each PR simultaneously.

elecmonkey:
> > why conflict
> 
> I think you give me the same shared utils issue for all of my PRs, and I fixed the same issue in each PR simultaneously.

it's a sad story. consider using stack PR?

swwind:
> > > why conflict
> > 
> > 
> > I think you give me the same shared utils issue for all of my PRs, and I fixed the same issue in each PR simultaneously.
> 
> it's a sad story. consider using stack PR?

I think we can port the shared utils in eslint-plugin-react first, to avoid those fixes to utils in each rule port.

elecmonkey (dismissed):

elecmonkey (approved):

fansenze (approved):

PR #2011, opened 2026-09-02
feat: port rule react/prefer-read-only-props

swwind:
## Summary

Port the `react/prefer-read-only-props` rule from ESLint to rslint.

Enforce that React component props are declared read-only in TypeScript, with autofixes for missing `readonly` modifiers.

## Related Links

- ESLint rule: https://github.com/jsx-eslint/eslint-plugin-react/blob/v7.37.5/docs/rules/prefer-read-only-props.md
- Source code: https://github.com/jsx-eslint/eslint-plugin-react/blob/v7.37.5/lib/rules/prefer-read-only-props.js

## Checklist


internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props_extras_test.go
```diff
@@ -0,0 +1,266 @@
...
+			Tsx:    true,
+			Output: []string{`type Props = { readonly name: string }; class Hello extends React.Component<(Props)> { render() { return <div/>; } }`},
+			Errors: []rule_tester.InvalidTestCaseError{{MessageId: "readOnlyProp", Message: "Prop 'name' should be read-only."}},
+		},
+	})
+}
+
+func TestPreferReadOnlyPropsLongAliasChain(t *testing.T) {
+	var source strings.Builder
+	for i := 0; i < 34; i++ {
```
elecmonkey: [P1] Make the new test lint-clean

The required `Lint&Check` job fails on this loop with the `intrange` diagnostic, and the failure reproduces locally with `golangci-lint run ./internal/plugins/react/rules/prefer_read_only_props`. Use `for i := range 34` so this PR can pass its required checks.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
```diff
@@ -0,0 +1,534 @@
...
+
+	if !reactutil.IsStatelessReactComponentWithWrappers(node, pragma, nil, wrappers) {
+		return
+	}
+	params := reactutil.FunctionParameters(node)
+	if len(params) == 0 {
+		return
+	}
+	hasParameterType := false
+	if param := params[0].AsParameterDeclaration(); param != nil && param.Type != nil {
```
elecmonkey, line 216: [P1] Restrict identifier parameters to the upstream props shape

This accepts every typed first parameter, while upstream's `isAnnotatedFunctionPropsDeclaration` accepts an identifier only when it is named `props` (destructuring remains valid). For `function Hello(input: { name: string }) { return <div /> }`, upstream 7.37.5 emits nothing, but Rslint reports and autofixes `name`; with `const Hello: React.FC<P> = (input: Q) => <div />`, upstream checks `P` while Rslint checks and fixes `Q`. Apply the upstream parameter-name gate and retain the variable-annotation fallback when the parameter does not qualify.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
```diff
@@ -0,0 +1,534 @@
...
+func classPropsTypeArgument(node *ast.Node) *ast.Node {
+	heritage := ast.GetClassExtendsHeritageElement(node)
+	if heritage == nil {
+		return nil
+	}
+	extends := heritage.AsExpressionWithTypeArguments()
+	if extends == nil || extends.TypeArguments == nil || len(extends.TypeArguments.Nodes) == 0 {
+		return nil
+	}
+	propsType := extends.TypeArguments.Nodes[0]
```
elecmonkey: [P2] Match upstream's three-argument class type selection

The first superclass type argument is always selected here, but upstream's fallback when no Flow version is configured selects index 1 whenever the superclass has three or more arguments. For `class Hello extends React.Component<Props, State, unknown>`, upstream reports and fixes `State.count`, while Rslint reports and fixes `Props.name`; this changes both the diagnostic and the edited declaration.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
```diff
@@ -0,0 +1,534 @@
...
+			for _, member := range literal.Members.Nodes {
+				checkPropertySignature(member, report)
+			}
+		}
+	case ast.KindInterfaceDeclaration:
+		decl := node.AsInterfaceDeclaration()
+		if decl == nil {
+			return
+		}
+		if decl.HeritageClauses != nil {
```
elecmonkey, line 590: [P2] Preserve upstream interface override ordering

This traverses base interfaces before the current interface body, whereas upstream appends the current body first and traverses `extends` afterward; its `declaredPropTypes` object therefore lets the base entry win on duplicate names. With `interface A { name: string } interface B extends A { readonly name: string }`, upstream reports `A.name` but Rslint emits nothing, and the reverse readonly/mutable arrangement produces the opposite mismatch.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
```diff
@@ -0,0 +1,534 @@
...
+	if propsType == nil || propsType.Kind != ast.KindTypeReference {
+		return nil
+	}
+	return propsType
+}
+
+func checkFunction(node *ast.Node, pragma string, genericImports map[string]string, wrappers []reactutil.ComponentWrapperEntry, validateType func(*ast.Node, *ast.Node)) {
+	// The forwardRef arm is checked before component-return heuristics in the
+	// upstream collector. Its second type argument is the props type.
+	if call := parentCall(node); call != nil && call.TypeArguments != nil && len(call.TypeArguments.Nodes) >= 2 && isForwardRefCall(call) {
```
elecmonkey: [P2] Resolve forwardRef against the call-site component

The early `forwardRef` arm validates solely from the file-wide import map and a hard-coded `React` object, bypassing upstream component registration and scope resolution. Consequently an imported `forwardRef` shadowed by a local function is reported and autofixed by Rslint although upstream is silent; Rslint also reports `React.forwardRef` when `settings.react.pragma` is `R`, where upstream does not classify that call as a component.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
```diff
@@ -0,0 +1,534 @@
...
+		if !isImportedForwardRef(call, genericImports) {
+			return
+		}
+		validateType(node, call.TypeArguments.Nodes[1])
+		return
+	}
+
+	if !reactutil.IsStatelessReactComponentWithWrappers(node, pragma, nil, wrappers) {
```
elecmonkey: [P2] Exclude async generators from component detection

Upstream's `Components.detect` explicitly registers async generator functions with confidence 0, excluding them from the component list. This call has no corresponding guard, so `async function* Hello(props: Props) { yield 1; return <div /> }` is reported and autofixed by Rslint while upstream emits nothing.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
```diff
@@ -0,0 +1,534 @@
...
+				validateType(node, propsType)
+			}
+		},
+		ast.KindPropertyDeclaration: func(node *ast.Node) {
+			pd := node.AsPropertyDeclaration()
+			if pd == nil || pd.Type == nil || reactutil.EnclosingClass(node) == nil {
+				return
+			}
+			classNode := reactutil.EnclosingClass(node)
+			if !reactutil.ExtendsReactComponent(classNode, pragma) {
```
elecmonkey: [P2] Recognize JSDoc-declared React classes

Upstream's `isES6Component` also accepts classes marked with `@extends React.Component` or `@augments React.Component`, but this property path requires a syntactic `extends` clause. For `/** @extends React.Component */ class Hello { props: Props; render() { return <div /> } }`, upstream reports and fixes the mutable property in `Props`, while Rslint emits nothing.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
```diff
@@ -0,0 +1,534 @@
...
+var PreferReadOnlyPropsRule = rule.Rule{
+	Name:   "react/prefer-read-only-props",
+	Schema: rule.EmptyArraySchema,
+	Run:    runRule,
+}
+
+// runRule mirrors eslint-plugin-react's propTypes collector for the
+// TypeScript shapes understood by tsgo.
+func runRule(ctx rule.RuleContext, _ []any) rule.RuleListeners {
+	pragma := reactutil.GetReactPragma(ctx.Settings)
```
elecmonkey: [P2] Honor source-level JSX pragma annotations

Upstream obtains the pragma from `pragmaUtil.getFromContext`, where a file's first `@jsx` annotation takes precedence over settings. This uses the settings-only helper, so `/** @jsx R */ ... class Hello extends R.Component<Props>` is reported by upstream but silently skipped by Rslint. Use the context-aware pragma lookup.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
```diff
@@ -0,0 +1,534 @@
...
+	case ast.KindTypeReference:
+		ref := node.AsTypeReferenceNode()
+		if ref == nil {
+			return
+		}
+		if propsType := reactGenericPropsType(node, genericImports); propsType != nil {
+			validateTypeNode(propsType, aliases, genericImports, report, seen)
+			return
+		}
+		if ref.TypeName != nil && ref.TypeName.Kind == ast.KindIdentifier &&
```
elecmonkey, line 626: [P2] Resolve ReturnType type-query callbacks

The upstream collector handles `ReturnType<typeof mapStateToProps>` by finding the top-level function and following object or generic-call returns. This branch only accepts an inline function type, so `const makeProps = () => make<{ name: string }>(); function Hello(props: ReturnType<typeof makeProps>) { return <div /> }` is reported and fixed upstream but completely missed by Rslint.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
```diff
@@ -0,0 +1,534 @@
...
+			decl := node.AsTypeAliasDeclaration()
+			if decl != nil && decl.Name() != nil && decl.Name().Kind == ast.KindIdentifier {
+				types[decl.Name().AsIdentifier().Text] = append(types[decl.Name().AsIdentifier().Text], decl.Type)
+			}
+		case ast.KindInterfaceDeclaration:
+			decl := node.AsInterfaceDeclaration()
+			if decl != nil && decl.Name() != nil && decl.Name().Kind == ast.KindIdentifier {
+				types[decl.Name().AsIdentifier().Text] = append(types[decl.Name().AsIdentifier().Text], node)
+			}
+		case ast.KindImportDeclaration:
```
elecmonkey: [P2] Preserve source-order React import discovery

Collecting all React imports before traversal changes upstream's source-order behavior. For `const Hello: FC<Props> = (props) => <div />; import { FC } from 'react';`, upstream visits the component before registering `FC` and emits nothing, while Rslint sees the later import up front and reports/autofixes `Props`.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
```diff
@@ -0,0 +1,534 @@
...
+		return types, imports
+	}
+	source.Node.ForEachChild(func(node *ast.Node) bool {
+		switch node.Kind {
+		case ast.KindTypeAliasDeclaration:
+			decl := node.AsTypeAliasDeclaration()
+			if decl != nil && decl.Name() != nil && decl.Name().Kind == ast.KindIdentifier {
+				types[decl.Name().AsIdentifier().Text] = append(types[decl.Name().AsIdentifier().Text], decl.Type)
+			}
+		case ast.KindInterfaceDeclaration:
```
elecmonkey: [P2] Preserve default-export interface handling

The ts-go traversal exposes `export default interface Props` directly and this collector records it, but upstream's AST search only unwraps `ExportNamedDeclaration`; an `ExportDefaultDeclaration` is not considered a matching interface. As a result, a component using a mutable default-exported `Props` interface is silently accepted upstream but reported and autofixed by Rslint.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
```diff
@@ -0,0 +1,534 @@
...
+		ast.KindPropertyDeclaration: func(node *ast.Node) {
+			pd := node.AsPropertyDeclaration()
+			if pd == nil || pd.Type == nil || reactutil.EnclosingClass(node) == nil {
+				return
+			}
+			classNode := reactutil.EnclosingClass(node)
+			if !reactutil.ExtendsReactComponent(classNode, pragma) {
+				return
+			}
+			name := pd.Name()
```
elecmonkey, line 129: [P2] Preserve the upstream class props-field gate

The upstream collector recognizes a typed class field only when `props` is one of its first two tokens. This AST-name-only check also accepts fields such as `public readonly props: Props`; upstream emits nothing for that shape, while Rslint reports and autofixes the referenced `Props` declaration.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
```diff
@@ -0,0 +1,534 @@
...
+		if expr.Kind != ast.KindNoSubstitutionTemplateLiteral {
+			if name, ok := utils.GetStaticPropertyName(nameNode); ok {
+				return name, true
+			}
+		}
+		return "undefined", true
+	}
+	var name string
+	switch nameNode.Kind {
+	case ast.KindIdentifier:
```
elecmonkey, line 860: [P3] Preserve the upstream __proto__ property behavior

Upstream stores declared props in a normal JavaScript object, so assigning the key `__proto__` does not create an own enumerable property and `Object.keys` never reports it. Rslint treats it as a normal string property and emits an extra diagnostic/fix for `type Props = { '__proto__': string }`, violating the requested 1:1 behavior.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props_upstream_test.go
```diff
@@ -0,0 +1,200 @@
+// TestPreferReadOnlyPropsUpstream migrates the TypeScript cases from
```
elecmonkey: [P2] Inventory every omitted upstream test group

The v7.37.5 suite contains 29 source cases: 17 Flow cases, 2 parser-neutral valid cases, and 10 TypeScript cases. This file contains only the 10 TypeScript cases and gives no per-case or grouped skip rationale for the remaining inventory; the extras happen to cover one parser-neutral case but do not establish the required mapping. Keep the omitted cases as explicit skips with the ts-go parser reason, or document their exact equivalent local cases.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
swwind: Fixed in 710f63324. Typed parameters are now accepted only for `props` or destructured bindings, while the variable annotation fallback remains supported.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
swwind: Fixed in 710f63324. Interface members are collected before heritage declarations, matching upstream overwrite ordering.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
swwind: Fixed in 710f63324. The forwardRef path now uses call-site wrapper matching, checker-based shadowing, and the active JSX pragma.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props_extras_test.go
swwind: Fixed in 710f63324. The loop now uses `for i := range 34`.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
swwind: Fixed in 710f63324. Classes with three or more type arguments now use the second argument as the props type.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
swwind: Fixed in 710f63324. Async generators are excluded from component detection.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
swwind: Fixed in 710f63324. JSDoc `@extends`/`@augments React.Component` classes are now recognized using the canonical comment store.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
swwind: Fixed in 710f63324. Default-exported interfaces are excluded from the top-level type inventory.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
swwind: Fixed in 710f63324. Source-level `@jsx` annotations now take precedence over configured pragma settings.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
swwind: Fixed in 710f63324. Class props fields now follow upstream’s first-two-token gate.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
swwind: Fixed in 710f63324. `__proto__` now follows JavaScript object-key behavior and is not reported.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
swwind: Fixed in 710f63324. React generic imports are now discovered in source order.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
swwind: Fixed in 710f63324. `ReturnType<typeof callback>` now follows top-level callback returns, including generic calls.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props_upstream_test.go
swwind: The upstream mirror now contains only the TypeScript cases. Flow cases are omitted entirely because Flow is outside rslint’s supported syntax; they are not represented as behavior divergences, and no Flow tests were added.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
```diff
@@ -0,0 +1,762 @@
...
+	}
+
+	validateType := func(owner, node *ast.Node) {
+		validateTypeNode(node, typeAliases, genericImports, func(node *ast.Node, name string, readonly bool) {
+			report(owner, node, name, readonly)
+		}, map[*ast.Node]bool{})
+	}
+
+	return rule.RuleListeners{
+		ast.KindTypeAliasDeclaration: func(node *ast.Node) {
```
elecmonkey: [P1] Precollect type declarations before checking components

The declaration maps are populated during source-order traversal, but upstream searches the complete program body when resolving a referenced props type. For `function Hello(props: Props) { return <div/>; } type Props = { name: string };`, eslint-plugin-react 7.37.5 reports and fixes `name`, while Rslint emits nothing. Defer component validation or pre-index all top-level type declarations.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
```diff
@@ -0,0 +1,762 @@
...
+		name := param.Name()
+		if isDestructuredParameter(name) || isPropsParameter(name) {
+			hasParameterType = true
+			validateType(node, param.Type)
+		}
+	}
+
+	// For `const Component: React.FC<Props> = (...) => ...` without a parameter
+	// annotation, the upstream collector uses the variable's annotation.
+	parent := node.Parent
```
elecmonkey: [P2] Unwrap parentheses before reading the variable annotation

Plain parentheses are transparent in the upstream ESTree AST, but ts-go keeps them between the function and its variable declaration. Consequently `const Hello: React.FC<Props> = (((props) => <div/>));` is reported and fixed upstream, while Rslint never reaches the `VariableDeclaration` fallback and emits nothing.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
```diff
@@ -0,0 +1,762 @@
...
+		return nil
+	}
+	return propsType
+}
+
+func checkFunction(node *ast.Node, tc *checker.Checker, pragma string, genericImports map[string]string, wrappers []reactutil.ComponentWrapperEntry, validateType func(*ast.Node, *ast.Node)) {
+	if isAsyncGenerator(node) {
+		return
+	}
+	// The forwardRef arm is checked before component-return heuristics in the
```
elecmonkey, line 198: [P2] Stop forwardRef fallback when only one type argument exists

Upstream enters its `forwardRef` arm whenever type arguments exist and then returns even when the props argument at index 1 is absent. This condition instead falls through to the annotated-parameter path for `forwardRef<HTMLDivElement>((props: { name: string }, ref) => <div/>)`, causing Rslint to report and autofix `name` while upstream emits nothing.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
```diff
@@ -0,0 +1,762 @@
...
+		vd := parent.AsVariableDeclaration()
+		if vd != nil && vd.Type != nil {
+			if propsType := reactGenericPropsType(vd.Type, genericImports); propsType != nil {
+				validateType(node, propsType)
+			}
+		}
+	}
+}
+
+func isForwardRefWrapper(call *ast.CallExpression, fn *ast.Node, wrappers []reactutil.ComponentWrapperEntry, pragma string, tc *checker.Checker) bool {
```
elecmonkey, line 237: [P2] Preserve shadowing checks in source-only Programs

When `ctx.TypeChecker` is nil, the wrapper helper falls back to a file-wide import scan that cannot distinguish a locally shadowed `forwardRef`. In a source-only Program, `import { forwardRef } from 'react'; function Demo() { const forwardRef = value => value; return forwardRef<HTMLDivElement, Props>((props, ref) => <div/>); }` is reported and autofixed by Rslint, whereas upstream resolves the local binding and remains silent.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
```diff
@@ -0,0 +1,762 @@
...
+		return imported, ok && imported != "" && imported != "*" && isGenericTypeName(imported)
+	}
+	if name.Kind != ast.KindQualifiedName {
+		return "", false
+	}
+	qualified := name.AsQualifiedName()
+	if qualified == nil || qualified.Left == nil || qualified.Right == nil {
+		return "", false
+	}
+	left := reactutil.EntityNameRightmost(qualified.Left)
```
elecmonkey: [P2] Require the imported namespace to be leftmost

Using the rightmost identifier of the qualifier makes `Other.React.FC<Props>` appear to be an imported `React.FC`. With a normal React import, upstream rejects this annotation and emits nothing, while Rslint reports and fixes `Props`; the import check must use the leftmost qualifier, matching upstream's `getLeftMostTypeName` logic.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
```diff
@@ -0,0 +1,762 @@
...
+	}
+	if node == nil {
+		return
+	}
+	switch node.Kind {
+	case ast.KindFunctionType:
+		if function := node.AsFunctionTypeNode(); function != nil {
+			validateTypeNode(function.Type, aliases, genericImports, report, seen)
+		}
+	case ast.KindTypeQuery:
```
elecmonkey, line 665: [P2] Reject qualified ReturnType queries

`entityNameText` collapses `typeof ns.makeProps` to `makeProps` and then resolves an unrelated top-level variable with that name. For `const makeProps = () => make<{ name: string }>(); function Hello(props: ReturnType<typeof ns.makeProps>) { return <div/>; }`, upstream emits nothing, but Rslint reports and fixes the local callback's type argument.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
```diff
@@ -0,0 +1,762 @@
...
+			}
+		}
+	}
+}
+
+func functionReturnExpression(node *ast.Node) *ast.Node {
+	if node == nil {
+		return nil
+	}
+	switch node.Kind {
```
elecmonkey, line 725: [P2] Inspect returns from block-bodied arrow callbacks

For an arrow function, this returns the block node itself instead of locating its return expression. Therefore `const makeProps = () => { return make<{ name: string }>(); }; function Hello(props: ReturnType<typeof makeProps>) { return <div/>; }` is reported and fixed upstream but completely missed by Rslint.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
```diff
@@ -0,0 +1,762 @@
...
+	if node == nil {
+		return
+	}
+	if ast.IsFunctionLike(node) {
+		node = functionReturnExpression(node)
+	}
+	if node == nil {
+		return
+	}
+	if node.Kind == ast.KindCallExpression {
```
elecmonkey, line 695: [P2] Traverse generic calls inside returned object spreads

The upstream `ReturnType<typeof callback>` collector also handles object returns containing spread calls with type arguments. For `const makeProps = () => ({ ...make<{ name: string }>() }); function Hello(props: ReturnType<typeof makeProps>) { return <div/>; }`, upstream reports and fixes `name`, while this direct-`CallExpression`-only branch produces no diagnostic.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
```diff
@@ -0,0 +1,762 @@
...
+		}
+		statement := node.AsVariableStatement()
+		if statement == nil || statement.DeclarationList == nil {
+			return false
+		}
+		declarationList := statement.DeclarationList.AsVariableDeclarationList()
+		if declarationList == nil || declarationList.Declarations == nil {
+			return false
+		}
+		for _, declaration := range declarationList.Declarations.Nodes {
```
elecmonkey, line 801: [P3] Preserve upstream multi-declarator ReturnType lookup

Upstream selects a whole variable statement when any declarator matches the queried name, then processes every initializer in that statement. For `const other = () => make<{ other: string }>(), makeProps = () => make<{ name: string }>();` referenced through `ReturnType<typeof makeProps>`, upstream reports both `other` and `name`, while Rslint filters to `makeProps` and reports only `name`.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props_upstream_test.go
```diff
@@ -0,0 +1,202 @@
...
```
elecmonkey: [P2] Map the omitted parser-neutral upstream cases

The v7.37.5 suite has 29 source cases: 17 Flow cases, 10 TypeScript cases, and 2 parser-neutral valid cases. The 10 TypeScript cases are present and the Flow group is explained, but the parser-neutral `Hello.propTypes = { name: PropTypes.string }` case has neither a migrated case nor an explicit equivalent mapping, so the required complete upstream test inventory is still incomplete.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
swwind: Fixed in f2b47cd10. Top-level type aliases and interfaces are now pre-indexed before traversal, while React generic imports remain source-order-sensitive. Added regression coverage for declarations appearing after the component.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
swwind: Fixed in f2b47cd10. Variable annotation lookup now unwraps transparent expression wrappers before reaching the VariableDeclaration fallback. Added the nested-parentheses regression.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
swwind: Fixed in f2b47cd10. The forwardRef arm now returns whenever type arguments exist and only validates the second argument when it is present.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
swwind: Fixed in f2b47cd10. The source-only pragma binding fallback now checks enclosing lexical bindings before the file-wide import scan. Added a permanent source-only shadowing regression.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
swwind: Fixed in f2b47cd10. Qualified React generic annotations now require the imported namespace to be the leftmost identifier.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
swwind: Fixed in f2b47cd10. Qualified ReturnType queries are now rejected instead of being collapsed to an unrelated local identifier.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
swwind: Fixed in f2b47cd10. Block-bodied arrow callbacks now resolve their return expression before collecting returned generic calls.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
swwind: Fixed in f2b47cd10. ReturnType collection now traverses generic calls inside returned object spread assignments.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
swwind: Fixed in f2b47cd10. ReturnType lookup now processes every initializer in a variable statement when any declarator matches the queried name.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props_upstream_test.go
swwind: Fixed in f2b47cd10. The upstream mirror now explicitly retains both parser-neutral valid cases, including Hello.propTypes, and documents the 17 intentionally omitted Flow cases.

internal/plugins/react/reactutil/pragma_binding.go
```diff
@@ -111,6 +105,92 @@ func IsDestructuredFromPragmaImport(ident *ast.Node, pragma string, tc *checker.
...
+					continue
+				}
+				parameter := param.AsParameterDeclaration()
+				if parameter != nil && bindingPatternBindsName(parameter.Name(), name) {
+					return true
+				}
+			}
+		}
+		switch current.Kind {
+		case ast.KindBlock, ast.KindCaseBlock, ast.KindModuleBlock:
```
elecmonkey: [P2] Cover every lexical binding in source-only shadow checks

When no TypeChecker is available, this scan only recognizes selected block declarations and object bindings, missing `for...of` declarations, array destructuring, and `catch` parameters. For example, with an imported `forwardRef` shadowed by `for (const forwardRef of [...])`, eslint-plugin-react 7.37.5 emits nothing, while Rslint treats the local function as React's wrapper and reports/autofixes the mutable props type. Please use binder-backed resolution or cover every lexical declaration and binding-pattern shape.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
```diff
@@ -0,0 +1,810 @@
...
+	return nil
+}
+
+func functionBodyReturnExpression(body *ast.Node) *ast.Node {
+	if body == nil || body.Kind != ast.KindBlock {
+		return nil
+	}
+	var result *ast.Node
+	body.ForEachChild(func(node *ast.Node) bool {
+		if node.Kind == ast.KindReturnStatement {
```
elecmonkey: [P2] Match upstream ReturnType return selection

This traversal stops at the first direct return, whereas upstream searches backward and also descends into the final switch case. With `const makeProps = () => { return make<{ first: string }>(); return make<{ last: string }>(); }`, upstream reports and fixes `last`, but Rslint reports and fixes `first`; a switch-bodied callback is missed entirely. Please mirror upstream's last-return lookup, including its final-switch-case behavior.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
```diff
@@ -0,0 +1,810 @@
...
+		},
+		ast.KindClassDeclaration: func(node *ast.Node) {
+			if !isReactComponentClass(node, pragma, ctx) {
+				return
+			}
+			if propsType := classPropsTypeArgument(node); propsType != nil {
+				validateType(node, propsType)
+			}
+		},
+		ast.KindClassExpression: func(node *ast.Node) {
```
elecmonkey, line 117: [P2] Defer class-expression validation until imports are collected

Upstream intentionally postpones class-expression processing until `Program:exit`, so later React generic imports participate in their props types. For `const Hello = class extends React.Component<PropsWithChildren<{ name: string }>> { render(){return <div/>} }; import { PropsWithChildren } from 'react';`, eslint-plugin-react 7.37.5 reports and fixes `name`, while Rslint validates the class expression before seeing the import and emits nothing.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
```diff
@@ -0,0 +1,810 @@
...
+		return imported, ok && imported != "" && imported != "*" && isGenericTypeName(imported)
+	}
+	if name.Kind != ast.KindQualifiedName {
+		return "", false
+	}
+	qualified := name.AsQualifiedName()
+	if qualified == nil || qualified.Left == nil || qualified.Right == nil {
+		return "", false
+	}
+	left := entityNameLeftmost(qualified.Left)
```
elecmonkey: [P2] Reject extra qualifiers in React generic annotations

Using only the leftmost and rightmost identifiers accepts arbitrary nesting between them. Consequently, `import React from 'react'; type Props = { name: string }; const Hello: React.X.FC<Props> = (props) => <div/>;` is reported and autofixed by Rslint, while eslint-plugin-react 7.37.5 rejects that annotation and emits nothing. Please require the same direct `React.FC` qualified-name shape as upstream.

internal/plugins/react/reactutil/pragma_binding.go
swwind: Fixed in 2cf6f1fac. Source-only wrapper binding resolution now uses the shared lexical scope model, covering for-of declarations, array destructuring, and catch parameters. Added permanent regressions and matching corpus cases.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
swwind: Fixed in 2cf6f1fac. ReturnType callback resolution now selects the last top-level return and recursively inspects the final switch case, matching upstream. Added permanent regressions and matching corpus cases.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
swwind: Fixed in 2cf6f1fac. Class expressions are now queued and validated at Program exit, after source-order React generic imports have been collected. Added permanent regression and corpus case.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
swwind: Fixed in 2cf6f1fac. React generic annotations now require the direct two-part namespace shape (React.FC) and reject extra qualifiers such as React.X.FC. Added permanent regression and corpus case.

internal/plugins/react/reactutil/pragma_binding.go
```diff
@@ -111,65 +102,36 @@ func IsDestructuredFromPragmaImport(ident *ast.Node, pragma string, tc *checker.
...
+		declaration := reference.Resolved()
+		if declaration == nil {
+			return false
 		}
-		n.ForEachChild(func(child *ast.Node) bool {
-			visit(child)
-			return found
-		})
+		if declaration.Kind == scope.DefImport {
+			return importDeclBindsNameFromPragma(declaration.DefNode, ident.AsIdentifier().Text, pragmaLower)
```
elecmonkey, line 134: [P1] Guard import-equals definitions before casting

In a source-only Program, `import forwardRef = React.forwardRef; type Props = { a: string }; const Hello = forwardRef<HTMLDivElement, Props>((props, ref) => <div/>);` resolves `forwardRef` as `scope.DefImport`, but its `DefNode` is an `ImportEqualsDeclaration`. Passing it unconditionally to `importDeclBindsNameFromPragma` calls `AsImportDeclaration()` and panics, terminating the lint run. Check the node kind before casting and handle or reject import-equals declarations.

internal/plugins/react/reactutil/pragma_binding.go
```diff
@@ -111,65 +102,36 @@ func IsDestructuredFromPragmaImport(ident *ast.Node, pragma string, tc *checker.
...
-			if importDeclBindsNameFromPragma(n, name, pragmaLower) {
-				found = true
-				return
-			}
-		case ast.KindVariableDeclaration:
-			if variableDeclBindsNameFromPragma(n, name, pragma, pragmaLower) {
-				found = true
-				return
-			}
+		declaration := reference.Resolved()
```
elecmonkey: [P2] Resolve the latest declaration of merged bindings

The new source-only path calls `Resolved()`, which explicitly returns the first declaration, while upstream uses the latest variable definition. With `var forwardRef = React.forwardRef; var forwardRef = other;` Rslint treats the call as React's wrapper and reports/fixes props although upstream is silent; reversing those declarations makes Rslint miss the diagnostic that upstream emits. Select the last applicable definition from `reference.Declarations` before checking its initializer.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
```diff
@@ -0,0 +1,828 @@
...
+		return
+	}
+	node = reactutil.SkipExpressionWrappers(node)
+	if ast.IsFunctionLike(node) {
+		node = functionReturnExpression(node)
+	}
+	if node == nil {
+		return
+	}
+	if node.Kind == ast.KindCallExpression {
```
elecmonkey, line 695: [P2] Restrict ReturnType collection to function initializers

Upstream processes a `ReturnType<typeof name>` initializer only when `name` is initialized with a function, then inspects that function's returned expression. This branch also interprets a direct generic call as the returned props type. For valid code such as `declare function makeFactory<T>(): () => T; const makeProps = makeFactory<{ a: string }>(); function Hello(props: ReturnType<typeof makeProps>) { return <div/>; }`, eslint-plugin-react 7.37.5 emits nothing, while Rslint reports and autofixes `a` inside the factory call.

internal/plugins/react/reactutil/pragma_binding.go
swwind: Fixed in 1b1c43892. Source-only pragma binding now rejects non-ImportDeclaration definitions before calling AsImportDeclaration, so import-equals bindings no longer panic. Added a permanent source-only regression and matching corpus case.

internal/plugins/react/reactutil/pragma_binding.go
swwind: Fixed in 1b1c43892. Both checker and source-only binding resolution now select the latest non-nil declaration before checking its initializer. Added regressions for both declaration orders and matching corpus cases.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
swwind: Fixed in 1b1c43892. ReturnType<typeof name> collection now only follows function-like top-level initializers; direct generic-call initializers are ignored, while generic calls returned by functions remain supported. Added a permanent regression and matching corpus case.

internal/plugins/react/rules/prefer_read_only_props/prefer_read_only_props.go
```diff
@@ -0,0 +1,832 @@
...
+		return imported, ok && imported != "" && imported != "*" && isGenericTypeName(imported)
+	}
+	if name.Kind != ast.KindQualifiedName {
+		return "", false
+	}
+	qualified := name.AsQualifiedName()
+	if qualified == nil || qualified.Left == nil || qualified.Right == nil {
+		return "", false
+	}
+	if qualified.Left.Kind != ast.KindIdentifier || qualified.Right.Kind != ast.KindIdentifier {
```
elecmonkey, line 515: [P2] Preserve nested React generic type traversal

This direct two-part qualified-name gate makes `reactGenericPropsType` reject nested references such as `React.X.FC<Props>` or `React.X.PropsWithChildren<{ name: string }>` when they are reached from a parameter type, class props wrapper, or referenced alias. Upstream `eslint-plugin-react@7.37.5` only applies the stricter shape check to the variable-annotation fallback; its general TS type walker uses the leftmost imported React name plus the rightmost allowed generic name, so `function Hello(props: React.X.FC<Props>) { return <div/>; }` reports and fixes `Props.name`, while this PR emits no diagnostic. The fix needs to keep the direct-shape rejection for variable annotations but allow the upstream leftmost/rightmost behavior inside the recursive TS prop-type traversal.

swwind: Fixed in 64baf69c8. The strict two-part React generic check remains for variable annotations, while recursive TypeScript prop traversal now follows upstream's leftmost imported namespace and rightmost allowed generic name. Added a regression for `React.X.FC<{ name: string }>` in a function parameter and retained the existing variable-annotation rejection case.

elecmonkey (approved):

PR #2012, opened 2026-09-02
feat: port rule react/jsx-one-expression-per-line

swwind:
## Summary

Port the `jsx-one-expression-per-line` rule from ESLint to rslint.

Require one JSX element or expression per line, with configurable `allow` modes and whitespace-preserving fixes. Five upstream output cases are marked skipped because the local RuleTester applies multipass fixes while ESLint records only the first pass.

## Related Links

- ESLint rule: https://github.com/jsx-eslint/eslint-plugin-react/blob/v7.37.5/docs/rules/jsx-one-expression-per-line.md
- Source code: https://github.com/jsx-eslint/eslint-plugin-react/blob/v7.37.5/lib/rules/jsx-one-expression-per-line.js

## Checklist


internal/plugins/react/rules/jsx_one_expression_per_line/jsx_one_expression_per_line.go
```diff
@@ -0,0 +1,357 @@
...
+	if start < 0 || end < start || end > len(source) {
+		return ""
+	}
+	return source[start:end]
+}
+
+func fixSource(source string) string {
+	source = strings.TrimLeft(source, " ")
+	if strings.HasSuffix(source, "\r\n") {
+		return strings.TrimRight(source[:len(source)-2], " ") + "\r\n"
```
elecmonkey: [P2] This CRLF-specific trim removes a trailing ASCII space before `\r\n`, but upstream's `details.source.replace(/(^ +| +(?=\n)*$)/g, '')` does not remove that space in CRLF JSX text. Repro: `<App>foo \r\n</App>\r\n`; ESLint 8.57.1 + eslint-plugin-react 7.37.5 fixes to `<App>\nfoo \r\n</App>\r\n`, while this PR fixes to `<App>\nfoo\r\n</App>\r\n`, changing the text content and breaking autofix parity.

internal/plugins/react/rules/jsx_one_expression_per_line/jsx_one_expression_per_line.go
```diff
@@ -0,0 +1,357 @@
...
+			Id:          "moveToNewLine",
+			Description: "`" + currentDetails.descriptor + "` must be placed on a new line",
+			Data:        map[string]string{"descriptor": currentDetails.descriptor},
+		}
+		child := currentDetails.child
+		ctx.ReportRangeWithDeferredFixes(core.NewTextRange(child.Pos(), child.End()), message, func() []rule.RuleFix {
+			fixedSource := fixSource(nodeSource(source, child))
+			leadingSpace := ""
+			if currentDetails.leadingSpace {
+				leadingSpace = "\n{' '}"
```
elecmonkey, line 406: [P2] Because Rslint's fixer applies adjacent replacement ranges in the same pass, this leading-space marker is kept even when the previous JSXText child is also fixed and has already trimmed that space away. Repro: `<App>foo <input /></App>`; upstream `verifyAndFix` returns `<App>\nfoo\n<input />\n</App>`, but this PR returns `<App>\nfoo\n{' '}\n<input />\n</App>`. The same happens for `<App>foo {value}</App>`, so `--fix` output is not 1:1 with upstream for common inline JSX children.

internal/plugins/react/rules/jsx_one_expression_per_line/jsx_one_expression_per_line.go
swwind: Fixed in 11e05ead7. `fixSource` now preserves an ASCII space before CRLF, matching the pinned ESLint implementation. Added a regression case for `<App>foo \r\n</App>\r\n`; the focused rule test passes.

internal/plugins/react/rules/jsx_one_expression_per_line/jsx_one_expression_per_line.go
swwind: Fixed in 11e05ead7. The fix now tracks the preceding JSXText node and omits `{' '}` when that text is also fixed in the same pass and its trailing gap is removed. Added regressions for `<App>text <Foo /></App>` and `<App>foo {value}</App>`; the focused rule test passes.

internal/plugins/react/rules/jsx_one_expression_per_line/jsx_one_expression_per_line.go
```diff
@@ -0,0 +1,375 @@
...
+func fixSource(source string) string {
+	source = strings.TrimLeft(source, " ")
+	if strings.HasSuffix(source, "\n") {
+		// The upstream regular expression trims spaces before LF, but not
+		// before CRLF because the carriage return sits between the spaces and
+		// the line feed.
+		if len(source) >= 2 && source[len(source)-2] == '\r' {
+			return source
+		}
+		return strings.TrimRight(source[:len(source)-1], " ") + source[len(source)-1:]
```
elecmonkey: [P2] Preserve spaces immediately before terminal line breaks

When a non-whitespace JSX text node ends with an ASCII space immediately before LF, this branch removes the space. For `<App>foo 
</App>`, eslint-plugin-react@7.37.5 with ESLint 8.57.1 fixes to `<App>
foo 
</App>`, while PR-head Rslint produces `<App>
foo
</App>`; the real Rslint CLI reproduces the same result. A bare CR has the same problem at line 66. The upstream regex only removes spaces at the actual end of the node text, so spaces preceding a line terminator must remain.

swwind: Fixed in a9858d7f7. `fixSource` now trims only ASCII spaces at the actual ends of the JSX text source, preserving spaces immediately before LF, CRLF, and CR, matching eslint-plugin-react@7.37.5. Added LF and bare-CR regressions; the focused rule test and the pinned ESLint 8.57.1 corpus differential pass.

internal/plugins/react/rules/jsx_one_expression_per_line/jsx_one_expression_per_line.go
```diff
@@ -0,0 +1,362 @@
...
+	leadingSpace     bool
+	leadingSpaceFrom []*ast.Node
+	leadingNewline   bool
+	trailingSpace    bool
+	trailingNewline  bool
+}
+
+func shouldKeepLeadingSpace(details *fixDetails, detailsByChild map[*ast.Node]*fixDetails) bool {
+	for _, previous := range details.leadingSpaceFrom {
+		if !isJSXText(previous) || detailsByChild[previous] == nil {
```
elecmonkey: [P2] Preserve spaces after fixed JSXText siblings

When the previous child is JSXText and is also fixed, this returns false for the following child's leading-space marker even if that previous text is between non-text siblings and its own replacement trims the trailing space. Repro: `<App>{x}a {y}</App>` with `react/jsx-one-expression-per-line`; ESLint 8.57.1 + eslint-plugin-react 7.37.5 `verifyAndFix` yields `<App>\n{x}\na \n{' '}\n{y}\n</App>`, while PR-head Rslint yields `<App>\n{x}\na\n{y}\n</App>`, dropping the rendered space between `a` and `{y}`.

swwind: Fixed in 94fcb5bc6. The fixer now distinguishes the first effective JSX child from JSXText between non-text siblings, preserving the rendered trailing space and {' '} marker for <App>{x}a {y}</App>. Added the regression to the persistent corpus and Go extras test. The ESLint 8.57.1 differential case and focused rule package pass.

internal/plugins/react/rules/jsx_one_expression_per_line/jsx_one_expression_per_line.go
```diff
@@ -0,0 +1,376 @@
...
+		message := rule.RuleMessage{
+			Id:          "moveToNewLine",
+			Description: "`" + currentDetails.descriptor + "` must be placed on a new line",
+			Data:        map[string]string{"descriptor": currentDetails.descriptor},
+		}
+		child := currentDetails.child
+		ctx.ReportRangeWithDeferredFixes(core.NewTextRange(child.Pos(), child.End()), message, func() []rule.RuleFix {
+			rawSource := nodeSource(source, child)
+			fixedSource := fixSource(rawSource)
+			if isJSXText(child) && !currentDetails.firstChild && hasTrailingTextSpace(rawSource) {
```
elecmonkey: [P2] Avoid duplicating a space represented by the trailing marker

When a non-first JSXText child ends with an ASCII space and also shares its line with the closing tag, this restores the trimmed space even though `trailingSpace` adds a `{' '}` marker. Repro: `<A>{x}a </A>` with `react/jsx-one-expression-per-line`; ESLint 8.57.1 + eslint-plugin-react 7.37.5 fixes to `<A>\n{x}\na\n{' '}\n</A>`, while PR-head Rslint produces `<A>\n{x}\na \n{' '}\n</A>`, duplicating the rendered space. Only restore the source space when `currentDetails.trailingSpace` is false; that guard also preserves the recently fixed `<A>{x}a {y}</A>` behavior.

swwind: Fixed in 72bbc5d62. The fixer now skips restoring the source trailing space when `trailingSpace` is already true, so it emits `<A>\n{x}\na\n{\x27 \x27}\n</A>` instead of duplicating the rendered space. Added the regression to the persistent corpus and Go extras test. The ESLint 10 compatibility driver and focused rule package pass.

internal/plugins/react/rules/jsx_one_expression_per_line/jsx_one_expression_per_line.go
```diff
@@ -0,0 +1,376 @@
...
+		message := rule.RuleMessage{
+			Id:          "moveToNewLine",
+			Description: "`" + currentDetails.descriptor + "` must be placed on a new line",
+			Data:        map[string]string{"descriptor": currentDetails.descriptor},
+		}
+		child := currentDetails.child
+		ctx.ReportRangeWithDeferredFixes(core.NewTextRange(child.Pos(), child.End()), message, func() []rule.RuleFix {
+			rawSource := nodeSource(source, child)
+			fixedSource := fixSource(rawSource)
+			if isJSXText(child) && !currentDetails.firstChild && !currentDetails.trailingSpace && hasTrailingTextSpace(rawSource) {
```
elecmonkey: [P2] Avoid preserving text spaces already emitted as markers

For a non-first JSXText child whose previous sibling is a JSX element, this branch restores the raw trailing space even when the next JSX child also emits a leading `{' '}` marker. Repro: `<App>foo <A /> bar <B /></App>` with ESLint 8.57.1 + eslint-plugin-react 7.37.5 fixes to `<App>\nfoo\n<A />\n{' '}\nbar\n<B />\n</App>`, but PR-head Rslint fixes to `<App>\nfoo\n<A />\n{' '}\nbar \n{' '}\n<B />\n</App>`, duplicating the rendered space before `<B />`. The previous guard fixed first-text-child cases, but this remaining path still breaks autofix parity for inline text between JSX siblings.

swwind:
In fact, I do not think it is necessary to port this style rule. We should re-evaluate #225 regarding rules of this kind.

~~This rule seems to have infinite false cases to fix~~

internal/plugins/react/rules/jsx_one_expression_per_line/jsx_one_expression_per_line.go
swwind: Fixed in 0682df2e1. The fixer now tracks suppressed leading-space markers across JSX siblings, so a JSXText trailing space is not restored when the following JSX element's marker already accounts for that gap. Added \u003cApp\u003efoo \u003cA /\u003e bar \u003cB /\u003e\u003c/App\u003e to the persistent corpus and Go extras regression; the ESLint 8.57.1 / eslint-plugin-react 7.37.5 differential and focused tests pass.

internal/plugins/react/rules/jsx_one_expression_per_line/jsx_one_expression_per_line.go
```diff
@@ -0,0 +1,402 @@
...
+		}
+	}
+	return false
+}
+
+func shouldKeepTrailingTextSpace(details *fixDetails, detailsByChild map[*ast.Node]*fixDetails) bool {
+	if details.firstChild || details.trailingSpace {
+		return false
+	}
+	return !hasSuppressedJSXPredecessor(details, detailsByChild)
```
elecmonkey: [P2] Avoid duplicating spaces after consecutive non-text children

When a trailing-space JSXText child follows at least two non-text children, this predicate fails to recognize that the following child’s marker accounts for the same gap. For `<A><B />{x}a {y}</A>`, eslint-plugin-react@7.37.5 fixes to `<A>\n<B />\n{x}\na\n{y}\n</A>`, while PR-head Rslint produces `<A>\n<B />\n{x}\na \n{' '}\n{y}\n</A>`, preserving the raw trailing space and adding another explicit space. The generated differential corpus reproduced this across elements, expressions, fragments, and every `allow` mode.

elecmonkey: Fixed in `b75c4e805`, by replacing the predicate with the rule it was approximating.

ESLint's `SourceCodeFixer` skips a report whose range starts exactly where the previously applied range ended, so when children abut, eslint-plugin-react only lands every other fix per pass; the skipped reports are rebuilt against the already rewritten source in a later pass. That parity is what decides the outcome: if a text child's own fix lands first, its raw trailing space is trimmed and the following sibling later sees no space, so no marker is emitted; if it is skipped, the sibling's fix inserts a line break behind the space first, so the space stops being trailing and survives every later pass alongside the marker.

The rule now computes that first-pass acceptance directly and keeps a raw trailing space only for a text child whose own fix would be skipped. Leading markers are also split by source: a space the child's own text starts with, or a gap between the two children, always survives, while a space contributed only by a preceding text sibling survives just while that sibling keeps it. `hasSuppressedJSXPredecessor` is gone.

`<A><B />{x}a {y}</A>` now fixes to `<A>\n<B />\n{x}\na\n{y}\n</A>`, matching eslint-plugin-react@7.37.5. Ten cases covering the parity — with and without a leading element, chained text/expression runs, own-leading-space text, whitespace-only sibling gaps, and trailing text before the closing tag — are added as regression tests with outputs taken from upstream, and the package tests pass.

elecmonkey (approved):

fansenze (approved):

PR #2013, opened 2026-09-02
feat: port rule react/no-arrow-function-lifecycle

swwind:
## Summary

Port the `react/no-arrow-function-lifecycle` rule from ESLint to rslint.

Disallow arrow functions for React lifecycle methods in class components and `createReactClass` objects, with autofixes to prototype/object methods.

## Related Links

- ESLint rule: https://github.com/jsx-eslint/eslint-plugin-react/blob/v7.37.5/docs/rules/no-arrow-function-lifecycle.md
- Source code: https://github.com/jsx-eslint/eslint-plugin-react/blob/v7.37.5/lib/rules/no-arrow-function-lifecycle.js

## Checklist

- [x] Tests updated.
- [x] Documentation updated.


internal/plugins/react/rules/no_arrow_function_lifecycle/no_arrow_function_lifecycle.go
```diff
@@ -0,0 +1,261 @@
...
+	}
+}
+
+// NoArrowFunctionLifecycleRule reports arrow-valued lifecycle members in
+// React class components and createReactClass object components.
+var NoArrowFunctionLifecycleRule = rule.Rule{
+	Name:   "react/no-arrow-function-lifecycle",
+	Schema: rule.EmptyArraySchema,
+	Run: func(ctx rule.RuleContext, options []any) rule.RuleListeners {
+		pragma := reactutil.GetReactPragma(ctx.Settings)
```
elecmonkey: Upstream `pragmaUtil.getFromContext` reads file-level `@jsx` comments before `settings.react.pragma`. With `/* @jsx Preact.h */ class Hello extends Preact.Component { render = () => null; }`, eslint-plugin-react@7.37.5 reports and fixes `render`, but this port keeps the settings-only pragma as `React` and skips the class. Please use the comment-aware pragma helper here so this rule preserves upstream component detection.

internal/plugins/react/rules/no_arrow_function_lifecycle/no_arrow_function_lifecycle.go
```diff
@@ -0,0 +1,261 @@
...
+		createClass := reactutil.GetReactCreateClass(ctx.Settings)
+		var visit ast.Visitor
+		visit = func(node *ast.Node) bool {
+			if node == nil {
+				return false
+			}
+			switch node.Kind {
+			case ast.KindClassDeclaration, ast.KindClassExpression:
+				if reactutil.ExtendsReactComponent(node, pragma) {
```
elecmonkey: Upstream `componentUtil.isES6Component` also treats classes annotated with `@extends React.Component` or `@augments React.Component` as React components even without a syntactic `extends React.Component`. For `/** @extends React.Component */ class Hello { render = () => null; }`, upstream reports and fixes `render`, but this branch only calls `ExtendsReactComponent`, so the port misses that valid upstream component shape.

internal/plugins/react/rules/no_arrow_function_lifecycle/no_arrow_function_lifecycle.go
swwind: Fixed: the rule now uses `reactutil.GetReactPragmaFromContext(ctx)`, so it prioritizes file-level `@jsx` annotations. Added a regression test for `/* @jsx Preact.h */`.

internal/plugins/react/rules/no_arrow_function_lifecycle/no_arrow_function_lifecycle.go
swwind: Fixed: added component detection aligned with upstream `componentUtil.isExplicitComponent` for `@extends/@augments React.Component/PureComponent`, with regression tests.

internal/plugins/react/rules/no_arrow_function_lifecycle/no_arrow_function_lifecycle.go
```diff
@@ -0,0 +1,304 @@
...
+}
+
+func parameterText(sf *ast.SourceFile, fn *ast.Node) string {
+	params := reactutil.FunctionParameters(fn)
+	if len(params) == 0 {
+		return ""
+	}
+	first := utils.TrimNodeTextRange(sf, params[0])
+	last := utils.TrimNodeTextRange(sf, params[len(params)-1])
+	return sourceSlice(sf, first.Pos(), last.End())
```
elecmonkey, line 265: This builds the replacement parameter list from the full source range of the first through last parameter, but eslint-plugin-react@7.37.5 uses `node.value.params.map((p) => p.name)` in `getRuleText`. For cases like `render = (value = 1, ...rest) => value` or `render: ({ value }, [item]) => value`, upstream reports the same diagnostic but fixes to `render(, ) { return value; }` / `function(, )`, while this PR fixes to valid full parameter text and even adds tests that lock that in. Since this PR is presented as a straight v7.37.5 port, please either match the upstream fix output or document this as an intentional divergence.

internal/plugins/react/rules/no_arrow_function_lifecycle/no_arrow_function_lifecycle.go
```diff
@@ -0,0 +1,304 @@
...
+			bodyEnd = rawValueEnd
+		}
+	}
+
+	params := parameterText(ctx.SourceFile, fn)
+	methodText := "(" + params + ") "
+	if member.Kind == ast.KindPropertyAssignment {
+		methodText = ": function(" + params + ") "
+	}
+	fixes := []rule.RuleFix{rule.RuleFixReplaceRange(core.NewTextRange(name.End(), headEnd), methodText)}
```
elecmonkey: For computed lifecycle keys, this replacement starts after the whole ts-go computed-name node, so `[render] = () => null` becomes the valid method `[render]() { return null; }`. Upstream uses `node.key.range[1]` where the key is the inner ESTree identifier, so the same inputs are reported but fixed to malformed text such as `[render() { return null; }` and `[render: function() { return null; }`. The extras tests currently lock in the improved output, but without an explicit accepted divergence this is another autofix mismatch from the declared upstream rule.

internal/plugins/react/rules/no_arrow_function_lifecycle/no_arrow_function_lifecycle.go
swwind: Thanks for catching this. I verified eslint-plugin-react@7.37.5 uses `node.value.params.map((p) => p.name)`, which produces malformed `render(, )` / `function(, )` for default, rest, and destructured parameters. I’m intentionally keeping rslint’s complete parameter text because it preserves valid syntax. I documented this deliberate divergence and retained regression coverage plus the corpus cases. Verified with `go test ./internal/plugins/react/rules/no_arrow_function_lifecycle`.

internal/plugins/react/rules/no_arrow_function_lifecycle/no_arrow_function_lifecycle.go
swwind: Thanks for catching this. I verified eslint-plugin-react@7.37.5 starts the replacement at the inner computed-key expression, which produces malformed `[render()` / `[render: function()` output. I’m intentionally replacing after the complete computed key so rslint preserves valid syntax. I documented this deliberate divergence and retained regression coverage plus the corpus cases. Verified with `go test ./internal/plugins/react/rules/no_arrow_function_lifecycle`.

internal/plugins/react/rules/no_arrow_function_lifecycle/no_arrow_function_lifecycle.go
```diff
@@ -0,0 +1,310 @@
...
+		}
+		for _, tag := range jd.Tags.Nodes {
+			if !ast.IsJSDocAugmentsTag(tag) {
+				continue
+			}
+			className := tag.AsJSDocAugmentsTag().ClassName
+			if className == nil {
+				continue
+			}
+			if isExplicitReactComponentType(className) {
```
elecmonkey: This path accepts `tag.AsJSDocAugmentsTag().ClassName` for both `/** @extends React.Component */` and `/** @extends {React.Component} */`, but eslint-plugin-react@7.37.5 uses doctrine's `tag.name` and only reports the unbraced form; doctrine parses the braced form as `tag.type.name` with `tag.name === null`. I verified the difference with `/** @extends {React.Component} */ class Hello { render = () => null; }`: upstream reports 0 diagnostics, while this port reports and fixes `render`, so users get a false positive for a form upstream ignores.

swwind: Fixed in 7379e7681. The rule now ignores braced JSDoc forms such as `@extends {React.Component}`, matching eslint-plugin-react@7.37.5's doctrine-based `tag.name` check. Added a valid regression test and corpus case. Focused Go tests, golangci-lint, and cspell all pass.

internal/plugins/react/rules/no_arrow_function_lifecycle/no_arrow_function_lifecycle.go
```diff
@@ -0,0 +1,330 @@
...
+	// them as ParenthesizedExpression nodes, so consume their closing suffix
+	// together with an expression body when converting the member.
+	if rawValue != nil {
+		rawValueEnd := utils.TrimNodeTextRange(ctx.SourceFile, rawValue).End()
+		if rawValueEnd > bodyEnd {
+			bodyEnd = rawValueEnd
+		}
+	}
+
+	params := parameterText(ctx.SourceFile, fn)
```
elecmonkey, line 346: [P1] For an async lifecycle such as `componentDidMount = async () => { await connect(); }`, this replacement emits `componentDidMount() { await connect(); }`, which is invalid syntax; even without `await`, dropping `async` changes the method's return semantics. Preserve the arrow's async modifier on the generated method/function, or decline to autofix these members.

elecmonkey, line 346: [P1] For `render = <T,>(x: T): T => x`, `parameterText` retains `x: T` while the head replacement deletes `<T,>`, producing `render(x: T) { return x; }` with an undefined `T`. Preserve the arrow's type parameters when constructing the method signature, or suppress the fix when they are present.

internal/plugins/react/rules/no_arrow_function_lifecycle/no_arrow_function_lifecycle.go
```diff
@@ -0,0 +1,330 @@
...
+	methodText := "(" + params + ") "
+	if member.Kind == ast.KindPropertyAssignment {
+		methodText = ": function(" + params + ") "
+	}
+	// NOTE: Unlike ESLint v7.37.5, replace after the complete computed key.
+	// The upstream range can remove the closing bracket and produce invalid
+	// JavaScript for keys such as `[render]`.
+	fixes := []rule.RuleFix{rule.RuleFixReplaceRange(core.NewTextRange(name.End(), headEnd), methodText)}
```
elecmonkey: [P1] Because the replacement begins after the member name, field-only modifiers remain untouched: valid inputs such as `readonly render = () => null` and `accessor render = () => null` become invalid `readonly render()` and `accessor render()` methods. The fixer should remove incompatible modifiers or refrain from fixing these declarations.

internal/plugins/react/rules/no_arrow_function_lifecycle/no_arrow_function_lifecycle.go
```diff
@@ -0,0 +1,330 @@
...
+	}
+	between := sourceSlice(sf, tagName.End(), className.Pos())
+	return !strings.Contains(between, "{")
+}
+
+func isExplicitReactComponentType(node *ast.Node) bool {
+	if node == nil {
+		return false
+	}
+	if node.Kind == ast.KindExpressionWithTypeArguments {
```
elecmonkey, line 237: [P2] Unwrapping `ExpressionWithTypeArguments` makes `/** @extends React.Component<Props> */ class A { render = () => null }` a component, so Rslint reports and fixes it, while eslint-plugin-react@7.37.5 reports nothing because doctrine does not produce the exact `tag.name === 'React.Component'` required by upstream. The JSDoc check must reject generic and otherwise decorated names rather than normalizing them to `React.Component`.

internal/plugins/react/rules/no_arrow_function_lifecycle/no_arrow_function_lifecycle.go
```diff
@@ -0,0 +1,330 @@
...
+
+// hasExplicitReactComponentJSDoc reports whether a class has an unbraced JSDoc
+// `@extends React.Component` / `@augments React.Component` tag. This matches
+// eslint-plugin-react's doctrine-based check, whose tag.name is nil for the
+// braced `@extends {React.Component}` form.
+func hasExplicitReactComponentJSDoc(sf *ast.SourceFile, classNode *ast.Node) bool {
+	if classNode == nil {
+		return false
+	}
+	for _, doc := range classNode.JSDoc(nil) {
```
elecmonkey: [P2] Iterating every entry returned by `classNode.JSDoc(nil)` accepts older comments that upstream's `sourceCode.getJSDocComment(node)` does not select. For `/** @extends React.Component */ /** other */ class A { render = () => null }`, upstream reports zero diagnostics while this implementation reports and fixes `render`; limit detection to the equivalent nearest attached JSDoc comment.

internal/plugins/react/rules/no_arrow_function_lifecycle/no_arrow_function_lifecycle.go
```diff
@@ -0,0 +1,330 @@
...
+// other rules use the first argument as their component-shape boundary.
+func isCreateReactClassObjectComponent(obj *ast.Node, pragma, createClass string) bool {
+	if obj == nil || obj.Kind != ast.KindObjectLiteralExpression {
+		return false
+	}
+	arg := obj
+	for arg.Parent != nil && arg.Parent.Kind == ast.KindParenthesizedExpression {
+		arg = arg.Parent
+	}
+	if arg.Parent == nil || arg.Parent.Kind != ast.KindCallExpression {
```
elecmonkey: [P2] Restricting the parent to `CallExpression` misses `new createReactClass({ render: () => null })`. eslint-plugin-react@7.37.5 treats that object as an ES5 component because its parent has the matching identifier callee, and reports/fixes `render`; this port emits no diagnostic. Handle the corresponding `NewExpression` shape as well.

internal/plugins/react/rules/no_arrow_function_lifecycle/no_arrow_function_lifecycle.go
```diff
@@ -0,0 +1,330 @@
...
+		return nil
+	}
+
+	name := member.Name()
+	if name == nil {
+		return nil
+	}
+	bodyRange := utils.TrimNodeTextRange(ctx.SourceFile, body)
+	bodyStart, bodyEnd := bodyRange.Pos(), bodyRange.End()
+	previousToken, hasPreviousToken := utils.TokenBeforePosition(ctx.SourceFile, bodyStart)
```
elecmonkey, line 307: [P3] The comment scan runs for block-bodied arrows even though upstream only calls `getCommentsBefore` and `getCommentsAfter` for expression bodies. Consequently, `render = () => /*keep*/ {}` becomes `render() /*keep*/ {}` here but `render() {}` in eslint-plugin-react@7.37.5. Either restrict this handling to expression bodies or explicitly document and test it as another intentional autofix divergence.

elecmonkey (commented):
I found several behavior and autofix issues that should be addressed before this port is considered aligned and safe.

internal/plugins/react/rules/no_arrow_function_lifecycle/no_arrow_function_lifecycle.go
swwind: Fixed. The fixer now preserves  when converting lifecycle arrow properties to methods/functions. Added regression coverage for async class fields.

internal/plugins/react/rules/no_arrow_function_lifecycle/no_arrow_function_lifecycle.go
swwind: Fixed in `3c628c69f`. The rule still reports `readonly` and `accessor` fields but suppresses their autofixes, avoiding invalid method syntax. Added regression tests and documented this behavior.

internal/plugins/react/rules/no_arrow_function_lifecycle/no_arrow_function_lifecycle.go
swwind: Fixed in `3c628c69f`. JSDoc detection now rejects generic and otherwise decorated component names, matching doctrine's exact-name behavior. Added valid regression coverage.

internal/plugins/react/rules/no_arrow_function_lifecycle/no_arrow_function_lifecycle.go
swwind: Fixed in `3c628c69f`. Detection now uses the nearest attached JSDoc block, matching `SourceCode#getJSDocComment`. Added regression coverage for an older adjacent marker.

internal/plugins/react/rules/no_arrow_function_lifecycle/no_arrow_function_lifecycle.go
swwind: Fixed in `3c628c69f`. Matching `NewExpression` component objects are now detected and fixed, alongside ordinary calls. Added regression coverage.

internal/plugins/react/rules/no_arrow_function_lifecycle/no_arrow_function_lifecycle.go
swwind: Fixed in `3c628c69f`. Comment preservation is now restricted to expression-bodied arrows, matching eslint-plugin-react v7.37.5. Added regression coverage.

internal/plugins/react/rules/no_arrow_function_lifecycle/no_arrow_function_lifecycle.go
swwind: Fixed in `3c628c69f`. The fixer now preserves `async`: class fields become `async componentDidMount()`, and object properties become `: async function(...)`. Added regression coverage and verified the focused rule tests and corpus.

swwind: Fixed in `3c628c69f`. The fixer now preserves authored arrow type parameters, including `<T,>`, when generating the method or function signature. Added regression coverage and verified the focused rule tests and corpus.

elecmonkey (approved):

PR #2023, opened 2026-09-02
refactor(rstest): match each plugin-managed API by its own call shape

elecmonkey:
## Summary

`importActual` and `requireActual` are rewritten through a bracketed string key and through an optional call, and a receiver the file declares itself takes the call back. The other twelve plugin-managed members are rewritten on the name as written and nothing else. rslint matched all fourteen with one shape, so the shape now lives per member in a table that `ParseRstestPluginManagedCall` applies.

`require-mock-type-parameters` under `checkImportFunctions` therefore stops reporting three shapes Rstest never rewrites — `rs['importMock']('./m')`, `rs.importMock?.('./m')` and ``rs[`importActual`]('./m')`` — and starts reporting `rs.importMock('./m')` under a local `const rs = { … }`, which the build rewrites regardless.

Measured on `@rstest/core` 0.11.11, one file per shape:

| shape | `importActual` / `requireActual` | the other twelve |
| --- | --- | --- |
| `rs.x(…)` | rewritten | rewritten |
| `rs['x'](…)` | rewritten | throws |
| ``rs[`x`](…)`` | throws | throws |
| `rs.x?.(…)` | rewritten | throws |
| `rs?.x(…)` | throws | throws |

A renamed binding is rewritten for no member.

## Related Links

Related #935

## Checklist


swwind (approved):
No issues found here.

PR #2027, opened 2026-09-02
feat(rstest): add no-restricted-rstest-methods rule

elecmonkey:
## Summary

Disallow members of the Rstest utilities object that a team has decided not to use, each with an optional message naming what to reach for instead. Nothing is disallowed by default, and the rule stays out of the recommended preset.

Because the member names come from configuration, the rule cannot know which half of the object it will be handed, so it looks each one up in the shared table and matches it the way Rstest reaches it. The two halves pull in opposite directions. `fn`, `spyOn`, `mocked` and the other ordinary functions follow the binding, so `import { rs as mocker }` is reported and a local `const rs = { … }` is not, and every call shape counts because they all really call the function. The module mock APIs are read as written, so a renamed binding is not reported — it throws where it stands rather than mocking anything — and a local declaration of the receiver is reported, because the build rewrites the call regardless. For the same reason `rs['fn']()` is reported and `rs['mock']()` is not: the first really calls `fn`, the second calls nothing.

`IsUtilitiesObject` also learns one receiver shape here: `import.meta.rstest.rs.fn()`, which reaches the same object through the module namespace.

The rule does not check where the call stands or what it is passed. Those decide whether a call that is already a use of the API is hoisted or valid, which is `hoisted-apis-on-top`'s question, not this one's.

## Related Links

Related #935

## Checklist

### DIFF
--- a/internal/plugins/rstest/rules/no_restricted_rstest_methods/no_restricted_rstest_methods.go
+++ b/internal/plugins/rstest/rules/no_restricted_rstest_methods/no_restricted_rstest_methods.go
@@ -1,0 +1,171 @@
+package no_restricted_rstest_methods
+
+import (
+	_ "embed"
+
+	"github.com/microsoft/typescript-go/shim/ast"
+	rstestUtils "github.com/web-infra-dev/rslint/internal/plugins/rstest/utils"
+	"github.com/web-infra-dev/rslint/internal/rule"
+	internalUtils "github.com/web-infra-dev/rslint/internal/utils"
+)
+
+//go:embed no_restricted_rstest_methods.schema.json
+var schemaJSON []byte
+
+// restriction is one entry of the option object: the member is disallowed, and
+// the author may have written a message saying what to reach for instead.
+type restriction struct {
+	Message    string
+	HasMessage bool
+}
+
+func parseOptions(rawOptions []any) map[string]restriction {
+	if len(rawOptions) == 0 {
+		return nil
+	}
+	raw, ok := rawOptions[0].(map[string]any)
+	if !ok {
+		return nil
+	}
+
+	restricted := make(map[string]restriction, len(raw))
+	for member, rawMessage := range raw {
+		if rawMessage == nil {
+			restricted[member] = restriction{}
+			continue
+		}
+		message, ok := rawMessage.(string)
+		if !ok {
+			continue
+		}
+		if message == "" {
+			restricted[member] = restriction{}
+			continue
+		}
+		restricted[member] = restriction{Message: message, HasMessage: true}
+	}
+	return restricted
+}
+
+func restrictedMethodMessage(member string) rule.RuleMessage {
+	return rule.RuleMessage{
+		Id:          "restrictedRstestMethod",
+		Description: "Use of `" + member + "` is disallowed",
+		Data:        map[string]string{"restriction": member},
+	}
+}
+
+func restrictedMethodWithMessage(member string, message string) rule.RuleMessage {
+	return rule.RuleMessage{
+		Id:          "restrictedRstestMethodWithMessage",
+		Description: message,
+		Data: map[string]string{
+			"message":     message,
+			"restriction": member,
+		},
+	}
+}
+
+var NoRestrictedRstestMethodsRule = rule.Rule{
+	Name:   "rstest/no-restricted-rstest-methods",
+	Schema: rule.NewSchema(schemaJSON),
+	Run: func(ctx rule.RuleContext, options []any) rule.RuleListeners {
+		restricted := parseOptions(options)
+		if len(restricted) == 0 {
+			return rule.RuleListeners{}
+		}
+
+		report := func(memberNode *ast.Node, member string) {
+			entry := restricted[member]
+			if entry.HasMessage {
+				ctx.ReportNode(memberNode, restrictedMethodWithMessage(member, entry.Message))
+				return
+			}
+			ctx.ReportNode(memberNode, restrictedMethodMessage(member))
+		}
+
+		return rule.RuleListeners{
+			ast.KindCallExpression: func(node *ast.Node) {
+				// The two kinds of member on the utilities object are reached
+				// in opposite ways, and which kind a name is cannot be known
+				// until the option object is read, so both are tried here.
+				if managed := rstestUtils.ParseRstestPluginManagedCall(node); managed != nil {
+					if _, ok := restricted[managed.Member]; !ok {
+						return
+					}
+					// The build hands the two members that resolve their
+					// receiver back to a local declaration, so a call on the
+					// file's own object is that object's, not Rstest's.
+					if managed.API.ResolvesReceiver &&
+						rstestUtils.ReceiverIsLocallyDeclared(ctx, managed.NamespaceNode) {
+						return
+					}
+					report(managed.MemberNode, managed.Member)
+					return
+				}
+
+				memberNode, member, receiver := calledMember(node)
+				if memberNode == nil {
+					return
+				}
+				// A plugin-managed member that did not parse above is written
+				// in a shape the build does not rewrite, so the call reaches
+				// the stub that throws rather than the API being restricted.
+				if rstestUtils.IsPluginManagedAPI(member) {
+					return
+				}
+				if _, ok := restricted[member]; !ok {
+					return
+				}
+				if !rstestUtils.IsUtilitiesObject(ctx, receiver) {
+					return
+				}
+				report(memberNode, member)
+			},
+		}
+	},
+}
+
+// calledMember returns the member a call reaches its callee through, the name
+// it is written as, and the expression it is read off.
+//
+// Every shape counts here, because the members matched this way are ordinary
+// functions on an ordinary object: `rs['fn']()`, `rs.fn?.()` and `rs?.fn()`
+// all really call `fn`. That is the opposite of the plugin-managed members,
+// where the shape decides whether anything runs at all.
+func calledMember(node *ast.Node) (*ast.Node, string, *ast.Node) {
+	call := node.AsCallExpression()
+	if call == nil {
+		return nil, "", nil
+	}
+	callee := internalUtils.SkipAssertionsAndParens(call.Expression)
+	if callee == nil {
+		return nil, "", nil
+	}
+
+	switch callee.Kind {
+	case ast.KindPropertyAccessExpression:
+		access := callee.AsPropertyAccessExpression()
+		if access == nil {
+			return nil, "", nil
+		}
+		name := access.Name()
+		if name == nil || name.Kind != ast.KindIdentifier {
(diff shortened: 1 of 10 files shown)


internal/plugins/rstest/rules/no_restricted_rstest_methods/no_restricted_rstest_methods.go
```diff
@@ -0,0 +1,171 @@
...
+				// A plugin-managed member that did not parse above is written
+				// in a shape the build does not rewrite, so the call reaches
+				// the stub that throws rather than the API being restricted.
+				if rstestUtils.IsPluginManagedAPI(member) {
+					return
+				}
+				if _, ok := restricted[member]; !ok {
+					return
+				}
+				if !rstestUtils.IsUtilitiesObject(ctx, receiver) {
```
swwind, line 120: [P2] Recognize direct CommonJS namespace bindings

`IsUtilitiesObject` does not recognize a direct CommonJS namespace binding here. `ResolveFunctionIdentifierReferenceFromSymbolModules` only handles `BindingElement` require declarations, so `const rs = require('@rstest/core'); rs.fn();` resolves as a local declaration and produces no diagnostic even though the rule documentation explicitly promises that `require` forms are reported. Please handle the `VariableDeclaration` require form and add a regression test.

internal/plugins/rstest/rules/no_restricted_rstest_methods/no_restricted_rstest_methods.go
```diff
@@ -0,0 +1,171 @@
...
+				if memberNode == nil {
+					return
+				}
+				// A plugin-managed member that did not parse above is written
+				// in a shape the build does not rewrite, so the call reaches
+				// the stub that throws rather than the API being restricted.
+				if rstestUtils.IsPluginManagedAPI(member) {
+					return
+				}
+				if _, ok := restricted[member]; !ok {
```
swwind, line 117: [P2] Ignore unknown utility member names

The documentation says that a name that is not a member of the utilities object never matches, but this path treats any configured non-plugin-managed property as a valid ordinary member. For options `{ "notAnRstestMember": null }`, `rs.notAnRstestMember()` is reported. Please either gate ordinary names against Rstest's actual utility API set or remove and correct this documented contract.

swwind (commented):
Found two behavioral issues:

1. Direct CommonJS namespace bindings are not recognized.
2. Unknown configured utility member names are incorrectly reported.

internal/plugins/rstest/rules/no_restricted_rstest_methods/no_restricted_rstest_methods.go
elecmonkey: `const rs = require('@rstest/core')` binds the module namespace, not the utilities object: `fn` is not a top-level export of `@rstest/core`, it lives on the `rs` and `rstest` objects the module exports (`RstestUtilities` in the package's types). So `rs.fn()` there calls nothing, and reporting it would be a false positive — the shape that reaches the utilities object through that binding is `rs.rs.fn()`.

The require forms the documentation promises are already handled, they were just untested: `const { rs } = require('@rstest/core'); rs.fn()` and `const { rs: mocker } = require('@rstest/core'); mocker.fn()` through the require binding, and `const core = require('@rstest/core'); core.rs.fn()` through the module namespace. bad34f28 adds all three as regression tests, and the whole-module require as a valid case so the distinction stays pinned.

internal/plugins/rstest/rules/no_restricted_rstest_methods/no_restricted_rstest_methods.go
elecmonkey: You are right that the two disagree. Corrected on the documentation side in bad34f28: the Options section no longer claims that a name the utilities object does not carry never matches, and a test now pins the actual behaviour — `rs.notAnRstestMember()` is reported when configured, `helpers.notAnRstestMember()` is not.

Gating ordinary names against a member allowlist would mean carrying a copy of the full utilities surface here and syncing it with every Rstest release; a member added upstream and missed here would silently stop being restrictable, which trades a harmless report for a false negative that shows up on upgrade. The names come from the configuration, so a name that is not a member matches nothing a test would really write.

internal/plugins/rstest/utils/utilities_object.go
```diff
@@ -41,14 +41,24 @@ func IsUtilitiesObject(ctx rule.RuleContext, receiver *ast.Node) bool {
...
+	if namespace == nil {
+		return false
+	}
+	if isImportMetaRstest(namespace) {
+		return true
+	}
+	if namespace.Kind != ast.KindIdentifier {
 		return false
 	}
 	return testFramework.IsModuleNamespaceSymbolModules(
```
swwind, line 64: [P2] Preserve require namespace bindings through TS assertions

`IsUtilitiesObject` now supports CommonJS namespace bindings, but `IsModuleNamespaceSymbolModules` only recognizes a direct `require(...)` initializer. Therefore both `const core = require('@rstest/core') as any; core.rs.fn();` and `const core = require('@rstest/core'); (core as any).rs.fn();` are silently missed, even though the assertion is erased before runtime. Please unwrap TypeScript assertions when checking the require binding and add a regression test.

swwind, line 64: [P2] Recognize TypeScript import-equals namespaces

`import core = require('@rstest/core')` creates a module-namespace binding, but `IsModuleNamespaceSymbolModules` only checks `NamespaceImport` and `VariableDeclaration`. As a result, `import core = require('@rstest/core'); core.rs.fn();` is silently missed. This is another `require` form covered by the rule's documentation; please handle `ImportEqualsDeclaration` or document the limitation and add a regression test.

swwind (commented):
Found two remaining false-negative cases in CommonJS/TypeScript namespace resolution; please address before merge.

internal/plugins/rstest/utils/utilities_object.go
elecmonkey: Half of this is right. `const core = require('@rstest/core') as any; core.rs.fn();` is missed, because `IsModuleRequireCallModules` only ran `ast.SkipParentheses` over the initializer, so an `AsExpression` never reached `IsRequireCall`. Fixed in e263c72b by skipping assertions there instead, and the shape is now a regression test.

The second shape already worked: `IsUtilitiesObject` runs `SkipAssertionsAndParens` over the namespace it reads off the member, so `(core as any).rs.fn()` resolves `core` and reports. I have not added a test for it, since it would pin behaviour the existing namespace cases already cover.

elecmonkey: Agreed, and the branch was already inconsistent about it: `ReceiverIsLocallyDeclared` lists `ImportEqualsDeclaration` among the kinds that count as an import, while the resolver it sits next to did not know the form at all.

e263c72b teaches `IsModuleNamespaceSymbolModules` the `ImportEqualsDeclaration` form alongside the namespace import and the require initializer, with `import core = require('@rstest/core'); core.rs.fn();` as a regression test. The same shape was pinned as a valid case in `require_mock_type_parameters` with a comment naming this limitation; it moves to the invalid cases in the same commit.

internal/utils/test_framework/module_reference.go
```diff
@@ -221,11 +221,38 @@ func IsModuleNamespaceSymbolModules(symbol *ast.Symbol, importModules []string)
 				return true
 			}
 		}
+
+		// `import core = require('m')` binds the module namespace the same way
+		// a namespace import does.
+		if declaration.Kind == ast.KindImportEqualsDeclaration {
```
swwind, line 240: [P2] Exclude type-only import-equals from namespace resolution

`import type core = require('@rstest/core')` does not create a runtime namespace, but this branch does not check `IsTypeOnly`, so `core.rs.fn()` is reported incorrectly. The same issue affects `ReceiverIsLocallyDeclared` on the `*Actual` path. Please exclude type-only import-equals declarations and add a regression test.

internal/utils/test_framework/module_reference.go
```diff
@@ -291,7 +318,9 @@ func IsModuleRequireCallModules(node *ast.Node, importModules []string) bool {
 		return false
 	}
 
-	node = ast.SkipParentheses(node)
+	// TypeScript assertions are erased before runtime, so
+	// `require('m') as any` still binds the module.
+	node = internalUtils.SkipAssertionsAndParens(node)
```
swwind, line 336: [P2] Unwrap assertions on require arguments

The initializer is unwrapped for TypeScript assertions, but the require argument still uses `ast.SkipParentheses`. As a result, `const core = require('@rstest/core' as any); core.rs.fn();` is silently missed even though the assertion is erased at runtime. Please use `SkipAssertionsAndParens` for the argument and add a regression test.

swwind (commented):
Found two P2 issues that should be addressed before merge.

internal/utils/test_framework/module_reference.go
elecmonkey: The namespace half is right, and it went wider than the import-equals form: `import type core = require('@rstest/core')` and `import type * as core from '@rstest/core'` both reached the module namespace, because neither `IsModuleNamespaceSymbolModules` nor the resolver beside it asked whether the declaration binds a value at all. 46545808 skips type-only import declarations in both, so a namespace that nothing binds at runtime resolves to nothing.

The `*Actual` half reads the other way. A type-only import does not declare a local `rs` at runtime — it declares nothing — so under `globals: true` the call reaches the global utilities object and the build rewrites it. That is the same reading `import type { beforeAll } from 'rstack/test'; beforeAll(() => {})` already gets in the registration tests. `ReceiverIsLocallyDeclared` returning false there is the answer that matches; treating an erased name as a local declaration would silence a call that really is Rstest's. Both shapes are pinned now — the erased namespaces as valid cases, `import type rs = require('@rstest/core'); rs.importActual('./m')` as an invalid one.

The same erasure also had the two spellings of a type-only named import disagreeing with each other: `import type { rs } from '@rstest/core'; rs.fn()` was read as a real import binding, while `import { type rs }` was read as a local declaration and reported nothing. Both are the global now, and both are tests.

internal/utils/test_framework/module_reference.go
elecmonkey: Right that the shape is missed, though `SkipParentheses` on the argument was not what rejected it: `ast.IsRequireCall(node, true /* requireStringLiteralLikeArgument */)` gates on a string-literal-like argument first, so `require('@rstest/core' as any)` never reached that line. 46545808 takes the argument check off `IsRequireCall` — it is passed `false` now — and unwraps assertions along with parentheses, so the specifier is read after erasure and the literal-kind switch below still decides what matches. `const core = require('@rstest/core' as any); core.rs.fn();` is a regression test.

swwind (approved):
LGTM!!!

PR #2029, opened 2026-09-02
feat(rstest): add no-importing-rstest-globals and prefer-importing-rstest-globals rules

elecmonkey:
## Summary

Add `no-importing-rstest-globals` and `prefer-importing-rstest-globals` to enforce either global Rstest APIs or explicit imports. Both rules recognize all 13 Rstest globals, support `@rstest/core` and `rstack/test` imports and requires, and provide autofixes while preserving aliases, comments, and non-call references.

## Credits

Adapted from [`eslint-plugin-vitest`](https://github.com/vitest-dev/eslint-plugin-vitest)'s `no-importing-vitest-globals` and [`eslint-plugin-jest`](https://github.com/jest-community/eslint-plugin-jest)'s `prefer-importing-jest-globals`.

- **Use Rstest's exact global API set.** The rules include Rstest-specific `rs`, `rstest`, `onTestFinished`, and `onTestFailed` globals and exclude APIs that Rstest does not expose globally.

- **Recognize both Rstest entry points.** Imports and CommonJS requires from `@rstest/core` and `rstack/test` are treated equivalently.

- **Keep unsafe removals diagnostic-only.** Aliased bindings, non-invocation references, and destructuring with defaults or rest elements are reported without an autofix.

- **Handle namespace-object uses.** The importing rule detects usages such as `rs.fn()` and `expect.any()` in addition to direct calls.

## Related Links

Related #935

## Checklist


internal/plugins/rstest/rules/no_importing_rstest_globals/no_importing_rstest_globals.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+	allGlobals := true
+	for _, element := range elements {
+		if !rstestUtils.IsRstestGlobal(rstestUtils.ImportedSpecifierName(element)) {
+			allGlobals = false
+			break
+		}
+	}
+	clause := declaration.ImportClause.AsImportClause()
+	if allGlobals && clause.Name() == nil {
```
swwind: **P1:** This branch only proves that the current specifier is removable, but `allGlobals` is computed across every specifier. For example, `import { expect, it as test } from '@rstest/core'; expect(1); test('x');` removes the whole import and leaves `test` undefined. The `require` branch has the same issue. Please remove the whole declaration only when every binding is independently safe to remove; otherwise remove just this specifier.

internal/plugins/rstest/rules/prefer_importing_rstest_globals/prefer_importing_rstest_globals.go
```diff
@@ -0,0 +1,268 @@
...
+		return []rule.RuleFix{rule.RuleFixReplaceRange(core.NewTextRange(0, 0), createImport(isModule, defaultModule, names.sortedJoined()))}
+	}
+	statements := ctx.SourceFile.Statements.Nodes
+	first := statements[0]
+	if isUseStrictDirective(first) {
+		return []rule.RuleFix{rule.RuleFixInsertAfter(first, "\n"+lineIndent(ctx, first)+createImport(isModule, defaultModule, names.sortedJoined()))}
+	}
+	if existing := findImport(statements); existing != nil {
+		collectImportNames(existing.node.AsImportDeclaration(), names)
+		return []rule.RuleFix{rule.RuleFixReplace(ctx.SourceFile, existing.node, createImport(isModule, existing.module, names.sortedJoined()))}
```
swwind: **P1:** Replacing the whole existing import while collecting only named specifiers drops existing default imports and inline comments. For example, `import core, { defineConfig } ...` loses `core`, and `import { defineConfig /* keep */ } ...` loses the comment. Please preserve the default binding and original trivia, or insert/replace only the named-specifier portion.

internal/plugins/rstest/rules/prefer_importing_rstest_globals/prefer_importing_rstest_globals.go
```diff
@@ -0,0 +1,268 @@
...
+		return "import { " + names + " } from '" + module + "';"
+	}
+	return "const { " + names + " } = require('" + module + "');"
+}
+
+func isUseStrictDirective(statement *ast.Node) bool {
+	if statement == nil || statement.Kind != ast.KindExpressionStatement {
+		return false
+	}
+	expression := ast.SkipParentheses(statement.AsExpressionStatement().Expression)
```
swwind: **P2:** `SkipParentheses` plus `GetStaticStringValue` treats `` `use strict` `` and `('use strict')` as a directive, although only a string-literal expression statement is in the directive prologue. This causes the fixer to insert after a non-directive statement and differs from ESLint's placement. Please check the expression kind before treating it as a strict directive.

swwind (commented):
Found three fixer correctness issues; details are inline.

internal/plugins/rstest/rules/no_importing_rstest_globals/no_importing_rstest_globals.go
elecmonkey: Fixed in c1628b00. `allBindingsRemovable` now requires every binding of the declaration to be an rstest global that is itself safe to remove, so whole-declaration removal no longer fires when a sibling is aliased or referenced outside an invocation; those cases fall back to removing just this specifier. The `require` branch shares the same helper. Regression tests cover `import { expect, it as test }`, its `require` equivalent, and a sibling used as a value.

internal/plugins/rstest/rules/prefer_importing_rstest_globals/prefer_importing_rstest_globals.go
elecmonkey: Fixed in c1628b00. The fix now inserts `, <names>` after the last named specifier instead of regenerating the declaration, so a default binding, existing aliases, and comments inside the braces all survive. Tests added for `import core, { defineConfig }` and `import { defineConfig /* keep */ }`.

internal/plugins/rstest/rules/prefer_importing_rstest_globals/prefer_importing_rstest_globals.go
elecmonkey: Fixed in c1628b00. `isUseStrictDirective` now requires a bare string-literal expression statement, so `` `use strict` `` and `('use strict')` are ordinary statements and the import is inserted before them. Tests added for both shapes.

internal/plugins/rstest/utils/specifier_removal.go
```diff
@@ -0,0 +1,102 @@
...
+// SpecifierRemovalRange spans the specifier at index together with the comma
+// that separates it from its neighbour. Comments stay with the specifier on
+// their side of the separating comma.
+func SpecifierRemovalRange(sourceFile *ast.SourceFile, elements []*ast.Node, index int) core.TextRange {
+	elementRange := internalUtils.TrimNodeTextRange(sourceFile, elements[index])
+	if index > 0 {
+		text := sourceFile.Text()
+		previousEnd := internalUtils.TrimNodeTextRange(sourceFile, elements[index-1]).End()
+		start := previousEnd
+		if comma := strings.LastIndexByte(text[previousEnd:elementRange.Pos()], ','); comma >= 0 {
```
swwind: **P1:** `LastIndexByte` searches the raw text between specifiers, so it can select a comma inside a comment instead of the separator. For example, `import { defineConfig, /* comma, in comment */ expect } ...` produces a fix beginning at the comment's comma and leaves `import { defineConfig, /* comma } ...`, which is syntactically invalid. Please locate the actual separator while skipping comment trivia (or derive the range from the AST).

internal/plugins/rstest/rules/prefer_importing_rstest_globals/prefer_importing_rstest_globals.go
```diff
@@ -0,0 +1,273 @@
...
+		imported := rstestUtils.RequireBindingImportedName(element)
+		if imported == "" {
+			continue
+		}
+		local := ""
+		if be.Name() != nil && be.Name().Kind == ast.KindIdentifier {
+			local = be.Name().AsIdentifier().Text
+		}
+		if local != "" && local != imported {
+			imported += ": " + local
```
swwind: **P1:** When a module-mode file has an existing destructured require with an alias, this string is later passed to `createImport`. `const { it: testCase } = require('@rstest/core'); expect(value);` is rewritten as `import { expect, it: testCase } ...`, which TypeScript cannot parse. The converted import needs `it as testCase`, or the fixer should preserve the require form/insert only missing names.

internal/plugins/rstest/rules/prefer_importing_rstest_globals/prefer_importing_rstest_globals.go
```diff
@@ -0,0 +1,273 @@
...
+	for _, name := range collected {
+		names.add(name)
+	}
+	isModule := ctx.LanguageOptions.EffectiveSourceType() == "module"
+	if ctx.SourceFile == nil || ctx.SourceFile.Statements == nil || len(ctx.SourceFile.Statements.Nodes) == 0 {
+		return []rule.RuleFix{rule.RuleFixReplaceRange(core.NewTextRange(0, 0), createImport(isModule, defaultModule, names.sortedJoined()))}
+	}
+	statements := ctx.SourceFile.Statements.Nodes
+	first := statements[0]
+	if isUseStrictDirective(first) {
```
swwind: **P1:** This checks only `statements[0]`. In CommonJS source with multiple directives, e.g. `"use asm"; "use strict"; expect(value);`, the fixer inserts `const` before the first statement, so `"use strict"` is no longer a directive and strict-mode semantics change. Please find the end of the whole directive prologue and insert after it when it contains `"use strict"`.

internal/plugins/rstest/rules/prefer_importing_rstest_globals/prefer_importing_rstest_globals.go
```diff
@@ -0,0 +1,273 @@
...
+	return rule.RuleMessage{
+		Id:          "preferImportingRstestGlobals",
+		Description: "Import `" + names + "` from `@rstest/core`.",
+		Data:        map[string]string{"names": names},
+	}
+}
+
+func shouldReportIdentifier(ctx rule.RuleContext, node *ast.Node) bool {
+	if node == nil || node.Kind != ast.KindIdentifier || !rstestUtils.IsRstestGlobal(node.AsIdentifier().Text) ||
+		ast.IsIdentifierName(node) || ast.IsDeclarationNameOrImportPropertyName(node) {
```
swwind: **P2:** `ast.IsIdentifierName` is also true for shorthand object values. Thus `const value = { expect };` is ignored even though `expect` is a runtime reference, so no import is requested. Please distinguish the key side of a shorthand `PropertyAssignment` from its value side and add a regression test.

internal/plugins/rstest/utils/specifier_removal.go
elecmonkey: Fixed in c73b8bd8. `SpecifierRemovalRange` now finds the separator with a trivia-aware scan (`separatorComma`) instead of `LastIndexByte`, so a comma inside a comment is skipped; the first-element branch uses the same scan. Tests added for `import { defineConfig, /* comma, in comment */ expect }` and for the comment sitting before the separator.

internal/plugins/rstest/rules/prefer_importing_rstest_globals/prefer_importing_rstest_globals.go
elecmonkey: Fixed in c73b8bd8. The require branch no longer regenerates the statement as an import: `mergeIntoRequire` inserts the missing names after the last binding, mirroring `mergeIntoImport`, so `const { it: testCase } = require(...)` keeps its own syntax and comments survive. A rest element stays last. Tests added for the aliased binding and for a trailing rest element; the existing merge-into-require case now expects the require form.

internal/plugins/rstest/rules/prefer_importing_rstest_globals/prefer_importing_rstest_globals.go
elecmonkey: Fixed in c73b8bd8. `strictDirectivePrologue` walks the whole prologue and returns its last statement when any directive is `use strict`, so the insertion lands after the prologue. Test added for `"use asm"; 'use strict'; expect(value);` in CommonJS.

internal/plugins/rstest/rules/prefer_importing_rstest_globals/prefer_importing_rstest_globals.go
elecmonkey: Fixed in c73b8bd8, with one correction to the diagnosis: in typescript-go `IsIdentifierName` has no `ShorthandPropertyAssignment` case, so it returns false here. What dropped the reference is `IsDeclarationNameOrImportPropertyName` on the same line, because a shorthand assignment is a declaration whose name is that identifier. `isShorthandPropertyRead` now exempts the value side of a shorthand from both name filters, while still skipping a shorthand that is written to (`({ expect } = holder)`, a `for` destructuring head, or one with an assignment initializer). Tests added for `const value = { expect };` and the assignment shape.

internal/plugins/rstest/rules/no_importing_rstest_globals/no_importing_rstest_globals.go
```diff
@@ -0,0 +1,256 @@
...
+	if clause.Name() == nil && allBindingsRemovable(ctx, elements, rstestUtils.ImportedSpecifierName) {
+		return []rule.RuleFix{rule.RuleFixRemove(ctx.SourceFile, statement)}
+	}
+	if len(elements) > 1 {
+		return []rule.RuleFix{rule.RuleFixRemoveRange(rstestUtils.SpecifierRemovalRange(ctx.SourceFile, elements, index))}
+	}
+	if clause.Name() != nil && clause.NamedBindings != nil {
+		start := internalUtils.TrimNodeTextRange(ctx.SourceFile, clause.Name()).End()
+		end := internalUtils.TrimNodeTextRange(ctx.SourceFile, clause.NamedBindings).End()
+		return []rule.RuleFix{rule.RuleFixRemoveRange(core.NewTextRange(start, end))}
```
swwind, line 183: **P1:** This fixer produces invalid syntax for `import core, { expect } from '@rstest/core'; expect(value);`: it becomes `import core } from '@rstest/core'; expect(value);`. Please preserve the default import while removing the complete named-binding range, including its braces and preceding comma.

elecmonkey: I could not reproduce this one. For `import core, { expect } from '@rstest/core'; expect(value);` the fixer emits `import core from '@rstest/core'; expect(value);`. The range does already include the braces: it ends at `TrimNodeTextRange(clause.NamedBindings).End()`, and `NamedBindings` is the `NamedImports` node, whose end is past the closing brace, while it starts at the end of the default binding, which swallows the preceding comma.

I added that exact case as a regression test in 4f0abd17, plus a multi-line variant, so the branch cannot drift into invalid syntax later. If you saw the broken output on a different shape, please paste it and I will dig further.

internal/plugins/rstest/rules/prefer_importing_rstest_globals/prefer_importing_rstest_globals.go
```diff
@@ -0,0 +1,337 @@
...
+		return true
+	}
+	name := node.AsIdentifier().Text
+	resolved, _, mode := testFramework.ResolveFunctionIdentifierReferenceFromSymbolModules(
+		name, node, symbol, ctx.SourceFile, rstestUtils.RstestCoreImportModules,
+	)
+	if mode == testFramework.ReferenceModeImport && resolved != "" {
+		return false
+	}
+	return !internalUtils.IsValueSymbolDeclaredInFile(symbol, ctx.SourceFile)
```
swwind: **P1:** `expect = value;` reaches this path and receives an autofix that inserts `import { expect } from '@rstest/core';`. That changes an assignment to a global into an assignment to an imported binding, which is read-only and can throw in module code. Please exclude write references (including updates and destructuring targets), or withhold the autofix for them.

swwind (commented):
Found three correctness and compatibility issues.

internal/plugins/rstest/rules/prefer_importing_rstest_globals/prefer_importing_rstest_globals.go
```diff
@@ -0,0 +1,337 @@
...
+		return owner.AsForInOrOfStatement().Initializer != object
+	}
+	return true
+}
+
+func shouldReportIdentifier(ctx rule.RuleContext, node *ast.Node) bool {
+	if node == nil || node.Kind != ast.KindIdentifier || !rstestUtils.IsRstestGlobal(node.AsIdentifier().Text) {
+		return false
+	}
+	if !isShorthandPropertyRead(node) && (ast.IsIdentifierName(node) || ast.IsDeclarationNameOrImportPropertyName(node)) {
```
swwind: **P2:** This check treats any identifier with an Rstest-global spelling as a value reference after the syntactic name filters. A type-only use such as `const value: expect = input;` is therefore reported and receives a runtime import, even though it does not use the API at runtime. Please restrict this to value references.

internal/plugins/rstest/utils/module_bindings.go
```diff
@@ -0,0 +1,73 @@
...
+	if binding == nil || binding.DotDotDotToken != nil {
+		return ""
+	}
+	name := binding.PropertyName
+	if name == nil {
+		name = binding.Name()
+	} else if name.Kind == ast.KindComputedPropertyName {
+		name = ast.SkipParentheses(name.AsComputedPropertyName().Expression)
+	}
+	if name == nil || (name.Kind != ast.KindIdentifier && !ast.IsStringLiteralLike(name)) {
```
swwind: **P2:** `RequireBindingImportedName` treats static string/computed property keys as Rstest globals, so `const { 'expect': expect } = require('@rstest/core')` is reported. The upstream `no-importing-vitest-globals` rule treats this shape as valid, and the rule-specific differential corpus now diverges here. Please either preserve upstream behavior or add an explicit documented contract and regression coverage for this intentional extension.

internal/plugins/rstest/rules/prefer_importing_rstest_globals/prefer_importing_rstest_globals.go
elecmonkey: Fixed in 7a96ae82. `shouldReportIdentifier` now bails out on a write reference (`IsWriteReference` covers plain and compound assignment, `++`/`--`, and destructuring targets), so `expect = value;` is no longer reported and gets no import. Regression tests added for `expect = value;`, `expect += 1;`, `expect++;`, and `[expect] = holder;`.

internal/plugins/rstest/rules/prefer_importing_rstest_globals/prefer_importing_rstest_globals.go
elecmonkey: Fixed in 7a96ae82. The check now skips `ast.IsPartOfTypeNode` and `ast.IsPartOfTypeQuery` identifiers, so a type position never asks for a runtime import. Tests added for `const value: expect = input;` and `type Alias = typeof expect;`.

internal/plugins/rstest/utils/module_bindings.go
elecmonkey: Addressed in 7a96ae82, taking the second option you offered for the string-literal key and treating the computed key as a real bug.

The computed branch was wrong: it read the expression text as the imported name, so `const { [expect]: local } = require('@rstest/core')` was reported even though `expect` there is a value being read, not a property name, and the module has no export matching it. A computed key now counts only when it is a static string literal.

For `const { 'expect': expect } = require('@rstest/core')` I kept the report. The property key is what decides which export the binding pulls in, and a string-literal key names the same export as an identifier key, so this really does import the global the rule forbids; the autofix removes the binding and leaves the global in place, which is the same output the identifier form produces. Dropping it would be a false negative on a genuine violation. It is now stated as an explicit contract in `no_importing_rstest_globals.md`, alongside the computed-key rule, and both shapes have regression tests (`'expect': expect` and `['expect']: expect` reported, `[expect]` and `[key]` valid).

internal/plugins/rstest/rules/prefer_importing_rstest_globals/prefer_importing_rstest_globals.go
```diff
@@ -0,0 +1,348 @@
...
+		return true
+	}
+	name := node.AsIdentifier().Text
+	resolved, _, mode := testFramework.ResolveFunctionIdentifierReferenceFromSymbolModules(
+		name, node, symbol, ctx.SourceFile, rstestUtils.RstestCoreImportModules,
+	)
+	if mode == testFramework.ReferenceModeImport && resolved != "" {
+		return false
+	}
+	return !internalUtils.IsValueSymbolDeclaredInFile(symbol, ctx.SourceFile)
```
swwind: **P2:** A type-only import does not create a runtime binding, but `IsValueSymbolDeclaredInFile` treats it as a local value declaration here. Therefore both `import { type expect } from '@rstest/core'; expect(value);` and `import type { expect } from '@rstest/core'; expect(value);` produce no diagnostic. Please ignore type-only imports when checking for local value declarations and add regressions for both forms.

elecmonkey: Fixed in 8b4c642d.

The local-declaration check now goes through `IsRuntimeValueSymbolDeclaredInFile`, a new sibling of `IsValueSymbolDeclaredInFile` that additionally skips `ast.IsTypeOnlyImportDeclaration` declarations, so a type-only binding no longer counts as providing or shadowing the value. Both helpers share one private implementation and `IsValueSymbolDeclaredInFile` is byte-for-byte unchanged in behaviour, so its other callers are untouched — its doc comment states a deliberately narrower contract ("only interfaces and type aliases are skipped"), and I did not want to change that out from under ~27 call sites in a rule-port PR. I put the narrowing in a shared sibling rather than in the rule itself because the ambient-augmentation part of the check is unexported, and duplicating it in the rule would have either dropped it or copied it.

One deliberate difference from a plain "report it" fix: the diagnostic carries no autofix when the name already has a type-only binding of its own. Both `import { type expect, expect } from '@rstest/core'` and a second `import { expect } ...` declaration beside an `import type` clause bind `expect` twice, so every edit available here is a duplicate identifier. The edit a reader would actually make is to turn the existing binding into a value import, and that is not a choice this rule is in a position to make.

Regression tests added for both forms reported without a fix, plus a case where the type-only import names something else (`import type { Mock }`) and the normal insertion still applies.

fansenze (dismissed):

internal/plugins/rstest/rules/no_importing_rstest_globals/no_importing_rstest_globals.go
```diff
@@ -0,0 +1,256 @@
...
+		node.ForEachChild(func(child *ast.Node) bool {
+			visit(child)
+			return used
+		})
+	}
+	visit(sourceFile.AsNode())
+	return used
+}
+
+func bindingCanBeRemoved(ctx rule.RuleContext, element *ast.Node, importedName string) bool {
```
fansenze, line 137: **[P1] Preserve exported require bindings**

The fixer removes this statement entirely:

```ts
export const { expect } = require('@rstest/core');
```

This removes the module’s `expect` export and breaks consumers importing it. `exportUsesName()` only checks export specifiers such as `export { expect }`, so it misses exported variable declarations. Please treat the `export` modifier as a surviving use and withhold removal.

internal/plugins/rstest/rules/prefer_importing_rstest_globals/prefer_importing_rstest_globals.go
```diff
@@ -0,0 +1,395 @@
...
+	additions := newNameSet(len(names.order))
+	for _, name := range names.order {
+		if _, exists := present[name]; !exists {
+			additions.add(name)
+		}
+	}
+	if len(additions.order) == 0 {
+		return nil
+	}
+	// A rest element has to stay last, so names go in front of it.
```
fansenze: **[P1] Preserve the properties of a rest binding**

For this input:

```ts
const { ...core } = require('@rstest/core');
expect(1);
core.expect(1);
```

The fixer changes the declaration to:

```ts
const { expect, ...core } = require('@rstest/core');
```

Adding `expect` before the rest binding excludes that property from `core`, so the existing `core.expect(1)` call breaks. Keeping rest last preserves syntax, but changes the resulting object. Please insert a separate declaration or withhold the fix when merging would change a rest binding’s properties.

internal/plugins/rstest/rules/prefer_importing_rstest_globals/prefer_importing_rstest_globals.go
```diff
@@ -0,0 +1,395 @@
...
+	}
+	// A type position never touches the value at runtime, so `const value:
+	// expect = input;` does not need the API imported.
+	if ast.IsPartOfTypeNode(node) || ast.IsPartOfTypeQuery(node) {
+		return false
+	}
+	// A write such as `expect = value;` assigns to the global. Importing the
+	// name would turn the target into a read-only module binding, so the write
+	// is not a use that the import can satisfy.
+	if internalUtils.IsWriteReference(node) {
```
fansenze: **[P1] Check writes across all uses before importing a name**

Skipping individual write references does not protect a global that is also read:

```ts
expect = customExpect;
expect(1);
```

The read still triggers this fix:

```ts
import { expect } from '@rstest/core';
expect = customExpect;
expect(1);
```

The assignment now targets a read-only import binding and fails. Please check whether any reference writes to each collected name before offering an autofix for that name.

internal/plugins/rstest/rules/no_importing_rstest_globals/no_importing_rstest_globals.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+	if clause.Name() != nil && clause.NamedBindings != nil {
+		start := internalUtils.TrimNodeTextRange(ctx.SourceFile, clause.Name()).End()
+		end := internalUtils.TrimNodeTextRange(ctx.SourceFile, clause.NamedBindings).End()
+		return []rule.RuleFix{rule.RuleFixRemoveRange(core.NewTextRange(start, end))}
+	}
+	return nil
+}
+
+func enclosingVariableStatement(node *ast.Node) (*ast.Node, *ast.VariableDeclarationList) {
```
fansenze: **[P1] Keep require removal within its own declaration**

This ancestor walk can cross a function boundary and select an unrelated outer variable statement:

```ts
const run = () => {
  for (const { expect } = require('@rstest/core'); ready;) {
    expect(1);
  }
};
run();
```

The fixer removes the entire `run` declaration, leaving only:

```ts
run();
```

Please verify that the variable statement directly owns the declaration being removed. A declaration in a `for` initializer should not cause an enclosing function declaration to be deleted.

internal/plugins/rstest/rules/prefer_importing_rstest_globals/prefer_importing_rstest_globals.go
```diff
@@ -0,0 +1,395 @@
...
+// bindsTypeOnly reports whether the file already imports one of names from an
+// rstest module as a type. Such a binding provides no value, so the rule still
+// asks for the import, but every fix it could write would declare the name a
+// second time: adding a value specifier beside the type-only one, or a second
+// declaration next to a whole `import type` clause, is a duplicate identifier
+// either way. Turning the existing binding into a value import is the edit a
+// reader would make, and it is not one this rule is in a position to choose,
+// so the diagnostic is reported without a fix.
+func bindsTypeOnly(statements []*ast.Node, names *nameSet) bool {
+	for _, statement := range statements {
```
fansenze: **[P2] Cover all type-only bindings before inserting an import**

The shared runtime-symbol helper excludes all type-only imports, but this collision check only covers named imports from Rstest modules. For example:

```ts
import type { expect } from './types';
expect(1);
```

The fixer inserts a second binding named `expect`:

```ts
import { expect } from '@rstest/core';
import type { expect } from './types';
expect(1);
```

Default, namespace, and import-equals type imports have the same gap. Please check all colliding type-only bindings, regardless of their module or import form, before inserting a value import.

internal/plugins/rstest/rules/prefer_importing_rstest_globals/prefer_importing_rstest_globals.go
```diff
@@ -0,0 +1,395 @@
...
+			return false
+		}
+		if parent.Kind == ast.KindSourceFile || parent.Kind == ast.KindBlock || ast.IsExpression(parent) {
+			break
+		}
+	}
+	if ctx.Refs == nil {
+		return true
+	}
+	symbol := ctx.Refs.Resolve(node)
```
fansenze: **[P2] Exclude labels before resolving global references**

This valid labeled loop is reported as requiring a Rstest import:

```ts
test: for (const item of items) {
  if (item) break test;
}
```

Here, `test` is only a label. `Refs.Resolve()` returns `nil` for non-reference positions too, but this branch treats every `nil` result as an unresolved global. Please use the shared reference-position classification to exclude labels before applying the global-name check.

fansenze (commented):
Six rule-level correctness issues are detailed inline, with reproductions and suggested fixes.

internal/plugins/rstest/rules/no_importing_rstest_globals/no_importing_rstest_globals.go
elecmonkey: Fixed in 4665b1d0. `exportUsesName` only recognises `export { expect }`, which names a binding that already exists; `export const { expect } = require('@rstest/core')` declares the export inline, so nothing saw it. `isExportedBinding` now walks to the enclosing variable statement and treats an `export` modifier as a use that keeps the binding, which withholds both the whole-statement removal and the single-specifier removal — dropping just `expect` out of `export const { expect, test }` would delete that export as well. Regression tests added for both shapes.

internal/plugins/rstest/rules/no_importing_rstest_globals/no_importing_rstest_globals.go
elecmonkey: Fixed in 4665b1d0. `enclosingVariableStatement` is replaced by `owningVariableStatement`, which requires the declaration's direct parent chain to be `VariableDeclarationList` -> `VariableStatement` and that statement to declare nothing else. A loop-head declaration fails the first check, so no statement is removed for it. I also moved the whole-statement branch so that failing it falls through to the specifier removal instead of returning no fix at all. Regression test added for your exact case.

internal/plugins/rstest/rules/prefer_importing_rstest_globals/prefer_importing_rstest_globals.go
elecmonkey: Fixed in a3a86e77. You are right that skipping the write reference alone does not protect the name. The identifier listener now classifies each reference as irrelevant, a write, or a read, and collects the written names for the whole file; `buildAutofix` withholds the fix as soon as any name it is about to import is written somewhere. The diagnostic stays, since the read really is a use of the global. Regression tests added for `expect = customExpect; expect(1);`, the reverse order, the destructuring form `({ expect } = holder); expect(1);`, and a case where a second name that is only read still gets its import.

internal/plugins/rstest/rules/prefer_importing_rstest_globals/prefer_importing_rstest_globals.go
elecmonkey: Fixed in a3a86e77. `mergeIntoRequire` now returns no fix when the pattern contains a rest element. Keeping the rest element last preserved the syntax but, as you say, changed the object it binds, and there is no position in that pattern where a new name is harmless. Withholding rather than emitting a separate declaration keeps a single `require` of the module in the output. Regression tests added for `const { ...core } = require('@rstest/core')` with `core.expect(1)`, and the previous `const { defineConfig, ...others }` case now expects no fix.

internal/plugins/rstest/rules/prefer_importing_rstest_globals/prefer_importing_rstest_globals.go
elecmonkey: Fixed in a3a86e77. The check is now `bindsNameAsTypeOnly` and scans every top-level import regardless of module: a type-only clause's default binding, a type-only namespace binding, `import type x = require(...)`, a whole `import type { ... }` clause, and an inline `{ type x }` specifier. That matches the range `IsRuntimeValueSymbolDeclaredInFile` excludes, so the two no longer disagree. A binding that does survive to runtime never reaches this point, since the identifier resolves to it and is not reported. Regression tests added for all five forms plus a type-only import of a different name, which still merges normally.

internal/plugins/rstest/rules/prefer_importing_rstest_globals/prefer_importing_rstest_globals.go
elecmonkey: Fixed in a3a86e77. The rule now classifies identifiers through `IsReadReference`, the shared decision about what counts as a reference position, so a label in `test: for (...) { break test; }` is excluded before the global-name check ever runs. That also folds in the property-key, declaration-name and type-position cases the rule had been spelling out by hand, and it keeps a shorthand property value a reference, which `IsWriteReference` then splits into the read form `{ expect }` and the destructuring form `({ expect } = holder)`. Regression tests added for a labeled `for ... of` with `break` and a labeled `while` with `continue`.

fansenze (approved):

PR #2030, opened 2026-09-02
feat(rstest): add prefer-rs-mocked rule

elecmonkey:
## Summary

Report Rstest mock type assertions that erase the original callable signature and prefer `rs.mocked()`.

Resolve `Mock`, `Mocked`, and `MockInstance` through `@rstest/core` and `rstack/test`, including type-only imports. Offer an autofix when `rs` or `rstest` is safe to reference, an `import.meta.rstest.rs.mocked()` suggestion when that is the only known entry point, and otherwise report without an edit. An imported namespace is written under the name the import binds, so `import { rs as r }` fixes to `r.mocked(value)`.

## Type-only imports

`ImportSpecifier.IsTypeOnly` records only the inline `type` modifier of `import { type rs }`; for `import type { rs }` the flag sits on the enclosing `ImportClause`. Two neighbouring places read the specifier flag alone and so took a type-only import for a runtime binding, and both are fixed here as well: `consistent-rstest-namespace`, when deciding whether an existing declaration blocks its fix, and `resolveModuleImportSpecifier` in `internal/utils/test_framework`, when resolving a function reference. All three now ask `ast.IsTypeOnlyImportDeclaration`. A type-only specifier for the module is skipped rather than treated as a local declaration, so a call it appears to bind keeps reading as the global registration.

## Intentional limitations

- Namespace-qualified mock types are not matched because selecting a safe runtime namespace from that type namespace is ambiguous.
- Unrelated bindings that capture the namespace name suppress automatic edits.

## Checklist


internal/plugins/rstest/rules/prefer_rs_mocked/prefer_rs_mocked.go
```diff
@@ -0,0 +1,269 @@
...
+		state.used = true
+	}
+}
+
+func isRstestUtilitiesImportName(identifier *ast.Node) bool {
+	if identifier == nil || identifier.Parent == nil || identifier.Parent.Kind != ast.KindImportSpecifier {
+		return false
+	}
+	specifier := identifier.Parent.AsImportSpecifier()
+	if specifier == nil || specifier.IsTypeOnly || specifier.Name() != identifier {
```
swwind: In `import type { rs, Mock } from '@rstest/core'`, ts-go sets `ImportSpecifier.IsTypeOnly` to false; the type-only status is stored on the enclosing `ImportClause`. The current implementation therefore treats `rs` as a runtime import and autofixes to `rs.mocked(value)`, even though that runtime binding does not exist. Please check `ImportClause.IsTypeOnly` and add a regression test.

internal/plugins/rstest/rules/prefer_rs_mocked/prefer_rs_mocked.go
```diff
@@ -0,0 +1,269 @@
...
+
+	var visit func(*ast.Node)
+	visit = func(node *ast.Node) {
+		if node == nil {
+			return
+		}
+		if rstestUtils.IsImportMetaRstest(node) {
+			file.hasImportMeta = true
+		}
+		if node.Kind == ast.KindIdentifier {
```
swwind, line 89: For `import { rs as r } from '@rstest/core'`, the scan only tracks local names spelled `rs`/`rstest`, so it misses the alias `r` and then defaults to the bare `rs.mocked(value)`, producing an incorrect fix. Please use the actual local alias (`r.mocked(value)`), or withhold the autofix when the namespace cannot be resolved safely.

internal/plugins/rstest/rules/prefer_rs_mocked/prefer_rs_mocked.go
elecmonkey: Fixed in 31543162. `ImportSpecifier.IsTypeOnly` only carries the inline `type` modifier, so the check now goes through `ast.IsTypeOnlyImportDeclaration`, which reads the enclosing `ImportClause` as well. The specifier then makes the name unavailable and no edit is offered; regression test added for `import type { rs, Mock } from '@rstest/core'`.

The same confusion sat in two more places, fixed alongside. `consistent-rstest-namespace` used the specifier flag when deciding whether an existing `rs` declaration blocks its fix, so `import type { rs }` plus a global `rstest.mock()` produced a fix naming a type-only binding. And in e2a163ed, `resolveModuleImportSpecifier` resolved `import type { describe } from '@rstest/core'` as a runtime import of the framework API; it now skips such a specifier entirely, so the call still reads as the global registration that `rstack_test_module_test.go` documents rather than as a shadowed one.

internal/plugins/rstest/rules/prefer_rs_mocked/prefer_rs_mocked.go
elecmonkey: Fixed in 2b930364. The scan collected identifiers by their text, so an alias was never recorded. It now walks the `@rstest/core` / `rstack/test` import declarations first and follows each namespace under the local name it binds, so `import { rs as r }` fixes to `r.mocked(value)` and `import { rstest as helper }` to `helper.mocked(value)`. The alias is also the name watched for shadowing, so a binding that captures it withholds the edit as before. Tests cover all three.

internal/plugins/rstest/rules/prefer_rs_mocked/prefer_rs_mocked.go
```diff
@@ -0,0 +1,338 @@
...
+	}
+	local := specifier.Name()
+	if local == nil || local.Kind != ast.KindIdentifier {
+		return
+	}
+	imported := specifier.PropertyName
+	if imported == nil {
+		imported = local
+	}
+	if imported.Kind != ast.KindIdentifier {
```
swwind: `PropertyName` can be a string literal in valid imports such as `import { "rs" as r } from "@rstest/core"`. This branch skips the runtime namespace, so the autofix falls back to `rs.mocked(...)` instead of the bound `r.mocked(...)`. Please handle string-literal `PropertyName` here and in `isRstestUtilitiesImportName`, with a regression test.

elecmonkey: Fixed in f7a93a81. Both places that matched the imported name required an identifier, so a specifier whose `PropertyName` is a string literal was skipped: the namespace collection kept `rs` as the local name and wrote a fix naming a binding the file does not have, and `isRstestUtilitiesImportName` took the import for an unrelated declaration and withheld the edit. Both now read the name with `Text()`, the way `resolveModuleImportSpecifier` already does, so `import { "rs" as r }` fixes to `r.mocked(value)` and `import { "rstest" as rstest }` to `rstest.mocked(value)`. Regression tests cover both.

internal/plugins/rstest/rules/prefer_rs_mocked/prefer_rs_mocked.go
```diff
@@ -0,0 +1,342 @@
...
+			return false
+		})
+	}
+	visit(file.ctx.SourceFile.AsNode())
+}
+
+// collectImports records which namespaces the file imports from Rstest core and
+// under which local name, because an edit has to write the name the file
+// actually binds rather than the spelling the namespace is exported under.
+func (file *fileNamespaces) collectImports() {
```
swwind, line 115: collectImports only walks ESM named imports, but Rstest also supports CommonJS bindings such as `const { rs } = require('@rstest/core')` and `const { rstest: helper } = require('rstack/test')`. The first form receives no autofix; the aliased form incorrectly falls back to `rs.mocked(...)`, even though only `helper` is bound. Please resolve these bindings with `resolveModuleRequireBinding`, or explicitly withhold autofixes for them and add regression tests.

elecmonkey: Fixed in 09bcf41f. `collectImports` walked only ESM named imports, so a namespace destructured out of `require('@rstest/core')` was invisible to it: `const { rs } = require('@rstest/core')` read as an unrelated declaration capturing the name and withheld the edit, and `const { rstest: helper } = require('rstack/test')` left `helper` unseen, so the fix named `rstest`, a binding the file does not have. Both forms are now collected the way named imports are, under the local name the pattern binds, and such a binding no longer makes the name unavailable — a renamed property is read with `Text()`, so a string-literal one works here too. A rest element, a nested pattern and a `require` of an unrelated module bind nothing the rule may reach for and are left alone. Regression tests cover all three.

swwind (changes requested):

internal/plugins/rstest/rules/prefer_rs_mocked/prefer_rs_mocked.go
```diff
@@ -0,0 +1,455 @@
...
+	visit(file.ctx.SourceFile.AsNode())
+}
+
+// collectImports records which namespaces the file imports from Rstest core and
+// under which local name, because an edit has to write the name the file
+// actually binds rather than the spelling the namespace is exported under. Both
+// ESM named imports and the CommonJS `const { rs } = require(...)` form bind a
+// namespace, so both are collected.
+func (file *fileNamespaces) collectImports() {
+	for _, statement := range file.ctx.SourceFile.Statements.Nodes {
```
swwind: [P1] Collect require bindings outside top-level statements

`collectImports` only scans `SourceFile.Statements`, so require bindings declared inside a function or block are missed. For `const { rstest: helper } = require('@rstest/core')`, the fixer falls back to `rs.mocked(...)`, even though only `helper` is bound. Please collect these bindings throughout the file or withhold the fix, and add a regression test.

internal/plugins/rstest/rules/prefer_rs_mocked/prefer_rs_mocked.go
```diff
@@ -0,0 +1,455 @@
...
+	binding := element.AsBindingElement()
+	if binding == nil || binding.DotDotDotToken != nil {
+		return "", ""
+	}
+	name := binding.Name()
+	if name == nil || name.Kind != ast.KindIdentifier {
+		return "", ""
+	}
+	if binding.PropertyName != nil {
+		return name.Text(), binding.PropertyName.Text()
```
swwind: [P1] Guard computed binding property names before calling Text

For valid code such as `const { ["rs"]: helper } = require('@rstest/core')`, `binding.PropertyName` is a `ComputedPropertyName`, and calling `Text()` panics when edits are materialized. Please extract only supported static names or reject computed names before calling `Text()`, with a regression test.

swwind (changes requested):

internal/plugins/rstest/rules/prefer_rs_mocked/prefer_rs_mocked.go
elecmonkey: Fixed in 548d1cc8. The require scan read `SourceFile.Statements`, so a binding declared inside a function or a block was invisible: the local name was never collected, and `const { rstest: helper } = require('@rstest/core')` in a block wrote a fix naming a namespace the file does not bind. A `require` may sit wherever a declaration may, so the whole file is now walked for one, and the ESM scan stays on the statement list, which is the only place a named import appears. Regression test covers the binding inside a function body.

internal/plugins/rstest/rules/prefer_rs_mocked/prefer_rs_mocked.go
elecmonkey: Fixed in 548d1cc8, and it did panic — the added case reproduced `Unhandled case in Node.Text: *ast.ComputedPropertyName` when the fix was materialized. Unlike an import specifier, a binding element may name the renamed property with a computed expression, which carries no text of its own, so the name is no longer read off the node directly. It is resolved when a static string spells it, so `const { ["rs"]: helper } = require('@rstest/core')` fixes to `helper.mocked(value)`; anything else yields no namespace, and the local name then blocks the edit the way any unrelated binding does, so `const { [key]: rs } = require('@rstest/core')` reports without one. Regression tests cover both.

internal/plugins/rstest/rules/prefer_rs_mocked/prefer_rs_mocked.go
elecmonkey: Fixed in 548d1cc8. `collectImports` no longer reads only `SourceFile.Statements`: it walks the whole file for a `VariableDeclaration` whose initializer is an rstest `require`, so a binding declared inside a function or a block is collected too. A regression test covers `const { rstest: helper } = require('@rstest/core')` inside a function body.

internal/plugins/rstest/rules/prefer_rs_mocked/prefer_rs_mocked.go
elecmonkey: Fixed in 548d1cc8. `requireBindingNames` now routes `binding.PropertyName` through `bindingPropertyName`, which accepts only an identifier or a static string (`StringLiteral` / `NoSubstitutionTemplateLiteral`) and otherwise returns not-ok, so `Text()` is never called on a `ComputedPropertyName`. A binding whose property name cannot be resolved withholds the namespace, which blocks the edit the same way any unrelated local binding does. Regression test added for `const { ["rs"]: helper } = require('@rstest/core')`.

fansenze (approved):

PR #2034, opened 2026-09-03
feat: port rule react/sort-comp

swwind:
## Summary

Port the `react/sort-comp` rule from ESLint to rslint.

Enforce a consistent order for methods and properties in React components.

## Related Links

- ESLint rule: https://github.com/jsx-eslint/eslint-plugin-react/blob/v7.37.5/docs/rules/sort-comp.md
- Source code: https://github.com/jsx-eslint/eslint-plugin-react/blob/v7.37.5/lib/rules/sort-comp.js

## Checklist


internal/plugins/react/rules/sort_comp/sort_comp.go
```diff
@@ -0,0 +1,536 @@
...
+		expanded := expandedOrder(opts)
+		order := make([]pattern, len(expanded))
+		for index, entry := range expanded {
+			order[index] = parsePattern(entry)
+		}
+		pragma := reactutil.GetReactPragmaFromContext(ctx)
+		createClass := reactutil.GetReactCreateClass(ctx.Settings)
+		errors := map[int]*storedError{}
+		checkClass := func(node *ast.Node) {
+			if reactutil.ExtendsReactComponent(node, pragma) {
```
elecmonkey: [P2] Recognize React components declared through JSDoc

When a class is declared as a React component only through `/** @extends React.Component */` or `/** @augments React.Component */`, this check skips the class entirely. For example, `/** @extends React.Component */ class Foo { render() {} componentDidMount() {} }` reports `render should be placed after componentDidMount` with eslint-plugin-react@7.37.5, while Rslint reports no diagnostic. Upstream `Components.detect` recognizes both annotations through `isExplicitComponent`, so this class-detection path needs equivalent JSDoc handling.

swwind: Fixed in bae43add4. react/sort-comp now recognizes classes explicitly marked with @extends React.Component or @augments React.PureComponent. Added focused regressions and updated the persistent corpus. The affected rule and reactutil package tests pass.

internal/plugins/react/rules/sort_comp/sort_comp.go
```diff
@@ -0,0 +1,536 @@
...
+		expanded := expandedOrder(opts)
+		order := make([]pattern, len(expanded))
+		for index, entry := range expanded {
+			order[index] = parsePattern(entry)
+		}
+		pragma := reactutil.GetReactPragmaFromContext(ctx)
+		createClass := reactutil.GetReactCreateClass(ctx.Settings)
+		errors := map[int]*storedError{}
+		checkClass := func(node *ast.Node) {
+			if reactutil.ExtendsReactComponent(node, pragma) || reactutil.IsExplicitReactComponent(node) {
```
elecmonkey, line 530: [P2] Recognize computed React component inheritance

This component check misses `class Foo extends React[Component] { render() {} componentDidMount() {} }` because `ExtendsReactComponent` only accepts `PropertyAccessExpression`. eslint-plugin-react@7.37.5 treats this as a `MemberExpression` and reports `render should be placed after componentDidMount`, while Rslint reports nothing. Please also handle identifier-based `ElementAccessExpression` here or in `ExtendsReactComponent`.

internal/plugins/react/rules/sort_comp/sort_comp.go
```diff
@@ -0,0 +1,536 @@
...
+}
+
+func reportErrors(ctx rule.RuleContext, errors map[int]*storedError) {
+	keys := make([]int, 0, len(errors))
+	for key := range errors {
+		keys = append(keys, key)
+	}
+	sort.Ints(keys)
+	for _, key := range keys {
+		stored := errors[key]
```
elecmonkey: [P2] Preserve snapshot semantics while deduplicating errors

Reloading each entry from the mutated map and skipping deleted entries differs from upstream’s `Object.entries(errors)` snapshot semantics. With `{ order: ['lifecycle', 'render'] }`, `class Foo extends React.Component { render() {} static foo() {} componentDidMount() {} onClick() {} displayName() {} }` produces only `displayName should be placed before componentDidMount` upstream, but Rslint additionally reports `render should be placed after componentDidMount`. Please snapshot both keys and values before deletion so every original entry still participates in deduplication.

internal/plugins/react/rules/sort_comp/sort_comp.go
swwind: Fixed in b037e699a. reactutil.ExtendsReactComponent now recognizes identifier-based ElementAccessExpression forms such as React[Component]. Added a focused regression test and persistent corpus case. The focused Go test and isolated ESLint corpus differential test pass.

internal/plugins/react/rules/sort_comp/sort_comp.go
swwind: Fixed in b037e699a. reportErrors now snapshots both error keys and values before deduplication, matching upstream's Object.entries(errors) semantics. Added a focused regression test and persistent corpus case. The focused Go test and isolated ESLint corpus differential test pass.

internal/plugins/react/rules/sort_comp/sort_comp.go
```diff
@@ -0,0 +1,540 @@
...
+	if first < 0 || first == last || last == 0 {
+		return pattern{text: text}
+	}
+	flags := text[last+1:]
+	flagEnd := 0
+	for flagEnd < len(flags) && strings.ContainsRune("gimsuy", rune(flags[flagEnd])) {
+		flagEnd++
+	}
+	flags = flags[:flagEnd]
+	re, err := esregexp.Compile(text[first+1:last], flags)
```
elecmonkey, line 257: [P2] Reject invalid slash-regex order entries

When an order entry looks like the upstream slash-regex form but is invalid, this branch silently falls back to treating it as a literal method name. Upstream eslint-plugin-react@7.37.5 calls new RegExp(...) and throws for inputs like `{ order: ["/[/", "render"] }`; Rslint instead accepts the config and then misses `class Foo extends React.Component { render() {} a() {} }` entirely because `a` no longer matches the intended first group. This makes malformed configs produce false negatives instead of the same configuration failure users get from ESLint.

swwind: Fixed in 10d1e69e6. `parsePattern` now returns compilation errors instead of treating invalid slash-regex entries as literal names, and `react/sort-comp` rejects them during rule initialization. Added `TestSortCompRejectsInvalidRegexPattern` and appended `invalid-regex-pattern.ts` to the persistent corpus. Focused Go tests, the corpus differential run, and the eslint-plugin-react@7.37.5 validation all pass.

elecmonkey (approved):

PR #2035, opened 2026-09-03
feat: port rule react/no-adjacent-inline-elements

swwind:
## Summary

Port the `react/no-adjacent-inline-elements` rule from ESLint to rslint.

Disallow adjacent inline HTML elements without whitespace between them.

## Related Links

- ESLint rule: https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-adjacent-inline-elements.md
- Source code: https://github.com/jsx-eslint/eslint-plugin-react/blob/master/lib/rules/no-adjacent-inline-elements.js

## Checklist


internal/plugins/react/rules/no_adjacent_inline_elements/no_adjacent_inline_elements.go
```diff
@@ -0,0 +1,162 @@
...
+			ast.KindJsxElement: func(node *ast.Node) {
+				jsx := node.AsJsxElement()
+				if jsx.Children == nil {
+					return
+				}
+				validate(ctx, node, jsx.Children.Nodes)
+			},
+			ast.KindCallExpression: func(node *ast.Node) {
+				call := node.AsCallExpression()
+				if !reactutil.IsCreateElementCallWithChecker(call.Expression, pragma, ctx.TypeChecker) {
```
elecmonkey, line 150: This makes the rule depend on `IsCreateElementCallWithChecker`’s syntax-only fallback when no type checker is available, which can over-match shadowed bare `createElement` bindings. For example, with `import { createElement } from "react"; function f(createElement) { createElement("div", null, [foo("a"), foo("span")]); }`, upstream `eslint-plugin-react@7.37.5` reports 0 diagnostics because the parameter shadows the import, but this PR reports `inlineElement` in source-only API mode. Since this rule is not marked `RequiresTypeInfo`, it should either avoid the imprecise fallback for bare callees or add scope-aware shadowing protection before treating the call as React’s `createElement`.

swwind: Fixed in commit 412b8dd82. In source-only mode, react/no-adjacent-inline-elements now skips bare createElement calls when no TypeChecker is available, avoiding the imprecise import fallback and the shadowing false positive. Added a permanent regression test and extended the corpus.

elecmonkey:
conflict

elecmonkey (dismissed):
LGTMMMMMM

internal/plugins/react/rules/no_adjacent_inline_elements/no_adjacent_inline_elements.go
swwind: Resolved in merge commit `e150d824d`. I preserved `origin/main`'s `ESTreeCallCallee`/JSDoc normalization and the PR's computed `createElement` handling; the source-only shadowing regression and the full ESLint corpus differential test pass.

swwind:
Resolved by merging `origin/main` into the PR branch in `e150d824d`; the conflict in `internal/plugins/react/reactutil/create_element.go` is fixed.

elecmonkey (dismissed):

elecmonkey (approved):

PR #2036, opened 2026-09-03
feat: port rule react/static-property-placement

swwind:
## Summary

Port the `react/static-property-placement` rule from ESLint to rslint.

Enforce where React component static properties are declared in an ES6 class.

## Related Links

- ESLint rule: https://github.com/jsx-eslint/eslint-plugin-react/blob/v7.37.5/docs/rules/static-property-placement.md
- Source code: https://github.com/jsx-eslint/eslint-plugin-react/blob/v7.37.5/lib/rules/static-property-placement.js

## Checklist

- [x] Tests updated.
- [x] Documentation updated.


internal/plugins/react/rules/static_property_placement/static_property_placement.go
```diff
@@ -0,0 +1,496 @@
...
+	symbol := ctx.Refs.Resolve(root)
+	if symbol != nil && symbol.ValueDeclaration != nil {
+		declaration := symbol.ValueDeclaration
+		if isReactClass(resolveComponentPath(declaration, path), pragma) {
+			return true
+		}
+		// A successfully resolved local binding is authoritative. Falling
+		// back to a same-name class elsewhere in the file would cross a
+		// shadowing boundary and report a non-component binding.
+		return false
```
elecmonkey, line 293: [P2] Handle components created through member assignment

For `let Box = {}; Box.C = class extends React.Component {}; Box.C.propTypes = {};`, eslint-plugin-react@7.37.5 reports `notStaticClassProp`, while this implementation reports nothing. `getRelatedComponent` also scans references to `Box` and finds the `Box.C = class ...` assignment, but this branch only evaluates `Box`'s `{}` initializer and then returns false. This misses static-property placement checks for components created by assigning a class expression to an object member. Please include matching reference assignments when resolving the related component.

swwind: Thanks for catching this. Fixed in 5686a3d0c by checking references to the resolved binding for matching member assignments such as `Box.C = class extends React.Component {}`. Added a permanent regression test and appended the case to the persistent ESLint corpus. The focused rule tests and all four corpus differential runs pass. The Windows Go CI job was cancelled by the 60-minute workflow timeout, which also caused CI Done to fail; no test failure output was produced.

internal/plugins/react/rules/static_property_placement/static_property_placement.go
```diff
@@ -0,0 +1,541 @@
...
+	if !ok || ctx.Refs == nil {
+		return false
+	}
+	symbol := ctx.Refs.Resolve(root)
+	if symbol != nil && symbol.ValueDeclaration != nil {
+		declaration := symbol.ValueDeclaration
+		if isReactClass(resolveComponentPath(declaration, path), pragma) {
+			return true
+		}
+		if relatedReactClassAssignment(ctx, symbol, path, pragma) {
```
elecmonkey: [P2] Don’t use later assignments for earlier property writes

When the checked property write appears before a later component assignment, this new `relatedReactClassAssignment` scan still finds that later reference and reports, diverging from eslint-plugin-react@7.37.5. For `let Box = { C: {} }; Box.C.propTypes = {}; Box.C = class extends React.Component {};`, upstream emits no diagnostics because `getRelatedComponent` stops on the current `Box.C.propTypes` reference before reaching the later `Box.C = class ...`; the PR reports `notStaticClassProp`. Please make this lookup preserve upstream reference order/source-position behavior so future assignments do not classify earlier member writes.

swwind: Fixed in `75f3e9704`. `relatedReactClassAssignment` now only considers matching member assignments whose reference starts before the current property write, matching eslint-plugin-react@7.37.5's reference-order behavior. Added the regression test and appended the case to the persistent ESLint corpus. Focused rule tests and all four corpus differential runs pass.

internal/plugins/react/rules/static_property_placement/static_property_placement.go
```diff
@@ -0,0 +1,546 @@
...
+			parent = parent.Parent
+		}
+		if parent == nil || parent.Kind != ast.KindBinaryExpression {
+			continue
+		}
+		binary := parent.AsBinaryExpression()
+		if binary.OperatorToken == nil || binary.OperatorToken.Kind == ast.KindCommaToken || binary.Right == nil {
+			continue
+		}
+		if isReactClass(reactutil.SkipExpressionWrappers(binary.Right), pragma) {
```
elecmonkey, line 354: [P2] Preserve upstream first-match reference behavior

`relatedReactClassAssignment` keeps scanning earlier matching references until it finds a React class assignment, but upstream `getRelatedComponent` stops at the first matching `Box.C` reference. For `let Box = {}; Box.C = {}; Box.C = class extends React.Component {}; Box.C.propTypes = {};`, `eslint-plugin-react@7.37.5` reports no diagnostics because the first `Box.C` reference resolves to `{}`, while this implementation skips past it and reports `notStaticClassProp` from the later class assignment. Please make the reference lookup stop once the first same-path reference before the current write has been considered, even when it is not a React class.

swwind: Fixed in `411c8bf99`. `relatedReactClassAssignment` now stops at the first matching same-path reference before the property write, including non-React assignments, matching eslint-plugin-react@7.37.5. Added a permanent regression test and appended the case to the persistent corpus. The focused rule tests and all four corpus differential runs pass.

internal/plugins/react/rules/static_property_placement/static_property_placement.go
```diff
@@ -0,0 +1,547 @@
...
+	})
+}
+
+func isReactClass(node *ast.Node, pragma string) bool {
+	if node == nil {
+		return false
+	}
+	switch node.Kind {
+	case ast.KindClassDeclaration, ast.KindClassExpression:
+		return reactutil.ExtendsReactComponent(node, pragma)
```
elecmonkey: [P2] Recognize JSDoc-declared React components

eslint-plugin-react@7.37.5 treats a class with `/** @extends React.Component */` or `/** @augments React.Component */` as an ES6 component before checking its `extends` clause, but this new rule only calls `reactutil.ExtendsReactComponent`, which is heritage-based. As a result, `/** @extends React.Component */ class C { static get propTypes() { return {}; } }` produces upstream `notStaticClassProp` while Rslint reports nothing, and the same gap applies to post-class `C.propTypes = {}` placement checks for JSDoc-only components.

swwind: Fixed in `0a79d0493`. The rule now recognizes `@extends React.Component` and `@augments React.Component` JSDoc declarations for both class-member and post-class assignment checks, while requiring the explicit `React.Component`/`React.PureComponent` form. Added permanent regressions and appended both cases to the persistent corpus. The focused rule tests and all four corpus differential runs pass.

internal/plugins/react/rules/static_property_placement/static_property_placement.go
```diff
@@ -0,0 +1,604 @@
...
+	property := node.AsPropertyAccessExpression()
+	object := ast.SkipParentheses(property.Expression)
+	name := property.Name()
+	return object != nil && object.Kind == ast.KindIdentifier && object.AsIdentifier().Text == "React" &&
+		name != nil && name.Kind == ast.KindIdentifier &&
+		(name.AsIdentifier().Text == "Component" || name.AsIdentifier().Text == "PureComponent")
+}
+
+func parentReactClass(node *ast.Node, pragma string) *ast.Node {
+	if component := reactutil.GetParentReactComponentScopeBased(node, pragma, ""); component != nil {
```
elecmonkey: [P2] Don’t treat createReactClass scopes as ES6 class parents

The class-member listener upstream uses `componentUtil.getParentES6Component` only, so a helper class declared inside `createReactClass({ render() { ... } })` is not checked unless that helper class itself extends React. With this `parentReactClass` helper, `GetParentReactComponentScopeBased` can return the surrounding ES5 `createReactClass` object before the loop reaches the nested helper class, so `const C = createReactClass({ render() { class Helper { static get propTypes() { return {}; } } return null; } });` with `property assignment` is reported as `declareOutsideClass` by Rslint while eslint-plugin-react@7.37.5 reports no diagnostics. Please keep the class-member path ES6-class-only and use the JSDoc extension check only for the nearest class, not the surrounding ES5 component scope.

elecmonkey: Fixed in `cd64ef5f0`. `parentReactClass` no longer falls back to the scope-based component lookup, whose ES5 path could resolve the surrounding `createReactClass` object literal. It now walks out to the nearest enclosing class only and treats it as the component just when that class extends React or carries an explicit `@extends`/`@augments React.Component` JSDoc tag, matching `componentUtil.getParentES6Component`.

Checked against eslint-plugin-react@7.37.5: `const C = createReactClass({ render() { class Helper { static get propTypes() { return {}; } } return null; } });` with `property assignment` reports nothing on both sides, while the same helper extending `React.Component` still reports `declareOutsideClass` on both sides. Both cases are added as regression tests, and the react rule tests pass.

elecmonkey (approved):

fansenze (approved):

PR #2039, opened 2026-09-03
feat: port rule react/function-component-definition

elecmonkey:
## Summary

Port the `react/function-component-definition` rule from ESLint to rslint.

The rule enforces a consistent function type for React function components — function declaration, function expression, or arrow function — separately for named and unnamed components, with an autofix that rewrites the function's head.

## Related Links

- ESLint rule: https://github.com/jsx-eslint/eslint-plugin-react/blob/v7.37.5/docs/rules/function-component-definition.md
- Source code: https://github.com/jsx-eslint/eslint-plugin-react/blob/v7.37.5/lib/rules/function-component-definition.js

## Checklist

### DIFF
--- a/internal/plugins/react/rules/function_component_definition/function_component_definition.go
+++ b/internal/plugins/react/rules/function_component_definition/function_component_definition.go
@@ -75,25 +75,25 @@ func parseOptions(opts []any) options {
 		return parsed
 	}
 	optsMap, _ := opts[0].(map[string]any)
-	if list := readStringOrList(optsMap["namedComponents"]); len(list) > 0 {
+	if list, ok := readStringOrList(optsMap["namedComponents"]); ok {
 		parsed.named = list
 	}
-	if list := readStringOrList(optsMap["unnamedComponents"]); len(list) > 0 {
+	if list, ok := readStringOrList(optsMap["unnamedComponents"]); ok {
 		parsed.unnamed = list
 	}
 	return parsed
 }
 
 // readStringOrList mirrors JS's `[].concat(value)` coercion for the
-// `string | string[]` option shape. An empty array stays empty, which lets
-// the caller fall back to the default exactly like upstream's `|| default`
-// does for `undefined` — note that upstream keeps an explicitly configured
-// empty array as-is, so we only substitute the default when the key is
-// absent or not a recognized shape.
-func readStringOrList(raw any) []string {
+// `string | string[]` option shape. The bool result says whether the key was
+// present in a recognized shape at all: upstream's `configuration.x || default`
+// substitutes the default only for `undefined`, and an explicitly configured
+// `[]` is truthy in JS and therefore kept as an empty list. So an empty array
+// returns `(nil, true)` and must NOT fall back to the default.
+func readStringOrList(raw any) ([]string, bool) {
 	switch v := raw.(type) {
 	case string:
-		return []string{v}
+		return []string{v}, true
 	case []any:
 		out := make([]string, 0, len(v))
 		for _, item := range v {
@@ -101,9 +101,9 @@ func readStringOrList(raw any) []string {
 				out = append(out, s)
 			}
 		}
-		return out
+		return out, true
 	}
-	return nil
+	return nil, false
 }
 
 var FunctionComponentDefinitionRule = rule.Rule{
@@ -241,11 +241,17 @@ func (w *walker) validate(node *ast.Node, functionType string) {
 	}
 
 	named := hasName(node)
-	if named && !contains(w.cfg.named, functionType) {
+	// An explicitly configured empty list allows nothing, so upstream reaches
+	// `report({ messageId: config[0] })` with `config[0] === undefined` and
+	// ESLint aborts the whole lint run with "Missing `message` property in
+	// report() call". rslint has no equivalent of that crash; staying silent is
+	// the closest safe behavior and, unlike falling back to the default list,
+	// it never invents a diagnostic upstream would not have produced.
+	if named && len(w.cfg.named) > 0 && !contains(w.cfg.named, functionType) {
 		target := w.cfg.named[0]
 		w.report(node, target, namedFunctionTemplates[target], w.namedFixRange(node))
 	}
-	if !named && !contains(w.cfg.unnamed, functionType) {
+	if !named && len(w.cfg.unnamed) > 0 && !contains(w.cfg.unnamed, functionType) {
 		target := w.cfg.unnamed[0]
 		w.report(node, target, unnamedFunctionTemplates[target], utils.TrimNodeTextRange(w.ctx.SourceFile, node))
 	}
@@ -470,10 +476,19 @@ func (w *walker) paramsText(node *ast.Node) string {
 
 // bodyText mirrors upstream's `getBody`: a block body is copied verbatim,
 // while an arrow's expression body is wrapped into an explicit block.
+//
+// Upstream slices `node.body.range`, and ESTree has no node for parentheses,
+// so `const Hello = (props) => (cond ? <A /> : null);` yields the body text
+// WITHOUT the source parentheses. tsgo keeps a ParenthesizedExpression node,
+// so the outer parens have to be peeled back explicitly to produce the same
+// fix.
 func (w *walker) bodyText(node *ast.Node) (string, bool) {
 	body := reactutil.FunctionBody(node)
 	if body == nil {
 		return "", false
+	}
+	if body.Kind != ast.KindBlock {
+		body = ast.SkipParentheses(body)
 	}
 	bodyRange := utils.TrimNodeTextRange(w.ctx.SourceFile, body)
 	if body.Kind != ast.KindBlock {
--- a/internal/plugins/react/rules/function_component_definition/function_component_definition.go
+++ b/internal/plugins/react/rules/function_component_definition/function_component_definition.go
@@ -1,0 +1,530 @@
+package function_component_definition
+
+import (
+	_ "embed"
+	"strings"
+
+	"github.com/microsoft/typescript-go/shim/ast"
+	"github.com/microsoft/typescript-go/shim/core"
+	"github.com/microsoft/typescript-go/shim/scanner"
+	"github.com/web-infra-dev/rslint/internal/plugins/react/reactutil"
+	"github.com/web-infra-dev/rslint/internal/rule"
+	"github.com/web-infra-dev/rslint/internal/utils"
+)
+
+//go:embed function_component_definition.schema.json
+var schemaJSON []byte
+
+// The three function forms the rule can enforce. These double as messageIds,
+// matching upstream's `messages` map keys.
+const (
+	typeFunctionDeclaration = "function-declaration"
+	typeFunctionExpression  = "function-expression"
+	typeArrowFunction       = "arrow-function"
+)
+
+// namedFunctionTemplates / unnamedFunctionTemplates mirror upstream's
+// NAMED_FUNCTION_TEMPLATES / UNNAMED_FUNCTION_TEMPLATES.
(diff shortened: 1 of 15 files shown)


internal/plugins/react/rules/function_component_definition/function_component_definition.go
```diff
@@ -0,0 +1,530 @@
...
+	// Upstream: `[].concat(configuration.namedComponents || 'function-declaration')`.
+	parsed := options{
+		named:   []string{typeFunctionDeclaration},
+		unnamed: []string{typeFunctionExpression},
+	}
+	if len(opts) == 0 {
+		return parsed
+	}
+	optsMap, _ := opts[0].(map[string]any)
+	if list := readStringOrList(optsMap["namedComponents"]); len(list) > 0 {
```
elecmonkey: [P2] Preserve explicit empty component option arrays

The `len(list) > 0` guard makes an explicitly configured empty array fall back to the defaults, so `{ namedComponents: [] }` accepts `function Hello() { return <div/>; }` and `{ unnamedComponents: [] }` accepts returned function expressions. In eslint-plugin-react@7.37.5 those empty arrays are schema-valid and are kept as empty by `[].concat(configuration.namedComponents || "function-declaration")`; when a matching component is reached, upstream does not silently use the default and instead reports through an undefined target, throwing `Missing message property in report() call`. I verified this with upstream ESLint and with a temporary rslint rule test, which produced no diagnostics for both empty-array configurations, so this currently hides a misconfiguration/behavior difference rather than matching upstream.

internal/plugins/react/rules/function_component_definition/function_component_definition.go
```diff
@@ -0,0 +1,530 @@
...
+	// Confidence 0: `Components.detect`'s FunctionExpression /
+	// FunctionDeclaration listeners permanently ban an async generator.
+	if node.Kind != ast.KindArrowFunction && reactutil.IsAsyncGeneratorFunction(node) {
+		return false
+	}
+	wrapper := reactutil.OutermostComponentWrapperCall(node, w.pragma, w.wrappers, w.ctx.TypeChecker)
+	if wrapper != nil && reactutil.WrapperWrapsKnownSiblingComponent(wrapper, node) {
+		return false
+	}
+	if !reactutil.IsStatelessReactComponentWithWrappers(node, w.pragma, w.ctx.TypeChecker, w.wrappers) {
```
swwind, line 529: Could this detection keep TypeScript assertion wrappers opaque? ESLint does not report `function Hello() { return <A /> as ReactNode; }`, but rslint reports it when `namedComponents` is `arrow-function` or `function-expression`. `isJSXExpression` currently unwraps `as`, causing this false positive.

swwind, line 529: `const Hello = (props) => <div /> satisfies ReactNode;` is valid upstream, but rslint reports it as an unnamed component. The detector unwraps `SatisfiesExpression`, while `declaratorParent` does not, so the node is detected but loses its variable name. Please keep these wrapper semantics consistent with ESLint.

swwind, line 529: Please distinguish `export =` from `export default` here. ESLint does not report `export = function Hello() { return <div />; }`, while rslint reports it under `arrow-function`. The shared detector treats both forms as `ast.KindExportAssignment`.

swwind (commented):
Found three behavior mismatches against ESLint in the TSX differential corpus.

internal/plugins/react/rules/function_component_definition/function_component_definition.go
```diff
@@ -0,0 +1,530 @@
...
+}
+
+// bodyText mirrors upstream's `getBody`: a block body is copied verbatim,
+// while an arrow's expression body is wrapped into an explicit block.
+func (w *walker) bodyText(node *ast.Node) (string, bool) {
+	body := reactutil.FunctionBody(node)
+	if body == nil {
+		return "", false
+	}
+	bodyRange := utils.TrimNodeTextRange(w.ctx.SourceFile, body)
```
swwind, line 493: [P2] Preserve parentheses in arrow expression bodies during fixes

ESLint preserves the source parentheses here because it uses `node.body.range`. For example, converting `const Hello = (props) => (condition ? <A /> : null);` produces `return (condition ? <A /> : null)`, while `TrimNodeTextRange` removes the parentheses and rslint produces `return condition ? <A /> : null`. Please use the actual body range/text for arrow expression bodies.

swwind (commented):
Full-corpus review found no additional false negatives. The existing review comments cover the `as`, `satisfies`, `export =`, and empty-option-array mismatches. This review adds the autofix mismatch where parentheses around arrow expression bodies are removed.

internal/plugins/react/rules/function_component_definition/function_component_definition.go
elecmonkey: Fixed in c220fbf2.

`readStringOrList` now returns a "key was present" flag, so an explicitly configured `[]` stays empty instead of falling back to the default list, and the two report sites skip a category whose list is empty. I confirmed upstream behavior with eslint-plugin-react v7.37.5 + @typescript-eslint/parser: both `{ namedComponents: [] }` and `{ unnamedComponents: [] }` abort the run with "Missing `message` property in report() call". rslint cannot reproduce that crash, so it now stays silent for that category — the important part being that it no longer invents a diagnostic from the default list. Added valid cases for all three shapes and a note in the rule doc.

internal/plugins/react/rules/function_component_definition/function_component_definition.go
elecmonkey: Fixed in c220fbf2. Confirmed the false positive: `function Hello() { return <A /> as ReactNode; }` was reported under both `arrow-function` and `function-expression`, while upstream reports nothing.

`isJSXExpression` now uses paren-only skipping (`ast.SkipParentheses`) instead of `reactutil.SkipExpressionWrappers`. That matches upstream: ESTree has no node for parentheses, but TSESTree does keep `TSAsExpression` / `TSSatisfiesExpression` / `TSTypeAssertion` / `TSNonNullExpression`, and `isJSXValue` falls through to `default: return false` for all of them.

This is the shared JSX-return predicate, so it also affects no-unstable-nested-components: two cases there asserted the unwrapping behavior. I checked both against upstream (`return <div /> satisfies React.ReactNode` and `return React.createElement("div", null)!` inside a nested function) — upstream reports nothing for either, so those cases moved to valid.

elecmonkey: Fixed in c220fbf2, but the mechanism is not the one described here, so recording what I measured.

`satisfies` binds inside the arrow body: in `const Hello = (props) => <div /> satisfies ReactNode;` the body is the SatisfiesExpression and the arrow’s parent is still the declarator. rslint reported it as a **named** component ("Function component is not a function declaration"), not as an unnamed one, and no name was lost.

The `declaratorParent` half is already consistent with ESLint: `const Hello = ((props) => <div />) satisfies ReactNode;` — the shape where the wrapper really does sit between the arrow and the declarator — produces no diagnostic today, because `isInAllowedPositionForComponent` only looks through parentheses and a SatisfiesExpression parent is not an accepted position. That matches upstream, which is also silent there.

So this was the same root cause as the `as` comment (the wrapper being peeled back in the JSX-return predicate) and is covered by the same fix, with a valid case added for the unparenthesized form.

elecmonkey: I could not reproduce this one, so I have not changed anything here.

Against this branch I ran, in a `.tsx` file with `namedComponents: "arrow-function"` and with `"function-expression"`:

```ts
export = function Hello() { return <div />; };
export = function Hello() { return <div />; }   // no semicolon
export = function () { return <div />; };
```

All of them are silent in rslint, and silent in eslint-plugin-react v7.37.5 as well. The only case that reports in the same batch is `export default function Hello() { return <div />; }`, which upstream also reports.

I agree that `ast.KindExportAssignment` not distinguishing `IsExportEquals` is worth tightening on principle, but I could not find an input where it produces a diagnostic difference. If you have a snippet where rslint reports `export =`, please paste it and I will fix it.

internal/plugins/react/rules/function_component_definition/function_component_definition.go
elecmonkey: There is a real fix mismatch here, but the direction is the other way round, so the suggested remedy would have kept the bug.

Measured on `const Hello = (props) => (condition ? <A /> : null);`:

- eslint-plugin-react v7.37.5 output: `function Hello(props) {\n  return condition ? <A /> : null\n}` — the parentheses are **dropped**. `getBody` slices `node.body.range`, and ESTree has no node for parentheses, so `node.body` is the ConditionalExpression and its range starts after the `(`.
- this branch before the fix: `function Hello(props) {\n  return (condition ? <A /> : null)\n}` — the parentheses were **kept**, because tsgo keeps a ParenthesizedExpression node and `TrimNodeTextRange` spans it from `(` to `)`.

Fixed in c220fbf2 by peeling back the parentheses of a non-block arrow body before slicing, which reproduces the upstream output. Added invalid cases for both the `function-declaration` and `function-expression` targets.

internal/plugins/react/rules/function_component_definition/function_component_definition.go
```diff
@@ -0,0 +1,545 @@
...
+	// Confidence 0: `Components.detect`'s FunctionExpression /
+	// FunctionDeclaration listeners permanently ban an async generator.
+	if node.Kind != ast.KindArrowFunction && reactutil.IsAsyncGeneratorFunction(node) {
+		return false
+	}
+	wrapper := reactutil.OutermostComponentWrapperCall(node, w.pragma, w.wrappers, w.ctx.TypeChecker)
+	if wrapper != nil && reactutil.WrapperWrapsKnownSiblingComponent(wrapper, node) {
+		return false
+	}
+	if !reactutil.IsStatelessReactComponentWithWrappers(node, w.pragma, w.ctx.TypeChecker, w.wrappers) {
```
swwind, line 529: [P2] Exclude `export =` from component detection

`isDetectedComponentNode` delegates to `IsStatelessReactComponentWithWrappers`, whose `ExportAssignment` branch treats every tsgo `KindExportAssignment` as an ES `export default`. In tsgo, `export =` uses the same kind but has `node.AsExportAssignment().IsExportEquals == true`; ESLint does not classify `export = function Hello() { return <div />; }` as a component. The rule-specific corpus cases `case-65.tsx`, `case-85.tsx`, and `case-86.tsx` are valid upstream but currently report here. Please distinguish `IsExportEquals` before classifying this node and add a regression test.

elecmonkey: Fixed in 684527d3.

Root-caused it one level deeper than the rule: the mismatch lives in the shared
component-detection helper, not in `isDetectedComponentNode`. `reactutil` matched
`KindExportAssignment` by kind alone in two places — the export-default branch of
`isStatelessReactComponentCore`, and the allowed-position list in
`isInAllowedPositionForComponent`. Both now go through a new
`isExportDefaultAssignment` helper that rejects `IsExportEquals`, so `export =`
falls through to the position check and is never classified. That second site
matters: with only the branch guarded, `export = function Hello() { return <div/>; }`
would still have been classified by the trailing named-function-expression check.

Fixing it centrally is the correct scope — typescript-eslint parses `export =` as
`TSExportAssignment`, which eslint-plugin-react's detection never matches, so no
rule built on this helper should treat it as a component. No existing react rule
test encoded the old behavior, and the full `internal/plugins/react/...` suite passes.

Added regression cases for both the named function expression and the arrow form.

swwind (approved):

PR #2040, opened 2026-09-03
test: preserve ESLint JSON test case fields

swwind:
## Summary

Extend the ESLint-format JSON RuleTester bridge to preserve test-case metadata:

- support the `tsx` field for TSX parsing
- preserve `filename` for source-file setup
- preserve `globals` for rule context construction
- add regression coverage for both valid and invalid test cases

## Related Links

N/A

## Validation

- `go test ./internal/rule_tester`

## Checklist

- [x] Tests updated.


elecmonkey (approved):
LGTM

PR #2043, opened 2026-09-03
feat: port rule unicorn/no-null

shiyudesu:
## Summary

Port `unicorn/no-null` from `eslint-plugin-unicorn@v74.0.0`.

The rule disallows the `null` literal and encourages using `undefined` as the
single value for missing data.

- Support the `checkArguments` and `checkStrictEquality` options.
- Preserve the upstream exceptions for `Object.create()`, React `useRef()`,
  and `insertBefore()` calls.
- Autofix loose equality comparisons by replacing `null` with `undefined`.
- Provide suggestions for replacing `null`, removing return values, and
  removing mutable variable initializers.
- Ignore `null` in TypeScript type positions.
- Preserve upstream behavior for parentheses, optional calls, TypeScript
  expression wrappers, diagnostic ranges, and suggestion ranges.
- Register the rule in the native Unicorn catalog and enable it in the Unicorn
  recommended preset.
- Migrate the complete upstream v74.0.0 test suite.
- Add coverage for tsgo/ESTree differences, option combinations, edit demand,
  and real-world cases from upstream issues.

## Related Links

- Documentation: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v74.0.0/docs/rules/no-null.md
- Source code: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v74.0.0/rules/no-null.js
- Upstream tests: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v74.0.0/test/no-null.js

## Validation

- `go test -count=1 ./internal/plugins/unicorn/rules/no_null ./internal/plugins/unicorn ./internal/rules ./internal/config`
- `pnpm --dir packages/rslint-test-tools exec rs test --testTimeout=10000 no-null`
- `pnpm typecheck`
- `pnpm lint`
- `pnpm -w run check-spell`
- `pnpm format:check`
- `golangci-lint v2.12.2`: 0 issues

## Checklist

- [x] Tests updated.
- [x] Documentation updated.

### DIFF
--- a/internal/plugins/unicorn/rules/no_null/no_null.go
+++ b/internal/plugins/unicorn/rules/no_null/no_null.go
@@ -202,7 +202,7 @@ func reportNull(ctx rule.RuleContext, node *ast.Node, opts options) {
 		declarationList := parent.Parent
 		if declaration != nil && declarationList != nil &&
 			declarationList.Kind == ast.KindVariableDeclarationList &&
-			declarationList.Flags&ast.NodeFlagsConst == 0 &&
+			!ast.IsVarConst(declarationList) &&
 			utils.ESTreeRuntimeExpression(declaration.Initializer) == node {
 			start := variableIDEnd(ctx, declaration)
 			if start > 0 {
--- a/internal/plugins/unicorn/rules/no_null/no_null.go
+++ b/internal/plugins/unicorn/rules/no_null/no_null.go
@@ -65,7 +65,8 @@ func equalityKind(node *ast.Node) ast.Kind {
 
 func directCallOrNewArgumentParent(node *ast.Node) *ast.Node {
 	parent := utils.ESTreeParent(node)
-	if parent == nil || (!ast.IsCallExpression(parent) && !ast.IsNewExpression(parent)) {
+	if parent == nil || ast.IsImportCall(parent) ||
+		(!ast.IsCallExpression(parent) && !ast.IsNewExpression(parent)) {
 		return nil
 	}
 	for _, argument := range parent.Arguments() {
--- a/internal/plugins/unicorn/rules/no_null/no_null.go
+++ b/internal/plugins/unicorn/rules/no_null/no_null.go
@@ -1,0 +1,245 @@
+package no_null
+
+import (
+	_ "embed"
+
+	"github.com/microsoft/typescript-go/shim/ast"
+	"github.com/microsoft/typescript-go/shim/core"
+	"github.com/web-infra-dev/rslint/internal/plugins/unicorn/unicornutil"
+	"github.com/web-infra-dev/rslint/internal/rule"
+	"github.com/web-infra-dev/rslint/internal/utils"
+)
+
+const (
+	messageIDError   = "error"
+	messageIDReplace = "replace"
+	messageIDRemove  = "remove"
+)
+
+var (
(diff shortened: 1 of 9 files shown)


internal/plugins/unicorn/rules/no_null/no_null.go
```diff
@@ -0,0 +1,245 @@
...
+		})
+		return
+	}
+
+	if parent != nil && parent.Kind == ast.KindVariableDeclaration {
+		declaration := parent.AsVariableDeclaration()
+		declarationList := parent.Parent
+		if declaration != nil && declarationList != nil &&
+			declarationList.Kind == ast.KindVariableDeclarationList &&
+			declarationList.Flags&ast.NodeFlagsConst == 0 &&
```
elecmonkey: For `async function f() { await using value = null; }`, Unicorn v74 emits both the `remove` and `replace` suggestions, while this implementation only emits `replace`. ts-go encodes `await using` as `NodeFlagsConst | NodeFlagsUsing`, so this raw `NodeFlagsConst` bit check treats it as a `const` declaration. Using `ast.IsVarConst(declarationList)` here would exclude only actual `const` declarations and preserve upstream suggestion behavior.

shiyudesu: Good catch, thanks. I changed the raw NodeFlagsConst check to ast.IsVarConst so await using keeps both the remove and replace suggestions, and added regression cases for both using and await using. The targeted Go tests and golangci-lint v2.12.2 pass.

swwind (dismissed):

internal/plugins/unicorn/rules/no_null/no_null.go
```diff
@@ -0,0 +1,245 @@
...
+	parent := utils.ESTreeParent(node)
+	if parent == nil || parent.Kind != ast.KindBinaryExpression {
+		return ast.KindUnknown
+	}
+	return parent.AsBinaryExpression().OperatorToken.Kind
+}
+
+func directCallOrNewArgumentParent(node *ast.Node) *ast.Node {
+	parent := utils.ESTreeParent(node)
+	if parent == nil || (!ast.IsCallExpression(parent) && !ast.IsNewExpression(parent)) {
```
elecmonkey: With `checkArguments: false`, this helper treats every ts-go `CallExpression` argument as an ESLint direct call/new argument and suppresses it. Dynamic `import(null)` is a ts-go `CallExpression`, but in ESTree/Unicorn v74 it is an `ImportExpression`, so upstream still reports the `null` literal. I verified `import(null)` with `{checkArguments:false}` reports one `unicorn/no-null` diagnostic in ESLint + eslint-plugin-unicorn@74.0.0, while this port reports none. Please exclude `ast.IsImportCall(parent)` here so the option only ignores ordinary function/constructor arguments.

shiyudesu: Good catch, thanks. I excluded ast.IsImportCall from the direct call/new argument detection, so checkArguments: false no longer suppresses import(null). I also added a regression case that asserts the diagnostic and replacement suggestion. The targeted Go tests and golangci-lint v2.12.2 pass.

elecmonkey (approved):
LGTM

PR #2045, opened 2026-09-03
feat: port rule unicorn/no-document-cookie

shiyudesu:
## Summary

Port `unicorn/no-document-cookie` from `eslint-plugin-unicorn@v74.0.0`.

The rule disallows assigning to `document.cookie` directly while allowing
cookie reads.

- Follow aliases of the global `document` object.
- Support `globalThis.document`, `window.document`, `self.document`, and
  `global.document`.
- Support statically computed properties such as
  `document["coo" + "kie"]`.
- Preserve upstream behavior for shadowed and modified globals, conditional
  and logical expressions, TypeScript wrappers, converging aliases, and
  diagnostic ranges.
- Register the rule in the native Unicorn catalog and enable it in the Unicorn
  recommended preset.
- Migrate the complete upstream v74.0.0 test suite.
- Add coverage for tsgo/ESTree differences, destructuring, nested aliases,
  static and dynamic properties, and real-world upstream cases.

## Related Links

- Documentation: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v74.0.0/docs/rules/no-document-cookie.md
- Source code: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v74.0.0/rules/no-document-cookie.js
- Upstream tests: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v74.0.0/test/no-document-cookie.js

## Validation

- `go test -count=1 ./internal/plugins/unicorn/rules/no_document_cookie`
- `pnpm --dir packages/rslint-test-tools exec rs test --testTimeout=10000 no-document-cookie`
- `pnpm typecheck`
- `pnpm lint`
- `pnpm -w run check-spell`
- `pnpm format:check`
- `golangci-lint v2.13.2`: 0 issues
- Differential checks against the upstream ESLint rule for JavaScript and
  TypeScript edge cases

## Checklist

- [x] Tests updated.
- [x] Documentation updated.

### DIFF
--- a/internal/plugins/unicorn/rules/no_document_cookie/no_document_cookie.go
+++ b/internal/plugins/unicorn/rules/no_document_cookie/no_document_cookie.go
@@ -17,16 +17,6 @@ const (
 	tracedGlobalObject
 )
 
-type tracedVariable struct {
-	symbol *ast.Symbol
-	value  tracedValue
-}
-
-type tracedGlobal struct {
-	name  string
-	value tracedValue
-}
-
 // documentReferenceTracker follows the global document object through the
 // flow-insensitive aliases supported by @eslint-community/eslint-utils'
 // ReferenceTracker. The active stacks prevent cycles without suppressing the
@@ -35,8 +25,8 @@ type documentReferenceTracker struct {
 	ctx               rule.RuleContext
 	identifiersByName map[string][]*ast.Node
 	propertyEvaluator *utils.StaticStringEvaluator
-	variableStack     map[tracedVariable]bool
-	globalStack       map[tracedGlobal]bool
+	variableStack     map[*ast.Symbol]bool
+	globalStack       map[string]bool
 	problems          []*ast.Node
 }
 
@@ -282,29 +272,25 @@ func (tracker *documentReferenceTracker) trackIdentifier(identifier *ast.Node, v
 }
 
 func documentBindingSymbol(identifier *ast.Node) *ast.Symbol {
-	if identifier == nil || identifier.Kind != ast.KindIdentifier || identifier.Parent == nil {
+	if identifier == nil || identifier.Kind != ast.KindIdentifier ||
+		!utils.IsDeclarationIdentifier(identifier) {
 		return nil
 	}
-	declaration := identifier.Parent
-	if declaration.Name() != identifier {
-		return nil
-	}
-	return declaration.Symbol()
+	return identifier.Parent.Symbol()
 }
 
 func (tracker *documentReferenceTracker) trackVariable(symbol *ast.Symbol, value tracedValue) {
 	if tracker.ctx.Refs == nil || symbol == nil {
 		return
 	}
-	key := tracedVariable{symbol: symbol, value: value}
-	if tracker.variableStack != nil && tracker.variableStack[key] {
+	if tracker.variableStack != nil && tracker.variableStack[symbol] {
 		return
 	}
 	if tracker.variableStack == nil {
-		tracker.variableStack = make(map[tracedVariable]bool)
+		tracker.variableStack = make(map[*ast.Symbol]bool)
 	}
-	tracker.variableStack[key] = true
-	defer delete(tracker.variableStack, key)
+	tracker.variableStack[symbol] = true
+	defer delete(tracker.variableStack, symbol)
 
 	for _, reference := range tracker.ctx.Refs.References(symbol) {
 		if !ast.IsWriteOnlyAccess(reference) {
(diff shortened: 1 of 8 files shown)


internal/plugins/unicorn/rules/no_document_cookie/no_document_cookie.go
```diff
@@ -0,0 +1,392 @@
...
+	}
+}
+
+func (tracker *documentReferenceTracker) accessExpressionStaticName(node *ast.Node) (string, bool) {
+	if node.Kind == ast.KindElementAccessExpression {
+		argument := node.AsElementAccessExpression().ArgumentExpression
+		if tracker.propertyEvaluator == nil {
+			tracker.propertyEvaluator = utils.NewStaticStringEvaluatorWithoutScope()
+		}
+		return tracker.propertyEvaluator.Eval(argument)
```
elecmonkey: [P2] Coerce computed property values to strings

`Eval` only accepts values already typed as strings, while upstream's `getPropertyName` uses `getStringIfConstant`, which applies JavaScript `String(...)` coercion to any statically known value. Consequently, `document[["cookie"]] = "x"` produces one diagnostic with `eslint-plugin-unicorn@74.0.0` but none in Rslint; the same issue prevents computed destructuring keys such as `const {[["document"]]: doc} = globalThis` from propagating the alias. Use the evaluator's string-coercing path for computed property names.

internal/plugins/unicorn/rules/no_document_cookie/no_document_cookie.go
```diff
@@ -0,0 +1,392 @@
...
+		tracker.trackGlobalVariable(name, value)
+	}
+}
+
+func documentBindingSymbol(identifier *ast.Node) *ast.Symbol {
+	if identifier == nil || identifier.Kind != ast.KindIdentifier || identifier.Parent == nil {
+		return nil
+	}
+	declaration := identifier.Parent
+	if declaration.Name() != identifier {
```
elecmonkey: [P2] Limit the symbol fallback to actual declarations

`documentBindingSymbol` accepts any parent whose `Name()` equals the identifier, including a shorthand property in a destructuring assignment. For `({document} = globalThis); document.cookie = "x"`, this selects the shorthand property's symbol instead of following the configured global variable, so upstream reports the assignment while Rslint reports nothing. Restrict this fallback to declaration-name node kinds; ordinary shorthand assignment targets should continue through global/local reference resolution.

internal/plugins/unicorn/rules/no_document_cookie/no_document_cookie.go
```diff
@@ -0,0 +1,392 @@
...
+		return nil
+	}
+	return declaration.Symbol()
+}
+
+func (tracker *documentReferenceTracker) trackVariable(symbol *ast.Symbol, value tracedValue) {
+	if tracker.ctx.Refs == nil || symbol == nil {
+		return
+	}
+	key := tracedVariable{symbol: symbol, value: value}
```
elecmonkey: [P2] Key cycle detection only by variable identity

Upstream's `ReferenceTracker` cycle guard tracks only the variable currently being traversed, but this key includes both the symbol and its inferred value role. That permits the same variable to be re-entered as a different role: for `let a = globalThis; let b = a.document; a = b; a.cookie = "x"`, upstream reports nothing while Rslint incorrectly reports `a.cookie`. The active traversal guard needs to stop re-entry by symbol regardless of `tracedValue`.

internal/plugins/unicorn/rules/no_document_cookie/no_document_cookie.go
shiyudesu: Fixed in `453ce704`. Computed access and destructuring keys now use `EvalToString`, matching `getStringIfConstant`'s JavaScript string coercion. I also added regression cases for `document[["cookie"]]` and the computed destructuring alias.

internal/plugins/unicorn/rules/no_document_cookie/no_document_cookie.go
shiyudesu: Fixed in `453ce704`. The fallback is now restricted with `utils.IsDeclarationIdentifier`, so shorthand destructuring assignment targets continue through normal global/local reference resolution. The reported `({document} = globalThis)` case is covered by a regression test.

internal/plugins/unicorn/rules/no_document_cookie/no_document_cookie.go
shiyudesu: Fixed in `453ce704`. The active cycle guards are now keyed only by variable identity (symbol/name), independent of the traced value role. I added the reported alias cycle as a valid regression case.

elecmonkey (approved):
LGTM. The rule implementation and migrated upstream test suite look good overall.

One follow-up parity gap remains around statically computed property names: upstream reports cases such as `document["cookie".toString()] = "x"`, while Rslint currently does not. This comes from incomplete safe built-in call coverage in the shared `StaticStringEvaluator`, rather than this rule’s tracking logic. Let’s follow up in a separate PR by extracting/completing the shared JavaScript-semantic utility and adding `WithoutScope` evaluator coverage, so this and other rules can reuse it.

elecmonkey:
> LGTM. The rule implementation and migrated upstream test suite look good overall.
> 
> One follow-up parity gap remains around statically computed property names: upstream reports cases such as `document["cookie".toString()] = "x"`, while Rslint currently does not. This comes from incomplete safe built-in call coverage in the shared `StaticStringEvaluator`, rather than this rule’s tracking logic. Let’s follow up in a separate PR by extracting/completing the shared JavaScript-semantic utility and adding `WithoutScope` evaluator coverage, so this and other rules can reuse it.

I’ll implement this after this PR is merged.

PR #2049, opened 2026-09-04
feat: port rule unicorn/no-this-outside-of-class

shiyudesu:
## Summary

Port the `unicorn/no-this-outside-of-class` rule from ESLint to rslint.

Disallow `this` outside of classes. `this` should only be used when JavaScript class syntax or an explicit TypeScript `this` parameter defines the receiver.

## Related Links

- ESLint rule: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-this-outside-of-class.md
- Source code: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/rules/no-this-outside-of-class.js

## Checklist

### DIFF
--- a/internal/plugins/unicorn/rules/no_this_outside_of_class/no_this_outside_of_class.go
+++ b/internal/plugins/unicorn/rules/no_this_outside_of_class/no_this_outside_of_class.go
@@ -1,0 +1,140 @@
+package no_this_outside_of_class
+
+import (
+	"github.com/microsoft/typescript-go/shim/ast"
+	"github.com/web-infra-dev/rslint/internal/rule"
+	"github.com/web-infra-dev/rslint/internal/utils"
+)
+
+const messageID = "no-this-outside-of-class"
+
+var defaultMessage = rule.RuleMessage{
+	Id:          messageID,
+	Description: "Do not use `this` outside of classes.",
+}
+
+// NoThisOutsideOfClassRule disallows using `this` outside of classes.
+//
+// https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v72.0.0/rules/no-this-outside-of-class.js
+var NoThisOutsideOfClassRule = rule.Rule{
+	Name:   "unicorn/no-this-outside-of-class",
+	Schema: rule.EmptyArraySchema,
+	Run:    run,
+}
+
+func run(ctx rule.RuleContext, options []any) rule.RuleListeners {
+	checkThis := func(node *ast.Node) {
+		if utils.IsInJsxTagName(node) {
+			return
+		}
+		if node.Parent != nil && ast.IsThisParameter(node.Parent) && node.Parent.Name() == node {
+			return
+		}
+		if !isAllowedThisBinding(node) {
+			ctx.ReportNode(node, defaultMessage)
+		}
+	}
+
+	return rule.RuleListeners{
+		ast.KindThisKeyword: checkThis,
+		ast.KindThisType: func(node *ast.Node) {
+			if isTypeQueryRoot(node) {
+				checkThis(node)
+			}
+		},
+		ast.KindIdentifier: func(node *ast.Node) {
+			if node.AsIdentifier().Text == "this" && isTypeQueryRoot(node) {
+				checkThis(node)
+			}
+		},
+	}
+}
+
+func isAllowedThisBinding(node *ast.Node) bool {
+	child := node
+	parent := node.Parent
+
+	for parent != nil {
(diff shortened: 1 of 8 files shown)


internal/plugins/unicorn/rules/no_this_outside_of_class/no_this_outside_of_class.go
```diff
@@ -0,0 +1,140 @@
...
+	default:
+		return false
+	}
+}
+
+func isInsideMethodFunction(method *ast.Node, child *ast.Node) bool {
+	if child == nil {
+		return false
+	}
+	if child.Kind == ast.KindParameter {
```
elecmonkey: Upstream treats an ESTree class method's FunctionExpression as the `this` binding for the whole implemented method, so `typeof this` in method return types/generic constraints is allowed, while body-less signatures still report. This helper only allows direct parameters and the body, which flips both cases: Rslint reports `class C { method(): typeof this.foo { return null as any; } }` and suppresses `class C { method(value: typeof this.foo): void; method(value: unknown) {} }`. Please account for method type parameters/return types only when the method has a body, and keep overload/abstract signatures reportable.

shiyudesu: Good catch, thank you! Updated to require an implemented body (`method.Body() != nil`) so that:
1. Implemented class methods allow `typeof this` in return types and generic type parameter constraints (mirroring ESTree's `FunctionExpression` scope).
2. Body-less overload signatures and abstract methods do not bind `this` and remain reportable.

Added the corresponding test cases in `_extras_test.go`.

internal/plugins/unicorn/rules/no_this_outside_of_class/no_this_outside_of_class.go
```diff
@@ -0,0 +1,140 @@
...
+	parent := node.Parent
+
+	for parent != nil {
+		if parent.Kind == ast.KindArrowFunction {
+			child = parent
+			parent = parent.Parent
+			continue
+		}
+
+		if parent.Kind == ast.KindFunctionDeclaration || parent.Kind == ast.KindFunctionExpression {
```
swwind, line 64: Body-less TypeScript declarations are treated as valid solely because they have an explicit `this` parameter. For example, ESLint reports `this` in `declare function f(this: C): typeof this.foo;`, but this branch returns `hasThisParameter(parent)` without checking whether the function has a body. Please keep body-less overload/declaration signatures reportable.

internal/plugins/unicorn/rules/no_this_outside_of_class/no_this_outside_of_class.go
swwind: Could you distinguish an implemented method from a body-less signature here? `child.Kind == ast.KindParameter` currently returns true even when `method.Body() == nil`, so `typeof this` in `class C { method(value: typeof this.foo): void; method(value: unknown) {} }` is incorrectly ignored. Conversely, when the method has a body, `typeof this` in its return type or type-parameter constraint is reported even though ESLint allows it. Please make the binding cover the whole implemented method while keeping overload and abstract signatures reportable.

swwind (commented):
Found two correctness issues in the TypeScript binding handling.

internal/plugins/unicorn/rules/no_this_outside_of_class/no_this_outside_of_class.go
shiyudesu: Good catch! We added a `parent.Body() == nil` check in commit 9dc7aa84 so that body-less function declarations and overload signatures (which have no ESTree `FunctionExpression` representation) do not bind `this` and pass through to the enclosing scope (reporting at the top level).

Added dedicated tests for `declare function`, `declare class`, `interface`, and type literal signatures in `_extras_test.go`.

internal/plugins/unicorn/rules/no_this_outside_of_class/no_this_outside_of_class.go
shiyudesu: Thanks! This was aligned in commit 9dc7aa84 via `isInsideImplementedMethodFunction`, which requires `method.Body() != nil` and excludes the method name (`[this.key]`) and decorators. Overload and abstract signatures now remain reportable, while implemented method return types and type parameter constraints are allowed.

Locked in via test cases in `_extras_test.go`.

swwind (approved):
LGTM

elecmonkey (approved):
lgtm

PR #2056, opened 2026-09-04
feat: port rule unicorn/prefer-string-trim-start-end

shiyudesu:
## Summary

Port `unicorn/prefer-string-trim-start-end` from eslint-plugin-unicorn v74.0.0.

The rule reports zero-argument `trimLeft()` and `trimRight()` calls and safely autofixes them to `trimStart()` and `trimEnd()`.

This change also:

- Preserves upstream optional chaining behavior: optional member access is reported, while optional calls are allowed.
- Skips receivers known not to be strings.
- Supports TypeScript annotations, assertions, aliases, interfaces, classes, and constrained type parameters even when no TypeChecker is available.
- Extracts the shared string receiver classification used by `no-unsafe-string-replacement` while preserving that rule's existing behavior.
- Registers the rule in the Unicorn catalog and recommended preset.
- Adds the complete upstream test suite, source-only regressions, AST edge cases, autofix boundary coverage, and JS integration tests.

## Related Links

- Tracking issue: #1309
- [Rule documentation](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v74.0.0/docs/rules/prefer-string-trim-start-end.md)
- [Upstream source](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v74.0.0/rules/prefer-string-trim-start-end.js)

## Verification

- `go test -count=1 ./internal/plugins/unicorn/rules/prefer_string_trim_start_end ./internal/plugins/unicorn/rules/no_unsafe_string_replacement ./internal/plugins/unicorn/unicornutil`
- `pnpm --dir packages/rslint-test-tools exec rs test --testTimeout=10000 prefer-string-trim-start-end`
- `pnpm typecheck`
- `pnpm lint`
- `pnpm -w run check-spell`
- `pnpm format:check`
- `golangci-lint v2.12.2`: 0 issues

## Checklist

- [x] Tests updated.
- [x] Documentation updated.

### DIFF
--- a/internal/plugins/unicorn/unicornutil/string_receiver.go
+++ b/internal/plugins/unicorn/unicornutil/string_receiver.go
@@ -1,0 +1,344 @@
+package unicornutil
+
+import (
+	"github.com/microsoft/typescript-go/shim/ast"
+	"github.com/microsoft/typescript-go/shim/checker"
+	"github.com/web-infra-dev/rslint/internal/rule"
+	"github.com/web-infra-dev/rslint/internal/utils"
+)
+
+// IsKnownNonString reports whether node's type is definitely not string-like.
+// It follows local TypeScript annotations and const initializers before using
+// the checker, matching Unicorn's source-only behavior for files that are not
+// part of a configured project. Unknown receivers return false so syntactic
+// string rules can still report them.
+func IsKnownNonString(ctx rule.RuleContext, node *ast.Node) bool {
+	return classifyStringReceiver(ctx, node, &stringReceiverWalkState{
+		visiting: map[*ast.Symbol]bool{},
+		memo:     map[*ast.Symbol]TypeClass{},
+	}) == TypeNonTarget
+}
+
+// IsKnownNonStringType reports whether the checker alone proves that node is
+// not string-like. Rules that require type information and intentionally do
+// not follow Unicorn's syntax classifier use this narrower helper.
+func IsKnownNonStringType(ctx rule.RuleContext, node *ast.Node) bool {
+	if ctx.TypeChecker == nil || node == nil {
+		return false
+	}
+
+	return classifyStringType(ctx, node) == TypeNonTarget
+}
+
+type stringReceiverWalkState struct {
+	visiting map[*ast.Symbol]bool
+	memo     map[*ast.Symbol]TypeClass
+	depth    int
+	steps    int
+}
+
+const (
+	maxStringReceiverDepth = 1024
+	maxStringReceiverSteps = 4096
+)
+
+func classifyStringReceiver(ctx rule.RuleContext, node *ast.Node, state *stringReceiverWalkState) TypeClass {
+	if node == nil || state.depth >= maxStringReceiverDepth || state.steps >= maxStringReceiverSteps {
+		return TypeUnknown
+	}
+	node = ast.SkipParentheses(node)
+	if node == nil {
+		return TypeUnknown
+	}
+
+	state.depth++
+	state.steps++
+	defer func() { state.depth-- }()
+
+	// Unicorn ignores a satisfies annotation and classifies the expression.
+	if node.Kind == ast.KindSatisfiesExpression {
+		return classifyStringReceiver(ctx, node.Expression(), state)
+	}
+	// An informative assertion wins. `any` and `unknown` fall through to the
+	// wrapped expression, matching upstream's getTypeFromExpression.
+	if node.Kind == ast.KindAsExpression || node.Kind == ast.KindTypeAssertionExpression {
+		if class := classifyStringTypeNode(ctx, node.Type(), map[*ast.Symbol]bool{}); class != TypeUnknown {
+			return class
+		}
+		return classifyStringReceiver(ctx, node.Expression(), state)
+	}
+	if node.Kind == ast.KindNonNullExpression {
+		return classifyStringReceiver(ctx, node.Expression(), state)
+	}
+
+	if ast.IsIdentifier(node) {
+		if class := classifyStringIdentifier(ctx, node, state); class != TypeUnknown {
+			return class
+		}
+	}
+	if node.Kind == ast.KindConditionalExpression {
+		conditional := node.AsConditionalExpression()
+		class := combineStringUnion([]TypeClass{
+			classifyStringReceiver(ctx, conditional.WhenTrue, state),
+			classifyStringReceiver(ctx, conditional.WhenFalse, state),
+		})
+		if class != TypeUnknown {
+			return class
+		}
+	}
+	if ast.IsBinaryExpression(node) &&
+		node.AsBinaryExpression().OperatorToken.Kind == ast.KindCommaToken {
+		if class := classifyStringReceiver(ctx, node.AsBinaryExpression().Right, state); class != TypeUnknown {
+			return class
+		}
+	}
+
+	if ctx.TypeChecker != nil {
+		return classifyStringType(ctx, node)
+	}
+	return TypeUnknown
+}
+
+func classifyStringIdentifier(ctx rule.RuleContext, node *ast.Node, state *stringReceiverWalkState) TypeClass {
+	if ctx.Refs == nil {
+		return TypeUnknown
+	}
+	symbol := ctx.Refs.ResolveInFile(node)
+	if symbol == nil || len(symbol.Declarations) != 1 {
+		return TypeUnknown
+	}
+	if class, ok := state.memo[symbol]; ok {
+		return class
+	}
+	if state.visiting[symbol] {
+		return TypeUnknown
+	}
+	state.visiting[symbol] = true
+	defer delete(state.visiting, symbol)
+
+	declaration := symbol.Declarations[0]
+	if declaration == nil {
+		return TypeUnknown
+	}
+	if class := classifyStringTypeNode(ctx, stringBindingTypeAnnotation(declaration), map[*ast.Symbol]bool{}); class != TypeUnknown {
+		state.memo[symbol] = class
+		return class
+	}
+	if !ast.IsVariableDeclaration(declaration) {
+		return TypeUnknown
+	}
+
+	declarationList := declaration.Parent
+	variable := declaration.AsVariableDeclaration()
+	if declarationList == nil || !ast.IsVariableDeclarationList(declarationList) ||
+		ast.IsVarUsing(declarationList) || ast.IsVarAwaitUsing(declarationList) ||
+		declarationList.Flags&ast.NodeFlagsConst == 0 || variable.Name() == nil ||
+		!ast.IsIdentifier(variable.Name()) || variable.Initializer == nil {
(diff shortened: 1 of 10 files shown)


internal/plugins/unicorn/unicornutil/string_receiver.go
```diff
@@ -0,0 +1,344 @@
...
+		switch declaration.Kind {
+		case ast.KindTypeAliasDeclaration:
+			return classifyStringTypeNode(ctx, declaration.AsTypeAliasDeclaration().Type, visitedSymbols)
+		case ast.KindTypeParameter:
+			return classifyStringTypeNode(ctx, declaration.AsTypeParameterDeclaration().Constraint, visitedSymbols)
+		case ast.KindInterfaceDeclaration:
+			return classifyStringInterface(ctx, declaration, visitedSymbols)
+		case ast.KindClassDeclaration, ast.KindClassExpression:
+			return classifyStringClass(ctx, declaration, visitedSymbols)
+		}
```
elecmonkey, line 247: **[P2] Follow const aliases when classifying base classes**

In source-only TypeScript, this resolver ignores variable declarations, so `class Base {}; const Alias = Base; class Derived extends Alias {}; function f(value: Derived) { value.trimLeft(); }` is classified as unknown. Unicorn 74 follows the const initializer and emits no diagnostic, while Rslint reports and offers an autofix. Please follow unmodified const aliases during class-heritage resolution before treating the receiver as unknown.

internal/plugins/unicorn/unicornutil/string_receiver.go
```diff
@@ -0,0 +1,344 @@
...
+}
+
+func classifyStringType(ctx rule.RuleContext, node *ast.Node) TypeClass {
+	t := ctx.TypeChecker.GetTypeAtLocation(node)
+	return ClassifyType(ctx, t, TypeClassifierOptions{
+		HeritageSymbolFlags:          ast.SymbolFlagsClass | ast.SymbolFlagsInterface,
+		NonTargetSymbolLessTypeFlags: checker.TypeFlagsObject,
+		IsTargetType: func(t *checker.Type) bool {
+			return utils.IsTypeFlagSet(t, checker.TypeFlagsStringLike)
+		},
```
elecmonkey, line 504: **[P2] Preserve unknown status for symbol-less literal types**

When a TypeChecker is available, this classifier marks symbol-less primitive and object types as definitely non-string, but Unicorn leaves them unknown unless its syntax classifier proves otherwise. For example, with a typed project, Unicorn 74 reports and fixes both `(1).trimLeft()` and `const value = 1; value.trimLeft()`, whereas Rslint emits no diagnostic. Please preserve upstream-compatible unknown handling for these inferred or literal types instead of skipping them as non-targets.

internal/plugins/unicorn/unicornutil/string_receiver.go
shiyudesu: Fixed in d0c1ea56. Source-only class heritage resolution now follows local const aliases, including aliases initialized with class expressions. Added regression coverage for both `const Alias = Base` and `const Alias = class {}` inheritance shapes.

internal/plugins/unicorn/unicornutil/string_receiver.go
shiyudesu: Fixed in d0c1ea56. The checker path now preserves Unicorn's unknown classification for symbol-less literal and object types instead of using the broader required-type-info classifier. Added regression coverage for direct numeric literals, const numeric aliases, and readonly tuples.

internal/plugins/unicorn/unicornutil/string_receiver.go
```diff
@@ -0,0 +1,465 @@
...
+	ctx rule.RuleContext,
+	node *ast.Node,
+	visitedSymbols map[*ast.Symbol]bool,
+) TypeClass {
+	node = ast.SkipParentheses(node)
+	if node == nil {
+		return TypeUnknown
+	}
+	if node.Kind == ast.KindClassDeclaration || node.Kind == ast.KindClassExpression {
+		return classifyStringClass(ctx, node, visitedSymbols)
```
elecmonkey, line 322: **[P2] Keep direct class-expression heritage unknown**

In source-only TypeScript, `class Derived extends class {} {}; function f(value: Derived) { value.trimLeft(); }` is treated as a known non-string receiver and skipped. Unicorn 74 only resolves an `extends` target when the superclass is an identifier, so it reports and fixes this call. Please leave direct class-expression heritage unknown unless it is reached through the const-alias path.

internal/plugins/unicorn/unicornutil/string_receiver.go
```diff
@@ -0,0 +1,465 @@
...
+		}
+		switch declaration.Kind {
+		case ast.KindClassDeclaration, ast.KindClassExpression:
+			return classifyStringClass(ctx, declaration, visitedSymbols)
+		case ast.KindVariableDeclaration:
+			declarationList := declaration.Parent
+			variable := declaration.AsVariableDeclaration()
+			if declarationList == nil || !ast.IsVariableDeclarationList(declarationList) ||
+				declarationList.Flags&ast.NodeFlagsConst == 0 || variable.Name() == nil ||
+				!ast.IsIdentifier(variable.Name()) || variable.Initializer == nil {
```
elecmonkey, line 350: **[P2] Exclude await-using bindings from const alias resolution**

This const check also accepts an `await using` declaration in ts-go. Consequently, `await using Alias = class {}; class Derived extends Alias {}; function f(value: Derived) { value.trimLeft(); }` is skipped by Rslint, while Unicorn 74 reports and fixes it because the binding kind is not `const`. Please reject `ast.IsVarUsing` and `ast.IsVarAwaitUsing` here as the existing variable-classification path does.

shiyudesu: Fixed in db5d0b27. Class-heritage alias resolution now explicitly rejects both `using` and `await using` declaration lists before checking the const flag. Added source-only regression cases for both binding forms.

internal/plugins/unicorn/unicornutil/string_receiver.go
shiyudesu: Fixed in db5d0b27. Direct class-expression heritage now stays unknown because the superclass must be an identifier before class-reference resolution begins. Const aliases initialized with class expressions still follow the separate alias path. Added source-only regression coverage for both shapes.

internal/plugins/unicorn/unicornutil/string_receiver.go
```diff
@@ -0,0 +1,473 @@
...
+		return classifyStringCheckerType(ctx, constraint)
+	}
+	if stringTypeHasTargetBase(ctx, t) {
+		return TypeTarget
+	}
+	if utils.IsIntrinsicType(t) {
+		return TypeNonTarget
+	}
+	if _, ok := TypeSymbolName(t); !ok {
+		return TypeUnknown
```
elecmonkey, line 473: **[P2] Classify checker boolean literals as non-string**

In a typed project, inferred or computed boolean-literal types fall through this intrinsic check because ts-go excludes `TypeFlagsBooleanLiteral` from `TypeFlagsIntrinsic`. Rslint therefore reports and autofixes cases such as `const value = true; value.trimLeft()`, `({flag: true} as const).flag.trimLeft()`, and indexed or conditional types resolving to `true`/`false`, while Unicorn 74 treats their `intrinsicName` as a known non-string and emits no diagnostic. Please classify boolean-literal types as non-targets before the symbol-less fallback.

internal/plugins/unicorn/unicornutil/string_receiver.go
```diff
@@ -0,0 +1,473 @@
...
+	}
+	if ast.IsBinaryExpression(node) &&
+		node.AsBinaryExpression().OperatorToken.Kind == ast.KindCommaToken {
+		if class := classifyStringReceiver(ctx, node.AsBinaryExpression().Right, state); class != TypeUnknown {
+			return class
+		}
+	}
+
+	if ctx.TypeChecker != nil {
+		return classifyStringType(ctx, node)
```
elecmonkey, line 100: **[P2] Apply syntax targets before checker classification**

When a TypeChecker is present, this fallback can classify a receiver before applying Unicorn's syntax-level string targets. For example, `{ function String(): number { return 1 } String().trimLeft() }` and a shadowed `String.fromCharCode()` returning a number or object are still reported and fixed by Unicorn 74 because those call shapes are unconditional syntax targets, but Rslint skips them as known non-strings. Please recognize `String(...)`, `String.fromCharCode(...)`, and `String.fromCodePoint(...)` before consulting the checker, matching upstream's `getType` ordering.

internal/plugins/unicorn/unicornutil/string_receiver.go
```diff
@@ -0,0 +1,473 @@
...
+	}
+}
+
+func classifyStringTypeNode(
+	ctx rule.RuleContext,
+	node *ast.Node,
+	visitedSymbols map[*ast.Symbol]bool,
+) TypeClass {
+	if node == nil {
+		return TypeUnknown
```
elecmonkey, line 165: **[P2] Ignore synthesized JSDoc type annotations**

This classifier consumes ts-go's synthesized JSDoc type nodes even though typescript-eslint does not expose them as ESTree `typeAnnotation` nodes. Consequently, source-only JavaScript such as `/** @type {number} */ let value; value.trimLeft()` and JSDoc casts such as `(/** @type {number} */ (value)).trimLeft()` are skipped by Rslint but reported and fixed by Unicorn 74. Please return unknown for `utils.IsJSDocSyntaxNode(node)` so both declaration and cast paths preserve the upstream boundary.

shiyudesu: Fixed in 4d852dcf. Source type classification now ignores nodes reported by `utils.IsJSDocSyntaxNode`, preserving the ESTree boundary for synthesized JSDoc annotations. Added source-only JavaScript coverage for both a JSDoc declaration annotation and a JSDoc cast.

internal/plugins/unicorn/unicornutil/string_receiver.go
shiyudesu: Fixed in 4d852dcf. `String(...)`, `String.fromCharCode(...)`, and `String.fromCodePoint(...)` are now recognized as syntax targets before consulting the checker, including shadowed bindings returning numbers or objects. Added regression coverage for all three call shapes.

internal/plugins/unicorn/unicornutil/string_receiver.go
shiyudesu: Fixed in 4d852dcf. The checker path now classifies `TypeFlagsBooleanLiteral` as a non-target before the symbol-less fallback. Added coverage for a direct const boolean, an `as const` property, and an indexed-access type resolving to a boolean literal.

elecmonkey (approved):
LGTM

PR #2063, opened 2026-09-05
feat: port rule unicorn/no-await-expression-member

shiyudesu:
## Summary

Port `unicorn/no-await-expression-member` from eslint-plugin-unicorn v74.0.0.

- Report property and element access directly from an `await` expression, including optional member access.
- Autofix simple variable declarations using array or object destructuring.
- Register the rule and enable it in the Unicorn recommended preset.
- Add documentation, upstream test coverage, edge-case regressions, and JS integration tests.

## Safety difference from upstream

Unlike upstream v74.0.0, `using` and `await using` declarations are reported without autofixes. Resource declarations cannot use destructuring, so applying the upstream fix would produce invalid code.

This difference is documented and covered by regression tests.

## Validation

- Go upstream, edge-case, edit-demand, and rule catalog tests.
- JS rule integration and preset tests.
- Go lint, TypeScript type checking, formatting, and spell checking.
- Differential checks against v74.0.0: 673 inputs, including 76 real source files, plus 838 additional edge-shape inputs. Only the documented resource-declaration autofix differences remain.
- Verified that `fix: true` leaves resource declarations unchanged while still fixing ordinary declarations.

## Related Links

- Tracking issue: #1309
- [Upstream documentation](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v74.0.0/docs/rules/no-await-expression-member.md)
- [Upstream implementation](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v74.0.0/rules/no-await-expression-member.js)

## Checklist

- [x] Tests updated.
- [x] Documentation updated.

### DIFF
--- a/internal/plugins/unicorn/rules/no_await_expression_member/no_await_expression_member.go
+++ b/internal/plugins/unicorn/rules/no_await_expression_member/no_await_expression_member.go
@@ -1,0 +1,112 @@
+package no_await_expression_member
+
+import (
+	"github.com/microsoft/typescript-go/shim/ast"
+	"github.com/microsoft/typescript-go/shim/core"
+	"github.com/web-infra-dev/rslint/internal/rule"
+	"github.com/web-infra-dev/rslint/internal/utils"
+)
+
+var message = rule.RuleMessage{
+	Id:          "noAwaitExpressionMember",
+	Description: "Do not access a member directly from an await expression.",
+}
+
+// NoAwaitExpressionMemberRule disallows member access directly from await.
+//
+// https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v74.0.0/rules/no-await-expression-member.js
+var NoAwaitExpressionMemberRule = rule.Rule{
+	Name:   "unicorn/no-await-expression-member",
+	Schema: rule.EmptyArraySchema,
+	Run: func(ctx rule.RuleContext, options []any) rule.RuleListeners {
+		check := func(node *ast.Node) {
+			object := node.Expression()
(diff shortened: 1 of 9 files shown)


fansenze (approved):

internal/plugins/unicorn/rules/no_await_expression_member/no_await_expression_member.go
```diff
@@ -0,0 +1,112 @@
...
+
+import (
+	"github.com/microsoft/typescript-go/shim/ast"
+	"github.com/microsoft/typescript-go/shim/core"
+	"github.com/web-infra-dev/rslint/internal/rule"
+	"github.com/web-infra-dev/rslint/internal/utils"
+)
+
+var message = rule.RuleMessage{
+	Id:          "noAwaitExpressionMember",
```
elecmonkey, line 11: **[P2] Preserve the upstream message ID**

For `const value = (await promise).value`, `eslint-plugin-unicorn@74.0.0` reports `messageId: "no-await-expression-member"`, while Rslint currently exposes `"noAwaitExpressionMember"` through its programmatic API. Other ported Unicorn rules in Rslint preserve their upstream message IDs verbatim—for example, `no-array-from-fill`, `no-await-in-promise-methods/error`, `missing-message`, and `member-access`. Since message IDs are externally visible, this mismatch breaks compatibility for consumers that identify diagnostics by `messageId`. Please use `"no-await-expression-member"` here and update the corresponding Go and JS expectations.

PR #2075, opened 2026-09-05
feat(jest,rstest): add padding rules

elecmonkey:
## Summary

Add seven focused padding rules and `padding-around-all` for both Jest and Rstest.
Share the padding engine while keeping framework-specific API tables, tests, and documentation.

## Intentional limitations

- Namespace-qualified and renamed API calls are not matched because classification follows first-token syntax.
- Rstest excludes Jest-only aliases and `expectTypeOf`.

## Related Links

Related #476.
Related #935.

## Checklist


internal/plugins/jest/utils/padding.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+func MakePaddingRule(name string, kind PaddingRuleKind) rule.Rule {
+	configs := paddingConfigs(kind)
+	priority := 0
+	if kind == PaddingAll {
+		priority = 100
+	}
+	return padding.NewRule(padding.Definition{
+		Name: name, Family: "jest", Priority: priority,
```
fansenze, line 49: **[P2] Adding a warning rule can make failing lint pass**

Enable these two rules:

```js
rules: {
  'jest/padding-around-before-all-blocks': 'warn',
  'jest/padding-around-before-each-blocks': 'error',
}
```

Then lint two hooks with no blank line between them:

```js
beforeAll(connect);
beforeEach(reset);
```

- **Expected:** The missing blank line before `beforeEach` produces an error. Exit code: **1**.
- **Actual:** Only the `beforeAll` warning is reported. Exit code: **0**.

Turning the first rule **off** restores the error. In other words, adding a warning rule makes code that previously failed lint pass.

Both rules detect the same missing blank line, but deduplication keeps the warning and drops the error. Rstest has the same problem. Please retain the error when these reports overlap and add a regression test for this configuration.

elecmonkey: Confirmed, fixed in dc00cd43e. The engine lets only the first matching rule of a family report a boundary, and the ordering key `family → priority → name` did not include severity, so the warning rule won the boundary by name order and suppressed the error. The key is now `family → severity → priority → name`, so the strictest configured rule owns a shared boundary.

That required a small addition to the rule package: `RuleContext.Severity()`. Severity is not a property of a rule definition — the linter attaches it per rule per file at `WithDiagnosticConsumer`, where it lands in the unexported `ruleContextReporter` and is unreadable from inside a rule. Threading it through `padding.Definition` does not work, because a definition is built once at registration while severity varies per configuration; deduplicating downstream in the linter does not work either, because the deferred fix is owned by the engine and the linter has no notion of two rules describing the same blank line. The arbitration has to stay in the engine, and it needs severity there. The accessor only exposes state the context already carries, alongside the existing `Program()`, and changes no behaviour for any other rule.

Regression tests added.

fansenze (approved):

PR #2081, opened 2026-09-07
ci: report binary size change on pull requests

### LINKED ISSUE #2059: [Proposal]: Establish binary size monitoring for Rslint
Rslint’s binary size is growing rapidly, while contributors and reviewers currently have no visibility into how much each pull request affects it.

Rspack monitors binary size by comparing pull requests against a baseline, reporting the difference, and retaining historical data from the main branch. Rslint could adopt a similar system, initially covering the stripped native executable.

Each pull request would show its binary size change against the base branch, and measurements from the main branch would be retained to show the long-term trend. This would give contributors and reviewers direct feedback on the binary size impact of their changes.

After @swwind reduced the install size by 10.95 MiB in v0.6.5 through #1168, it grew by 7.84 MiB (+23.26%) through v0.9.1.

For reference, Oxlint v1.81.0 is 17.41 MiB and oxlint-tsgolint v7.0.2001 is 21.26 MiB, totaling 38.67 MiB; Biome v2.5.12 is 61.43 MiB; Rslint v0.9.1 is 41.56 MiB (full installation size, Linux x64 glibc). Rslint’s rule coverage is still expanding, with many more rules expected to be ported in the foreseeable future, so we need to control binary size growth before it becomes increasingly difficult to manage.

I’d appreciate everyone’s feedback, and I’d be happy to work on establishing this system.

elecmonkey:
## Summary

Nothing in CI tells a reviewer how much a pull request grows the binary. This reports it, with no added build and no repository configuration.

### How it works

- **Measure** — `Test npm packages` already compiles everything behind `./cmd/rslint`, so a release relink right after `Build` is link-only off the warm cache: 49s against 5m12s for a full build. It relinks rather than measuring the binary that job already produced, which is 52.20 MiB unstripped where the released one is 37.10 MiB.
- **Store** — every run, `main` included, uploads its measurement as a workflow artifact. No repository writes, no token beyond the built-in `GITHUB_TOKEN`.
- **Compare** — a pull request's baseline is the artifact of the `main` run for its base commit. The new `Binary Size` job reads both, writes the comparison to the job summary, and updates one pull request comment in place.

The job is deliberately absent from `CI Done` — porting a rule legitimately grows the binary, so this informs a review rather than blocking a merge.

### Limits

- A fork's token is read-only, so the comment step fails there by design and the job summary is the report.
- Artifacts expire, so this covers the per-pull-request comparison; the long-term `main` trend still needs somewhere durable to write.

## Related Links

Closes #2059

## Checklist

> **Token scope in the `Binary Size` job**
>
> The trigger is `pull_request`, never `pull_request_target`: a fork's token is capped read-only regardless of the `permissions:` block, and a same-repo author already has push access. The `pull-requests: write` token crosses no privilege boundary.

### DIFF
--- a/scripts/binary-size.mjs
+++ b/scripts/binary-size.mjs
@@ -1,0 +1,320 @@
+#!/usr/bin/env node
+/**
+ * Binary size monitoring for the `rslint` executable.
+ *
+ *   node scripts/binary-size.mjs record <binary> <out.json>
+ *   node scripts/binary-size.mjs base-run                    # -> $GITHUB_OUTPUT
+ *   node scripts/binary-size.mjs report <head.json> [base.json]
+ *   node scripts/binary-size.mjs comment <report.md>
+ *
+ * The measurement itself is produced inside the existing `Test npm packages`
+ * job, which has already compiled every package behind ./cmd/rslint — so the
+ * release relink it measures is a link-only action off a warm Go build cache
+ * rather than a second build. Each run stores its number as a workflow
+ * artifact; a pull request reads back the artifact of the main run for its
+ * base commit. That keeps the whole system on the built-in GITHUB_TOKEN: no
+ * data branch, no personal access token, no repository writes.
+ */
+import { execFileSync } from 'node:child_process';
+import fs from 'node:fs';
+import path from 'node:path';
+
+const SCHEMA_VERSION = 1;
+
+// Ties an update to the comment this workflow posted last time, instead of
+// opening a new thread on every push.
+const COMMENT_MARKER = '<!-- rslint-binary-size -->';
+
+const apiBase = process.env.GITHUB_API_URL || 'https://api.github.com';
+const repository = process.env.GITHUB_REPOSITORY || '';
+
+async function api(endpoint) {
+  const response = await fetch(`${apiBase}${endpoint}`, {
+    headers: githubHeaders(),
+  });
+  if (!response.ok) {
+    throw new Error(
+      `GET ${endpoint} failed: ${response.status} ${response.statusText}`,
+    );
+  }
+  return response.json();
+}
+
+function githubHeaders() {
+  const token = process.env.GITHUB_TOKEN;
+  if (!token) throw new Error('GITHUB_TOKEN is not set');
+  return {
+    accept: 'application/vnd.github+json',
+    authorization: `Bearer ${token}`,
+    'content-type': 'application/json',
+    'x-github-api-version': '2022-11-28',
+  };
+}
+
+function readEvent() {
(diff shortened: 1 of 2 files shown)


elecmonkey:
@codex review

scripts/binary-size.mjs
```diff
@@ -0,0 +1,320 @@
...
+ * channel. That is why the caller runs this step with `continue-on-error`.
+ */
+async function comment(reportPath) {
+  const body = fs.readFileSync(reportPath, 'utf8');
+  const issueNumber = readEvent().pull_request?.number;
+  if (!issueNumber) throw new Error('no pull request number in the event');
+
+  const existing = await api(
+    `/repos/${repository}/issues/${issueNumber}/comments?per_page=100`,
+  );
```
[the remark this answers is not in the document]
elecmonkey: fixed

scripts/binary-size.mjs
```diff
@@ -0,0 +1,368 @@
...
+/**
+ * Locate the workflow run that measured the pull request's base commit.
+ *
+ * The base commit is on main, so its own CI run holds its measurement.
+ * A miss is not an error: the base may predate this workflow, its run may
+ * still be going, or its artifact may have expired. The report then just
+ * states the current size.
+ */
+async function findBaseRun() {
+  const baseSha = readEvent().pull_request?.base?.sha;
```
fansenze: Please derive the baseline from the **first parent of the merge commit being measured**, and record that SHA in the measurement artifact.

CI builds the PR merged into main, so we need to compare `size(main B + PR) - size(main B)`. If the event's `base.sha` points to older main A, this also counts the changes from A to B. For example, if A = 38 MiB, B = 40 MiB, and B + PR = 41 MiB, the PR adds **1 MiB**, but the current lookup could report **3 MiB**.

The mismatch already appears in [this run](https://github.com/web-infra-dev/rslint/actions/runs/34079799840): the measured merge commit `f5f60cc` has first parent `a8bed38`, while `base-run` looks up `f13d1e9`. That run had no baseline artifact, so it did not emit an incorrect delta, but using the measured merge's parent would keep future comparisons consistent.

elecmonkey: Sorry, I misread `base.sha`.

`base-run` now takes `parents[0]` of the measured merge commit as the baseline, falling back to `base.sha` only for non-merge builds.

Didn't put it in the artifact: it's resolved in the `Binary Size` job, which already has the measurement, and doing it at measure time would need `fetch-depth: 2` on `Test npm packages` just for `git rev-parse HEAD^1`.

fansenze (approved):

PR #2091, opened 2026-09-07
feat(rstest): add prefer-equality-matcher rule

elecmonkey:
## Summary

Add `rstest/prefer-equality-matcher` to replace boolean assertions over strict comparisons with direct equality assertions. For example, `expect(actual === expected).toBe(true)` can be changed to `expect(actual).toBe(expected)`. The rule handles `===`, `!==`, boolean expectations, existing `not`, and promise modifiers, and offers three suggestions using `toBe`, `toEqual`, and `toStrictEqual`. It is registered in the Rstest plugin but stays out of the `recommended` preset, matching upstream.

Assertions resolve through the shared Rstest call analysis, covering globals, imports and aliases, namespace and `require` forms, `import.meta.rstest`, Playwright, TestContext `expect`, and assertion factories. Only the first called equality matcher is considered; loose comparisons, Chai assertions, dynamic matcher names, foreign expect functions, and local shadows are ignored. The suggestions retain the assertion root, a second expect message, promise modifiers, accessor spelling, surrounding trivia, and trailing commas, and they are built only when requested.

The strict-comparison predicate, boolean truth table, and suggestion planning now live under `internal/utils/test_framework/rules/prefer_equality_matcher` and are shared with the existing Jest rule. A framework-neutral accessor insertion helper is also extracted for placing `not` safely across dot, element, and optional access.

## Credits

Port from [`@vitest/eslint-plugin`](https://github.com/vitest-dev/eslint-plugin-vitest)'s `prefer-equality-matcher`.

- **Rstest suggestions preserve authored accessors and comments.** String and template accessors retain their delimiters, existing `not` members are removed in place, and comments outside the strict-comparison span survive. If rewriting the comparison would delete a comment or an accessor cannot be edited safely, the diagnostic is retained without suggestions.

- **TypeScript-wrapped boolean literals are recognized.** Parentheses and basic type assertions around `true` or `false` do not change the value being asserted, so these forms participate in the same equality truth table.

## Related Links

Related #935

## Checklist

### DIFF
--- a/internal/plugins/rstest/rules/prefer_equality_matcher/prefer_equality_matcher.go
+++ b/internal/plugins/rstest/rules/prefer_equality_matcher/prefer_equality_matcher.go
@@ -1,0 +1,114 @@
+package prefer_equality_matcher
+
+import (
+	"github.com/microsoft/TypeScript/tsc/shim/ast"
+	"github.com/microsoft/TypeScript/tsc/shim/scanner"
+	rstestUtils "github.com/web-infra-dev/rslint/internal/plugins/rstest/utils"
+	"github.com/web-infra-dev/rslint/internal/rule"
+	"github.com/web-infra-dev/rslint/internal/utils"
+	testFramework "github.com/web-infra-dev/rslint/internal/utils/test_framework"
+	shared "github.com/web-infra-dev/rslint/internal/utils/test_framework/rules/prefer_equality_matcher"
+)
+
+var PreferEqualityMatcherRule = shared.NewRule(shared.Config{
+	Name: "rstest/prefer-equality-matcher",
+	Prepare: func(ctx rule.RuleContext) shared.Runtime {
+		analysis := rstestUtils.GetRstestCallAnalysis(ctx)
+		return shared.Runtime{
+			Parse: func(node *ast.Node) *shared.ExpectCall {
+				parsed := analysis.ParseExpectCall(node)
+				if parsed == nil ||
+					parsed.Reason != rstestUtils.RstestExpectParseReasonNone ||
+					parsed.Head == nil ||
+					parsed.MatcherEntry == nil ||
+					len(parsed.Matchers) == 0 ||
+					parsed.Matchers[0].Kind != rstestUtils.RstestExpectMatcherCall ||
+					testFramework.IsComputedIdentifierAccessor(parsed.MatcherEntry.Node) {
+					return nil
+				}
+
+				matcherCall := rstestUtils.MatcherCall(parsed.MatcherEntry)
+				if matcherCall == nil {
+					return nil
+				}
+
+				return &shared.ExpectCall{
+					HeadCall:     parsed.Head,
+					MatcherCall:  matcherCall,
+					MatcherEntry: *parsed.MatcherEntry,
+					Matcher:      parsed.Matcher,
+					Modifiers:    parsed.ModifierEntries,
+				}
+			},
+		}
+	},
+	BuildFixes: buildSuggestionFixes,
+})
+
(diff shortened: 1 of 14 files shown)


internal/plugins/rstest/rules/prefer_equality_matcher/prefer_equality_matcher.go
```diff
@@ -0,0 +1,114 @@
...
+	)
+	if !ok {
+		return nil
+	}
+
+	fixes := []rule.RuleFix{
+		rule.RuleFixReplace(
+			ctx.SourceFile,
+			match.Comparison,
+			scanner.GetSourceTextOfNodeFromSourceFile(ctx.SourceFile, ast.SkipParentheses(match.Left), false),
```
fansenze: **[P2] Preserve parentheses around comma expressions**

This turns `expect((f(), a) === b).toBe(true)` into `expect(f(), a).toBe(b)`, so the assertion compares `f()`'s return value with `b` instead of comparing `a` with `b`. Please preserve the parentheses when moving a comma expression into either argument list.

internal/plugins/rstest/rules/prefer_equality_matcher/prefer_equality_matcher.go
```diff
@@ -0,0 +1,114 @@
...
+		),
+		rule.RuleFixReplaceRange(matcherRange, matcherText),
+		rule.RuleFixReplace(
+			ctx.SourceFile,
+			match.MatcherArgument,
+			scanner.GetSourceTextOfNodeFromSourceFile(ctx.SourceFile, ast.SkipParentheses(match.Right), false),
```
fansenze: **[P2] Include the type assertion in the replacement range**

Only replacing the boolean literal turns `expect(a === b).toBe(true as const)` into `expect(a).toBe(b as const)`, which fails with TS1355 when `b` is a variable. Please replace the full type-asserted argument, or omit the suggestion if it cannot be rewritten safely.

internal/plugins/rstest/rules/prefer_equality_matcher/prefer_equality_matcher.go
elecmonkey: Agreed, fixed in 8a8fb304d.

`SkipParentheses` dropped parentheses that are load-bearing around a comma expression, so `expect((f(), a) === b).toBe(true)` became `expect(f(), a).toBe(b)`. Operands are now re-parenthesized when they are comma expressions, on both sides: the suggestions are `expect((f(), a)).toBe(b)` and `expect(a).toBe((f(), b))`.

The edit lives in the shared package, so `jest/prefer-equality-matcher` is fixed as well, and both rules gained test cases for either side.

internal/plugins/rstest/rules/prefer_equality_matcher/prefer_equality_matcher.go
elecmonkey: Agreed, fixed in the same commit.

Recognizing `true as const` but replacing only the literal left an assertion attached to an operand it was never written for. The replaced span now extends to the outermost type assertion around the literal, so `expect(a === b).toBe(true as const)` suggests `expect(a).toBe(b)` — the assertion is dropped with the literal it belonged to. Plain parentheses are still preserved, so `toBe((true))` still suggests `toBe((b))`.

Since the replaced span is wider, the Rstest comment guard now covers it too: `toBe(true /* keep */ as const)` keeps the diagnostic and drops the suggestions rather than deleting the comment. `jest/prefer-equality-matcher` inherits the fix; its `toBe(true as boolean)` case now suggests `toBe(b)`.

fansenze (approved):

PR #2092, opened 2026-09-07
feat(rstest): add require-to-throw-message rule

elecmonkey:
## Summary

Add `rstest/require-to-throw-message` to report `toThrow()` and `toThrowError()` assertions that do not specify the expected error. An assertion with any argument is accepted, as is a negated assertion, while `rejects.toThrow()` is still checked. The rule reports only and is registered in the Rstest plugin without being added to the `recommended` preset, matching upstream.

The rule recognizes global, imported, required, namespace, `import.meta.rstest`, Playwright, and TestContext forms of `expect`. It runs with or without type information and distinguishes local or foreign assertion functions from Rstest bindings in source-only programs. Assertion factories and promise modifiers are supported, while Chai's `throw` and `throws` matchers are outside the rule.

The matcher-independent predicate and diagnostic now live under `internal/utils/test_framework/rules/require_to_throw_message`, with thin Jest and Rstest parser adapters. The Jest rule keeps its existing diagnostics, and the Rstest suite mirrors the complete Vitest fixture set alongside source-only, provenance, parser-boundary, range, and real-user coverage.

## Credits

Port from [`@vitest/eslint-plugin`](https://github.com/vitest-dev/eslint-plugin-vitest)'s `require-to-throw-message`.

## Related Links

Related #935

## Checklist


internal/plugins/rstest/rules/require_to_throw_message/require_to_throw_message.go
```diff
@@ -0,0 +1,96 @@
...
+					parsed.Reason != rstestUtils.RstestExpectParseReasonNone ||
+					parsed.Head == nil ||
+					parsed.MatcherEntry == nil ||
+					len(parsed.Matchers) == 0 ||
+					parsed.Matchers[0].Kind != rstestUtils.RstestExpectMatcherCall {
+					return nil
+				}
+				if isSourceOnlyLocalExpect(node, parsed, ctx) {
+					return nil
+				}
```
fansenze: **[P2] Do not skip `expect` from Rstest test callbacks**

This filter also skips Rstest’s own `expect` when linting without type information:

```ts
test('throws', ({ expect }) => {
  expect(() => {
    throw new Error('boom');
  }).toThrow();
});
```

This should report the missing expected error, but currently produces no diagnostic. Without a type checker, `parsed.FromTestContext` stays `false`, so the filter treats the callback’s `expect` as an unrelated local function.

Please recognize Rstest test-context bindings before filtering out local `expect` functions, and add a regression test without type information.

elecmonkey: Agreed, fixed in 564686938.

The diagnosis is exactly right: `ContextExpectNames` is only populated when a `TypeChecker` is available, so without one a destructured TestContext `expect` reached the local-value branch and the assertion went unreported, even though the same code is reported with type information.

The TestContext bindings a test callback receives are already collected from syntax alone; only their symbol identities needed the checker. `collectRstestTestCallbacks` now also records the declaration node of each destructured `expect`, and the local-value check consults that index before rejecting. The lookup runs only on the rejection path, so nothing extra is collected for the common case.

Added source-only regression tests for `test('throws', ({ expect }) => …)`, which is now reported, and for a `const expect = …` shadow inside a test callback, which still is not.

One related limitation is unchanged and not specific to this rule: without type information the shared parser only accepts an expect root spelled `expect`, so the receiver form `test('x', (ctx) => ctx.expect(fn).toThrow())` is still not recognized in a source-only program.

fansenze (approved):

PR #2100, opened 2026-09-07
feat(rstest): add no-async-mock-factory rule

elecmonkey:
## Summary

Add `rstest/no-async-mock-factory`, which reports a mock factory for `rs.mock`, `rs.doMock`, `rs.mockRequire` or `rs.doMockRequire` that can only return a promise — Rstest calls the factory while the module graph is being built and throws `[Rstest] An async mock factory is not supported.` The type system does not catch it and the throw surfaces only once the mocked module is loaded, so the rule enters `recommended` at `error`.

## Related Links

Related #935

## Checklist

### DIFF
--- a/internal/plugins/rstest/rules/no_async_mock_factory/no_async_mock_factory.go
+++ b/internal/plugins/rstest/rules/no_async_mock_factory/no_async_mock_factory.go
@@ -5,6 +5,7 @@ import (
 	"github.com/microsoft/TypeScript/tsc/shim/checker"
 	"github.com/microsoft/TypeScript/tsc/shim/core"
 	rstestUtils "github.com/web-infra-dev/rslint/internal/plugins/rstest/utils"
+	lintprogram "github.com/web-infra-dev/rslint/internal/program"
 	"github.com/web-infra-dev/rslint/internal/rule"
 	"github.com/web-infra-dev/rslint/internal/utils"
 	"github.com/web-infra-dev/rslint/internal/utils/ecmascript"
@@ -199,6 +200,16 @@ func classifyFunction(ctx rule.RuleContext, fn *ast.Node) verdict {
 		return verdictPromise
 	}
 
+	// A path that runs off the end of a block body hands back `undefined`,
+	// which the runtime's `instanceof Promise` check lets through: such a
+	// factory only sometimes returns a promise, and this rule reports the ones
+	// that certainly do. `async` is settled above, where the implicit
+	// `undefined` is wrapped like every other value the body hands back.
+	if body := fn.Body(); body != nil && body.Kind == ast.KindBlock &&
+		utils.IsFunctionEndReachable(fn) {
+		return verdictUnknown
+	}
+
 	returned, ok := returnedExpressions(fn)
 	if !ok || len(returned) == 0 {
 		return verdictUnknown
@@ -212,9 +223,10 @@ func classifyFunction(ctx rule.RuleContext, fn *ast.Node) verdict {
 }
 
 // classifyThroughDeclaration follows an identifier to a declaration in this
-// file. A name that resolves to more than one declaration, to another file, or
-// to anything but a `const` function or a function declaration is left to the
-// type layer, where a reassignment or a re-export is accounted for.
+// file. A name that resolves to more than one declaration, to another file, to
+// anything but a `const` function or a function declaration, or to a function
+// declaration that is written to, is left to the type layer, where a
+// reassignment or a re-export is accounted for.
 func classifyThroughDeclaration(ctx rule.RuleContext, factory *ast.Node) verdict {
 	if factory.Kind != ast.KindIdentifier || ctx.Refs == nil {
 		return verdictUnknown
@@ -230,6 +242,12 @@ func classifyThroughDeclaration(ctx rule.RuleContext, factory *ast.Node) verdict
 
 	switch declaration.Kind {
 	case ast.KindFunctionDeclaration:
+		// A function declaration's name is a writable binding, and an
+		// assignment to it adds no declaration of its own: the body below is
+		// the installed factory only while nothing writes to the name.
+		if isWrittenTo(ctx, symbol) {
+			return verdictUnknown
+		}
 		return classifyFunction(ctx, declaration)
 	case ast.KindVariableDeclaration:
 		if !ast.IsVarConst(declaration) {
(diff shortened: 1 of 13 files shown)


internal/plugins/rstest/rules/no_async_mock_factory/no_async_mock_factory.go
```diff
@@ -0,0 +1,631 @@
...
+	}
+	returned, complete := returnedExpressions(factory)
+	if !complete {
+		return nil
+	}
+
+	if ast.GetFunctionFlags(factory)&ast.FunctionFlagsAsync != 0 {
+		return removeAsyncSuggestion(ctx, factory, returned)
+	}
+	return unwrapPromiseResolveSuggestion(ctx, factory, returned)
```
fansenze, line 516: **[P2] Keep explicit return types valid after applying suggestions**

For `async (): Promise<{ sum: number }> => ({ sum: 0 })`, removing `async` leaves the Promise return annotation in place, so the suggested code fails with TS2353. Unwrapping `Promise.resolve()` has the same problem when the factory has an explicit Promise return type. Please suppress these suggestions when the annotation would become invalid, or update the annotation as part of the suggestion.

internal/plugins/rstest/rules/no_async_mock_factory/no_async_mock_factory.go
```diff
@@ -0,0 +1,631 @@
...
+
+	found := false
+	var walk func(node *ast.Node) bool
+	walk = func(node *ast.Node) bool {
+		if node == nil || found {
+			return found
+		}
+		if ast.IsFunctionLike(node) {
+			return false
+		}
```
fansenze, line 661: **[P2] Check computed method names for await**

Consider `async () => ({ [await Promise.resolve('sum')]() { return 0; } })`. The computed name executes in the factory's async context, but skipping the entire method node hides this await. The rule then suggests removing `async`, producing a syntax error. Please inspect computed names before skipping nested function bodies.

internal/plugins/rstest/rules/no_async_mock_factory/no_async_mock_factory.go
```diff
@@ -0,0 +1,631 @@
...
+		return false
+	}
+
+	resolved := t
+	if utils.IsTypeFlagSet(resolved, checker.TypeFlagsObject) &&
+		checker.Type_objectFlags(resolved)&checker.ObjectFlagsReference != 0 {
+		resolved = resolved.Target()
+	}
+	if symbol := checker.Type_symbol(resolved); symbol != nil && symbol.Name == "Promise" {
+		return true
```
fansenze: **[P2] Do not treat every type named Promise as the native Promise**

A local `class Promise { sum = 0; }` makes `rs.doMockRequire('./sum', () => new Promise())` return a synchronous object, which Rstest accepts. This check still reports it because the type is named Promise. Please verify that the symbol refers to the global Promise type or a subclass, rather than accepting any class or interface with that name.

internal/plugins/rstest/rules/no_async_mock_factory/no_async_mock_factory.go
```diff
@@ -0,0 +1,631 @@
...
+		return verdictUnknown
+	}
+	declaration := symbol.Declarations[0]
+	if declaration == nil || ast.GetSourceFileOfNode(declaration) != ctx.SourceFile {
+		return verdictUnknown
+	}
+
+	switch declaration.Kind {
+	case ast.KindFunctionDeclaration:
+		return classifyFunction(ctx, declaration)
```
fansenze, line 251: **[P2] Account for reassignment before inspecting a function declaration**

Function declarations can be reassigned. For example, `function factory(): unknown { return Promise.resolve({}); } factory = () => ({}); rs.doMock('./m', factory);` uses a synchronous factory, but this branch reports it based on the original declaration. Please check for writes before trusting the declaration's body; otherwise, fall back to the type layer.

internal/plugins/rstest/rules/no_async_mock_factory/no_async_mock_factory.go
```diff
@@ -0,0 +1,631 @@
...
+	if flags&ast.FunctionFlagsGenerator != 0 {
+		return verdictSync
+	}
+	if flags&ast.FunctionFlagsAsync != 0 {
+		return verdictPromise
+	}
+
+	returned, ok := returnedExpressions(fn)
+	if !ok || len(returned) == 0 {
+		return verdictUnknown
```
fansenze, line 215: **[P2] Include implicit returns when classifying the factory**

For `rs.doMockRequire('./m', () => { if (flag) return Promise.resolve({}); })`, a false flag makes the factory return undefined synchronously, which Rstest accepts for a CommonJS mock. Only explicit returns are collected here, so this factory is incorrectly classified as always returning a Promise. Please account for paths that reach the end without a return before producing `verdictPromise`.

fansenze (commented):
Five issues found: three false positives and two problems with the suggested fixes.

internal/plugins/rstest/rules/no_async_mock_factory/no_async_mock_factory.go
elecmonkey: Fixed. A block body whose end is reachable now yields verdictUnknown (utils.IsFunctionEndReachable) — the fall-through path returns undefined, which the runtime accepts. Limited to block bodies, since the binder sets the same flag on concise arrows; the async branch is settled earlier and unchanged.

internal/plugins/rstest/rules/no_async_mock_factory/no_async_mock_factory.go
elecmonkey: Fixed. The function-declaration branch now bails to the type layer when ctx.Refs.References shows a write to the name (utils.IsWriteReference). The type layer sees `() => unknown` and stays silent.

internal/plugins/rstest/rules/no_async_mock_factory/no_async_mock_factory.go
elecmonkey: Fixed. isPromiseInstanceType now also requires the symbol to come from the default library (utils.IsSymbolFromDefaultLibrary), so a local `class Promise` no longer matches. Subclasses of the global Promise still report through the base-type branch.

internal/plugins/rstest/rules/no_async_mock_factory/no_async_mock_factory.go
elecmonkey: Fixed by suppressing both suggestions when the factory has a return type annotation. Rewriting the annotation would mean handling PromiseLike, unions and aliases, and a wrong rewrite breaks the code again — not worth it for a suggestion. The diagnostic is still reported, just without one.

internal/plugins/rstest/rules/no_async_mock_factory/no_async_mock_factory.go
elecmonkey: Fixed. bodyAwaits now walks every child of a nested function-like node except its body, so computed member names and decorators are inspected in the context where they run. Parameter defaults are safe to include: `await` is an early error there.

fansenze (approved):

Issue #475, opened 2026-02-12
[Tracking]: eslint-plugin-n rules


fansenze:
This is a tracking issue for porting the existing rules from [`eslint-plugin-n`](https://github.com/eslint-community/eslint-plugin-n?tab=readme-ov-file#-rules).

There are **41** rules (+ 2 deprecated).

| Category | ✅ Done | 🚫 Won't | Total |
| :--- | :---: | :---: | :---: |
| Recommended | 14 | 1 | 15 |
| Not recommended | 24 | 0 | 24 |
| Deprecated | 0 | 2 | 2 |
| **Total** | **38** | **3** | **41** |

> ✅ = Implemented &nbsp;|&nbsp; 🚫 = Not intending to implement

---

## Recommended rules

<details open>
<summary>✅ 14 &nbsp;|&nbsp; 🚫 1 &nbsp;/&nbsp; Total: 15</summary>

| Status | Name |
| :----: | :--- |
| ✅ | [n/hashbang](https://github.com/eslint-community/eslint-plugin-n/blob/master/docs/rules/hashbang.md) |
| ✅ | [n/no-deprecated-api](https://github.com/eslint-community/eslint-plugin-n/blob/master/docs/rules/no-deprecated-api.md) |
| ✅ | [n/no-exports-assign](https://github.com/eslint-community/eslint-plugin-n/blob/master/docs/rules/no-exports-assign.md) |
| ✅ | [n/no-extraneous-import](https://github.com/eslint-community/eslint-plugin-n/blob/master/docs/rules/no-extraneous-import.md) |
| ✅ | [n/no-extraneous-require](https://github.com/eslint-community/eslint-plugin-n/blob/master/docs/rules/no-extraneous-require.md) |
| ✅ | [n/no-missing-import](https://github.com/eslint-community/eslint-plugin-n/blob/master/docs/rules/no-missing-import.md) |
| ✅ | [n/no-missing-require](https://github.com/eslint-community/eslint-plugin-n/blob/master/docs/rules/no-missing-require.md) |
| ✅ | [n/no-process-exit](https://github.com/eslint-community/eslint-plugin-n/blob/master/docs/rules/no-process-exit.md) |
| ✅ | [n/no-unpublished-bin](https://github.com/eslint-community/eslint-plugin-n/blob/master/docs/rules/no-unpublished-bin.md) |
[...]

PR #2103, opened 2026-09-08
feat: add node/hashbang rule

fansenze:
## Summary

Add the bundled `node` plugin and `node/hashbang`, based on eslint-plugin-n 18.3.0. Package executables require the configured hashbang, BOM removal and LF ending; library files report unwanted hashbangs. Support `convertPath`, `additionalExecutables`, `executableMap` and publication filtering, with native/JS plugin registration and documentation.

Reuse tsgo's scanner and path/package services, rslint's Program cache, ECMAScript helpers and minimatch. Pure Git ignore parsing lives in `internal/utils/gitignore`, with separate APIs for pattern arrays and file text. Config collection retains its existing glob projection, matching engine and directory ownership. Node package metadata, bin aliases, path conversion and publication policy live in `internal/plugins/node/nodeutil`; the rule no longer imports config internals. No dependencies are added.

Preserve intended glob and package semantics when upstream has bugs: character-class negation and backslash quoting keep their meaning, array entries remain intact, and publication paths use the containing package. This also fixes `..hidden.js` being treated as an ancestor, cwd-dependent README checks and converted paths entering nested packages. Existing headers are replaced safely instead of inserting a second hashbang. The rule's Differences section explains remaining deviations with concrete user examples; emoji and Kelvin-sign matching differences are removed.

## Validation

- Scoped Go tests pass for the rule, Node utilities, Git ignore parser/collector, config ignore consumers and both existing minimatch consumer rules (`import/order`, `import/no-restricted-paths`). Race tests pass for shared matchers, concurrent package/config caches and the rule.
- Rebuilt the native binary. The rule and selected Git ignore CLI suites pass 109 JS tests; the 69 upstream rule cases were rerun after the final publication change.
- 1067 native/upstream diagnostic and fix comparisons meet their intended semantic expectations; 29 intentional upstream differences are accounted for. Added semantic cases verify exact fixed output and clean re-lint. Another 71 rooted/unrooted exclusion cases and 23 Unicode, conversion and safe-fix cases pass.
- Git `check-ignore` independently verifies valid pattern semantics. Branch-filtered Go lint, explicit-path spellcheck, formatting and staged diff checks pass. Local verification ran on macOS arm64; no whole-repository or whole-plugin suites were run locally.

## Performance

Warm Go rule benchmarks, three samples compared with review baseline `f446357c`:

| Case | Before | After | Allocated before → after |
| --- | ---: | ---: | ---: |
| Publication filtering | 2.65 µs | 1.50 µs | 2640 → 1280 B |
| Additional executables | 1.44 µs | 1.26 µs | 1232 → 856 B |
| 288 KiB library | 0.77 µs | 0.71 µs | 704 → 624 B |

Compiled publication/pattern data is reused per Program and configuration. The existing minimatch engine avoids regexp work for ASCII case-insensitive literals while preserving its default behavior and Unicode fallback. These measurements exclude Program construction and do not imply cold-start or whole-command speedups. Benchmark files and ad hoc audit scripts are excluded from the PR.

## Related Links

- Node plugin implementation progress: #475 (`node/hashbang`).
- [Pinned upstream rule](https://github.com/eslint-community/eslint-plugin-n/blob/v18.3.0/lib/rules/hashbang.js)
- Shared matching support: #2102

## Checklist

- [x] Tests updated.
- [x] Documentation updated.


packages/rslint/src/config/define-config.ts
```diff
@@ -35,6 +36,7 @@ const NATIVE_PLUGIN_DECL_ALIASES = [
   'eslint-plugin-import',
   'eslint-plugin-jest',
   'eslint-plugin-jsx-a11y',
+  'eslint-plugin-n',
```
chenjiahan: Since `eslint-plugin-n` is a fork of `eslint-plugin-node`, should we rename it to `eslint-plugin-node`? The name `n` may not be clear to users.

elecmonkey: Most users configurate `eslint-plugin-n` now, so keeping the n name may make migration easier

elecmonkey:
should we rename PR title `feat: add n/hashbang rule` or `feat: port n/hashbang rule`? Including plugin name in the PR title follows the history convention.

packages/rslint/src/config/define-config.ts
chenjiahan: Migration can be done by coding agents, I’d prioritize getting the naming right for the long term

fansenze: Renamed the plugin to `node` and the rule to `node/hashbang`, matching [Oxlint](https://oxc.rs/docs/guide/usage/linter/plugins). The port still follows `eslint-plugin-n`.

internal/plugins/node/rules/hashbang/paths.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+	var includes, excludes []string
+	var extendedIncludes, extendedExcludes []string
+	for _, file := range files {
+		pattern, ok := file.(string)
+		if !ok || pattern == "" {
+			continue
+		}
+		negated := strings.HasPrefix(pattern, "!")
+		body := strings.TrimRight(path.Clean(strings.TrimPrefix(strings.TrimPrefix(pattern, "!"), "/")), "/")
```
elecmonkey: With `{"bin":"lib/foo.js","files":["lib","!/foo.js"]}` and `ignoreUnpublished: true`, upstream removes only the leading `!`, so `/foo.js` remains root-anchored and `lib/foo.js` is still checked, producing `expectedHashbangNode`. This expression removes both `!` and `/`, turns the exclusion into the unrooted `foo.js`, and incorrectly treats `lib/foo.js` as unpublished, so the rule emits no diagnostic. Please preserve the leading slash after removing the negation marker.

elecmonkey (approved):
no more issues found

PR #2110, opened 2026-09-08
feat(rstest): add consistent-test-it rule

elecmonkey:
## Summary

Add `rstest/consistent-test-it`, which enforces a consistent choice between `test` and `it` when registering tests. By default it prefers `test` outside `describe` calls and `it` inside them, while `fn` and `withinDescribe` can be used to require a single convention or override the suite convention.

The rule resolves Rstest globals, imports from `@rstest/core` and `rstack/test`, namespace imports, destructured `require` bindings, `import.meta.rstest`, const aliases, fixture extension chains, modifiers, and parameterized `.each` / `.for` registrations. It reports non-registration references without treating imports as standalone violations, preserves existing bindings when fixing calls, and withholds fixes when the preferred name would be captured by another binding or when the call's binding could have been reassigned.

`@rstest/playwright` test registrations are exempt because that package exposes `test` but no `it`, which the shared Rstest call parser now reports as provenance. Spellings shadowed by a local TypeScript namespace are filtered by the rule rather than by the parser, so no other parser user pays for that scope walk.

## Credits

Port from [`eslint-plugin-jest`](https://github.com/jest-community/eslint-plugin-jest)'s and [`@vitest/eslint-plugin`](https://github.com/vitest-dev/eslint-plugin-vitest)'s `consistent-test-it`.

## Related Links

Related #935

## Checklist


internal/plugins/rstest/rules/consistent_test_it/consistent_test_it.go
```diff
@@ -0,0 +1,166 @@
...
+		if statement.Kind != ast.KindImportDeclaration {
+			continue
+		}
+		declaration := statement.AsImportDeclaration()
+		if !rstestUtils.IsRstestCoreImportModule(declaration.ModuleSpecifier.Text()) {
+			continue
+		}
+		for _, element := range rstestUtils.NamedImportElements(declaration) {
+			if rstestUtils.ImportedSpecifierName(element) == preferred && element.Name().Text() == preferred {
+				return []rule.RuleFix{fix}
```
fansenze, line 173: **[P2] Check static-block shadowing before reusing an import**

When both `it` and `test` are imported, `class C { static { if (true) { var test = () => {}; } it('case', () => {}); } }` is fixed to call the local `test` function, so the test is no longer registered. `IsNameShadowedBetween` misses the `var` binding hoisted to the static block, and this early return skips the later `IsShadowed` check. Please account for static-block bindings before reusing the import.

internal/plugins/rstest/rules/consistent_test_it/consistent_test_it.go
```diff
@@ -0,0 +1,166 @@
...
+		return nil
+	}
+	// Only a direct import/global can be replaced by the other base API.
+	// `const test = base.extend(...)` must retain its fixtures.
+	symbol := ctx.Refs.Resolve(root)
+	name, _, mode := testFramework.ResolveFunctionIdentifierReferenceFromSymbolModules(
+		root.Text(), root, symbol, ctx.SourceFile, rstestUtils.RstestCoreImportModules,
+	)
+	if name != parsed.Name {
+		return nil
```
fansenze, line 129: **[P2] Preserve fixtures on reassigned CommonJS bindings**

With an existing `test` import, `let { it } = require('@rstest/core'); it = it.extend({ account: {} }); it('case', ({ account }) => {});` is fixed to call the imported `test`. That switches to the base API and loses the `account` fixture. The declaration still looks like a direct CommonJS import, but its value has changed. Please check for subsequent assignments, or withhold the fix for mutable CommonJS bindings.

internal/plugins/rstest/utils/parse_rstest_fn.go
```diff
@@ -353,6 +354,14 @@ func resolveRstestRoot(
 
 	localName := root.AsIdentifier().Text
 	symbol := resolveRstestRootSymbol(ctx, root)
+	// The source-only binder does not resolve every value-producing TypeScript
+	// declaration (notably namespaces), and can resolve through a declaration
+	// in a nearer module block to an outer import. Do not treat either shape as
+	// an Rstest API merely because its spelling is `test`, `it`, or `describe`.
+	if isRstestRootShadowedByModuleBlock(root, ctx.SourceFile, localName, symbol) ||
+		(symbol == nil && internalUtils.IsShadowed(root, localName)) {
+		return rstestResolvedAPI{}, 0, false
```
fansenze: **[P2] Avoid rescanning the file for every global test call**

Without type information, each global `test` or `it` call has a nil symbol and runs `IsShadowed`, repeatedly scanning the file's declarations and AST. This introduces quadratic work. A benchmark with 5,000 global registrations, enabling only the existing `no-focused-tests` rule, increased from about 2.4 ms with the parent parser to 578 ms with this change. Please cache shadowing results by scope and name, or index declarations once per file.

fansenze (commented):
Found three issues with autofix safety and parser performance.

internal/plugins/rstest/utils/parse_rstest_fn.go
elecmonkey: Confirmed, fixed in 49c4cacb.

The guard was in the shared parser, so every rule paid for a strictness requirement only this one needs. `resolveRstestRoot` is now unchanged — `parse_rstest_fn.go` adds nothing but the `IsPlaywright` field — and the filter moved into this rule, where a one-pass index of the file's declaration names keeps the scope walk off the common path.

On your benchmark shape, 1250/2500/5000 registrations went from 39.2 / 156.3 / 603.3 ms to 2.0 / 2.3 / 5.6 ms.

internal/plugins/rstest/rules/consistent_test_it/consistent_test_it.go
elecmonkey: Confirmed, fixed in 49c4cacb.

`IsNameShadowedBetween` reads the static block's body as a plain `Block`, and a static block is not function-like, so its hoisted-`var` arm never runs. The ancestor walk in `callFixes` now handles `ClassStaticBlockDeclaration`, before the import-reuse loop — reordering the checks would not work, since at file level `IsShadowed` counts the `import { test }` itself.

Your case is now reported without a fix, with a regression test.

internal/plugins/rstest/rules/consistent_test_it/consistent_test_it.go
elecmonkey: Confirmed, fixed in 49c4cacb.

The require-binding resolver never checks whether the binding is `const`, so a later assignment is invisible to it. The fix is now withheld for any CommonJS destructured binding declared with `let` or `var`; `const { it } = require('@rstest/core')` still gets its fix. Regression tests and the Autofix docs cover it.

internal/plugins/rstest/rules/consistent_test_it/consistent_test_it.go
```diff
@@ -0,0 +1,291 @@
...
+	root := parsed.Head.Local.Node
+	if root == nil || root.Kind != ast.KindIdentifier {
+		return false
+	}
+	name := root.Text()
+	symbol := ctx.Refs.Resolve(root)
+	if isShadowedByModuleBlock(root, ctx.SourceFile, name, symbol) {
+		return true
+	}
+	return symbol == nil && scan.declaresName(name) && utils.IsShadowed(root, name)
```
fansenze: **[P2] Cache shadowing checks when the name index matches**

An unrelated `const helper = { test() {} };` makes `declaresName('test')` true for the whole file. Every global `test(...)` then runs the full scope scan again, on both entry and exit. With only this rule enabled and no type information, 5,000 registrations take about **1.48 s instead of 2.8 ms**. Please cache the result by scope and name, and cover unrelated parameter/method names in a regression benchmark.

internal/plugins/rstest/rules/consistent_test_it/consistent_test_it_extras_test.go
```diff
@@ -0,0 +1,307 @@
...
+			// CommonJS can reuse the existing named base API.
+			{Code: "import { test } from '@rstest/core'; const { it } = require('@rstest/core'); it('case');", Output: []string{"import { test } from '@rstest/core'; const { it } = require('@rstest/core'); test('case');"}, Errors: []rule_tester.InvalidTestCaseError{{MessageId: "consistentMethod", Message: "Prefer using 'test' instead of 'it'", Line: 1, Column: 78, EndLine: 1, EndColumn: 80}}},
+			// A reassignable CommonJS binding no longer holds what its pattern declares.
+			{Code: "import { test } from '@rstest/core'; let { it } = require('@rstest/core'); it = it.extend({ account: {} }); it('case', ({ account }) => {});", Output: []string{}, Errors: []rule_tester.InvalidTestCaseError{{MessageId: "consistentMethod", Message: "Prefer using 'test' instead of 'it'", Line: 1, Column: 109, EndLine: 1, EndColumn: 111}}},
+			// A reassignable CommonJS binding no longer holds what its pattern declares.
```
fansenze: **Fix the spelling-check failure**

CSpell rejects `reassignable` in these two comments, causing the latest `Lint&Check` job to fail. Please replace it with `mutable`, which is already used elsewhere in this PR. [CI log](https://github.com/web-infra-dev/rslint/actions/runs/34247151707/job/102299914421).

fansenze (commented):
The two unsafe autofix cases now pass, and the shared parser's performance has recovered. Two follow-ups remain:

internal/plugins/rstest/rules/consistent_test_it/consistent_test_it.go
elecmonkey: Confirmed, fixed in 5bf7b882.

Member names no longer enter the declaration-name index, and the scope walk is memoized per (scope, name). Your shape, 5000 registrations: 1195 ms → 4.2 ms; an unrelated parameter or local behaves the same.

`BenchmarkConsistentTestItRegistrations` covers all five shapes, and `TestConsistentTestItSourceOnlyShadowing` pins that separate scopes do not share an answer.

internal/plugins/rstest/rules/consistent_test_it/consistent_test_it_extras_test.go
elecmonkey: Fixed in 5bf7b882 — both comments now read `mutable`.

fansenze (approved):

PR #2111, opened 2026-09-08
feat: port rule unicorn/prefer-date-now

shiyudesu:
## Summary

Port `unicorn/prefer-date-now` from eslint-plugin-unicorn v74.0.0.

The rule prefers `Date.now()` over creating a `Date` object solely for its numeric timestamp, with automatic fixes for method calls, numeric conversions, and arithmetic expressions.

This change also:

- Registers the rule in the Unicorn catalog and adds the JS integration suite.
- Adds the complete upstream test suite, regression coverage, and rule documentation.
- Fixes keyword spacing through JSDoc cast wrappers in the shared Unicorn fixer, with coverage for its existing consumer.

## Validation

- Targeted Go tests, scoped Go lint, formatting, and spelling checks passed.
- All 116 differential cases matched upstream v74.0.0.
- JS integration and rule catalog checks were not completed locally and remain to be verified in CI.

## Related Links

- Tracking issue: #1309
- [Upstream documentation](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v74.0.0/docs/rules/prefer-date-now.md)
- [Upstream source](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v74.0.0/rules/prefer-date-now.js)

## Checklist

### DIFF
--- a/internal/plugins/unicorn/all.go
+++ b/internal/plugins/unicorn/all.go
@@ -36,6 +36,7 @@ import (
 	"github.com/web-infra-dev/rslint/internal/plugins/unicorn/rules/prefer_array_flat_map"
 	"github.com/web-infra-dev/rslint/internal/plugins/unicorn/rules/prefer_array_some"
 	"github.com/web-infra-dev/rslint/internal/plugins/unicorn/rules/prefer_blob_reading_methods"
+	"github.com/web-infra-dev/rslint/internal/plugins/unicorn/rules/prefer_date_now"
 	"github.com/web-infra-dev/rslint/internal/plugins/unicorn/rules/prefer_node_protocol"
 	"github.com/web-infra-dev/rslint/internal/plugins/unicorn/rules/prefer_number_properties"
 	"github.com/web-infra-dev/rslint/internal/plugins/unicorn/rules/prefer_set_has"
@@ -85,6 +86,7 @@ func GetAllRules() []rule.Rule {
 		prefer_array_flat_map.PreferArrayFlatMapRule,
 		prefer_array_some.PreferArraySomeRule,
 		prefer_blob_reading_methods.PreferBlobReadingMethodsRule,
+		prefer_date_now.PreferDateNowRule,
 		prefer_node_protocol.PreferNodeProtocolRule,
 		prefer_number_properties.PreferNumberPropertiesRule,
 		prefer_set_has.PreferSetHasRule,
--- a/internal/plugins/unicorn/rules/prefer_date_now/prefer_date_now.go
+++ b/internal/plugins/unicorn/rules/prefer_date_now/prefer_date_now.go
@@ -1,0 +1,109 @@
+package prefer_date_now
+
+import (
+	"github.com/microsoft/TypeScript/tsc/shim/ast"
+	"github.com/web-infra-dev/rslint/internal/plugins/unicorn/unicornutil"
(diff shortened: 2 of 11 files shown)


internal/plugins/unicorn/all.go
```diff
@@ -85,6 +86,7 @@ func GetAllRules() []rule.Rule {
 		prefer_array_flat_map.PreferArrayFlatMapRule,
 		prefer_array_some.PreferArraySomeRule,
 		prefer_blob_reading_methods.PreferBlobReadingMethodsRule,
+		prefer_date_now.PreferDateNowRule,
```
elecmonkey, line 89: [P1] Enable the ported rule in the recommended preset

Registering the rule here makes explicit configuration work, but `packages/rslint/src/config/presets/unicorn.ts` still leaves `unicorn/prefer-date-now` commented out as ‘not implemented.’ With `unicornPlugin.configs.recommended`, `const timestamp = new Date().getTime()` produces no Rslint diagnostic while eslint-plugin-unicorn v74.0.0 reports one; enable the rule in the preset as prior recommended Unicorn ports do.

internal/plugins/unicorn/rules/prefer_date_now/prefer_date_now.go
```diff
@@ -0,0 +1,109 @@
...
+		zeroArguments := 0
+		return rule.RuleListeners{
+			ast.KindCallExpression: func(node *ast.Node) {
+				if method, ok := unicornutil.MatchDotMethodCall(node, unicornutil.DotMethodCallOptions{
+					Methods:         []string{"getTime", "valueOf"},
+					ArgumentsLength: &zeroArguments,
+				}); ok && isNewDate(utils.ESTreeRuntimeExpression(method.Object)) {
+					report(method.Property, node, rule.RuleMessage{
+						Id:          "prefer-date-now-over-methods",
+						Description: "Prefer `Date.now()` over `Date#" + method.Property.Text() + "()`.",
```
elecmonkey: [P2] Preserve the method placeholder data

For both `getTime` and `valueOf`, upstream reports structured message data such as `{method: "getTime"}`, but this message only supplies the rendered description, leaving `RuleMessage.Data` nil. That loses the repository's structured placeholder metadata contract and differs from neighboring Unicorn ports; populate `Data` with the matched method name.

internal/plugins/unicorn/all.go
shiyudesu: Fixed in bcdd461ea45aec921c083319a8e738ee67ac5e40. `unicorn/prefer-date-now` is now enabled at `error` in the recommended preset. Added a regression assertion in `presets.test.ts` and documented the preset inclusion. The source-level preset check passed; the JS preset suite was not run locally and remains to be checked in CI.

internal/plugins/unicorn/rules/prefer_date_now/prefer_date_now.go
shiyudesu: Fixed in bcdd461ea45aec921c083319a8e738ee67ac5e40. Both method diagnostics now populate `RuleMessage.Data` with `{method: methodName}`. Extended the edit-demand tests to assert the complete message, including `Data`, for both `getTime` and `valueOf` under all four edit-demand modes. The targeted Go rule tests and scoped Go lint passed.

elecmonkey (approved):

PR #2113, opened 2026-09-08
feat: add node/no-extraneous-import rule

fansenze:
## Summary

Adds `node/no-extraneous-import`, porting eslint-plugin-n v18.3.0 as part of [#475](https://github.com/web-infra-dev/rslint/issues/475). Reports imports, re-exports, and literal dynamic imports of installed packages that are not declared by the package or its nearest matching workspace. Supports type-only dependency exemptions and shared Node settings.

Node-specific package checks, runtime import resolution, and tsconfig lookups live in `internal/plugins/node/nodeutil`. They reuse the existing Program filesystem and generation cache, rslint's glob utilities, and tsgo's module traversal, compiler host, path helpers, extension mapping, and config parser. The PR has no changes under `internal/program`. Repeated imports reuse results and retain a diagnostic for every occurrence. Custom module directories and symlinked packages preserve runtime lookup.

The rule documentation describes remaining differences through concrete configurations, diagnostic outcomes, and migration guidance: `resolverConfig` supports `modules` only; numeric brace ranges and caret character classes in workspace patterns can change whether a root dependency is accepted.

## Validation

- `go test ./internal/plugins/node/nodeutil ./internal/plugins/node/rules/no_extraneous_import`: 30 upstream rule cases and 124 extras pass, along with utility tests for runtime paths, custom directories, symlinks, cache generations, and compiler options.
- `go test ./internal/rules -run '^TestAllContainsEveryGoRuleExactlyOnce$'` passes; registration is unchanged by the follow-up fixes.
- Rebuilt with `pnpm --filter @rslint/core build:bin`, then ran `CI=true pnpm --dir packages/rslint-test-tools exec rs test run tests/eslint-plugin-node/rules/no-extraneous-import.test.ts`: 30 tests pass, no skips. The two existing Node integration consumers also passed when the shared fixture wrapper changed.
- Pinned-upstream comparisons pass for 63 configuration boundary cases, 90 resolution cases, and 25 review cases. The two documented caret-class diagnostic differences are explicitly asserted. A real-source comparison covers 14 repository files plus a positive fixture, checking file/rule coverage, messages, severity, full ranges, and absent fixes/suggestions.
- Scoped Go lint, changed-file spell check, and `pnpm run format:check` pass. Temporary Go benchmarks cover eight cold/warm scenarios with five repetitions and diagnostic assertions. The final code was also checked against the previous commit with alternating benchmark runs; benchmark files are excluded from the commit.

## Related Links

- Related to #475
- [Pinned upstream rule](https://github.com/eslint-community/eslint-plugin-n/blob/v18.3.0/lib/rules/no-extraneous-import.js)

## Checklist


internal/plugins/node/nodeutil/imports.go
```diff
@@ -0,0 +1,142 @@
...
+		Extensions: StringListSetting("tryExtensions", options, settings),
+		Paths:      StringListSetting("resolvePaths", options, settings),
+		Conditions: []string{"node", "require", "import"},
+	}
+	if typeOnly {
+		result.Conditions = append(result.Conditions, "types")
+	}
+	for _, value := range settingValues("resolverConfig", options, settings) {
+		if config, ok := value.(map[string]any); ok {
+			result.Modules = stringArray(config["modules"])
```
elecmonkey, line 92: [P2] Preserve string-valued `resolverConfig.modules`

`enhanced-resolve` accepts `modules` as either a string or an array, but `stringArray` returns `nil` for a string and silently restores the default `node_modules` lookup. With `resolverConfig: { modules: "bower_components" }` and `bower_components/string-mod/index.js`, eslint-plugin-n@18.3.0 reports `import "string-mod"` as extraneous, while Rslint cannot resolve it and reports nothing. Please normalize a scalar module directory to a one-element list.

internal/plugins/node/nodeutil/imports.go
```diff
@@ -0,0 +1,142 @@
...
+		aliases[emitted] = append(aliases[emitted], extension)
+	}
+	return aliases
+}
+
+// ImportModuleName returns the npm package root after removing loader params.
+// Builtins, relative/absolute paths, import maps and URL imports have no npm name.
+func ImportModuleName(specifier string) (name, resource string) {
+	resource, _, _ = strings.Cut(specifier, "!")
+	if core.NodeCoreModules()[resource] || strings.HasPrefix(resource, "data:") ||
```
elecmonkey: [P2] Match Node's complete builtin-module set

Upstream filters imports with `node:module.isBuiltin`, whose set includes legacy internal builtins such as `_http_agent`, while `core.NodeCoreModules()` omits them. If `node_modules/_http_agent/index.js` exists, eslint-plugin-n@18.3.0 ignores `import "_http_agent"`, but Rslint resolves the shadow package and reports `"_http_agent" is extraneous.` Please use a builtin set compatible with Node's `isBuiltin` behavior.

elecmonkey:
Document the additional workspace-glob difference

Upstream evaluates `workspaces` patterns with globrex, while this uses minimatch3, which also differs for character classes beyond the documented numeric brace-range divergence. For example, with `workspaces: ["packages/[^a]*"]`, a child at `packages/app`, and a root-only `workspace-dep`, upstream treats the child as part of the workspace and reports nothing, while Rslint reports `"workspace-dep" is extraneous.` If exact parity is not intended here, could we document and test this additional divergence so users do not assume numeric brace ranges are the only difference?

elecmonkey (approved):

PR #2115, opened 2026-09-08
feat(rstest): add prefer-to-be rule

elecmonkey:
## Summary

Add `rstest/prefer-to-be` to replace equality assertions with their clearest matcher. Primitive literals use `toBe()`, while `null`, `undefined`, and `NaN` use `toBeNull()`, `toBeUndefined()` / `toBeDefined()`, and `toBeNaN()`. The rule is registered in the Rstest plugin but stays out of the `recommended` preset, matching upstream.

Assertions resolve through the shared Rstest call analysis, covering globals, named and renamed imports, namespace and `require` forms, `rstack/test`, `import.meta.rstest`, Playwright, TestContext `expect`, and the `soft` and `poll` assertion factories. Chai assertions, browser-only `expect.element()` assertions, dynamic matcher names, foreign expect functions, and local shadows are ignored.

The matcher-independent decision tree now lives under `internal/utils/test_framework/rules/prefer_to_be` and is shared with the existing Jest rule. Matcher-call lookup, argument-list ranges, and basic TypeScript assertion unwrapping are also shared through `internal/utils/test_framework`. The Jest adapter retains the `eslint-plugin-jest@29.16.1` contract, including computed accessor fixes and parenthesized optional-chain boundaries, while both adapters construct autofixes only when requested.

## Credits

- Port from [`@vitest/eslint-plugin@1.6.27`'s `prefer-to-be` documentation](https://github.com/vitest-dev/eslint-plugin-vitest/blob/v1.6.27/docs/rules/prefer-to-be.md) and [source](https://github.com/vitest-dev/eslint-plugin-vitest/blob/v1.6.27/src/rules/prefer-to-be.ts).
- Align the shared Jest implementation with [`eslint-plugin-jest@29.16.1`'s `prefer-to-be` documentation](https://github.com/jest-community/eslint-plugin-jest/blob/v29.16.1/docs/rules/prefer-to-be.md) and [source](https://github.com/jest-community/eslint-plugin-jest/blob/v29.16.1/src/rules/prefer-to-be.ts).

- **Negative fractional numeric literals are left unchanged.** Upstream leaves positive fractional literals unchanged but reports negative ones after unwrapping the unary minus. Rstest applies the same floating-point policy in both directions because these assertions should use `toBeCloseTo()` when approximate comparison is intended.

- **Negated undefined equality assertions preserve their meaning.** `expect(value).not.toEqual(undefined)` is fixed to `expect(value).toBeDefined()` by removing `not`; upstream changes the matcher without removing the modifier.

- **Locally shadowed `undefined` and `NaN` are not treated as built-ins.** A local runtime declaration with either name prevents the rule from selecting the corresponding dedicated matcher.

- **Browser element assertions are excluded.** `expect.element()` exposes browser-element matchers rather than the value matchers targeted by this rule.

- **Autofixes preserve authored accessor spelling.** Quoted and template-literal accessors retain their delimiters instead of being normalized to single quotes.

- **Comments inside a removed argument list suppress the autofix.** The diagnostic is still reported, but Rstest does not delete the comment while converting the assertion to a no-argument matcher.

- **Parenthesized optional matcher calls are recognized.** Rstest treats `(expect(value)?.toEqual)(1)` as the same assertion as the unparenthesized optional-chain form.

- **Computed matcher accessors keep a valid key.** Upstream's `replaceAccessorFixer` rewrites an identifier accessor in place, so `expect(value)[toEqual](1)` becomes a reference to an undeclared `toBe`. The fix writes a quoted key instead.

- **Type arguments are removed with the value arguments.** `toBeNull`, `toBeNaN`, `toBeUndefined` and `toBeDefined` declare no type parameters, so `expect(null).toEqual<null>(null)` is fixed to `expect(null).toBeNull()`. Leaving the type argument behind fails with TS2558. `toBe` shares `toEqual`'s single type parameter, so its type argument stays.

- **Each assertion is reported once.** An outer call in the same chain resolves back to the matcher call, so upstream reports `expect(promise).resolves.toEqual(1).then(done)` twice at the same location.

## Related Links

Related #935

## Checklist

### DIFF
--- a/internal/plugins/jest/rules/prefer_to_be/prefer_to_be.go
+++ b/internal/plugins/jest/rules/prefer_to_be/prefer_to_be.go
@@ -50,25 +50,31 @@ func buildFixes(ctx rule.RuleContext, match shared.Match) []rule.RuleFix {
 		return nil
 	}
 
-	var renameFix rule.RuleFix
-	if match.Expect.MatcherEntry.Node.Kind == ast.KindIdentifier {
-		// Match replaceAccessorFixer exactly: an identifier accessor remains an
-		// identifier even in `expect(x)[toEqual](1)`.
-		renameFix = rule.RuleFixReplace(ctx.SourceFile, match.Expect.MatcherEntry.Node, replacement)
-	} else {
-		var ok bool
-		renameFix, ok = jestUtils.ReplaceMemberNameFix(
-			ctx,
-			&match.Expect.MatcherEntry,
-			replacement,
-		)
-		if !ok {
-			return nil
-		}
+	// NOTE: replaceAccessorFixer keeps an identifier accessor bare, which turns
+	// `expect(x)[toEqual](1)` into a reference to an undeclared `toBe`. A
+	// computed key is quoted instead so the fixed code still runs.
+	renameFix, ok := jestUtils.ReplaceMemberNameFix(
+		ctx,
+		&match.Expect.MatcherEntry,
+		replacement,
+	)
+	if !ok {
+		return nil
 	}
 	fixes := []rule.RuleFix{renameFix}
 
 	arguments := match.Expect.MatcherCall.Arguments()
+	if match.Kind != shared.KindToBe {
+		// NOTE: the dedicated matchers declare no type parameters, so type
+		// arguments written for toBe/toEqual have to go with the value
+		// arguments. Upstream leaves them behind and produces TS2558.
+		if typeArgumentsRange, hasTypeArguments := testFramework.CallTypeArgumentListRange(
+			ctx.SourceFile,
+			match.Expect.MatcherCall,
+		); hasTypeArguments {
+			fixes = append(fixes, rule.RuleFixRemoveRange(typeArgumentsRange))
+		}
+	}
 	if match.Kind != shared.KindToBe && len(arguments) > 0 {
 		argumentsRange, ok := testFramework.CallArgumentListRange(
 			ctx.SourceFile,
--- a/internal/plugins/jest/rules/prefer_to_be/prefer_to_be.go
+++ b/internal/plugins/jest/rules/prefer_to_be/prefer_to_be.go
@@ -1,267 +1,138 @@
 package prefer_to_be
 
 import (
-	"slices"
-
 	"github.com/microsoft/TypeScript/tsc/shim/ast"
-	"github.com/web-infra-dev/rslint/internal/plugins/jest/utils"
+	"github.com/microsoft/TypeScript/tsc/shim/core"
+	jestUtils "github.com/web-infra-dev/rslint/internal/plugins/jest/utils"
 	"github.com/web-infra-dev/rslint/internal/rule"
+	internalUtils "github.com/web-infra-dev/rslint/internal/utils"
+	testFramework "github.com/web-infra-dev/rslint/internal/utils/test_framework"
+	shared "github.com/web-infra-dev/rslint/internal/utils/test_framework/rules/prefer_to_be"
 )
 
-// Message Builders
-
-func buildUseToBeErrorMessage() rule.RuleMessage {
-	return rule.RuleMessage{
-		Id:          "useToBe",
-		Description: "Use `toBe` when expecting primitive literals",
+func message(kind shared.Kind) rule.RuleMessage {
+	switch kind {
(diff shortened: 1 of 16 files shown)


internal/plugins/jest/rules/prefer_to_be/prefer_to_be.go
```diff
@@ -1,267 +1,138 @@
...
-	}
-	if arg.Kind == ast.KindPrefixUnaryExpression {
-		unary := arg.AsPrefixUnaryExpression()
-		if unary.Operator == ast.KindMinusToken {
-			arg = utils.UnwrapBasicTypeAssertions(unary.Operand)
+	var renameFix rule.RuleFix
+	if match.Expect.MatcherEntry.Node.Kind == ast.KindIdentifier {
+		// Match replaceAccessorFixer exactly: an identifier accessor remains an
+		// identifier even in `expect(x)[toEqual](1)`.
+		renameFix = rule.RuleFixReplace(ctx.SourceFile, match.Expect.MatcherEntry.Node, replacement)
```
fansenze: **[P2] Avoid renaming computed-key variables**

With `const toEqual = 'toEqual'`, this changes `expect(value)[toEqual](1)` to `expect(value)[toBe](1)`, which throws `ReferenceError` when `toBe` is undeclared. Please skip computed identifiers or emit a quoted matcher name, as the previous implementation did.

internal/plugins/rstest/rules/prefer_to_be/prefer_to_be.go
```diff
@@ -0,0 +1,141 @@
...
+			match.Expect.MatcherCall,
+		)
+		if !ok || internalUtils.HasCommentInSpan(
+			ctx.Comments.All(),
+			argumentsRange.Pos(),
+			argumentsRange.End(),
+		) {
+			return nil
+		}
+		fixes = append(fixes, rule.RuleFixRemoveRange(argumentsRange))
```
fansenze, line 94: **[P2] Remove type arguments when switching to a non-generic matcher**

`expect(null).toEqual<null>(null)` is fixed to `expect(null).toBeNull<null>()`, which fails TypeScript with TS2558 (`Expected 0 type arguments, but got 1`). Please remove the type arguments along with the value arguments. The Jest adapter has the same issue.

internal/plugins/jest/rules/prefer_to_be/prefer_to_be.go
```diff
@@ -1,267 +1,138 @@
...
+		return shared.Runtime{Parse: func(node *ast.Node) *shared.ExpectCall {
+			parsed := jestUtils.ParseJestFnCall(node, ctx)
+			if parsed == nil ||
+				parsed.Kind != jestUtils.JestFnTypeExpect ||
+				parsed.MatcherEntry == nil ||
+				hasParenthesizedOptionalChainBoundary(parsed.MemberEntries) {
+				return nil
+			}
+			matcherCall := testFramework.InvokedAccessorCall(parsed.MatcherEntry)
+			if matcherCall == nil {
```
fansenze, line 132: **[P3] Report each matcher call only once**

`expect(promise).resolves.toEqual(1).then(done)` now reports `useToBe` twice at the same location: both the matcher call and the outer `.then()` resolve to this `matcherCall`. Please require `node == matcherCall` or deduplicate by matcher call.

internal/plugins/jest/rules/prefer_to_be/prefer_to_be.go
elecmonkey: Fixed in d9a5e929a — computed keys are now quoted, so the fix emits expect(value)['toBe'](1).

internal/plugins/rstest/rules/prefer_to_be/prefer_to_be.go
elecmonkey: Fixed in d9a5e929a — type arguments are removed with the value arguments in both adapters.

internal/plugins/jest/rules/prefer_to_be/prefer_to_be.go
elecmonkey: Fixed in d9a5e929a — each matcher call is reported once.

fansenze (approved):

PR #2120, opened 2026-09-09
feat: port node/no-extraneous-require

fansenze:
## Summary

Port eslint-plugin-n v18.3.0's rule as `node/no-extraneous-require`, contributing to [the Node plugin tracking issue #475](https://github.com/web-infra-dev/rslint/issues/475). It reports installed packages loaded through `require()` or `require.resolve()` when they are missing from the applicable `package.json` dependencies.

Reuse the existing Node package, workspace, and resolution helpers. Share require-call discovery and alias propagation through `nodeutil.CollectRequireCalls`, using `ReferenceIndex` and `RefStore` for reference queries. The rule owns its dependency checks and reporting. Cover destructuring, computed and optional calls, constant module arguments, and TypeScript syntax. Include the option schema, catalog registration, and user documentation; the differences section explains resolver overrides, workspace glob forms, and compound BigInt expressions with concrete examples.

## Validation

- All 23 upstream cases in Go and JS, plus 91 Go extras for the new rule; no skipped upstream cases.
- Targeted Go tests for the rule, Node resolution helpers, and `no-extraneous-import`; catalog registration test also passed.
- Rebuilt the native binary and verified the generated schema/types. JS integration covers both extraneous rules and `require-unicode-regexp`.
- Matched the pinned upstream on 12 unchanged ESLint source files plus a positive control, in two dependency-metadata scenarios, including full diagnostic ranges and edit behavior.
- Temporary Go benchmarks covered eight workloads; the benchmark file is excluded from this PR.
- Scoped Go lint, spelling checks, and `pnpm run format:check` passed.

## Related Links

- [Upstream rule](https://github.com/eslint-community/eslint-plugin-n/blob/v18.3.0/lib/rules/no-extraneous-require.js)

## Checklist


elecmonkey (dismissed):

internal/plugins/node/rules/no_extraneous_require/no_extraneous_require.md
```diff
@@ -0,0 +1,119 @@
+# node/no-extraneous-require
```
elecmonkey:
```suggestion
# no-extraneous-require
```

elecmonkey (approved):

PR #2125, opened 2026-09-10
perf(rstest): cache call-chain candidate roots

elecmonkey:
## Motivation

Rstest call analysis checks the first identifier of each call expression before consulting its parse caches. In a nested call chain such as:

```ts
builder().step().step().step();
```

each inner call walks back through the chain again, making candidate filtering quadratic in the chain length.

Expect parsing has a similar repeated upward walk when determining whether each call is the outermost call in its callee chain. Rstest registration factory chains such as repeated `test.extend({})` also fully parse factory calls that cannot themselves be final registrations.

## Changes

- Cache expression-to-first-identifier resolution in `RstestCallAnalysis`, including negative results.
- Reuse the cache from `ParseFnCall`, `ParseExpectCall`, `IsExpectCall`, alias collection, and the Rstest `expect-expect` candidate gate.
- Avoid allocating the cache for ordinary unchained identifier calls.
- Replace repeated top-most expect-call searches with an outer-callee-call existence check.
- Reject `.extend()`, `.each()`, `.for()`, `.runIf()`, and `.skipIf()` factory invocations before full registration parsing.
- Add regression coverage for aliases, TestContext expect, `import.meta.rstest`, source-only and type-aware programs, static accessors, optional chains, and factory boundaries.

## Performance

Measurements use `--timing all --singleThreaded`, one warmup, and three interleaved runs per version. The table reports median total rule execution time for the Rstest recommended preset.

| Scenario | Size | Before | After |
| --- | ---: | ---: | ---: |
| Ordinary non-Rstest chain | 500 | 18.6 ms | 0.2 ms |
| Ordinary non-Rstest chain | 1,000 | 74.3 ms | 0.5 ms |
| Ordinary non-Rstest chain | 2,000 | 290.6 ms | 1.7 ms |
| Rstest `extend` chain | 500 | 31.3 ms | 0.4 ms |
| Rstest `extend` chain | 1,000 | 122.5 ms | 0.7 ms |
| Rstest `extend` chain | 2,000 | 472.1 ms | 1.6 ms |
| Ordinary tests | 1,000 | 2.9 ms | 2.8 ms |
| Ordinary tests | 2,000 | 5.7 ms | 5.7 ms |
| Ordinary tests | 4,000 | 12.1 ms | 12.0 ms |

The ordinary and `extend` chain benchmarks change from quadratic growth to approximately linear growth.

Seven interleaved timing runs across the six-repository fixture produced:

| Repository | Before | After |
| --- | ---: | ---: |
| rspack | 3.5 ms | 3.6 ms |
| rsbuild | 16.7 ms | 17.3 ms |
| rslib | 6.8 ms | 6.8 ms |
| rspress | 10.3 ms | 10.9 ms |
| rsdoctor | 6.9 ms | 7.3 ms |
| rstest | 50.8 ms | 50.2 ms |

All synthetic and repository runs produced identical diagnostics.

### DIFF
--- a/internal/plugins/rstest/rules/expect_expect/expect_expect.go
+++ b/internal/plugins/rstest/rules/expect_expect/expect_expect.go
@@ -39,7 +39,8 @@ var ExpectExpectRule = shared.NewRule(shared.Config{
 			}
 		}
 		return shared.Runtime{
-			IsAssertion: isAssertion,
+			IsAssertion:     isAssertion,
+			FirstIdentifier: analysis.FirstIdentifier,
 			ClassifyTest: func(node *ast.Node) shared.TestClassification {
 				parsed := analysis.ParseTestCall(node)
 				if parsed == nil {
--- a/internal/plugins/rstest/utils/call_analysis.go
+++ b/internal/plugins/rstest/utils/call_analysis.go
@@ -8,13 +8,14 @@ import (
 )
 
 type RstestCallAnalysis struct {
-	ctx         rule.RuleContext
-	candidates  map[string]rstestCandidateKind
-	fnCalls     map[*ast.Node]*ParsedRstestFnCall
-	expectCalls map[*ast.Node]*ParsedRstestExpectCall
-	isExpect    map[*ast.Node]bool
-	expectRoots map[*ast.Symbol]rstestExpectRoot
-	calls       []*ast.Node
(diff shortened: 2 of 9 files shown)


elecmonkey:
Closing because this introduce substantial complexity for a pattern rarely seen in real-world code.

PR #2127, opened 2026-09-10
feat(rstest): add prefer-to-contain rule

elecmonkey:
## Summary

Add `rstest/prefer-to-contain` to replace equality assertions over `includes()` boolean results with clearer containment assertions.

## Credits

- Port from [`@vitest/eslint-plugin@1.6.27`'s `prefer-to-contain` documentation](https://github.com/vitest-dev/eslint-plugin-vitest/blob/v1.6.27/docs/rules/prefer-to-contain.md), [source](https://github.com/vitest-dev/eslint-plugin-vitest/blob/v1.6.27/src/rules/prefer-to-contain.ts), and tests.

- **Explicit `NaN` items are excluded.** `Array.prototype.includes()` uses SameValueZero and matches `NaN`, while Rstest's current `toContain()` array path does not. Rewriting would reverse a passing assertion.

- **String literal receivers with regexp items are excluded.** Native `String.prototype.includes()` rejects regular expressions, while matcher behavior is not the same assertion shape.

- **Autofixes are withheld when moving the item may change evaluation order.** For example, function calls, property reads, custom `expect()` message arguments, and comments inside edited ranges still report but do not fix.

- **Assertion-factory and matcher type arguments are removed for Rstest autofixes.** These type arguments describe the original boolean assertion and may become invalid after the received value and matcher argument are replaced.

- **Accessor spelling is preserved.** Quoted and template-literal matcher accessors retain their delimiters instead of being normalized.

- **Type arguments containing comments are not fixed.** The rule still reports, but avoids deleting comments from either the assertion factory or matcher type arguments.

- **Chai `.deep` chains are excluded.** The flag changes `toContain()` to use deep equality, which does not match the identity comparison performed by `includes()`.

- **Assertion chains that continue after the equality matcher are excluded.** Replacing the received boolean would change the value observed by later matchers or chained members.

## Related Links

Related #935

## Checklist

- [x] Tests updated.
- [x] Documentation updated.

### DIFF
--- a/internal/plugins/rstest/rules/prefer_to_contain/prefer_to_contain.go
+++ b/internal/plugins/rstest/rules/prefer_to_contain/prefer_to_contain.go
@@ -67,7 +67,7 @@ func isSafeToMove(node *ast.Node) bool {
 		return true
 	case ast.KindPrefixUnaryExpression:
 		unary := node.AsPrefixUnaryExpression()
-		return unary != nil && unary.Operator != ast.KindPlusPlusToken && unary.Operator != ast.KindMinusMinusToken &&
+		return unary != nil && unary.Operator == ast.KindExclamationToken &&
 			isSafeToMove(unary.Operand)
 	case ast.KindArrayLiteralExpression:
 		array := node.AsArrayLiteralExpression()
--- a/internal/plugins/rstest/rules/prefer_to_contain/prefer_to_contain.go
+++ b/internal/plugins/rstest/rules/prefer_to_contain/prefer_to_contain.go
@@ -9,14 +9,80 @@ import (
 	shared "github.com/web-infra-dev/rslint/internal/utils/test_framework/rules/prefer_to_contain"
 )
 
-func isGlobalIdentifier(ctx rule.RuleContext, node *ast.Node, name string) bool {
+func isUnshadowedRuntimeGlobal(ctx rule.RuleContext, node *ast.Node, name string) bool {
 	if node == nil || node.Kind != ast.KindIdentifier || node.AsIdentifier().Text != name {
 		return false
 	}
 	if ctx.Refs != nil {
-		return ctx.Refs.IsGlobalNameReference(node, name, ast.SymbolFlagsValue|ast.SymbolFlagsAlias)
+		return ctx.Refs.ResolveInFileWithMeaning(node, ast.SymbolFlagsValue|ast.SymbolFlagsAlias) == nil &&
+			!ctx.Refs.HasImplicitWrapperBinding(name)
 	}
 	return !utils.IsShadowed(node, name)
+}
+
+type nanGlobalWritesCacheKey struct{}
+
+func nanGlobalWrites(ctx rule.RuleContext) map[string]bool {
+	return rule.CachedByFile(ctx, nanGlobalWritesCacheKey{}, func() map[string]bool {
+		writes := map[string]bool{}
+		if ctx.SourceFile == nil {
+			return writes
+		}
+		var visit func(*ast.Node)
+		visit = func(node *ast.Node) {
+			if node == nil {
+				return
+			}
+			if node.Kind == ast.KindIdentifier {
+				name := node.AsIdentifier().Text
+				if name == "Number" || name == "globalThis" {
+					affected := nanGlobalsWrittenFrom(node)
+					if len(affected) > 0 && isUnshadowedRuntimeGlobal(ctx, node, name) {
+						for _, globalName := range affected {
+							writes[globalName] = true
+						}
+					}
+				}
+			}
+			node.ForEachChild(func(child *ast.Node) bool {
+				visit(child)
+				return false
+			})
+		}
+		visit(ctx.SourceFile.AsNode())
+		return writes
+	})
+}
+
+func nanGlobalsWrittenFrom(node *ast.Node) []string {
+	if utils.IsWriteReference(node) {
+		return []string{node.AsIdentifier().Text}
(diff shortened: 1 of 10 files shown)


internal/plugins/rstest/rules/prefer_to_contain/prefer_to_contain.go
```diff
@@ -0,0 +1,173 @@
...
+		}
+		fixes := shared.BuildFixes(ctx, matched)
+		if len(fixes) == 0 {
+			return nil
+		}
+		if typeArguments, ok := testFramework.CallTypeArgumentListRange(ctx.SourceFile, matched.Expect.MatcherCall); ok {
+			if utils.HasCommentInSpan(ctx.Comments.All(), typeArguments.Pos(), typeArguments.End()) {
+				return nil
+			}
+			fixes = append(fixes, rule.RuleFixRemoveRange(typeArguments))
```
fansenze: **[P2] Remove stale type arguments from `expect()` too**

The fix removes matcher type arguments, but leaves the type argument on `expect()` even though its argument changes from a boolean to an array:

```ts
// Before: type-checks
expect<boolean>([1].includes(1)).toBe(true);

// Autofix: TS2345 — number[] is not assignable to boolean
expect<boolean>([1]).toContain(1);
```

`expect.soft<boolean>(...)` has the same problem. Please also handle `matched.Expect.HeadCall`'s type arguments, preserving comments or withholding the fix when they cannot be preserved.

internal/plugins/rstest/rules/prefer_to_contain/prefer_to_contain.go
```diff
@@ -0,0 +1,173 @@
...
+				parsed.Matchers[0].Kind != rstestUtils.RstestExpectMatcherCall ||
+				testFramework.IsComputedIdentifierAccessor(parsed.MatcherEntry.Node) {
+				return nil
+			}
+
+			for _, modifier := range parsed.Modifiers {
+				if modifier == "resolves" || modifier == "rejects" {
+					return nil
+				}
+			}
```
fansenze, line 158: **[P2] Skip Chai flags that change containment semantics**

`.deep` is a Chai flag, so it is absent from `parsed.Modifiers` and passes this guard. The fix preserves it, but Rstest's `toContain()` then uses deep equality instead of the identity comparison performed by `includes()`:

```js
// Before: passes — these are different object references
expect([{ a: 1 }].includes({ a: 1 })).deep.toBe(false);

// Autofix: fails — the objects are deeply equal
expect([{ a: 1 }]).deep.not.toContain({ a: 1 });
```

Please exclude assertion chains with flags such as `.deep` before recommending this transformation.

internal/plugins/rstest/rules/prefer_to_contain/prefer_to_contain.go
```diff
@@ -0,0 +1,173 @@
...
+		analysis := rstestUtils.GetRstestCallAnalysis(ctx)
+		return shared.Runtime{Parse: func(node *ast.Node) *shared.ExpectCall {
+			parsed := analysis.ParseExpectCall(node)
+			if parsed == nil ||
+				parsed.Reason != rstestUtils.RstestExpectParseReasonNone ||
+				parsed.Head == nil ||
+				(parsed.Entry != rstestUtils.RstestExpectEntryCall && parsed.Entry != rstestUtils.RstestExpectEntrySoft) ||
+				parsed.MatcherEntry == nil ||
+				len(parsed.Matchers) == 0 ||
+				parsed.Matchers[0].Kind != rstestUtils.RstestExpectMatcherCall ||
```
fansenze: **[P2] Leave multi-matcher assertion chains unchanged**

This accepts chains with multiple matchers but only inspects the first. Replacing the value passed to `expect()` also changes what later matchers receive:

```js
// Before: both assertions pass in Rstest
expect([1].includes(1)).toBe(true).toBe(true);

// Autofix: the second assertion receives [1] and fails
expect([1]).toContain(1).toBe(true);
```

Please require a single matcher before applying this transformation, including any trailing Chai assertion getters counted in `parsed.Matchers`.

fansenze (commented):
Found three reproducible autofix issues below. The existing targeted Rstest/Jest rule tests and rule registration check pass; additional cases expose these gaps. The runtime examples were verified with Rstest 0.11.11, and the generic-call examples with TypeScript 6.0.3.

internal/plugins/rstest/rules/prefer_to_contain/prefer_to_contain.go
elecmonkey: Fixed in 9811092. The autofix now removes type arguments from both `expect()`/`expect.soft()` and the matcher. If either list contains comments, the entire fix is withheld.

internal/plugins/rstest/rules/prefer_to_contain/prefer_to_contain.go
elecmonkey: Fixed in 9811092. Assertions using `.deep` are now excluded because it changes `toContain()` to use deep equality. Other unrelated Chai flags remain eligible.

internal/plugins/rstest/rules/prefer_to_contain/prefer_to_contain.go
elecmonkey: Fixed in 9811092. The rule now requires a single matcher that terminates the assertion chain, preventing later matchers or members from observing the rewritten subject.

fansenze (approved):

PR #2128, opened 2026-09-10
feat(unicorn): add no-unreadable-iife rule

HashemKhalifa:
## Motivation

Add native support for `unicorn/no-unreadable-iife`, part of #1309. Immediately invoked arrows with parenthesized expression bodies can obscure where the function body begins.

## Changes

Port detection, messages, source ranges and editor suggestions from `eslint-plugin-unicorn` v74.0.0. The suggestion replaces the parenthesized body with a return-containing block and is withheld when removing the parentheses would discard comments. The rule has no options and does not apply automatic fixes.

Register the native rule, enable it in the recommended Unicorn preset, and add rule documentation. Go tests cover the upstream tests and documentation examples, plus nested parentheses, comment boundaries, Unicode, CRLF, optional calls, JSDoc casts and edit-demand behavior. The JS mirror checks diagnostics and suggested output through the native service.

Validation:

- Owning Go rule package: passed.
- Native rule catalog registration test: passed.
- Native binary, rule schema generation and JavaScript build: passed.
- Scoped JavaScript integration mirror: passed (upstream tests and documentation examples).
- Differential comparison against pinned Unicorn v74.0.0: passed on 51 JavaScript inputs, including exact diagnostic ranges, suggestions, re-linted outputs and unchanged automatic-fix output.
- Eight result/evaluation-order comparisons: passed.
- Formatting and source spell-check: passed.

Related to #1309.

Follow-up validation: restored the existing third-party notice and removed the rule-local license as requested. All seven globals tests, including the published-notice assertion, pass.

### DIFF
--- a/internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
+++ b/internal/plugins/rstest/rules/prefer_called_exactly_once_with/prefer_called_exactly_once_with.go
@@ -161,12 +161,11 @@ type chainAssertion struct {
 // interest, which merges with a matching statement elsewhere in the block, or
 // one of each role, which already states both halves and merges with itself.
 type mergeCandidate struct {
-	hits      []chainAssertion
-	statement *ast.Node
-	arguments []*ast.Node
-	// position is the trimmed start offset of statement, used to order the
-	// pair and to bound the mock-reset search between them.
-	position int
+	hits           []chainAssertion
+	statement      *ast.Node
+	statementIndex int
+	arguments      []*ast.Node
+	position       int
 	// pairKey groups the assertions that may merge; see pairKey().
 	pairKey string
 	// fixable is false when the chain asserts more than the rule understands,
@@ -784,20 +783,19 @@ func onlyInertStatementsBetween(
 	statements []*ast.Node,
 	first, second *mergeCandidate,
 ) bool {
-	minPosition, maxPosition := first.position, second.position
-	if minPosition > maxPosition {
-		minPosition, maxPosition = maxPosition, minPosition
+	minIndex, maxIndex := first.statementIndex, second.statementIndex
+	if minIndex > maxIndex {
+		minIndex, maxIndex = maxIndex, minIndex
+	}
+	if maxIndex-minIndex == 1 {
+		return true
 	}
 	scanner := &betweenScanner{ctx: ctx, analysis: analysis, assertionNames: map[string]bool{}}
 	collectIdentifierNames(first.statement, scanner.assertionNames)
 	collectIdentifierNames(second.statement, scanner.assertionNames)
 
-	for _, statement := range statements {
+	for _, statement := range statements[minIndex+1 : maxIndex] {
 		if statement == nil {
-			continue
-		}
-		position := internalUtils.TrimNodeTextRange(ctx.SourceFile, statement).Pos()
-		if position <= minPosition || position >= maxPosition {
 			continue
 		}
 		if !scanner.isInertStatement(statement) && !scanner.isInertDeclaration(statement) {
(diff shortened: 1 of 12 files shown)


elecmonkey (commented):
It looks like you're already familiar with Rstack—thank you for your contribution to Rslint! The changes to `THIRD-PARTY-NOTICES.md` are causing some tests to fail, and as a repository convention, we do not include a separate `LICENSE` file in each rule directory. Could you please revert the changes to the legal files for now? We will reevaluate this separately and address it in a follow-up PR.

HashemKhalifa:
@elecmonkey Thanks for the clarification! 
Reverted the notice changes and removed the rule-local LICENSE. The previously failing test now passes locally.

elecmonkey (approved):
Thanks! No issues found here. Sorry for missing the upstream license, we’ll add them next release.

HashemKhalifa (commented):
## Code-quality review

Reviewed all eight changed-file diffs at `99f4d6ea601fae57d7f321b94ef74d0c4fa1d06e` against `main` at `4add5b1a73ed285fcfabf625bef3f5078e0e941b`, the existing maintainer discussion, repository guidance, relevant shared AST/comment helpers, and the pinned Unicorn v74.0.0 source/tests/documentation. Also verified the follow-up commit below and that the PR head is now `a38c48ed9487908c45d925b31cbe7065c95bbac5`.

**Assessment: no substantive correctness or structural blocker identified in this static review.**

### P3 — dangling license reference in the JS mirror (fixed)

`packages/rslint-test-tools/tests/eslint-plugin-unicorn/rules/no-unreadable-iife.test.ts:2` still pointed readers to `internal/plugins/unicorn/rules/no_unreadable_iife/LICENSE`, which was removed in response to the maintainer's request. The source-attribution comment therefore referenced a nonexistent file.

Fixed in [a38c48e](https://github.com/HashemKhalifa/rslint/commit/a38c48ed9487908c45d925b31cbe7065c95bbac5): point to the verified [upstream license at v74.0.0](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v74.0.0/license). This is a one-line comment-only change. It does not restore a rule-local LICENSE, change THIRD-PARTY-NOTICES.md, or reopen the deferred legal-file work.

### Structural assessment

- The 44-line native implementation is already direct and appropriately scoped. It reuses `ESTreeCallCallee`, `ESTreeRuntimeExpression`, `TrimNodeTextRange`, `CommentsInSpan`, and the deferred-suggestion API rather than introducing its own AST normalization, comment scanner, or edit-demand machinery. Authored TypeScript assertions remain intact at the shared normalization boundary.
- No behavior-preserving restructuring was substantiated that would remove meaningful complexity. Extracting the single-use suggestion body into another wrapper would add indirection without simplifying this rule.
- The custom JS integration assertions are justified here. I inspected the existing Unicorn `RuleTester`: although its types accept ranges, suggestions, and output expectations, its implementation does not assert those fields. Replacing the direct native-service checks with that runner would weaken this PR's coverage. A shared-runner upgrade would be separate infrastructure work, not a prerequisite for this bounded port.
- The upstream test inputs and documentation examples are represented in the Go upstream suite and JS mirror. The separate Go extras cover nested parentheses, comment boundaries, optional calls, JSDoc wrappers, Unicode/CRLF inputs, and suggestion-only edit demand. Registration, recommended-preset enablement, and test selection are scoped additions.

### Verification and limits

For the comment-only fix, I reconstructed the file and verified its original Git blob hash against the connector response, checked TypeScript syntax with TypeScript 5.8.3, and verified that before/after comment-stripped emitted JavaScript is identical. The returned content hash also matches the locally validated replacement. GitHub's compare confirms exactly one changed line in one file. No runtime regression test was added for a non-executable comment change.

I did **not** independently rerun the owning Go package, native JS integration, the reported 51-input differential comparison, or the repository pnpm formatting/spell-check commands: this environment could not clone the repository or obtain its dependencies. The PR description's earlier passing results remain author-reported evidence, not results reproduced by this review. This is a COMMENT review, not a claim of a newly verified native test pass.

chenjiahan (approved):

PR #2129, opened 2026-09-10
feat: port rule react/no-unused-prop-types

swwind:
## Summary

Port the `react/no-unused-prop-types` rule from eslint-plugin-react to rslint. The implementation covers runtime `propTypes`, nested `shape`/`exact` validators, aliases and destructuring, class/createReactClass/stateless components, TypeScript prop shapes, options, and React rule registration.

## Related Links

- ESLint rule: https://github.com/jsx-eslint/eslint-plugin-react/blob/v7.37.5/docs/rules/no-unused-prop-types.md
- Source code: https://github.com/jsx-eslint/eslint-plugin-react/blob/v7.37.5/lib/rules/no-unused-prop-types.js

## Verification

- Full supported upstream suite: 243 valid + 118 invalid cases.
- Four upstream parser-matrix cases are retained with explicit `Skip` because the same source has opposite expectations across parser variants.
- Go rule and React catalog tests pass.
- Typecheck, lint, spellcheck, formatting, and changed-package Go lint pass.

## Checklist

- [x] Tests updated.
- [x] Documentation updated.


internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
```diff
@@ -0,0 +1,1002 @@
...
+			if call.Arguments != nil && len(call.Arguments.Nodes) > 0 {
+				arg := unwrap(call.Arguments.Nodes[0])
+				if arg != nil && arg.Kind == ast.KindObjectLiteralExpression {
+					c := &component{node: arg, name: componentName(arg), used: map[string]bool{}, aliases: map[string][]string{}}
+					components = append(components, c)
+					nested[arg] = true
+				}
+			}
+		}
+		if isFunctionLike(node) && (reactutil.IsStatelessReactComponentWithChecker(node, pragma, ctx.TypeChecker) || (hasTypedFunctionParameters(node) && !hasKnownComponentAncestor(node, nested))) {
```
elecmonkey: [P1] Restrict the typed-parameter fallback to React components

The `hasTypedFunctionParameters` fallback classifies every top-level typed function as a component even when it cannot render JSX. For `function helper(props: { used: string; unused: string }) { return props.used.length; }`, eslint-plugin-react@7.37.5 emits nothing, but Rslint reports `unused`; ordinary typed helpers and event handlers therefore become false positives whenever this rule is enabled. Please keep component detection separate from extracting TypeScript prop declarations.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
```diff
@@ -0,0 +1,1002 @@
...
+				fullName = prefix + "." + name
+			}
+			result = append(result, newProp(name, fullName, reportNode))
+		}
+		return result
+	}
+	return nil
+}
+
+func collectTypeAliases(root *ast.Node) map[string][]*ast.Node {
```
elecmonkey, line 546: [P2] Resolve type references in their lexical scope

`collectTypeAliases` flattens every nested type declaration into one file-wide name map. If two functions each declare a local `Props`, resolving either component merges both declarations; eslint-plugin-react emits nothing, while Rslint reports members belonging to the other scope. Please resolve the referenced binding or maintain a lexical scope chain instead of grouping declarations by textual name alone.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
```diff
@@ -0,0 +1,1002 @@
...
+		return nil
+	}
+	name := propertyName(pa.Name())
+	if name == "" {
+		return nil
+	}
+	return append(parts, name)
+}
+
+func findComponent(components []*component, name string) *component {
```
elecmonkey: [P2] Resolve propTypes assignments by binding

`findComponent` returns the first component with the requested textual name without considering lexical scope. With two block-scoped `Foo` components, the second `Foo.propTypes` assignment can be attached to the first component; a differential case where each component reads only the other component's prop makes eslint-plugin-react report both declarations while Rslint reports only one. Please associate assignments with the actual declaration binding rather than the component name alone.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
```diff
@@ -0,0 +1,1002 @@
...
+					if callee != nil && callee.Kind == ast.KindPropertyAccessExpression {
+						method := propertyName(callee.AsPropertyAccessExpression().Name())
+						if method == "map" || method == "forEach" || method == "filter" {
+							if path, ok := pathFrom(callee.AsPropertyAccessExpression().Expression, c.aliases); ok {
+								prefix = append(path, "*")
+							}
+						}
+					}
+				}
+				bindingPath(name, prefix, c)
```
elecmonkey: [P2] Limit destructured callback parameters to known prop sources

This marks every object-binding parameter encountered inside a component as a root props path. For `function Foo(props) { [1].map(({ used }) => used); return <div />; } Foo.propTypes = { used: PropTypes.string };`, eslint-plugin-react reports `used` as unused, but Rslint emits nothing because the unrelated callback parameter is counted as component-prop usage. Only bind component parameters, lifecycle/setState parameters, or callbacks whose receiver is already known to originate from props.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
elecmonkey: [P2] Honor configured component wrapper functions during discovery

This uses `IsStatelessReactComponentWithChecker`, which recognizes only the default wrappers. With `settings.componentWrapperFunctions: ['observer']` and `const Foo = observer((props) => <div />); Foo.propTypes = { unused: PropTypes.string };`, eslint-plugin-react reports `unused`, while Rslint does not discover the component. Please pass `GetComponentWrapperFunctions(ctx.Settings, pragma)` to `IsStatelessReactComponentWithWrappers` so configured wrappers participate in discovery.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
```diff
@@ -0,0 +1,1002 @@
...
+			} else if name.Kind == ast.KindIdentifier {
+				text := name.AsIdentifier().Text
+				if isSetStateCallback && parameterIndex == 1 || text == "props" || strings.HasSuffix(text, "Props") {
```
elecmonkey: [P2] Gate lifecycle prop aliases by the configured React version

Parameters named `nextProps` or `prevProps` are accepted here without checking either the containing lifecycle method or `settings.react.version`. For `UNSAFE_componentWillReceiveProps(nextProps) { return nextProps.value; }` with React 16.2 configured, eslint-plugin-react reports `value` as unused while Rslint treats it as used. This also explains the four currently skipped upstream cases, so they are not an unavoidable parser-matrix difference; please apply upstream's lifecycle and React-version gates.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
```diff
@@ -0,0 +1,1002 @@
...
+			method := ""
+			if callee != nil && callee.Kind == ast.KindPropertyAccessExpression {
+				pa := callee.AsPropertyAccessExpression()
+				method = propertyName(pa.Name())
+				if object := unwrap(pa.Expression); object != nil && object.Kind == ast.KindIdentifier && slices.Contains(customValidators, object.AsIdentifier().Text) {
+					method = "custom"
+				}
+			}
+			switch method {
+			case "shape", "exact", "custom":
```
elecmonkey: [P2] Keep configured custom validators opaque

eslint-plugin-react treats calls on a configured custom-validator namespace as opaque and does not interpret their arguments as nested prop declarations. With `{ customValidators: ['CustomValidator'], skipShapeProps: false }`, `outer: CustomValidator.shape({ inner: CustomValidator.string })`, and only `props.outer` used, upstream emits nothing while Rslint reports `outer.inner`. Please keep the outer prop as a leaf when the callee object matches `customValidators`.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
```diff
@@ -0,0 +1,1002 @@
...
+					} else if typ != nil && typ.Kind == ast.KindTypeReference {
+						result = mergeProps(result, typeProps(typ, aliases, seen, prefix))
+					}
+				}
+			}
+		}
+		if decl.Members == nil {
+			return result
+		}
+		for _, member := range decl.Members.Nodes {
```
elecmonkey, line 380: [P2] Collect TypeScript method signatures as declared props

This traversal accepts only `PropertySignature`, so method-shaped props are silently omitted. For `interface Props { used: string; onClick(): void }` and a component that reads only `used`, eslint-plugin-react reports `onClick` as unused while Rslint emits nothing. Please handle `MethodSignature` here and in the equivalent type-literal branch, as the existing `react/prop-types` implementation already does.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
```diff
@@ -0,0 +1,1002 @@
...
+			switch method {
+			case "shape", "exact", "custom":
+				current.shape = method != "custom"
+				if call.Arguments != nil && len(call.Arguments.Nodes) > 0 {
+					current.children = propMap(call.Arguments.Nodes[0], customValidators, fullName)
+				}
+			case "arrayOf", "objectOf":
+				if call.Arguments != nil && len(call.Arguments.Nodes) > 0 {
+					current.children = validatorChildren(call.Arguments.Nodes[0], customValidators, fullName+".*")
+					child := unwrap(call.Arguments.Nodes[0])
```
elecmonkey, line 237: [P3] Report synthetic array children at the validator argument

For `prop2: PropTypes.arrayOf(Comp1.propTypes.prop1)`, eslint-plugin-react reports the `prop2.*` diagnostic at `Comp1.propTypes.prop1`, whereas Rslint highlights the outer `prop2` key because the synthetic child reuses `reportNode`. Please retain the `arrayOf`/`objectOf` argument as the child's report node; the three migrated cases covering this branch currently omit location assertions.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
```diff
@@ -0,0 +1,1002 @@
...
+				if len(parts) >= 2 && (parts[len(parts)-1] == "propTypes" || (len(parts) >= 3 && (parts[1] == "propTypes" || parts[len(parts)-2] == "propTypes"))) {
+					c := findComponent(components, parts[0])
+					if c != nil {
+						if len(parts) == 2 || (len(parts) == 3 && parts[1] == "prototype") {
+							appendDeclared(c, declaredProps(bin.Right, customValidators, wrappers))
+						} else {
+							var nestedProps []*prop
+							for i := 2; i < len(parts); i++ {
+								name := strings.Join(parts[2:i+1], ".")
+								nestedProps = append(nestedProps, newProp(parts[i], name, leftmostProperty(bin.Left)))
```
elecmonkey: [P3] Preserve the assignment range for nested prop declarations

For `Foo.propTypes.a.b.c = PropTypes.number`, eslint-plugin-react reports the `a.b.c` diagnostic over the complete assignment expression, but `leftmostProperty` stores only the terminal `c` node, so Rslint highlights that identifier alone. Please retain the assignment node for the synthesized terminal declaration; the migrated test currently leaves its location assertions unset and therefore does not catch this mismatch.

elecmonkey (commented):
Let’s keep our `react/` journey going! Your new avatar looks great.

elecmonkey:
golangci-lint failed~~ not flaky

swwind:
@elecmonkey fixed meow

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
```diff
@@ -0,0 +1,1188 @@
...
+				}
+				{
+					// Function parameters are the root aliases for SFCs.
+					if isFunctionLike(c.node) {
+						for _, parameter := range reactutil.FunctionParameters(c.node) {
+							if parameter.Kind != ast.KindParameter {
+								continue
+							}
+							decl := parameter.AsParameterDeclaration()
+							bindingPath(decl.Name(), nil, c)
```
elecmonkey: [P2] Only treat the first component parameter as props

The end-of-file pass calls `bindingPath` and `typeProps` for every function-component parameter, so later parameters are also treated as props. For `function Foo(props, options) { return <div>{options.used}</div>; } Foo.propTypes = { used: PropTypes.string };`, upstream reports `used` as unused while Rslint emits nothing; conversely, an inline type on a second `context` parameter can create Rslint-only prop diagnostics. Please limit root prop usage and declarations to the component's first parameter.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
```diff
@@ -0,0 +1,1188 @@
...
+				continue
+			}
+			setStateCallback := isSetStateCallback(node)
+			if setStateCallback && parameterIndex == 0 {
+				continue
+			}
+			isRootComponentParameter := node == c.node && parameterIndex == 0
+			isLifecycleParameter := parameterIndex == 0 && isLifecycleFunction(node, checkAsyncSafe)
+			isValidatorParameter := parameterIndex == 0 && isPropTypesValidatorCallback(node)
+			bindsProps := isRootComponentParameter || isLifecycleParameter || isValidatorParameter || (setStateCallback && parameterIndex == 1)
```
elecmonkey: [P2] Preserve upstream nested `props` parameter handling

The new `bindsProps` gate drops identifier parameters named `props` in nested helpers. For `function Foo() { function helper(props) { return props.unused; } return <div />; } Foo.propTypes = { unused: PropTypes.string };`, eslint-plugin-react@7.37.5 emits nothing, but Rslint reports `unused`; the same mismatch occurs for nested object and class methods. Upstream creates a scoped props alias for such function parameters, so the callback fix needs to reject unrelated destructuring without removing this identifier behavior.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
```diff
@@ -0,0 +1,1188 @@
...
+}
+
+func lifecycleFunctionName(node *ast.Node) string {
+	if node == nil {
+		return ""
+	}
+	if node.Kind == ast.KindConstructor {
+		return "constructor"
+	}
+	if node.Kind == ast.KindMethodDeclaration || node.Kind == ast.KindGetAccessor || node.Kind == ast.KindSetAccessor {
```
elecmonkey, line 1077: [P2] Require lifecycle methods to belong to the component

`isLifecycleFunction` classifies any method with a lifecycle name without checking that it is a method of the React component being analyzed. For `function Foo() { const helper = { componentDidUpdate(nextProps) { return nextProps.used; } }; return <div />; } Foo.propTypes = { used: PropTypes.string };`, upstream reports `used` as unused while Rslint treats the unrelated object's `nextProps.used` as component-prop usage.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
```diff
@@ -0,0 +1,1188 @@
...
+		return propertyName(node.Name())
+	}
+	if node.Kind != ast.KindArrowFunction && node.Kind != ast.KindFunctionExpression {
+		return ""
+	}
+	parent := node.Parent
+	for parent != nil && isExpressionWrapper(parent) {
+		parent = parent.Parent
+	}
+	if parent != nil && parent.Kind == ast.KindPropertyAssignment && parent.AsPropertyAssignment().Initializer == node {
```
elecmonkey: [P2] Recognize lifecycle functions stored in class fields

`lifecycleFunctionName` recognizes object-literal `PropertyAssignment` initializers but not class `PropertyDeclaration` initializers. Consequently `componentDidUpdate = (nextProps) => nextProps.used` and the React 16.3 form of `UNSAFE_componentWillReceiveProps = ...` count as usage upstream but make Rslint report `used` as unused; function-expression and destructured-parameter forms fail the same way.

elecmonkey (commented):
Round 2

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
```diff
@@ -0,0 +1,1188 @@
...
+	}
+	return false
+}
+
+func isSetStateCallback(node *ast.Node) bool {
+	if node == nil || node.Parent == nil || node.Parent.Kind != ast.KindCallExpression {
+		return false
+	}
+	callee := unwrap(node.Parent.AsCallExpression().Expression)
+	return callee != nil && callee.Kind == ast.KindPropertyAccessExpression && propertyName(callee.AsPropertyAccessExpression().Name()) == "setState"
```
elecmonkey, line 1135: [P2] Restrict setState props handling to the updater argument

`isSetStateCallback` checks only the callee name, so it also treats the completion callback in `this.setState({}, (state, props) => props.used)` as an updater and maps its second parameter to component props. eslint-plugin-react reports `used` as unused for this input, while Rslint emits nothing. Please additionally require that the function is the call's first argument, as upstream's `isSetStateUpdater` does.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
```diff
@@ -0,0 +1,1188 @@
...
+			if name == nil {
+				continue
+			}
+			setStateCallback := isSetStateCallback(node)
+			if setStateCallback && parameterIndex == 0 {
+				continue
+			}
+			isRootComponentParameter := node == c.node && parameterIndex == 0
+			isLifecycleParameter := parameterIndex == 0 && isLifecycleFunction(node, checkAsyncSafe)
+			isValidatorParameter := parameterIndex == 0 && isPropTypesValidatorCallback(node)
```
elecmonkey, line 1197: [P2] Do not alias identifier parameters of propTypes validators

The new validator branch treats an identifier parameter as the component's root props object. With `static propTypes = { a: PropTypes.string, b: (props) => props.b }` and only `this.props.a` used, upstream still reports `b` as unused, whereas Rslint suppresses it; upstream only extracts usages from destructured validator parameters in this path. Please avoid creating a root alias for identifier parameters of validator callbacks.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
```diff
@@ -0,0 +1,1188 @@
...
+		}
+		var key, value *ast.Node
+		switch member.Kind {
+		case ast.KindPropertyAssignment:
+			key = member.AsPropertyAssignment().Name()
+			value = member.AsPropertyAssignment().Initializer
+		case ast.KindShorthandPropertyAssignment:
+			key = member.AsShorthandPropertyAssignment().Name()
+			value = key
+		default:
```
elecmonkey, line 197: [P2] Collect method-style runtime propTypes declarations

`propMap` ignores object-literal methods because it only accepts property and shorthand assignments. For `function Foo() { return <div />; } Foo.propTypes = { unused() {} };`, eslint-plugin-react reports `unused` as unused, while Rslint never records the declaration and emits nothing. Method declarations need to contribute their key and function value just like ESTree's method-valued `Property` nodes.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
```diff
@@ -0,0 +1,1188 @@
...
+		return nil
+	}
+	if node.Kind == ast.KindPropertyAccessExpression {
+		return propertyAccessRoot(node.AsPropertyAccessExpression().Expression)
+	}
+	return node
+}
+
+func assignmentComponent(root *ast.Node, components []*component, resolve func(*ast.Node) *ast.Symbol) *component {
+	if root == nil || root.Kind != ast.KindIdentifier {
```
elecmonkey: [P2] Associate member-path components with their propTypes

`assignmentComponent` only accepts an identifier root, so it cannot associate a component stored as an object property. For `const obj = { Foo: function(props) { return <div />; } }; obj.Foo.propTypes = { unused: PropTypes.string };`, eslint-plugin-react reports `unused`, while Rslint emits nothing even though it discovers the `Foo` property function as a component. Please retain and resolve the complete component member path instead of rejecting every non-identifier component target.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
```diff
@@ -0,0 +1,1188 @@
...
+	if body.Kind == ast.KindObjectLiteralExpression {
+		var result []*prop
+		for _, member := range body.AsObjectLiteralExpression().Properties.Nodes {
+			if member != nil && member.Kind == ast.KindSpreadAssignment {
+				spread := unwrap(member.AsSpreadAssignment().Expression)
+				if spread != nil && spread.Kind == ast.KindCallExpression {
+					call := spread.AsCallExpression()
+					callee := unwrap(call.Expression)
+					if callee != nil && callee.Kind == ast.KindIdentifier && callee.AsIdentifier().Text == "bindActionCreators" && call.Arguments != nil && len(call.Arguments.Nodes) > 0 {
+						result = append(result, typePropsFromFunctionBody(call.Arguments.Nodes[0], aliases, seen, prefix)...)
```
elecmonkey: [P2] Read bindActionCreators props from its type arguments

For a spread `bindActionCreators<{ typedOnly: () => void }>({ runtimeOnly: fn }, dispatch)`, upstream derives `ReturnType` props from the explicit type argument and reports `typedOnly`, but this code recurses into the first runtime argument and makes Rslint report `runtimeOnly` instead. The existing migrated case happens to use matching names, which hides both this semantic mismatch and the remaining source-location mismatch.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
```diff
@@ -0,0 +1,1267 @@
...
+						}
+					}
+				}
+				{
+					// Function parameters are the root aliases for SFCs.
+					if isFunctionLike(c.node) {
+						parameters := reactutil.FunctionParameters(c.node)
+						if len(parameters) > 0 && parameters[0].Kind == ast.KindParameter {
+							decl := parameters[0].AsParameterDeclaration()
+							bindingPath(decl.Name(), nil, c)
```
elecmonkey: [P2] Preserve upstream identifier rules for component props

The first function-component parameter is always registered as a props alias, regardless of its name. For `function Foo(p) { return <div>{p.used}</div>; } Foo.propTypes = { used: PropTypes.string };`, eslint-plugin-react@7.37.5 reports `used` as unused, while Rslint emits nothing; typed parameters also produce the inverse mismatch because Rslint treats arbitrary identifiers as prop usage. Only register identifier parameters under the same conditions as upstream.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
```diff
@@ -0,0 +1,1267 @@
...
+			}
+			typ := parent.AsVariableDeclaration().Type
+			if typ.Kind == ast.KindTypeReference {
+				args := typ.AsTypeReferenceNode().TypeArguments
+				if args != nil && len(args.Nodes) > 0 {
+					return args.Nodes[0]
+				}
+			}
+			return nil
+		default:
```
elecmonkey, line 679: [P2] Extract props from generic forwardRef wrappers

`functionComponentType` stops when it reaches the enclosing `forwardRef` call, so props supplied only through `React.forwardRef<HTMLDivElement, Props>` are never declared. For `type Props = { used: string; unused: string }; const Demo = React.forwardRef<HTMLDivElement, Props>(({ used }, ref) => <div>{used}</div>);`, eslint-plugin-react@7.37.5 reports `unused`, while Rslint emits nothing. This also leaves the omitted upstream `React.memo(React.forwardRef(...))` cases unsupported.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
```diff
@@ -0,0 +1,1267 @@
...
+				argument := args.Nodes[0]
+				if argument.Kind == ast.KindFunctionType {
+					return typeProps(argument.AsFunctionTypeNode().Type, aliases, seen, prefix)
+				}
+				if argument.Kind == ast.KindTypeQuery {
+					query := argument.AsTypeQueryNode()
+					if query.ExprName != nil && query.ExprName.Kind == ast.KindIdentifier {
+						if initializer := reactutil.ResolveIdentifierInitializer(query.ExprName.AsNode(), nil); initializer != nil {
+							initializer = unwrap(initializer)
+							if initializer.Kind == ast.KindArrowFunction {
```
elecmonkey: [P2] Handle block-bodied ReturnType producers

The `ReturnType<typeof fn>` branch accepts only arrow functions and forwards their body directly to `typePropsFromFunctionBody`, which handles only object-literal bodies. For `const makeProps = () => { return { used: '', unused: '' }; }; type Props = ReturnType<typeof makeProps>; const Foo = (props: Props) => <div>{props.used}</div>;`, upstream reports `unused`, while Rslint emits nothing; function-expression producers fail the same way.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
```diff
@@ -0,0 +1,1267 @@
...
+					return c
+				}
+			}
+		}
+	}
+	// A reference resolver is unavailable for some syntax-only runs. Keep the
+	// fallback confined to the same lexical scope so sibling shadowed bindings
+	// cannot be associated by text alone.
+	for _, c := range components {
+		if slices.Equal(c.target, target) && (len(target) > 1 || (root != nil && componentScope(c.node) == componentScope(root))) {
```
elecmonkey: [P2] Keep member-path matching lexical-scope aware

The member-path fallback deliberately skips the lexical-scope check whenever the target contains multiple segments. With two blocks that each declare `const obj = { Foo: function(props) { ... } }`, the second `obj.Foo.propTypes` assignment is attached to the first component; when each component uses its own prop, upstream emits nothing but Rslint reports the second prop as unused. Resolve the root binding or apply the same scope constraint used for single-segment targets.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
```diff
@@ -0,0 +1,1267 @@
...
+	for current := node; current != nil && current.Parent != nil; current = current.Parent {
+		parent := current.Parent
+		if parent.Kind != ast.KindPropertyAssignment || parent.AsPropertyAssignment().Initializer != current {
+			continue
+		}
+		member := propertyName(parent.AsPropertyAssignment().Name())
+		if member == "" || parent.Parent == nil || parent.Parent.Kind != ast.KindObjectLiteralExpression {
+			break
+		}
+		object := parent.Parent
```
elecmonkey: [P2] Preserve complete component member paths

`componentTarget` records only a direct `{ Foo: fn }` object literal whose object is immediately assigned to a variable. Consequently `const root = { nested: { Foo: function() { return <div />; } } }; root.nested.Foo.propTypes = { unused: PropTypes.string };` produces the expected diagnostic upstream but none in Rslint; `obj.Foo = function() { ... }` followed by `obj.Foo.propTypes = ...` fails similarly. Retain the complete assignment/member path instead of falling back to the leaf name.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
```diff
@@ -0,0 +1,1267 @@
...
+
+func isReactClass(node *ast.Node, pragma string) bool {
+	return node != nil && (node.Kind == ast.KindClassDeclaration || node.Kind == ast.KindClassExpression) && reactutil.ExtendsReactComponent(node, pragma)
+}
+
+func isFunctionLike(node *ast.Node) bool {
+	if node == nil {
+		return false
+	}
+	switch node.Kind {
```
elecmonkey, line 137: [P2] Discover object-method components

Component discovery considers only function declarations, function expressions, and arrows even though the shared React detector supports object-literal methods. For `const obj = { Foo() { return <div />; } }; obj.Foo.propTypes = { unused: PropTypes.string };`, eslint-plugin-react@7.37.5 reports `unused`, while Rslint emits nothing because `Foo` is never registered as a component.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
```diff
@@ -0,0 +1,1267 @@
...
+						}
+					}
+				}
+				if bindsProps {
+					bindingPath(name, prefix, c)
+				}
+			case ast.KindIdentifier:
+				// Validator callbacks only contribute destructured paths. Other nested
+				// parameters named `props` retain eslint-plugin-react's scoped alias.
+				if !isValidatorParameter && (bindsProps || (name.AsIdentifier().Text == "props" && !isCallArgument(node))) {
```
elecmonkey: [P2] Retain nested props aliases in callback arguments

The `isCallArgument` exclusion drops nested parameters named `props` whenever their function is passed to a call. For `function Foo() { [1].map((props) => props.used); return <div />; } Foo.propTypes = { used: PropTypes.string };`, upstream treats `props.used` as usage and emits nothing, while Rslint reports `used`; the same mismatch occurs with `filter`, `then`, and function-expression callbacks.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
```diff
@@ -0,0 +1,1267 @@
...
+	default:
+		return false
+	}
+}
+
+func propKey(node *ast.Node) (string, *ast.Node) {
+	if node == nil {
+		return "", nil
+	}
+	name, ok := staticName(node)
```
elecmonkey, line 150: [P2] Preserve non-static computed prop declarations

Returning no key for a non-static computed property silently discards a declaration that upstream records under the key expression's identifier. For `const key = 'unused'; function Foo() { return <div />; } Foo.propTypes = { [key]: PropTypes.string, other: PropTypes.string };`, upstream reports both `key` and `other`, while Rslint reports only `other`.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types_upstream_test.go
```diff
@@ -0,0 +1,56 @@
...
+			PluginVersion string `json:"PluginVersion"`
+			Source        string `json:"Source"`
+		} `json:"Upstream"`
+		Valid   []rule_tester.ValidTestCase   `json:"Valid"`
+		Invalid []rule_tester.InvalidTestCase `json:"Invalid"`
+	}
+	assert.NilError(t, json.Unmarshal(upstreamSuiteJSON, &suite))
+	assert.Equal(t, suite.Upstream.Rule, "react/no-unused-prop-types")
+	assert.Equal(t, suite.Upstream.PluginVersion, "7.37.5")
+	assert.Equal(t, len(suite.Valid), 243)
```
elecmonkey: [P1] Restore every supported upstream test case

The asserted 243 valid and 118 invalid entries are not a complete mapping of the v7.37.5 suite: enumerating unique source cases after `parsers.all` finds four omitted valid expectations and two omitted invalid expectations. The missing cases include both runtime and TypeScript `React.memo(React.forwardRef(...))` forms, and the invalid typed form exposes the generic `forwardRef` false negative above, so the current count assertions certify an incomplete fixture rather than the claimed full supported suite.

elecmonkey (commented):
Round 3

swwind:
@elecmonkey You are absolutely right! fixed meow.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types_upstream_test.go
```diff
@@ -0,0 +1,56 @@
...
+			PluginVersion string `json:"PluginVersion"`
+			Source        string `json:"Source"`
+		} `json:"Upstream"`
+		Valid   []rule_tester.ValidTestCase   `json:"Valid"`
+		Invalid []rule_tester.InvalidTestCase `json:"Invalid"`
+	}
+	assert.NilError(t, json.Unmarshal(upstreamSuiteJSON, &suite))
+	assert.Equal(t, suite.Upstream.Rule, "react/no-unused-prop-types")
+	assert.Equal(t, suite.Upstream.PluginVersion, "7.37.5")
+	assert.Equal(t, len(suite.Valid), 247)
```
elecmonkey: [P1] Complete the upstream-suite fix

The previous completeness finding is still not fully resolved. Excluding Flow-only cases, v7.37.5 contains 245 supported valid and 118 supported invalid source cases; this fixture contains only 243 of those valid cases, omits the two `ConnectedProps` cases using untyped `bindActionCreators`, and reaches 247/120 by adding four valid and two invalid `memo(forwardRef(...))` cases that are not in this upstream suite. Please restore the two actual upstream cases and move independently authored coverage to the extras suite.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
```diff
@@ -0,0 +1,1339 @@
...
+	node = unwrap(node)
+	if node == nil {
+		return nil
+	}
+	if node.Kind == ast.KindCallExpression {
+		call := node.AsCallExpression()
+		callee := unwrap(call.Expression)
+		if callee != nil && callee.Kind == ast.KindPropertyAccessExpression {
+			method := propertyName(callee.AsPropertyAccessExpression().Name())
+			if method == "shape" || method == "exact" {
```
elecmonkey: [P2] Complete custom-validator handling in nested validators

The previous custom-validator fix only covers the outer validator. With `{ customValidators: ['Custom'], skipShapeProps: false }`, `outer: PropTypes.arrayOf(Custom.shape({ inner: Custom.string }))` produces no upstream diagnostic, but Rslint reports `outer.*.inner`; the same mismatch occurs inside `oneOfType`. Please apply the configured custom-validator namespace check in this recursive path as well.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
```diff
@@ -0,0 +1,1339 @@
...
+			result = append(result, newProp("__ANY_KEY__", "__ANY_KEY__", member))
+			continue
+		}
+		fullName := name
+		if prefix != "" {
+			fullName = prefix + "." + name
+		}
+		current := newProp(name, fullName, reportNode)
+		validator := unwrap(value)
+		if validator != nil && validator.Kind == ast.KindPropertyAccessExpression {
```
elecmonkey: [P2] Avoid regressing custom validators wrapped in isRequired

The custom-validator fix introduced a regression for `outer: Custom.shape({ inner: Custom.string }).isRequired`: with `customValidators: ['Custom']`, `skipShapeProps: false`, and only `outer` used, upstream reports `outer.inner`, while Rslint emits nothing. Upstream checks for a custom-validator call before unwrapping `.isRequired`, so this ordering must be preserved even though its behavior differs from the unwrapped form.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
```diff
@@ -0,0 +1,1339 @@
...
+		if node.Kind == ast.KindCallExpression && reactutil.IsCreateClassCall(node.AsCallExpression(), pragma, createClass) {
+			call := node.AsCallExpression()
+			if call.Arguments != nil && len(call.Arguments.Nodes) > 0 {
+				arg := unwrap(call.Arguments.Nodes[0])
+				if arg != nil && arg.Kind == ast.KindObjectLiteralExpression {
+					newComponent(arg)
+				}
+			}
+		}
+		if isFunctionLike(node) && reactutil.IsStatelessReactComponentWithWrappers(node, pragma, ctx.TypeChecker, componentWrappers) {
```
elecmonkey: [P2] Avoid discovering computed object properties as components

The member-path component fix introduced a false positive for `const obj = { ['Foo']: function() { return <div />; } }; obj.Foo.propTypes = { unused: PropTypes.string };`: upstream does not classify this computed property as a component and emits nothing, while Rslint reports `unused`. Please preserve upstream's computed-property exclusion before associating the member path with `propTypes`.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
```diff
@@ -0,0 +1,1339 @@
...
+				fullName = prefix + "." + name
+			}
+			p := newProp(name, fullName, reportNode)
+			p.never = typ != nil && typ.Kind == ast.KindNeverKeyword
+			p.children = typeProps(typ, aliases, seen, fullName)
+			result = append(result, p)
+		}
+		return result
+	case ast.KindTypeReference:
+		name := reactutil.EntityNameRightmost(node.AsTypeReferenceNode().TypeName)
```
elecmonkey, line 413: [P2] Traverse React prop-wrapper type arguments

For `import React from 'react'; type Props = { used: string; unused: string }; const Foo = ({ used }: React.PropsWithChildren<Props>) => <div>{used}</div>;`, upstream reports `unused`, while Rslint emits nothing. The imported `PropsWithChildren<Props>` and `forwardRef<HTMLDivElement, React.PropsWithChildren<Props>>` forms fail similarly; recognized React generic wrappers need to recurse into the type argument that contains the props.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
```diff
@@ -0,0 +1,1339 @@
...
+				if argument.Kind == ast.KindFunctionType {
+					return typeProps(argument.AsFunctionTypeNode().Type, aliases, seen, prefix)
+				}
+				if argument.Kind == ast.KindTypeQuery {
+					query := argument.AsTypeQueryNode()
+					if query.ExprName != nil && query.ExprName.Kind == ast.KindIdentifier {
+						if initializer := reactutil.ResolveIdentifierInitializer(query.ExprName.AsNode(), nil); initializer != nil {
+							initializer = unwrap(initializer)
+							if isFunctionLike(initializer) {
+								return typePropsFromFunctionBody(initializer.Body(), aliases, seen, prefix)
```
elecmonkey: [P2] Extract ReturnType props from directly returned generic calls

For `const makeProps = () => bindActionCreators<{ used: () => void; unused: () => void }>({}, dispatch); type Props = ReturnType<typeof makeProps>; const Foo = ({ used }: Props) => <div>{String(used)}</div>;`, upstream reports `unused`, while Rslint emits nothing. The ReturnType path handles object literals and object-spread calls, but it also needs the upstream branch that reads type arguments from a directly returned call expression.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
```diff
@@ -0,0 +1,1339 @@
...
+				switch c.node.Kind {
+				case ast.KindClassDeclaration, ast.KindClassExpression:
+					if typ := classType(c.node); typ != nil {
+						appendDeclared(c, typeProps(typ, aliases, map[string]bool{}, ""))
+					}
+					for _, member := range c.node.Members() {
+						if member.Kind == ast.KindPropertyDeclaration {
+							pd := member.AsPropertyDeclaration()
+							name := propertyName(pd.Name())
+							if name == "propTypes" && pd.Initializer != nil {
```
elecmonkey: [P2] Ignore computed propTypes declarations

For `class Foo extends React.Component { static ['propTypes'] = { unused: PropTypes.string }; render() { return <div />; } }`, upstream emits nothing, while Rslint reports `unused`; computed `['propTypes']` inside `createReactClass` behaves the same way. These declaration-collection branches should not use static evaluation to treat computed properties as ordinary `propTypes` declarations.

elecmonkey (commented):
Round 4: 2 incomplete fixes and 8 new findings across two review passes.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
```diff
@@ -0,0 +1,1339 @@
...
+			if name == "" {
+				continue
+			}
+			fullName := name
+			if prefix != "" {
+				fullName = prefix + "." + name
+			}
+			p := newProp(name, fullName, reportNode)
+			p.never = typ != nil && typ.Kind == ast.KindNeverKeyword
+			p.children = typeProps(typ, aliases, seen, fullName)
```
elecmonkey: [P2] Keep nested TypeScript properties opaque

For `type Props = { user: { used: string; unused: string } }; const Foo = (props: Props) => <div>{props.user.used}</div>;`, eslint-plugin-react@7.37.5 emits nothing, but Rslint reports `user.unused`, regardless of `skipShapeProps`. Upstream records TypeScript property signatures as leaves instead of recursively validating their member types, so recursively calling `typeProps` here introduces false positives for ordinary nested TypeScript props.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
```diff
@@ -0,0 +1,1339 @@
...
+			continue
+		case ast.KindVariableDeclaration:
+			if parent.AsVariableDeclaration().Initializer != current || parent.AsVariableDeclaration().Type == nil {
+				return nil
+			}
+			typ := parent.AsVariableDeclaration().Type
+			if typ.Kind == ast.KindTypeReference {
+				args := typ.AsTypeReferenceNode().TypeArguments
+				if args != nil && len(args.Nodes) > 0 {
+					return args.Nodes[0]
```
elecmonkey: [P2] Restrict variable type arguments to React component wrappers

For `type Extra = { phantom: string }; const Foo: Wrapper<Extra> = () => <div />;`, upstream emits nothing, while Rslint reports `phantom`. This path treats the first type argument of every generic variable annotation as props, even though upstream only unwraps recognized React generic types such as `FC` and `FunctionComponent`; arbitrary application types must not create prop declarations.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
```diff
@@ -0,0 +1,1339 @@
...
+		return nil
+	}
+	if node.Kind == ast.KindCallExpression && reactutil.IsPropWrapperCall(node, wrappers) {
+		call := node.AsCallExpression()
+		if call.Arguments != nil && len(call.Arguments.Nodes) > 0 {
+			return declaredProps(call.Arguments.Nodes[0], customValidators, wrappers)
+		}
+	}
+	if node.Kind == ast.KindIdentifier {
+		if initializer := reactutil.ResolveIdentifierInitializer(node, nil); initializer != nil && initializer != node {
```
elecmonkey: [P2] Resolve propTypes aliases in lexical scope

For `const shared = { first: PropTypes.string }; function Foo() { const shared = { second: PropTypes.string }; return <div />; } Foo.propTypes = shared;`, upstream reports `second`, while Rslint reports `first`. `declaredProps` resolves an identifier without the declaration site's lexical context, so a shadowed alias can resolve to the wrong initializer; the same mismatch occurs with multiple same-name declarations.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
```diff
@@ -0,0 +1,1339 @@
...
+}
+
+func bindingPath(name *ast.Node, prefix []string, c *component) {
+	if name == nil {
+		return
+	}
+	switch name.Kind {
+	case ast.KindIdentifier:
+		path := append([]string(nil), prefix...)
+		c.aliases[name.AsIdentifier().Text] = path
```
elecmonkey: [P2] Scope prop aliases to their declaring function

For `function Foo(props) { const x = props.a; function helper() { const x = props.b; return x.other; } return <div>{x.used}</div>; }` with matching nested shapes, upstream emits nothing, while Rslint reports `a.used`; conversely, an alias declared only inside one helper leaks into siblings and can suppress a real diagnostic. The component-wide textual alias map overwrites bindings without restoring them when leaving a nested scope.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
```diff
@@ -0,0 +1,1339 @@
...
+			path := append(append([]string(nil), prefix...), part)
+			c.mark(path)
+			bindingPath(key, path, c)
+		}
+	}
+}
+
+func (c *component) mark(path []string) {
+	if len(path) == 0 {
+		c.anyUsed = true
```
elecmonkey: [P2] Do not treat a bare props object as using every prop

For `class Foo extends React.Component { static propTypes = { unused: PropTypes.string }; render() { this.props; return <div />; } }`, upstream reports `unused`, while Rslint emits nothing. `pathFrom(this.props)` returns a successful empty path, and `mark` converts that empty path into `anyUsed`; merely evaluating the props object is not equivalent to spreading it or accessing every declaration.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
```diff
@@ -0,0 +1,1339 @@
...
+				c.mark(path)
+			}
+		}
+	}
+	if node.Kind == ast.KindJsxSpreadAttribute {
+		if _, ok := pathFrom(node.AsJsxSpreadAttribute().Expression, c.aliases); ok {
+			c.anyUsed = true
+		}
+	}
+	if node.Kind == ast.KindSpreadAssignment {
```
elecmonkey: [P2] Limit open-props handling to JSX spreads

For `function Foo(props) { const copy = { ...props }; return <div />; } Foo.propTypes = { unused: PropTypes.string };`, upstream reports `unused`, while Rslint emits nothing; `fn(...props)` and `[...props]` have the same mismatch. Upstream disables unused-prop validation for a non-object JSX spread, but ordinary object, array, and call spreads do not imply that every prop is used.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types_upstream_test.go
elecmonkey: Fixed in 7fd491dba. Restored the two missing ConnectedProps cases, moved the six independently authored memo/forwardRef cases to extras, and preserved all 30 upstream data.name assertions as exact messages. The fixture now maps every supported v7.37.5 case: 245 valid and 118 invalid.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
elecmonkey: Fixed in 7fd491dba. Custom validator namespaces remain opaque throughout recursive arrayOf, objectOf, and oneOfType handling, including .isRequired. Added regressions for nested custom and standard validators.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
elecmonkey: Fixed in 7fd491dba. React prop-wrapper type arguments are now traversed for qualified, named-import, and aliased-import forms, including PropsWithChildren inside forwardRef. Added differential regressions.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
elecmonkey: Fixed in 7fd491dba. ReturnType now extracts props from directly returned generic calls and generic object spreads, including block-bodied producers. Untyped calls no longer infer declarations from their runtime arguments.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
elecmonkey: Fixed in 7fd491dba. TypeScript property signatures are leaves rather than recursively inspected shapes, for both interfaces and type literals. Added regressions with default options and skipShapeProps: false.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
elecmonkey: Fixed in 7fd491dba. Variable annotations are unwrapped only for recognized React generic imports, with the appropriate props argument index. Arbitrary application generics and non-React FC imports no longer create prop declarations.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
elecmonkey: Addressed in 7fd491dba by resolving the actual lexical binding at the propTypes expression. In this example, shared refers to the outer declaration, so reporting first is intentional; copying upstream's second result would select an out-of-scope variable. Added regressions for shadowing, repeated var declarations, deferred getters, and alias cycles, and documented this semantic difference.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
elecmonkey: Fixed in 7fd491dba. Prop aliases are keyed by declaration symbols instead of textual names, so nested scopes cannot overwrite outer aliases or leak aliases into siblings. Added regressions for nested functions, block scopes, and parameter shadowing.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
elecmonkey: Fixed in 7fd491dba. An empty props path no longer marks every declaration as used; bare this.props expressions and unused aliases leave unused-prop diagnostics intact. Added regression coverage.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
elecmonkey: Fixed in 7fd491dba. Ordinary object, array, and call spreads no longer disable unused-prop validation; open-props handling is limited to non-object JSX spreads and open destructuring. Added regressions for each spread form.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
```diff
@@ -0,0 +1,1435 @@
...
+	root, member := name, ""
+	if name.Kind == ast.KindQualifiedName {
+		root = name.AsQualifiedName().Left
+		member = propertyName(name.AsQualifiedName().Right)
+	}
+	if root.Kind != ast.KindIdentifier {
+		return nil
+	}
+	imported := ""
+	for _, statement := range ast.GetSourceFileOfNode(name).Statements.Nodes {
```
elecmonkey, line 608: [P2] Resolve React generic wrappers by lexical binding

The new import scan matches wrapper names textually without checking whether the reference still resolves to that import. For `import type { FC } from 'react'; function scope() { type FC<T> = (props: T) => unknown; type Props = { phantom: string }; const Foo: FC<Props> = () => <div />; return Foo; }`, upstream emits nothing, while Rslint reports `phantom`; shadowing a namespace/default React import has the same result. Please resolve the type-name binding before treating it as an imported React wrapper.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
```diff
@@ -0,0 +1,1435 @@
...
+							appendDeclared(c, declaredProps(bin.Right, customValidators, wrappers, resolve))
+						} else {
+							parentExists := true
+							for i := propTypesIndex + 1; i < len(parts)-1; i++ {
+								if !declaredContains(c.declared, strings.Join(parts[propTypesIndex+1:i+1], ".")) {
+									parentExists = false
+									break
+								}
+							}
+							if parentExists {
```
elecmonkey: [P2] Inherit shape skipping for nested member assignments

With default options, `Foo.propTypes = { a: PropTypes.shape({ b: PropTypes.shape({}) }) }; Foo.propTypes.a.b.c = PropTypes.number;` produces no upstream diagnostic because the declaration remains under a skipped shape, but Rslint reports `a.b.c`. The newly appended terminal prop loses its ancestor shape context, so the recursive reporter cannot apply the default `skipShapeProps` behavior. Preserve the existing declaration hierarchy when adding the member assignment.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
elecmonkey: Fixed in 3c54aabc0. React generic wrappers now require the type reference to resolve to the matching React import binding, while the existing unbound global `React` compatibility path remains supported. Added regressions for named, default, namespace, and `forwardRef` shadowing, plus an unshadowed import control.

internal/plugins/react/rules/no_unused_prop_types/no_unused_prop_types.go
elecmonkey: Fixed in 3c54aabc0. Nested member assignments are now inserted into the existing declaration tree, so they inherit ancestor shape semantics and remain skipped by default while still reporting with `skipShapeProps: false`. Direct `Foo.propTypes.x` assignments remain supported.

elecmonkey (approved):

fansenze (approved):

PR #2131, opened 2026-09-10
feat: port rule unicorn/empty-brace-spaces

shiyudesu:
## Summary

Port `unicorn/empty-brace-spaces` from `eslint-plugin-unicorn@v73.0.0`.

The rule enforces no spaces between braces in empty blocks, class bodies, static blocks, and object literals, and autofixes them by removing the whitespace:

```js
// Incorrect
class Unicorn {
}

// Correct
class Unicorn {}
```

A comment between the braces counts as content, so the whitespace around it is left untouched. The rule has no options.

- Register the rule in the native Unicorn catalog.
- Enable it in the Unicorn recommended preset, matching upstream.
- Migrate the complete upstream v73.0.0 test suite.
- Add coverage for TypeScript-only class shapes, decorators, heritage clauses, labeled blocks, destructuring parameters, JSX expression containers, and non-ASCII ECMAScript whitespace.

Two tsgo/ESTree structural differences needed explicit handling. tsgo represents `static {}` as a `ClassStaticBlockDeclaration` that wraps a real `Block`, while ESTree has only `StaticBlock`; both nodes yield the identical brace span, so the nested `Block` is skipped to keep upstream's single diagnostic. tsgo also models a stray `;` between class members as a `SemicolonClassElement`, which ESTree drops, so these are filtered from the emptiness check.

## Related Links

- Documentation: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v73.0.0/docs/rules/empty-brace-spaces.md
- Source code: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v73.0.0/rules/empty-brace-spaces.js
- Upstream tests: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v73.0.0/test/empty-brace-spaces.js

## Validation

- `go test -count=1 ./internal/plugins/unicorn/rules/empty_brace_spaces`
- `pnpm --filter @rslint/core build:bin`
- `pnpm --dir packages/rslint-test-tools exec rs test --testTimeout=10000 empty-brace-spaces`
- `pnpm typecheck`
- `pnpm run format:check`
- `pnpm -w run check-spell`
- `golangci-lint run --new-from-merge-base=origin/main ./internal/plugins/unicorn/rules/empty_brace_spaces`: 0 issues

## Checklist

- [x] Tests updated.
- [x] Documentation updated.

### DIFF
--- a/internal/plugins/unicorn/rules/empty_brace_spaces/empty_brace_spaces.go
+++ b/internal/plugins/unicorn/rules/empty_brace_spaces/empty_brace_spaces.go
@@ -5,7 +5,6 @@ package empty_brace_spaces
 import (
 	"github.com/microsoft/TypeScript/tsc/shim/ast"
 	"github.com/microsoft/TypeScript/tsc/shim/core"
-	"github.com/microsoft/TypeScript/tsc/shim/scanner"
 	"github.com/web-infra-dev/rslint/internal/rule"
 	"github.com/web-infra-dev/rslint/internal/utils"
 	"github.com/web-infra-dev/rslint/internal/utils/ecmascript"
@@ -55,23 +54,33 @@ func hasChildren(node *ast.Node) bool {
 	}
 }
 
-// innerBraceRange returns the span strictly between the node's opening and
-// closing braces. The opening brace is located with a scan bounded by
-// node.End(), because a `static {}` block starts at the `static` keyword and
-// the enclosing class body's brace comes first. The closing brace is always
-// the node's last character for every node kind this rule inspects.
+// innerBraceRange returns the span strictly between the node's own braces. The
+// node's last token is always its closing brace, so the pair is located by
+// walking the node's tokens and remembering the `{` that leaves the nesting
+// depth at zero. Scanning for the first `{` instead would pick up a brace that
+// belongs to the node's header — `class A extends mixin({}) { }` — or to a
+// header type, leaving the real pair unreported.
 func innerBraceRange(node *ast.Node, sourceFile *ast.SourceFile) (core.TextRange, bool) {
-	start := scanner.SkipTrivia(sourceFile.Text(), utils.TrimNodeTextRange(sourceFile, node).Pos())
 	end := node.End()
+	depth := 0
+	opening := -1
 
-	s := scanner.GetScannerForSourceFile(sourceFile, start)
-	for s.TokenStart() < end {
-		if s.Token() == ast.KindOpenBraceToken {
-			return core.NewTextRange(s.TokenEnd(), end-1), true
+	utils.ForEachToken(node, func(token *ast.Node) {
+		switch token.Kind {
+		case ast.KindOpenBraceToken:
+			if depth == 0 {
+				opening = token.End()
+			}
+			depth++
+		case ast.KindCloseBraceToken:
+			depth--
 		}
-		s.Scan()
+	}, sourceFile)
+
+	if opening < 0 {
+		return core.TextRange{}, false
 	}
-	return core.TextRange{}, false
+	return core.NewTextRange(opening, end-1), true
 }
 
 // isWhitespaceOnly reports whether text is made up solely of ECMAScript
--- a/internal/plugins/unicorn/rules/empty_brace_spaces/empty_brace_spaces.go
+++ b/internal/plugins/unicorn/rules/empty_brace_spaces/empty_brace_spaces.go
@@ -1,0 +1,116 @@
+// Package empty_brace_spaces ports eslint-plugin-unicorn's
+// `empty-brace-spaces` rule.
+package empty_brace_spaces
+
+import (
+	"github.com/microsoft/TypeScript/tsc/shim/ast"
+	"github.com/microsoft/TypeScript/tsc/shim/core"
+	"github.com/microsoft/TypeScript/tsc/shim/scanner"
+	"github.com/web-infra-dev/rslint/internal/rule"
+	"github.com/web-infra-dev/rslint/internal/utils"
+	"github.com/web-infra-dev/rslint/internal/utils/ecmascript"
+)
+
+var messageEmptyBraceSpaces = rule.RuleMessage{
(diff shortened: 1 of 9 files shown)


internal/plugins/unicorn/rules/empty_brace_spaces/empty_brace_spaces.go
```diff
@@ -0,0 +1,116 @@
...
+// the enclosing class body's brace comes first. The closing brace is always
+// the node's last character for every node kind this rule inspects.
+func innerBraceRange(node *ast.Node, sourceFile *ast.SourceFile) (core.TextRange, bool) {
+	start := scanner.SkipTrivia(sourceFile.Text(), utils.TrimNodeTextRange(sourceFile, node).Pos())
+	end := node.End()
+
+	s := scanner.GetScannerForSourceFile(sourceFile, start)
+	for s.TokenStart() < end {
+		if s.Token() == ast.KindOpenBraceToken {
+			return core.NewTextRange(s.TokenEnd(), end-1), true
```
elecmonkey: **[P1] Anchor the scan at the actual class body brace**

When a class header contains an earlier brace, such as `class A extends mixin({}) { }`, this scan selects that brace instead of the class body's opener. The resulting range is not whitespace-only, so Rslint emits no diagnostic and leaves the outer class unchanged, while `eslint-plugin-unicorn@73.0.0` reports the outer space and fixes the code to `class A extends mixin({}) {}`. The same issue affects decorators containing object literals or function bodies and TypeScript type parameters containing object types.

internal/plugins/unicorn/rules/empty_brace_spaces/empty_brace_spaces_upstream_test.go
```diff
@@ -0,0 +1,182 @@
...
+		valid = append(valid, validCase(replacePlaceholder(code, strings.Repeat(" ", 3))))
+	}
+
+	var invalid []rule_tester.InvalidTestCase
+	for _, spaces := range []string{" ", "\t", " \t \t ", "\n\n", "\r\n"} {
+		for _, code := range allCases {
+			invalid = append(invalid, invalidCase(code, spaces))
+		}
+	}
+	invalid = append(invalid, invalidCase("with (foo) {/* */}", strings.Repeat(" ", 5)))
```
elecmonkey, line 179: **[P2] Migrate the standalone upstream snapshot case**

The v73.0.0 upstream test file has a separate multiline `catch (error)` snapshot case with seven spaces on an otherwise empty line, outside its generated invalid-case table. This test stops after the generated cases and `with`, so the stated complete upstream migration omits that exact input and its multiline location/fix assertions. Add the standalone case with its diagnostic range and output expectations.

elecmonkey (commented):
Thanks for the contribution! I found a few issues.

internal/plugins/unicorn/rules/empty_brace_spaces/empty_brace_spaces.go
shiyudesu: Fixed in fbfd29e0 — the anchor now comes from the node's own brace (token walk tracking nesting depth), so a header brace can no longer win.

internal/plugins/unicorn/rules/empty_brace_spaces/empty_brace_spaces_upstream_test.go
shiyudesu: Added in fbfd29e0 — the standalone `test.snapshot` case is migrated with its range and output assertions.

shiyudesu:
Both are valid — thanks for catching them. Fixed in fbfd29e0.

**Header braces.** This was worse than an edge case: the scan took the first `{`, so `class A extends mixin({}) { }`, `mixin({ a: 1 })`, `T extends {}`, and a decorated class with an object literal all selected a header brace, failed the whitespace check, and reported nothing. Now anchored on the node's own brace via a token walk that tracks nesting depth — one uniform rule that also removes the earlier `StaticBlock` workaround, since `static {}` was the same bug. All four now fix to `class A extends mixin({}) {}`. A header pair that is itself empty is reported separately, both fixes in one pass.

**Snapshot case.** Also right — the `test.snapshot` block after the generated table was missing. Added with its range and fix assertions (`line 3, col 18 -> line 5, col 1`, output `} catch (error) {}`).

Fixing the first one exposed three bugs in the extras test helpers (a compact `{}` counted as empty, a dropped span in the two-diagnostic case, wrong error order). Corrected, and the header cases are now covered in the Go and JS suites.

Verified: rule package tests, JS integration, scoped `golangci-lint` (0 issues), `format:check`, and `check-spell`.

internal/plugins/unicorn/rules/empty_brace_spaces/empty_brace_spaces_upstream_test.go
```diff
@@ -0,0 +1,201 @@
...
+
+	// Upstream's standalone `test.snapshot` case, kept outside the generated
+	// table. Its brace pair opens with a whitespace-only line, so the report
+	// covers the whole run of spaces and the fix removes the line outright.
+	invalid = append(invalid, rule_tester.InvalidTestCase{
+		Code:     "try {\n\tfoo();\n} catch (error) {\n" + strings.Repeat(" ", 7) + "\n}",
```
elecmonkey: **[P2] Preserve the upstream snapshot's tab indentation**

The upstream `outdent` result includes a tab before the seven explicit spaces on this blank line. Preserve that exact input so the migrated test covers the same mixed tab-and-space whitespace as upstream.

```suggestion
		Code:     "try {\n\tfoo();\n} catch (error) {\n\t" + strings.Repeat(" ", 7) + "\n}",
```

elecmonkey (approved):
Thanks again for your contributions to Rslint!

elecmonkey:
@fansenze need your approval because i'm the last pusher

fansenze (approved):

PR #2134, opened 2026-09-10
feat(unicorn): implement prefer-set-size rule

HashemKhalifa:
## Motivation

Contributes `unicorn/prefer-set-size` to the Unicorn rule-porting tracker #1309: replace array conversion used only to count a Set with its `.size` property.

Rslint does not provide a native implementation of `unicorn/prefer-set-size`. Counting a known Set by spreading it into an array or calling `Array.from()` creates an unnecessary array.

## Changes

Port the rule from eslint-plugin-unicorn v74.0.0 and enable it in the Unicorn preset. Report the `length` property and replace the conversion with `.size`, preserving required parentheses and withholding fixes that would discard comments.

Add the upstream cases, native regression coverage, and JavaScript integration assertions for diagnostics, ranges, fixes, and clean relinting.

Validation: scoped Go rule tests and rule-catalog completeness test; native binary and JavaScript builds; schema generation; focused JavaScript integration test (3 tests); formatting and spelling checks.


elecmonkey (commented):
Thanks! I found some issues.

internal/plugins/unicorn/rules/prefer_set_size/prefer_set_size.go
```diff
@@ -0,0 +1,239 @@
...
+				}
+			}
+		}
+	}
+
+	// eslint-plugin-unicorn uses parser type information after its syntax and
+	// scope checks. Restrict this fallback to TypeScript inputs: JavaScript's
+	// upstream parser has no TypeScript Program, so mutable JS bindings must not
+	// become matches merely because tsgo can infer their type.
+	if ctx.TypeChecker == nil || ast.IsInJSFile(node) {
```
elecmonkey: Returning when `ctx.TypeChecker` is nil drops upstream's syntax-only type classification. With `project: false`, Unicorn 74.0.0 reports and fixes all four upstream TypeScript cases, including `function getSize(set: Set<string>) { return Array.from(set).length; }`, while Rslint emits no diagnostic; the committed tests hide this because their shared config always supplies a tsconfig. Please inspect TypeScript annotations and aliases before requiring a checker.

internal/plugins/unicorn/rules/prefer_set_size/prefer_set_size.go
```diff
@@ -0,0 +1,239 @@
...
+	}
+	switch node.Kind {
+	case ast.KindIdentifier, ast.KindPropertyAccessExpression, ast.KindCallExpression,
+		ast.KindTemplateExpression, ast.KindThisKeyword, ast.KindArrayLiteralExpression,
+		ast.KindFunctionExpression:
+		return false
+	case ast.KindNewExpression:
+		range_ := utils.TrimNodeTextRange(ctx.SourceFile, node)
+		text := ctx.SourceFile.Text()
+		return range_.End() <= range_.Pos() || range_.End() > len(text) || text[range_.End()-1] != ')'
```
elecmonkey: Checking only whether a `NewExpression` ends in `)` mistakes the callee grouping in `new (Set)` for constructor argument parentheses. Unicorn 74.0.0 fixes `[...new (Set)].length` to `(new (Set)).size`, whereas Rslint emits `new (Set).size`, which constructs `Set.size` and throws `TypeError`; determine whether the closing parentheses belong to the argument list before omitting the wrapper.

internal/plugins/unicorn/rules/prefer_set_size/prefer_set_size.go
```diff
@@ -0,0 +1,239 @@
...
+	// This keeps `unknown satisfies Set` unknown while accepting a Set wrapped by
+	// `satisfies` or a non-null assertion, and recognizes compound expressions
+	// only when every possible result is a Set.
+	switch runtime.Kind {
+	case ast.KindSatisfiesExpression:
+		satisfies := runtime.AsSatisfiesExpression()
+		return satisfies != nil && isSet(ctx, satisfies.Expression, visiting)
+	case ast.KindNonNullExpression:
+		nonNull := runtime.AsNonNullExpression()
+		return nonNull != nil && isSet(ctx, nonNull.Expression, visiting)
```
elecmonkey: For a project-backed input such as `declare const set: Set<string> | null; [...set!].length`, Unicorn 74.0.0 uses the narrowed type of `set!` and reports with output `(set!).size`. Rslint recursively classifies the inner nullable binding and emits nothing; consult the checker for the non-null node before discarding its narrowing wrapper.

HashemKhalifa (commented):
## Current-head review: fixes required

Reviewed all eight changed files at `c8efbe1b17dcb86c5899434a01e992800223a56e` against `291f9748e42c50a38bd03368768e4e9e7ac56746`, the five existing review threads, tracker #1309, and the pinned Unicorn v74.0.0 rule/type helpers/test corpus. The intervening main-branch merge did not change this rule.

Three remaining findings are attached: repeated const-alias expansion, disabled checker heritage support, and incomplete source-only compound/derived type recognition. The first is a structural/performance issue; the latter two are source-traced upstream compatibility gaps. A focused implementation and regression tests are prepared in the existing rule/extras files; no shared-classifier rewrite or unrelated cleanup is needed.

The earlier spread-argument rejection, grouped `new (Set)` fix, non-null narrowing, and four basic project:false TypeScript cases are addressed by the current code and committed tests. I also checked the provenance correction: conditional/sequence/satisfies cases are not in v74.0.0's prefer-set-size test file, so their extras labeling is correct.

Validation limits: native Rslint Go tests and JS integration have **not** run in this review environment. GitHub cannot be resolved from the local execution container, so only an API-fetched source snapshot is available; pnpm is also unavailable. `gofmt`, `git diff --check`, clean patch application to the exact source snapshot, three isolated cache-helper/model checks, and TypeScript checking of 14 selected input/output fixtures passed. The isolated checks are not substitutes for Rslint end-to-end validation. No approval or merge is implied.

internal/plugins/unicorn/rules/prefer_set_size/prefer_set_size.go
```diff
@@ -0,0 +1,301 @@
...
+			}
+			if declaration != nil && declaration.Kind == ast.KindVariableDeclaration {
+				list := declaration.Parent
+				variable := declaration.AsVariableDeclaration()
+				if list != nil && list.Kind == ast.KindVariableDeclarationList &&
+					list.Flags&ast.NodeFlagsConst != 0 && variable != nil &&
+					ast.IsIdentifier(variable.Name()) && variable.Initializer != nil {
+					visiting[symbol] = true
+					result := isSet(ctx, variable.Initializer, visiting)
+					delete(visiting, symbol)
```
HashemKhalifa: ### [P1] Memoize completed const classifications instead of only guarding cycles

Deleting the symbol from `visiting` after each initializer walk means every later reference re-expands it. Together with the two recursive conditional branches, `const set0 = new Set(); const set1 = flag ? set0 : set0; const set2 = flag ? set1 : set1; ...; [...set30].length` has a linear-size source graph but approximately 2^30 leaf visits for one diagnostic. This can monopolize the linter on a small generated file; a cycle guard does not bound repeated work on an acyclic graph.

Use one per-classification walk state with active symbols and completed declaration results, following the existing source-only array receiver's visiting/memo pattern. Keep use-site TypeChecker results out of a symbol-wide cache because mutable bindings can narrow differently at each reference. Add a generated shared-branch const-chain regression to the existing extras suite, plus cycle and narrowing guards.

Evidence is the recursive call graph, not a measured native Rslint benchmark. An isolated model of the current expansion performed 2,097,151 classifications at depth 20; the extracted proposed cache helper performs 31 unique classifications at depth 30. This is a local performance requirement, not a claim that upstream avoids the same pattern.

internal/plugins/unicorn/rules/prefer_set_size/prefer_set_size.go
```diff
@@ -0,0 +1,301 @@
...
+	return isSetWithChecker(ctx, runtime)
+}
+
+func isSetWithChecker(ctx rule.RuleContext, node *ast.Node) bool {
+	if node == nil || ctx.TypeChecker == nil || ast.IsInJSFile(node) {
+		return false
+	}
+	return unicornutil.ClassifyType(ctx, ctx.TypeChecker.GetTypeAtLocation(node), unicornutil.TypeClassifierOptions{
+		TargetTypeNames: setTypeNames,
+	}) == unicornutil.TypeTarget
```
HashemKhalifa, line 212: ### [P2] Enable the shared classifier's inherited Set recognition

`ClassifyType` traverses base types only when `HeritageSymbolFlags` is nonzero. Passing just the target names misses ordinary project-backed derived Sets, for example `class StringSet extends Set<string> {} const set = new StringSet(); [...set].length`, and `interface StringSet extends ReadonlySet<string> {} declare function getSet(): StringSet; Array.from(getSet()).length`. Their type symbols are `StringSet`, not `Set`/`ReadonlySet`, so this path returns false.

Unicorn v74.0.0 defaults `checkClassHeritage` to true and visits base types in `getTypeScriptType` (rules/utils/type-helpers.js). Set `HeritageSymbolFlags: ast.SymbolFlagsClass | ast.SymbolFlagsInterface` here rather than implementing another checker traversal. Add focused regressions expecting `set.size` / `getSet().size`, with an Array subclass as a negative control. This finding is established by source tracing; the native regressions still need execution.

internal/plugins/unicorn/rules/prefer_set_size/prefer_set_size.go
```diff
@@ -0,0 +1,301 @@
...
+	for node.Kind == ast.KindParenthesizedType {
+		node = node.AsParenthesizedTypeNode().Type
+		if node == nil {
+			return false
+		}
+	}
+	if node.Kind != ast.KindTypeReference {
+		return false
+	}
+	return isSetTypeReference(ctx, node.AsTypeReferenceNode().TypeName, visiting)
```
HashemKhalifa: ### [P2] Preserve compound and derived annotations in project:false

The new syntax-only path still rejects every union/intersection before checking its parts, and its type-reference resolver only handles aliases/type parameters. With `project:false`, `function getSize(set: Set<string> | ReadonlySet<string>) { return Array.from(set).length; }` therefore has no diagnostic, as does `[...(set as Set<string> & { tag: string })].length` when `set` is unknown. Interface/class annotations derived from Set are also skipped without the checker.

Unicorn v74.0.0's `getTypeAnnotationType` combines unions with all-target semantics, intersections with any-target semantics, and resolves interface/class heritage through its type-reference helper. Extend the existing annotation walker accordingly, preserving cycle protection and memoizing shared alias graphs. Keep mixed Set/array and nullable unions unreported without narrowing, and ignore `implements` as runtime inheritance. Extend `TestPreferSetSizeProjectFalseTypeSyntax` with exact diagnostic/fix expectations and negative controls; do not duplicate these extras in the JS upstream mirror.

This is a remaining broader-syntax gap, not a repeat of the now-fixed four basic source-only TypeScript cases.

HashemKhalifa:
## Review fixes pushed: `09a9e9721b1467548c2c41d24bf6dc7aa5c82cba`

Implemented the three findings from [review 5177514343](https://github.com/web-infra-dev/rslint/pull/2134#pullrequestreview-5177514343) and fast-forwarded the existing `feat/unicorn-prefer-set-size` branch. The commit is based on the latest main-merge head `c8efbe1`; no force push, new PR, merge, or unrelated files.

Changes are limited to `prefer_set_size.go` and its existing extras suite:
- A single walk-state/cache mechanism handles active symbols and completed declaration classifications. Use-site checker results remain outside the symbol cache.
- The existing shared type classifier now receives the class/interface heritage flags.
- Source-only type classification handles all-Set unions, Set-bearing intersections, and interface/class `extends` annotations, including the distinct class heritage AST shape.
- Added regressions for shared const/type-alias graphs, derived constructor/call results, compound annotations, and mixed/nullable/non-Set/cyclic/narrowing controls. The upstream mirror remains untouched.

### Verification actually completed

- Original API-fetched source hashes matched the GitHub blobs.
- `gofmt` and `git diff --check`: passed.
- Patch application against the exact reviewed source snapshot: passed.
- Three isolated cache-helper/model tests: passed. A depth-30 shared graph required 31 unique classifications with the extracted helper; the un-memoized depth-20 model required 2,097,151. These are **not native Rslint timings or integration results**.
- Fourteen selected TypeScript input/output fixtures typechecked with the locally available TypeScript 5.8.3. This only checks those fixtures, not the Go port.
- Uploaded source/test blob hashes match the locally checked files; the final compare contains exactly the intended two files.

### Still required before approval

The local container cannot resolve GitHub, preventing a complete checkout/dependency setup. Native `go test` could not start from the source snapshot, and pnpm is unavailable. Native rule tests, native rebuild/JS integration, and repository spelling/format checks are therefore **not verified**. This commit and comment do not represent an approval.

Run the scoped checks in a provisioned checkout:

```sh
go test ./internal/plugins/unicorn/rules/prefer_set_size
pnpm --filter @rslint/core build:bin
CI=true pnpm --dir packages/rslint-test-tools exec rs test run tests/eslint-plugin-unicorn/rules/prefer-set-size.test.ts
pnpm run check-spell internal/plugins/unicorn/rules/prefer_set_size/prefer_set_size.go internal/plugins/unicorn/rules/prefer_set_size/prefer_set_size_extras_test.go
pnpm run format:check
```

**Status: fixes implemented and pushed; native validation remains outstanding.**

elecmonkey (approved):
LGTM

elecmonkey (approved):

PR #2136, opened 2026-09-10
feat(unicorn): implement no-new-buffer rule

HashemKhalifa:
## Motivation

Contributes `unicorn/no-new-buffer` to the Unicorn rule-porting tracker #1309: replace deprecated Buffer construction with the appropriate allocation or conversion API.

Rslint does not provide a native implementation of `unicorn/no-new-buffer`. The deprecated Buffer constructor uses the same call shape for allocation and conversion.

## Changes

Port the rule from eslint-plugin-unicorn v74.0.0 and enable it in the Unicorn preset. Replace known numeric arguments with `Buffer.alloc()` and known conversion arguments with `Buffer.from()`. Offer both replacements as suggestions when the argument cannot be classified.

Reuse the constructor-to-call fixer with `new-for-builtins` to preserve comments and required parentheses. Add upstream fixtures, regression cases for classification and edit handling, and JavaScript assertions for diagnostics, replacements, and suggestions.

Validation: scoped Go tests for this rule and `new-for-builtins`; rule-catalog completeness test; native binary and JavaScript builds; schema generation; focused JavaScript integration test (5 tests); formatting and spelling checks.


internal/plugins/unicorn/rules/no_new_buffer/no_new_buffer.go
```diff
@@ -0,0 +1,586 @@
...
+	messageIDError        = "error"
+	messageIDErrorUnknown = "error-unknown"
+	messageIDSuggestion   = "suggestion"
+)
+
+var NoNewBufferRule = rule.Rule{
+	Name:   "unicorn/no-new-buffer",
+	Schema: rule.EmptyArraySchema,
+	Run: func(ctx rule.RuleContext, options []any) rule.RuleListeners {
+		staticEvaluator := utils.NewStaticStringEvaluatorWithSourceFile(ctx.TypeChecker, ctx.SourceFile)
```
elecmonkey: **[P1] Pass `ctx.Refs` to the static evaluator**

In ordinary source-only API/CLI linting, `ctx.TypeChecker` is nil, but `ctx.Refs` can still resolve same-file bindings through the binder. Because this constructor does not receive `ctx.Refs`, upstream cases such as `const size = 10; new Buffer(size)`, `const string = "string"; new Buffer(string)`, and `const array = [1]; new Buffer(array)` fall back to `error-unknown` instead of the upstream autofixes. Construct it with `NewStaticStringEvaluatorWithReferenceResolver(ctx.TypeChecker, ctx.SourceFile, ctx.Refs)`: this uses `RefStore` first in source-only mode while retaining checker-backed resolution when type information is available. The control-flow layer must still reject mutable `let`/`var` bindings to match upstream.

internal/plugins/unicorn/rules/no_new_buffer/no_new_buffer.go
```diff
@@ -0,0 +1,586 @@
...
+		return getStaticValueForControlFlow(node.AsAsExpression().Expression, staticEvaluator, ctx)
+	case ast.KindSatisfiesExpression:
+		return getStaticValueForControlFlow(node.AsSatisfiesExpression().Expression, staticEvaluator, ctx)
+	case ast.KindTypeAssertionExpression:
+		return getStaticValueForControlFlow(node.AsTypeAssertion().Expression, staticEvaluator, ctx)
+	case ast.KindNonNullExpression:
+		return getStaticValueForControlFlow(node.AsNonNullExpression().Expression, staticEvaluator, ctx)
+	case ast.KindConditionalExpression:
+		conditional := node.AsConditionalExpression()
+		truthy, known := knownTruthy(conditional.Condition, ctx)
```
elecmonkey: **[P2] Move control-flow evaluation into the shared evaluator**

This rule-local `knownTruthy` implementation handles only a small literal subset, so expressions such as `new Buffer((1 === 1) ? 1 : value)`, `new Buffer(undefined ?? 1)`, and `new Buffer({} && 1)` lose the upstream `Buffer.alloc(...)` autofix. Add a public `StaticStringEvaluator` operation for conservative control-flow evaluation that reuses its existing `evalValue`, `staticValueTruthy`, and `staticValueNullish` logic while rejecting mutable bindings and unsafe member reads; then replace this rule-local conditional/logical evaluator with that shared operation.

internal/plugins/unicorn/unicornutil/new_to_call.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	source := sourceFile.Text()
+	removeEnd := nodeRange.Pos() + len("new")
+	for removeEnd < expressionRange.Pos() && isWhitespace(source[removeEnd]) {
+		removeEnd++
+	}
+
+	insertAfterExpression := ""
+	if newExpression.Arguments == nil {
```
elecmonkey, line 40: **[P1] Insert synthesized call parentheses after type arguments**

For a TypeScript constructor with type arguments but no argument list, this inserts `()` immediately after the callee. Thus `new Buffer<string>` is fixed to `Buffer.from()<string>` instead of upstream's `Buffer.from<string>()`; TypeScript parses the former as applying type arguments to the result of `Buffer.from()`, which can introduce type errors and change behavior. The shared helper likewise turns `new Symbol<string>` into `Symbol()<string>`.

elecmonkey (commented):
Thanks for contributing to Rslint! I found a few issues.

internal/plugins/unicorn/unicornutil/new_to_call.go
```diff
@@ -0,0 +1,136 @@
...
+	for _, char := range sourceFile.Text()[left:right] {
+		if char == '\n' || char == '\r' || char == '\u2028' || char == '\u2029' {
+			return false
+		}
+	}
+	return true
+}
+
+func isWhitespace(char byte) bool {
+	return char == ' ' || char == '\t' || char == '\n' || char == '\r'
```
elecmonkey: **[P2] Remove all JavaScript whitespace after `new`**

This helper recognizes fewer whitespace characters than upstream's `/^\s*/` removal and than the previous `new-for-builtins` implementation. Valid inputs such as `new\vBuffer(1)`, `new\fBuffer(1)`, and `new\u00A0Buffer(1)` retain that whitespace in Rslint's output, while upstream produces `Buffer.alloc(1)`; the first two are also regressions for the existing shared-helper consumer. Please use the scanner's JavaScript trivia handling or otherwise cover the complete ECMAScript whitespace set.

internal/plugins/unicorn/rules/no_new_buffer/no_new_buffer_upstream_test.go
```diff
@@ -0,0 +1,127 @@
...
+
+func newBufferError(code, expression, messageID, message string, suggestions ...rule_tester.InvalidTestCaseSuggestion) rule_tester.InvalidTestCaseError {
+	offset := strings.Index(code, expression)
+	if offset < 0 {
+		panic("expression not found: " + expression)
+	}
+	prefix := code[:offset]
+	line := strings.Count(prefix, "\n") + 1
+	column := offset - strings.LastIndex(prefix, "\n")
+	endPrefix := code[:offset+len(expression)]
```
elecmonkey: **[P1] Count ECMAScript line separators in expected locations**

This helper counts only `\n`, but the new extras cases contain U+2028 and U+2029, which Rslint correctly treats as line terminators. Consequently `go test ./internal/plugins/unicorn/rules/no_new_buffer` currently fails both cases because their actual end positions are on line 3 while this helper expects line 2. Please calculate positions with the same ECMAScript/UTF-16 location utility used by diagnostics.

internal/utils/static_string_evaluator.go
```diff
@@ -192,6 +193,65 @@ func (staticEvaluator *StaticStringEvaluator) EvalValue(node *ast.Node) (any, bo
...
+// EvalControlFlowArrayValue reports whether a conservatively evaluated value
+// is an array. It is the aggregate counterpart to EvalControlFlowValue.
+func (staticEvaluator *StaticStringEvaluator) EvalControlFlowArrayValue(node *ast.Node) (isArray bool, known bool) {
+	if staticEvaluator == nil || node == nil {
+		return false, false
+	}
+	previous := staticEvaluator.controlFlowOnly
+	staticEvaluator.controlFlowOnly = true
+	defer func() { staticEvaluator.controlFlowOnly = previous }()
+	result := staticEvaluator.evalValue(node)
```
elecmonkey: [P1] Reject unsafe member reads before classifying arrays

`EvalControlFlowArrayValue` bypasses the unsafe-member check used by `EvalControlFlowValue`, so `const object = {value: [1]}; Object.defineProperty(object, "value", {get() { return unknown; }}); new Buffer(object.value);` is automatically changed to `Buffer.from(object.value)`. eslint-plugin-unicorn@74.0.0 reports `error-unknown` with two suggestions because the getter can return a number at runtime. Apply the same conservative member-access validation before returning an array classification.

elecmonkey (commented):
Still one existing issue.

elecmonkey (commented):
The previous issues are fixed. I found two regressions introduced by the fixes. Could you address them?

internal/utils/static_string_evaluator.go
```diff
@@ -192,6 +193,51 @@ func (staticEvaluator *StaticStringEvaluator) EvalValue(node *ast.Node) (any, bo
...
+		return value, true
+	}
+	return result.value, true
+}
+
+func hasUnsafeControlFlowMemberAccess(node *ast.Node) bool {
+	if node == nil {
+		return false
+	}
+	if node.Kind == ast.KindPropertyAccessExpression || node.Kind == ast.KindElementAccessExpression {
```
elecmonkey: **[P2] Preserve safe static member evaluation**

This blanket rejection also covers member reads that upstream can prove safe. For example, eslint-plugin-unicorn@74.0.0 automatically fixes `new Buffer(({value: "x"}).value)` to `Buffer.from(({value: "x"}).value)` and `new Buffer([1][0])` to `Buffer.alloc([1][0])`, while Rslint reports `error-unknown` with two suggestions. The same mismatch affects `Object.freeze([1])` and computed safe globals such as `Math["PI"]`. Could you distinguish safe literal members, static global properties, and pass-through calls from potentially mutable getter-backed reads, as upstream's `hasPotentiallyMutableMemberAccess` does?

internal/utils/static_string_evaluator.go
```diff
@@ -192,6 +193,51 @@ func (staticEvaluator *StaticStringEvaluator) EvalValue(node *ast.Node) (any, bo
...
+// conditional or logical branch in an autofix: unlike EvalValue it never
+// treats an unwritten let or var binding as immutable.
+func (staticEvaluator *StaticStringEvaluator) EvalControlFlowValue(node *ast.Node) (any, bool) {
+	if staticEvaluator == nil || node == nil {
+		return nil, false
+	}
+	if hasUnsafeControlFlowMemberAccess(node) {
+		return nil, false
+	}
+	previous := staticEvaluator.controlFlowOnly
```
elecmonkey: **[P2] Reject side-effectful control-flow evaluation**

`EvalControlFlowValue` evaluates expressions without applying upstream's side-effect guard. With `let flag; new Buffer((flag = true) ? 1 : unknown)`, eslint-plugin-unicorn@74.0.0 reports `error-unknown` with both suggestions, but Rslint folds the assignment condition and automatically changes the call to `Buffer.alloc(...)`; the same happens when this expression is reached through a `const` initializer. Could you reject side-effectful expressions before static control-flow evaluation, matching upstream's `getStaticValueIfNoSideEffectsInternal` policy?

HashemKhalifa:
@elecmonkey  > The previous issues are fixed. I found two regressions introduced by the fixes. Could you address them?sure thing I will take care of it.

elecmonkey (approved):
Thanks

PR #2139, opened 2026-09-10
fix(react): align jsx-props-no-multi-spaces with upstream

elecmonkey:
## Summary

- use the parsed type argument boundary when checking spacing after generic JSX tag names, avoiding unsafe fixes when comments contain `>`
- make line-gap detection comment-aware so blank lines inside block comments are not treated as gaps between props
- recognize Unicode whitespace-only lines when checking for gaps between JSX props
- report the actual spread argument text in diagnostics instead of the generic `spread` placeholder
- add Go regression coverage for generic comments, block comments, Unicode whitespace, spread diagnostics, and autofix output

### DIFF
--- a/internal/plugins/react/rules/jsx_props_no_multi_spaces/jsx_props_no_multi_spaces.go
+++ b/internal/plugins/react/rules/jsx_props_no_multi_spaces/jsx_props_no_multi_spaces.go
@@ -20,7 +20,6 @@ var JsxPropsNoMultiSpacesRule = rule.Rule{
 		lineStarts := ctx.SourceFile.ECMALineMap()
 
 		check := func(node *ast.Node) {
-			comments := ctx.Comments.All()
 			var props []*ast.Node
 			var tagNameEnd int
 			var tagName *ast.Node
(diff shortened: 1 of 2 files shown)


internal/plugins/react/rules/jsx_props_no_multi_spaces/jsx_props_no_multi_spaces.go
```diff
@@ -20,6 +20,7 @@ var JsxPropsNoMultiSpacesRule = rule.Rule{
 		lineStarts := ctx.SourceFile.ECMALineMap()
 
 		check := func(node *ast.Node) {
+			comments := ctx.Comments.All()
```
elecmonkey: **[P2] Defer comment collection until checking a multiline gap**

Calling `ctx.Comments.All()` here materializes the full-file comment list as soon as the rule encounters any JSX opening element, even when the element has no props or all of its props are on one line. In a benchmark with one ordinary file-level comment and 10,000 single-line JSX elements, this increased median runtime from approximately 4.94 ms to 8.29 ms (+68%) and allocations from 11.20 MB to 13.05 MB. Moving `Comments.All()` into the multiline branch of `checkGap` restored the baseline cost while preserving the new behavior, so please defer comment collection until `hasEmptyLines` is actually needed.

elecmonkey: optimized.

fansenze (approved):

PR #2140, opened 2026-09-10
fix(react): align jsx-fragments with upstream

elecmonkey:
## Summary

- honor file-level `@jsx` pragmas when matching and expanding React fragments
- resolve local fragment aliases through the ESLint-compatible scope model in both source-only and type-aware linting
- preserve upstream first-definition semantics for redeclared fragment aliases
- prevent shadowed bindings without initializers from crashing the linter
- prevent shadowed or unrelated component bindings from being autofixed into fragment shorthand
- add Go and JS regression coverage for custom JSX runtimes, source-only scope resolution, shadowing, and redeclarations


internal/plugins/react/rules/jsx_fragments/jsx_fragments.go
```diff
@@ -238,19 +237,12 @@ func (m fragmentMatcher) collectImportDeclaration(node *ast.Node) {
...
-		if decl := declarationForIdentifier(utils.GetReferenceSymbol(ident, m.ctx.TypeChecker)); decl != nil {
-			return m.declarationRefersToReactFragment(decl)
-		}
-	}
-	if m.ctx.SourceFile == nil {
-		return false
-	}
-	return m.syntaxDeclarationRefersToReactFragment(m.ctx.SourceFile.AsNode(), name)
+	decl := declarationForIdentifier(m.ctx.Refs.Resolve(ident))
+	return decl != nil && m.declarationRefersToReactFragment(decl)
```
fansenze: **[P1] Guard missing initializers before inspecting the resolved binding**

Without a type checker, this now resolves `F` to the loop binding below. That declaration has no initializer, so `declarationRefersToReactFragment` reaches `ast.SkipParentheses(nil)` and panics, aborting linting.

```jsx
const F = React.Fragment;
for (const F of widgets) {
  <F />;
}
```

A shadowing `catch (F)` parameter triggers the same panic. Before this change, these examples were incorrectly reported but did not crash; eslint-plugin-react 7.37.5 reports nothing. Please check for a nil initializer **before** calling `SkipParentheses` and add source-only regression coverage for these bindings.

fansenze (commented):
Found two regressions in source-only linting. Reproductions and suggested fixes are inline.

internal/plugins/react/rules/jsx_fragments/jsx_fragments.go
```diff
@@ -238,19 +237,12 @@ func (m fragmentMatcher) collectImportDeclaration(node *ast.Node) {
...
-	if m.ctx.TypeChecker != nil {
-		if decl := declarationForIdentifier(utils.GetReferenceSymbol(ident, m.ctx.TypeChecker)); decl != nil {
-			return m.declarationRefersToReactFragment(decl)
-		}
-	}
-	if m.ctx.SourceFile == nil {
-		return false
-	}
-	return m.syntaxDeclarationRefersToReactFragment(m.ctx.SourceFile.AsNode(), name)
+	decl := declarationForIdentifier(m.ctx.Refs.Resolve(ident))
```
fansenze: **[P2] Preserve the variable definition when resolving redeclarations**

Without a type checker, switching to `Refs.Resolve` loses the fragment diagnostic for this `.jsx` input:

```jsx
function render() {
  var F = React.Fragment;
  function F() {}
  return <F />;
}
```

The assignment runs after function hoisting, so `<F />` uses `React.Fragment`. The previous implementation and eslint-plugin-react 7.37.5 both report `preferFragment`; this PR reports nothing because the resolved symbol selects the function declaration. Please preserve upstream's first-definition lookup for this case and add a source-only regression test.

internal/plugins/react/rules/jsx_fragments/jsx_fragments.go
elecmonkey: Fixed in 3d6b24479. The matcher now checks for a nil initializer before calling `ast.SkipParentheses` and resolves bare JSX identifiers through the ESLint-shaped scope model. I added source-only and regular regression coverage for `for...of`, destructured `for...of`, `catch`, and uninitialized bindings.

internal/plugins/react/rules/jsx_fragments/jsx_fragments.go
elecmonkey: Fixed in 3d6b24479. Bare JSX identifiers now use the source-ordered first definition from the ESLint-shaped scope model instead of TypeScript's `ValueDeclaration`. I added source-only, Go RuleTester, and JS integration coverage for `var`/function and repeated-`var` declaration ordering.

fansenze:
@elecmonkey Please resolve the merge conflicts. Thanks!

elecmonkey:
> @elecmonkey Please resolve the merge conflicts. Thanks!

resolved, thanks

fansenze (approved):

PR #2144, opened 2026-09-11
feat: port rule unicorn/require-post-message-target-origin

shiyudesu:
## Summary

Port `unicorn/require-post-message-target-origin` from eslint-plugin-unicorn v74.0.0.

The rule reports `.postMessage()` calls missing a `targetOrigin` argument and provides editor suggestions without automatic fixes.

- Preserve upstream matching behavior, diagnostic ranges, structured suggestion data, and suggestion order.
- Preserve comments, parentheses, and trailing commas when inserting suggested arguments.
- Register the rule in the native Unicorn catalog and explicitly disable it in the recommended preset, matching upstream.
- Migrate the complete upstream test suite and documentation examples.
- Add Go coverage for TypeScript, JSX, optional chains, Unicode, and edit-demand modes.
- Add JS integration tests that verify diagnostic ranges, suggestion data, messages and outputs, and the absence of automatic fixes.

Like upstream, the rule cannot distinguish window messaging from other APIs such as `Worker#postMessage()`. This limitation is documented. The rule has no options. Place explicit rule configuration after the recommended preset to enable it.

## Related Links

- Tracking issue: #1309
- [Upstream documentation](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v74.0.0/docs/rules/require-post-message-target-origin.md)
- [Upstream source](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v74.0.0/rules/require-post-message-target-origin.js)
- [Upstream tests](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v74.0.0/test/require-post-message-target-origin.js)

## Validation

- Go upstream, edge-case, edit-demand, and rule catalog tests passed.
- JS rule integration and targeted Unicorn preset tests passed.
- Core native and JS builds passed, including schema and public rule-option type generation.
- Scoped formatting and spelling checks, repository `pnpm run format:check`, and `git diff --check` passed.

Verification used reduced concurrency. No full-repository or whole-plugin test suites were run.

## Checklist

- [x] Tests updated.
- [x] Documentation updated.

### DIFF
--- a/internal/plugins/unicorn/all.go
+++ b/internal/plugins/unicorn/all.go
@@ -46,6 +46,7 @@ import (
 	"github.com/web-infra-dev/rslint/internal/plugins/unicorn/rules/prefer_then_catch"
 	"github.com/web-infra-dev/rslint/internal/plugins/unicorn/rules/require_array_join_separator"
 	"github.com/web-infra-dev/rslint/internal/plugins/unicorn/rules/require_number_to_fixed_digits_argument"
+	"github.com/web-infra-dev/rslint/internal/plugins/unicorn/rules/require_post_message_target_origin"
 	"github.com/web-infra-dev/rslint/internal/plugins/unicorn/rules/throw_new_error"
 	"github.com/web-infra-dev/rslint/internal/rule"
 )
@@ -97,6 +98,7 @@ func GetAllRules() []rule.Rule {
 		prefer_ternary.PreferTernaryRule,
 		require_array_join_separator.RequireArrayJoinSeparatorRule,
 		require_number_to_fixed_digits_argument.RequireNumberToFixedDigitsArgumentRule,
+		require_post_message_target_origin.RequirePostMessageTargetOriginRule,
 		throw_new_error.ThrowNewErrorRule,
 	}
 }
--- a/internal/plugins/unicorn/rules/require_post_message_target_origin/require_post_message_target_origin.go
+++ b/internal/plugins/unicorn/rules/require_post_message_target_origin/require_post_message_target_origin.go
@@ -1,0 +1,79 @@
+// Ported from eslint-plugin-unicorn v74.0.0 (MIT).
+package require_post_message_target_origin
+
+import (
+	"github.com/microsoft/TypeScript/tsc/shim/ast"
+	"github.com/microsoft/TypeScript/tsc/shim/core"
+	"github.com/microsoft/TypeScript/tsc/shim/scanner"
+	"github.com/web-infra-dev/rslint/internal/plugins/unicorn/unicornutil"
+	"github.com/web-infra-dev/rslint/internal/rule"
+	"github.com/web-infra-dev/rslint/internal/utils"
+)
+
+// https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v74.0.0/rules/require-post-message-target-origin.js
+var RequirePostMessageTargetOriginRule = rule.Rule{
+	Name:   "unicorn/require-post-message-target-origin",
+	Schema: rule.EmptyArraySchema,
+	Run: func(ctx rule.RuleContext, options []any) rule.RuleListeners {
+		oneArgument := 1
+		return rule.RuleListeners{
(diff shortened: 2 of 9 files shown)


internal/plugins/unicorn/rules/require_post_message_target_origin/require_post_message_target_origin.go
```diff
@@ -0,0 +1,79 @@
...
+					replacements = append(replacements, "'*'")
+
+					suggestions := make([]rule.RuleSuggestion, 0, len(replacements))
+					for _, replacement := range replacements {
+						text := ", " + replacement
+						if trailingComma {
+							text = " " + replacement + ","
+						}
+						suggestions = append(suggestions, rule.RuleSuggestion{
+							Message: rule.RuleMessage{Id: "suggestion", Description: "Use `" + replacement + "`."},
```
elecmonkey: [P2] Preserve suggestion payload data

For every reported call, upstream v74 emits each suggestion with `data: { code: replacement }`, but this message only embeds the replacement in `Description` and leaves `RuleMessage.Data` nil. A differential run over 43 reporting cases showed matching text and fixes but missing `suggestions[*].data` in Rslint; the API explicitly projects that map for structured consumers, so populate `Data` with the replacement and assert it in the integration test.

internal/plugins/unicorn/all.go
```diff
@@ -97,6 +98,7 @@ func GetAllRules() []rule.Rule {
 		prefer_ternary.PreferTernaryRule,
 		require_array_join_separator.RequireArrayJoinSeparatorRule,
 		require_number_to_fixed_digits_argument.RequireNumberToFixedDigitsArgumentRule,
+		require_post_message_target_origin.RequirePostMessageTargetOriginRule,
```
elecmonkey, line 101: [P2] Mirror the explicit preset disable

Once this rule enters the native catalog, `unicornPlugin.configs.recommended` still omits `unicorn/require-post-message-target-origin` because its preset entry remains commented as not implemented. Upstream v74 explicitly sets this non-recommended rule to `off`; when an earlier flat-config entry enables it and the recommended preset follows, upstream disables it but Rslint continues reporting. Add the explicit `off` entry and a preset assertion to preserve composition semantics.

internal/plugins/unicorn/rules/require_post_message_target_origin/require_post_message_target_origin.go
shiyudesu: Fixed in ac321202. Suggestions now include `data: { code: replacement }`. Added Go edit-demand assertions and end-to-end checks for all 33 suggestion payloads in the upstream/documentation cases. The rule's Go tests and JS integration tests pass.

internal/plugins/unicorn/all.go
shiyudesu: Fixed in ac321202. The recommended preset now explicitly sets `unicorn/require-post-message-target-origin` to `off`. Added a preset assertion and integration coverage for both config orders: a later preset disables an earlier setting, while an explicit setting after the preset re-enables the rule. The targeted Unicorn preset tests pass, and the documentation now explains the ordering.

elecmonkey:
The npm package CI failure is deterministic rather than flaky. The new preset test passes `RslintConfigEntry` values directly to the low-level `@rslint/core/internal` `lint()` API, whose `config` field expects an already-normalized `Record<string, unknown>[]`. This causes the three `TS2322` errors in `presets.test.ts`.

The Windows npm job passes because it does not run `pnpm typecheck`; it only builds and runs the tests, where this type mismatch has no runtime effect.

Please normalize each authored config before passing it to `lint()`, for example:

```ts
for (const authoredConfig of [[enabled], [preset, enabled]]) {
  const result = await lint({
    ...options,
    config: normalizeConfig(authoredConfig),
  });
  // ...
}

const result = await lint({
  ...options,
  config: normalizeConfig([enabled, preset]),
});
```

This matches the existing test infrastructure, where authored presets are normalized before being passed to the internal lint API.

elecmonkey (approved):

PR #2145, opened 2026-09-11
feat(react): add iframe-missing-sandbox rule

swwind:
## Motivation

Add parity for eslint-plugin-react’s iframe sandbox security rule. It catches iframe elements without a sandbox attribute, invalid sandbox tokens, and the unsafe allow-scripts + allow-same-origin combination.

## Changes

- Port `react/iframe-missing-sandbox` from eslint-plugin-react v7.37.5, including documentation, registration, and Go/JS test coverage.
- Preserve upstream React createElement recognition, including computed identifiers and destructured React imports.
- Extract direct JSX string-attribute entity decoding into reactutil and reuse it from react/no-invalid-html-attribute.
- Add regressions for entity-decoded sandbox tokens and delimiters.

### DIFF
--- a/internal/plugins/react/rules/iframe_missing_sandbox/iframe_missing_sandbox.go
+++ b/internal/plugins/react/rules/iframe_missing_sandbox/iframe_missing_sandbox.go
@@ -1,0 +1,186 @@
+package iframe_missing_sandbox
+
+import (
+	"strings"
+
+	"github.com/microsoft/TypeScript/tsc/shim/ast"
+	"github.com/web-infra-dev/rslint/internal/plugins/react/reactutil"
+	"github.com/web-infra-dev/rslint/internal/rule"
+	"github.com/web-infra-dev/rslint/internal/utils/ecmascript"
+)
+
+const (
+	msgAttributeMissing   = "An iframe element is missing a sandbox attribute"
+	msgInvalidValue       = "An iframe element defines a sandbox attribute with invalid value \"{{ value }}\""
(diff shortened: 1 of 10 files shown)


internal/plugins/react/rules/iframe_missing_sandbox/iframe_missing_sandbox.go
```diff
@@ -0,0 +1,186 @@
...
+				if len(call.Arguments.Nodes) > 1 {
+					props := ast.SkipParentheses(call.Arguments.Nodes[1])
+					if props != nil && props.Kind == ast.KindObjectLiteralExpression {
+						for _, member := range props.AsObjectLiteralExpression().Properties.Nodes {
+							value, hasSandbox := objectSandboxProperty(member)
+							if !hasSandbox {
+								continue
+							}
+							found = true
+							value = ast.SkipParentheses(value)
```
elecmonkey, line 172: [P1] Guard valueless sandbox properties before unwrapping

When the props object uses a shorthand, method, getter, or setter named `sandbox`—for example, `React.createElement("iframe", { sandbox })`—`objectSandboxProperty` returns `(nil, true)`, and this unconditional `ast.SkipParentheses(value)` dereferences nil and terminates the entire Rslint process. eslint-plugin-react@7.37.5 accepts these valid forms without diagnostics; please check `value != nil` before unwrapping it.

swwind: fixed

elecmonkey (approved):
LGTM

PR #2147, opened 2026-09-12
feat(react): add jsx-curly-newline rule

swwind:
## Summary
- add the `react/jsx-curly-newline` rule, schema, documentation, and React catalog registration
- migrate every eslint-plugin-react v7.37.5 upstream case (excluding Flow syntax) with exact runtime source and fixer output
- add tsgo edge-case, edit-demand, and JS integration coverage

## Verification
- `go test ./internal/plugins/react/rules/jsx_curly_newline`
- `golangci-lint run --new-from-merge-base=origin/main ./internal/plugins/react/rules/jsx_curly_newline ./internal/plugins/react`
- `pnpm run check-spell <changed files>`
- `pnpm run format:check`

## Remaining
- The JS integration test could not be rerun with a rebuilt native CLI: `pnpm --filter @rslint/core build:bin` failed because the environment hit its `/tmp` build disk quota.

### DIFF
--- a/internal/plugins/react/rules/jsx_curly_newline/jsx_curly_newline.go
+++ b/internal/plugins/react/rules/jsx_curly_newline/jsx_curly_newline.go
@@ -1,0 +1,186 @@
+package jsx_curly_newline
+
+import (
+	_ "embed"
+	"strings"
+
+	"github.com/microsoft/TypeScript/tsc/shim/ast"
+	"github.com/microsoft/TypeScript/tsc/shim/core"
+	"github.com/microsoft/TypeScript/tsc/shim/scanner"
+	"github.com/web-infra-dev/rslint/internal/rule"
+	"github.com/web-infra-dev/rslint/internal/utils"
+	"github.com/web-infra-dev/rslint/internal/utils/ecmascript"
+)
+
+//go:embed jsx_curly_newline.schema.json
+var schemaJSON []byte
+
+const (
+	consistent = "consistent"
+	require    = "require"
+	forbid     = "forbid"
+)
+
+type options struct {
+	singleline string
+	multiline  string
+}
+
+func parseOptions(raw []any) options {
+	result := options{singleline: consistent, multiline: consistent}
+	if len(raw) == 0 {
+		return result
+	}
+	if value, ok := raw[0].(string); ok && value == "never" {
+		return options{singleline: forbid, multiline: forbid}
+	}
+	if value, ok := raw[0].(map[string]any); ok {
+		if singleline, ok := value["singleline"].(string); ok {
+			result.singleline = singleline
+		}
+		if multiline, ok := value["multiline"].(string); ok {
+			result.multiline = multiline
+		}
+	}
+	return result
+}
+
+func requiresNewlines(option options, expression *ast.Node, lineMap []core.TextPos, hasLeftNewline bool) bool {
+	multiline := false
+	if expression != nil {
+		multiline = expression.Pos() != expression.End() && scanner.ComputeLineOfPosition(lineMap, expression.Pos()) != scanner.ComputeLineOfPosition(lineMap, expression.End())
+	}
+	mode := option.singleline
+	if multiline {
+		mode = option.multiline
+	}
+	switch mode {
+	case forbid:
+		return false
+	case require:
+		return true
+	default:
+		return hasLeftNewline
+	}
+}
+
+// JsxCurlyNewlineRule enforces matching line breaks immediately within JSX
+// expression-container braces. JSX spread children are excluded because the
+// upstream listener is JSXExpressionContainer-only.
+var JsxCurlyNewlineRule = rule.Rule{
+	Name:   "react/jsx-curly-newline",
+	Schema: rule.NewSchema(schemaJSON),
+	Run: func(ctx rule.RuleContext, rawOptions []any) rule.RuleListeners {
+		option := parseOptions(rawOptions)
+		text := ctx.SourceFile.Text()
+		lineMap := ctx.SourceFile.ECMALineMap()
+
+		return rule.RuleListeners{
+			ast.KindJsxExpression: func(node *ast.Node) {
+				expression := node.AsJsxExpression()
+				if expression == nil || expression.DotDotDotToken != nil {
+					return
+				}
+				trimmed := utils.TrimNodeTextRange(ctx.SourceFile, node)
+				openPos, closePos := trimmed.Pos(), trimmed.End()-1
+				if openPos < 0 || closePos <= openPos || closePos >= len(text) || text[openPos] != '{' || text[closePos] != '}' {
+					return
+				}
+
+				innerLow, innerHigh := openPos+1, closePos
+				firstToken := ecmascript.SkipLeadingWhitespace(text, innerLow, innerHigh)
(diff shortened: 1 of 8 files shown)


internal/plugins/react/rules/jsx_curly_newline/jsx_curly_newline.go
```diff
@@ -0,0 +1,186 @@
...
+			result.multiline = multiline
+		}
+	}
+	return result
+}
+
+func requiresNewlines(option options, expression *ast.Node, lineMap []core.TextPos, hasLeftNewline bool) bool {
+	multiline := false
+	if expression != nil {
+		multiline = expression.Pos() != expression.End() && scanner.ComputeLineOfPosition(lineMap, expression.Pos()) != scanner.ComputeLineOfPosition(lineMap, expression.End())
```
elecmonkey: [P1] Derive multiline status from the ESTree-equivalent expression span

This makes brace-adjacent trivia and ts-go-only wrapper nodes choose the option branch, although upstream classifies `node.expression.loc`: with `{ singleline: "forbid", multiline: "require" }`, `<App>{\nvalue}</App>` is a single-line expression upstream and is fixed to `<App>{value}</App>`, while Rslint reports `expectedBefore` and fixes it to `<App>{\nvalue\n}</App>`; parenthesized and TypeScript `as` expressions diverge similarly. The nil case also needs the authored `JSXEmptyExpression` span: upstream accepts `<App>{\n}</App>` under that option, whereas Rslint removes the newline. These ranges can be derived locally from the ts-go nodes and source text, so aligning them implements the documented distinction instead of reproducing an upstream bug.

internal/plugins/react/rules/jsx_curly_newline/jsx_curly_newline.go
```diff
@@ -0,0 +1,186 @@
...
+		}
+		position = ecmascript.SkipLeadingWhitespace(text, position, end)
+	}
+	return position
+}
+
+func skipTrailingComments(text string, position, start int) int {
+	for position > start {
+		lineStart := strings.LastIndexByte(text[start:position], '\n') + start + 1
+		if comment := strings.LastIndex(text[lineStart:position], "//"); comment >= 0 {
```
elecmonkey: [P2] Use parsed token and comment ranges for trailing trivia

The raw reverse searches treat token text as comments. Under `"never"`, both `<App>{'http://host'\n}</App>` and `<App>{/x*/\n}</App>` are safely autofixed upstream by removing the newline, but Rslint interprets `//` or `*/` inside the string or regexp as a trailing comment and suppresses the fix; a multiline template literal containing `http://host` can even receive a false `unexpectedBefore`. The existing parser-backed token and comment indexes distinguish these cases, so this can be fixed entirely in Rslint without a ts-go change.

internal/plugins/react/rules/jsx_curly_newline/jsx_curly_newline.go
```diff
@@ -0,0 +1,186 @@
...
+			for position+1 < end && (text[position] != '*' || text[position+1] != '/') {
+				position++
+			}
+			if position+1 >= end {
+				return end
+			}
+			position += 2
+		case '/':
+			position += 2
+			for position < end && text[position] != '\n' && text[position] != '\r' {
```
elecmonkey: [P2] End line comments at every ECMAScript line terminator

This scan recognizes only CR and LF, but U+2028 and U+2029 also terminate ECMAScript line comments. For `<App>{//a\u2028value}</App>` with the default option, upstream terminates the comment before `value` and reports `expectedBefore`, while Rslint consumes `value` as comment text and reports nothing; under `"never"` it emits an extra right-side diagnostic. `ecmascript.LineTerminatorSequenceAt` or the parser's comment ranges can handle all four terminators locally, so this is not a ts-go limitation.

swwind: fixed all

internal/plugins/react/rules/jsx_curly_newline/jsx_curly_newline.go
```diff
@@ -0,0 +1,164 @@
...
+				// ESLint's token APIs exclude comments. Use the parser-backed comment
+				// ranges rather than scanning text, which would mistake comment-like
+				// content in strings, regexps, and templates for trivia.
+				firstCode := skipLeadingComments(text, innerLow, innerHigh, comments)
+				lastCodeEnd := skipTrailingComments(text, innerLow, innerHigh, comments)
+				hasLeftNewline := ecmascript.ContainsLineTerminator(text, innerLow, firstCode)
+				hasRightNewline := ecmascript.ContainsLineTerminator(text, lastCodeEnd, innerHigh)
+				// ESTree expression ranges exclude brace-adjacent trivia and tsgo-only
+				// wrapper spans. Empty JSX expressions use their authored interior.
+				needsNewlines := requiresNewlines(option, firstCode, lastCodeEnd, lineMap, hasLeftNewline)
```
elecmonkey: [P1] Strip transparent parentheses before selecting the mode

For `<App>{(\nvalue\n)}</App>`, Espree and typescript-eslint expose the identifier itself as the expression, so it is single-line despite newlines inside transparent parentheses. Using the first and last non-comment token instead keeps those parentheses in the span: with `{ singleline: "require", multiline: "consistent" }`, upstream inserts newlines immediately inside the braces while Rslint reports nothing; with `{ singleline: "consistent", multiline: "require" }`, Rslint instead emits and applies two fixes that upstream does not. The updated 2,622-case differential matrix still has 30 diagnostic/fix mismatches, all from this remaining wrapper-span issue.

internal/plugins/react/rules/jsx_curly_newline/jsx_curly_newline.go
```diff
@@ -0,0 +1,164 @@
...
+				if openPos < 0 || closePos <= openPos || closePos >= len(text) || text[openPos] != '{' || text[closePos] != '}' {
+					return
+				}
+
+				innerLow, innerHigh := openPos+1, closePos
+				// ESLint's token APIs exclude comments. Use the parser-backed comment
+				// ranges rather than scanning text, which would mistake comment-like
+				// content in strings, regexps, and templates for trivia.
+				firstCode := skipLeadingComments(text, innerLow, innerHigh, comments)
+				lastCodeEnd := skipTrailingComments(text, innerLow, innerHigh, comments)
```
elecmonkey, line 93: [P2] Avoid rescanning every file comment for each JSX expression

Both helpers iterate from the beginning or end of the complete file-level comment slice for every JSX expression, making files with many JSX containers and comments quadratic. A targeted run with 32,000 repeated comment/container pairs took about 820 ms versus 37 ms for an equivalent comment-free control, and doubling from 16,000 to 32,000 raised the commented case from about 198 ms to 820 ms. Start from the relevant comment with binary search or maintain a source-ordered cursor so each container examines only adjacent comments.

swwind: fixed meow

internal/plugins/react/rules/jsx_curly_newline/jsx_curly_newline.go
swwind: fixed nyaa

elecmonkey (approved):
LGTM

PR #2148, opened 2026-09-12
feat: port rule unicorn/number-literal-case

shiyudesu:
## Summary

Port `unicorn/number-literal-case` from eslint-plugin-unicorn v74.0.0.

The rule enforces lowercase radix prefixes and exponent markers for numbers and BigInts, with configurable hexadecimal digit casing and automatic fixes.

- Support `hexadecimalValue: "uppercase" | "lowercase"`, defaulting to `"uppercase"`.
- Preserve numeric separators, BigInt suffixes, comments, and surrounding syntax when fixing literals.
- Register the rule and its options schema in the native Unicorn catalog.
- Migrate upstream tests and documentation examples.
- Add Go coverage for TypeScript, JSX, UTF-16 diagnostic positions, option validation, and edit-demand modes.
- Add JS integration tests and document parser limitations.

Seven Vue cases and two legacy numeric-literal cases are explicitly skipped in JS integration because of existing parser limitations. The legacy literals are additionally checked directly against the Go rule using parsed AST nodes.

The rule is enabled at `error` severity in the Unicorn recommended preset, matching upstream.

## Related Links

- Tracking issue: #1309
- [Upstream documentation](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v74.0.0/docs/rules/number-literal-case.md)
- [Upstream source](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v74.0.0/rules/number-literal-case.js)
- [Upstream tests](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v74.0.0/test/number-literal-case.js)

## Validation

- Go upstream, edge-case, schema, and edit-demand tests passed, with 100% statement coverage for the rule package.
- Rule catalog registration check passed.
- Targeted Unicorn recommended-preset test passed with one worker.
- JS rule integration tests passed, with the parser-related skips described above.
- Core native and JS builds passed, including schema and public rule-option type generation.
- Scoped Go lint, formatting, spelling checks, repository `pnpm run format:check`, and `git diff --check` passed.

Verification used reduced concurrency. No full-repository or whole-plugin test suites were run.

## Checklist

- [x] Tests updated.
- [x] Documentation updated.

### DIFF
--- a/internal/plugins/unicorn/all.go
+++ b/internal/plugins/unicorn/all.go
@@ -33,6 +33,7 @@ import (
 	"github.com/web-infra-dev/rslint/internal/plugins/unicorn/rules/no_unsafe_string_replacement"
 	"github.com/web-infra-dev/rslint/internal/plugins/unicorn/rules/no_useless_switch_case"
 	"github.com/web-infra-dev/rslint/internal/plugins/unicorn/rules/no_xor_as_exponentiation"
+	"github.com/web-infra-dev/rslint/internal/plugins/unicorn/rules/number_literal_case"
 	"github.com/web-infra-dev/rslint/internal/plugins/unicorn/rules/prefer_add_event_listener_options"
 	"github.com/web-infra-dev/rslint/internal/plugins/unicorn/rules/prefer_array_flat"
 	"github.com/web-infra-dev/rslint/internal/plugins/unicorn/rules/prefer_array_flat_map"
@@ -85,6 +86,7 @@ func GetAllRules() []rule.Rule {
 		no_unsafe_string_replacement.NoUnsafeStringReplacementRule,
 		no_useless_switch_case.NoUselessSwitchCaseRule,
 		no_xor_as_exponentiation.NoXorAsExponentiationRule,
+		number_literal_case.NumberLiteralCaseRule,
 		prefer_add_event_listener_options.PreferAddEventListenerOptionsRule,
 		prefer_array_flat.PreferArrayFlatRule,
 		prefer_array_flat_map.PreferArrayFlatMapRule,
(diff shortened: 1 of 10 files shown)


internal/plugins/unicorn/all.go
```diff
@@ -85,6 +86,7 @@ func GetAllRules() []rule.Rule {
 		no_unsafe_string_replacement.NoUnsafeStringReplacementRule,
 		no_useless_switch_case.NoUselessSwitchCaseRule,
 		no_xor_as_exponentiation.NoXorAsExponentiationRule,
+		number_literal_case.NumberLiteralCaseRule,
```
elecmonkey, line 89: **[P2] Enable this rule in the recommended preset**

Adding the rule to the native catalog makes explicit configurations work, but `packages/rslint/src/config/presets/unicorn.ts` still leaves it commented out as “not implemented.” Upstream v74 enables `unicorn/number-literal-case` in both the recommended and unopinionated presets. With the current built Rslint preset, `const foo = 0Xff;` produces no diagnostic, while the upstream recommended preset reports one. Please include this preset update and add the corresponding preset assertion:

```diff
-    // 'unicorn/number-literal-case': 'error', // not implemented
+    'unicorn/number-literal-case': 'error',
```

shiyudesu: Thanks for catching this. Fixed in 70823ff5.

`unicorn/number-literal-case` is now enabled as `error` in the Unicorn recommended preset, and the existing preset test asserts it. The core JS build and targeted Unicorn preset test passed (one test worker).

I also updated the PR description to reflect the preset change.

elecmonkey (approved):

PR #2149, opened 2026-09-12
feat(rstest): add no-unnecessary-assertion rule

elecmonkey:
## Motivation

TypeScript types can prove that some Rstest assertions cannot produce the asserted result. Such assertions either add no useful coverage or indicate inaccurate types.

## Changes

Add the type-aware `rstest/no-unnecessary-assertion` rule, based on `eslint-plugin-jest@29.16.0` and adapted to Rstest semantics.

- Check `toBeNull()`, `toBeUndefined()`, `toBeDefined()`, `toBeNaN()`, and Chai `.null`, `.undefined`, and `.NaN`.
- Resolve Rstest globals, imports, CommonJS bindings, `rstack/test`, `import.meta.rstest`, Playwright imports, Test Context `expect`, and `expect.soft()`.
- Analyze `expect.poll()` using the awaited return types of overloads callable without arguments, matching Rstest's `fn()` invocation.
- Resolve the first runtime subject across spread arguments and skip cases that cannot be determined safely.
- Account for structural numeric compatibility, generics, intersections, branded primitives, shared type graphs, narrowing, cross-file types, and ambient declarations.
- Skip `resolves`, `rejects`, `expect.element()`, and Chai property getters after `expect.poll()`.
- Require type information and warn when `strictNullChecks` is disabled.
- Add Go tests, compiler capability fixtures, JavaScript integration coverage, and documentation.

The rule is not included in the `recommended` preset.

## Credits

Ported from [`eslint-plugin-jest@29.16.0`](https://github.com/jest-community/eslint-plugin-jest/blob/v29.16.0/src/rules/no-unnecessary-assertion.ts).

- Supports Rstest-specific expect sources, `expect.soft()`, `expect.poll()`, and Chai property assertions.
- Models spread arguments and zero-argument poll overload resolution using TypeScript checker semantics.
- Avoids false positives for structural numeric types, unresolved generics, `void`, and branded or intersection types.
- Distinguishes shared type-graph nodes from recursive cycles.

## Related Links

Related #935.

## Checklist

- [x] Tests updated.
- [x] Documentation updated.


internal/plugins/rstest/rules/no_unnecessary_assertion/no_unnecessary_assertion.go
```diff
@@ -0,0 +1,189 @@
...
+		return true
+	}
+	seen[typeValue] = true
+
+	for _, part := range utils.UnionTypeParts(typeValue) {
+		flags := checker.Type_flags(part)
+		// TypeScript permits a value-returning function to be assigned to a
+		// void-returning signature. A call whose static type is void can therefore
+		// produce any runtime value, including the values checked by this rule.
+		if flags&(checker.TypeFlagsAny|checker.TypeFlagsUnknown|checker.TypeFlagsVoid|desiredFlags) != 0 {
```
fansenze: **[P2] Account for structural types that can contain numbers**

This valid TypeScript assertion is reported as `Unnecessary assertion, subject cannot be a number`:

```ts
function check(value: {}) {
  expect(value).toBeNaN();
}
check(NaN);
```

`{}` accepts primitive numbers, even though its type flags are not `NumberLike`. The same false positive occurs with `Number`, `{ toFixed(): string }`, and `T extends {}`. Please account for structural compatibility before ruling out numeric values, so assertions that can succeed are retained.

internal/plugins/rstest/rules/no_unnecessary_assertion/no_unnecessary_assertion.go
```diff
@@ -0,0 +1,189 @@
...
+					}
+				}
+
+				head := parsed.Head.AsCallExpression()
+				if head == nil || len(head.Arguments.Nodes) == 0 {
+					return
+				}
+				checked, canBeDesiredType := subjectCanBeDesiredType(
+					ctx.TypeChecker,
+					head.Arguments.Nodes[0],
```
fansenze: **[P2] Resolve the first runtime argument before checking its type**

A leading empty spread is currently treated as the assertion subject:

```ts
const prefix: [] = [];
expect(...prefix, null).toBeNull();
```

This valid assertion receives `null`, but the rule reads the spread element's `never` type and reports `subject cannot be null`. The same issue affects `expect.soft()` and the undefined/NaN matchers. Please find the first argument after expansion, or skip leading spreads when the subject cannot be determined safely.

internal/plugins/rstest/rules/no_unnecessary_assertion/no_unnecessary_assertion.go
```diff
@@ -0,0 +1,189 @@
...
+	if typeValue == nil {
+		return true
+	}
+	if seen == nil {
+		seen = make(map[*checker.Type]bool)
+	}
+	if seen[typeValue] {
+		return true
+	}
+	seen[typeValue] = true
```
fansenze: **[P2] Restrict cycle detection to the current recursion path**

This assertion produces no diagnostic, although both alternatives are strings and cannot be `null`:

```ts
type A = string & { a: true };
type B = string & { b: true };
declare const value: A | B;
expect(value).toBeNull();
```

The first intersection marks `string` as seen; the second encounters the same type and returns `true`, treating a shared constituent as a cycle. This also skips unions of type parameters sharing a constraint. Please remove types from `seen` when leaving their recursion path, or memoize the actual result.

fansenze (commented):
Found three reproducible edge cases: two false positives and one false negative. The existing targeted Go tests pass; 11 additional cases reproduce the issues described inline.

internal/plugins/rstest/rules/no_unnecessary_assertion/no_unnecessary_assertion.go
elecmonkey: Fixed in 7936d68e. NaN checks now consider whether `number` is assignable to the subject type, with regressions for compatible and incompatible structural types.

internal/plugins/rstest/rules/no_unnecessary_assertion/no_unnecessary_assertion.go
elecmonkey: Fixed in 7936d68e. The rule now resolves the first runtime subject across leading spreads and skips cases that cannot be determined safely.

internal/plugins/rstest/rules/no_unnecessary_assertion/no_unnecessary_assertion.go
elecmonkey: Fixed in 7936d68e. Recursion-path tracking is now separate from memoized results, so shared constituents are no longer treated as cycles. Shared-constraint and large-DAG regressions were added.

fansenze (approved):

PR #2151, opened 2026-09-12
feat(rstest): add valid-expect-with-promise rule

elecmonkey:
## Motivation

Assertions on Promise values need `resolves` or `rejects` to inspect the settled value rather than the Promise object itself. Conversely, applying either modifier to a non-promise subject fails at runtime.

Rstest adds framework-specific cases that require type-aware handling: `expect(...).rejects` invokes function subjects without arguments, `expect.soft` and Chai assertion chains share ordinary expect semantics, while `expect.poll` and `expect.element` provide separate asynchronous assertion APIs.

Related #935

## Changes

Add the type-aware `rstest/valid-expect-with-promise` rule and register it outside the `recommended` preset.

The rule recognizes Rstest globals, named and renamed imports, namespace and CommonJS forms, `rstack/test`, `import.meta.rstest`, Playwright, and Test Context expect calls. It checks ordinary `expect` and `expect.soft` assertions, including promise modifiers between Chai assertions, and excludes `expect.poll` and `expect.element`.

TypeScript expressions that preserve their runtime value—`as`, angle-bracket assertions, `satisfies`, non-null assertions, and parentheses—are handled consistently across member extraction, call-chain traversal, root resolution, and `import.meta.rstest` recognition. This behavior remains opt-in so existing parser consumers retain their established wrapper boundaries.

Chai subject transformations such as `property`, `ownProperty`, and `haveOwnProperty` are tracked in source order. The rule checks the subject used by each subsequent matcher or promise modifier while conservatively stopping when a transformed subject cannot be determined.

Native Promise instances and subclasses are recognized through the shared default-library symbol helper, while Promise constructors remain non-promise subjects. With `checkThenables: true`, custom thenables must expose a `then` method whose fulfillment and rejection parameters are both callable.

For callable subjects passed to `.rejects`, the rule uses TypeScript call resolution to inspect the return type selected by a zero-argument call, matching Rstest runtime behavior across overloaded and generic functions.

Add the complete `eslint-plugin-jest@29.16.0` test suite, Rstest-specific and tsgo edge cases, source-only and compiler-backed capability proofs, JavaScript integration coverage, runtime probes, differential checks, and user documentation.

## Credits

Adapted from [`eslint-plugin-jest@29.16.0`'s `valid-expect-with-promise` rule](https://github.com/jest-community/eslint-plugin-jest/blob/v29.16.0/docs/rules/valid-expect-with-promise.md).


fansenze (commented):
Found three reproducible false negatives. The existing targeted Go tests pass, but regression probes for the examples below each produce zero diagnostics.

internal/plugins/rstest/rules/valid_expect_with_promise/valid_expect_with_promise.go
```diff
@@ -0,0 +1,180 @@
...
+			arguments := parsed.Head.AsCallExpression().Arguments.Nodes
+			if len(arguments) == 0 {
+				return
+			}
+			subject := arguments[0]
+			modifier := parsed.PromiseModifierEntry()
+			typ := ctx.TypeChecker.GetTypeAtLocation(subject)
+			if modifier != nil {
+				var known bool
+				typ, known = promiseSubjectTypeAtModifier(ctx.TypeChecker, parsed, modifier, typ)
```
fansenze: [P2] Track subject changes even without a promise modifier

The subject type is updated only when `resolves` or `rejects` is already present, so this produces no diagnostic:

```ts
expect({ p: Promise.resolve(1) }).property('p').toEqual(1);
```

Chai's `property('p')` changes the subject to the Promise. The subsequent `toEqual(1)` therefore compares the Promise object with `1` and fails at runtime; inserting `.resolves` before `toEqual` makes it pass. Please track the subject used by subsequent matchers even when there is no modifier, so this reports `poorlyExpectedPromise`.

internal/plugins/rstest/utils/parse_rstest_expect.go
```diff
@@ -231,11 +255,29 @@ func parseRstestExpectCall(
...
+	if node == nil || node.Kind != ast.KindCallExpression || findTopMostCallExpression(node, throughTypeAssertions) != node {
 		return nil
 	}
-	expression := findTopMostRstestExpectExpression(node)
+	expression := findTopMostRstestExpectExpressionOptions(node, throughTypeAssertions)
 	entries := testFramework.GetMemberEntries(expression)
+	if throughTypeAssertions {
+		entries = testFramework.GetMemberEntriesThroughTypeAssertions(expression)
+	}
 	match := rstestExpectMemberMatch(node, entries, analysis)
```
fansenze: [P2] Preserve import.meta.rstest recognition through type assertions

Member extraction unwraps `as`, but `rstestExpectMemberMatch` still uses `isImportMetaRstestExpectCall` / `parseImportMetaRstestChain`, which stops at that wrapper. As a result, the following chain is discarded and produces no diagnostic:

```ts
(import.meta.rstest.expect(1) as any).resolves.toBe(1);
```

The subject is a number, so this should report `unneededRejectResolve`, just like the same expression without `as any`. Please make import-meta root recognition honor the same wrapper handling as member extraction.

internal/utils/test_framework/member_chain.go
```diff
@@ -11,14 +11,31 @@ import (
...
 	if node == nil {
 		return nil
 	}
-
+	if throughTypeAssertions {
+		switch node.Kind {
+		case ast.KindAsExpression:
+			return getMemberEntries(node.AsAsExpression().Expression, true)
+		case ast.KindTypeAssertionExpression:
+			return getMemberEntries(node.AsTypeAssertion().Expression, true)
```
fansenze: [P2] Include satisfies and non-null assertions in the opt-in traversal

The traversal handles `as` and angle-bracket assertions, but these two runtime-transparent wrappers still bypass the rule:

```ts
(expect(Promise.resolve(1)) satisfies unknown).toBe(1);
expect(Promise.resolve(1))!.toBe(1);
```

Both should report `poorlyExpectedPromise`, just like the expression without the wrapper, but regression probes produce zero diagnostics for each. Please unwrap `satisfies` and `!` consistently in the upward chain traversal and in member/root extraction.

elecmonkey: Fixed in `3189081d`. The opt-in traversal now consistently handles `as`, angle-bracket assertions, `satisfies`, and non-null assertions in both upward chain traversal and downward member/root extraction. Regression coverage includes each wrapper and nested combinations.

internal/plugins/rstest/rules/valid_expect_with_promise/valid_expect_with_promise.go
elecmonkey: Fixed in `3189081d`. The rule now tracks Chai subject transformations in source order, so a Promise selected by `property`, `ownProperty`, or `haveOwnProperty` is checked by subsequent matchers even when no promise modifier is present. The analysis remains a single linear pass and conservatively stops when the transformed subject cannot be determined.

internal/plugins/rstest/utils/parse_rstest_expect.go
elecmonkey: Fixed in `3189081d`. The opt-in expect parser now applies the same runtime-transparent wrapper policy to member extraction, call-chain traversal, root resolution, and `import.meta.rstest` provenance recognition, so the wrapped example reports `unneededRejectResolve`.

fansenze (approved):

PR #2153, opened 2026-09-13
feat(node): port process-exit-as-throw

fansenze:
## Motivation

`process.exit()` currently looks like a returning call to rslint's code-path rules: `no-unreachable` misses following statements, and `consistent-return` can reject a branch that exits the process.

Refs https://github.com/web-infra-dev/rslint/issues/475.

## Changes

- Port eslint-plugin-n v18.3.0 as `node/process-exit-as-throw`, following the existing Node rule prefix. Reuse rslint's ESTree expression helpers for parentheses, optional chains and JavaScript/TypeScript wrappers.
- Carry enabled call predicates through per-file configuration into the existing CFG. The Node rule identifies calls, the CFG handles throw completion, and consumer rules own diagnostics. Update existing CFG consumers and connect `no-unreachable` and `no-fallthrough` without changing the AST or TypeScript binder.
- Preserve all three upstream tests and the documentation example in Go and JS. Add Go regressions for evaluation order, escaped identifiers, catch/finally, nested roots and complete diagnostic ranges. Reuse recorded statement results to avoid repeated root lookup in `no-unreachable`.
- Document observable compatibility limits and examples: `array-callback-return`, `constructor-super`, `no-this-before-super` and third-party ESLint plugins do not yet receive this behavior. Also explain the class-member, loop-source and switch-case differences when comparing with the upstream rule on ESLint 7.

Validation: focused Go tests for the new rule, affected consumers, configuration, linter and catalog passed; scoped golangci-lint, changed-file spelling and formatting checks passed. After rebuilding the native binary, the JS upstream mirror passed 4/4 tests. Differential checks covered two real repository scripts, the original edge corpus and 438 additional file/rule/enablement comparisons, asserting full diagnostic details and accounting for the documented differences. Temporary Go benchmarks covered matching and `no-unreachable` execution; benchmark files are excluded from the commit.


internal/utils/cfg/expressions.go
```diff
@@ -283,7 +283,11 @@ func (b *Builder[E]) accessOrCall(node *ast.Node) {
 		b.expr(tagged.Template)
 	}
 
-	b.firstThrowableFork()
+	if node.Kind == ast.KindCallExpression && b.hooks.CallThrows != nil && b.hooks.CallThrows(node) {
```
elecmonkey, line 286: [P2] Preserve upstream completion for optional exit calls

With `node/process-exit-as-throw` and `no-unreachable` enabled, `process.exit?.(); after();` reports `after()` as unreachable in ESLint 7.32.0 with `eslint-plugin-n@18.3.0`, but Rslint reports nothing; `consistent-return` is likewise affected. The call matches the upstream rule syntactically, but `forkOptionalChain` has already retained a reachable skip edge before `makeThrow()` runs. Please either make a matched optional call terminate that continuation for upstream parity, or document and test this as an intentional semantic difference.

internal/plugins/node/rules/process_exit_as_throw/process_exit_as_throw.md
```diff
@@ -0,0 +1,60 @@
...
+`process.exit()` calls match; `process["exit"]()` and aliases do not. Like upstream,
+the match is syntactic, so a locally declared variable named `process` also matches.
+
+## Differences from upstream
+
+Use `node/process-exit-as-throw` in your rslint configuration instead of
+`n/process-exit-as-throw`. Enabling it affects supported built-in rules;
+it does not change TypeScript errors or diagnostics from third-party ESLint plugins.
+
+`array-callback-return`, `constructor-super`, and `no-this-before-super` do not
```
elecmonkey, line 36: Please also list `require-atomic-updates` as an unsupported consumer for now. For `let x = 0; async function f() { let old = x; await foo(); process.exit(); x = old; }`, enabling the upstream rule removes the `require-atomic-updates` diagnostic because the assignment is unreachable, while Rslint still reports `nonAtomicUpdate`. Its analyzer currently maintains an independent control-flow model and does not consume `ctx.CallThrows`. Integrating it with the shared CFG looks feasible, but preserving its await/yield, scope-escape, and assignment-exit semantics is substantial enough to handle in a separate PR.

elecmonkey:
This rule is not really a “rule” — it merely exists to use the extension point provided by ESLint. I generally agree with keeping it as a rule for now, so that enabling it can reproduce the original plugin’s behavior. However, I suspect there may be a better design for this in Rslint.

fansenze:
Closing because this rule specifically patches ESLint’s shared code-path analyzer, which rslint does not use.

PR #2154, opened 2026-09-13
feat(node): add no-missing-import rule

fansenze:
## Motivation

Add missing-module checks for Node imports and re-exports, as part of [#475](https://github.com/web-infra-dev/rslint/issues/475).

## Changes

- Add `node/no-missing-import`, ported from eslint-plugin-n v18.3.0, with static and literal dynamic imports, re-exports, TypeScript aliases, package entry points and shared Node settings.
- Reuse the Node helpers and tsgo resolver/config parser. Share literal import traversal and cache repeated lookups separately for type and runtime imports.
- Preserve package imports and self-references when external module directories are disabled. Stop at a found package's exports failure instead of accepting another copy from a later module directory or imports-array entry.
- Match TypeScript path wildcards anywhere in the alias and preserve export error details after imports-map redirects. Keep process-relative settings and escaped Unicode filename lookup consistent with upstream.
- Mirror upstream cases and documentation examples in Go/JS, add Go regressions for resolution boundaries, and normalize expected diagnostic paths in the shared Node RuleTester for Windows.
- Describe remaining differences by their user-visible effects: extra `resolverConfig` fields such as `alias` can cause missing-import reports, some invalid imports mappings have different error messages, and unpaired Unicode characters can display differently in diagnostic text.

Validation: the four affected Go packages passed, and scoped Go lint reports 0 issues. After rebuilding the binary, the six affected JS suites passed 220 tests with 3 explained skips (two filename-less cases and a filesystem case-sensitivity case). Real-source differential checks covered 12 files plus a missing-module control and found no differences. The 46 edge comparisons agree on diagnostic counts, IDs and locations; 41 match exactly, with five cases retaining the documented message differences. Formatting and changed-file spelling checks passed. Temporary Go benchmarks were not committed. The Windows path mismatch was confirmed from the failed CI log; a Windows runner must verify the updated assertions.


packages/rslint-test-tools/tests/eslint-plugin-node/rule-tester.ts
```diff
@@ -103,7 +106,11 @@ export class RuleTester {
               expect(diagnostic.ruleName).toBe(`node/${name}`);
               expect(diagnostic.suggestions).toBeUndefined();
               if (typeof expected !== 'string') {
-                expect(diagnostic.message).toBe(expected.message);
+                expect(diagnostic.message).toBe(
```
elecmonkey, line 109: **[P1] Normalize expected diagnostic paths on Windows**

The Windows npm CI job currently fails all 20 invalid `no-missing-import` cases because `root` comes from `path.join` with backslashes while native diagnostics use normalized forward slashes (`C:\...` expected versus `C:/...` received). Since rslint consistently exposes normalized paths in diagnostics, please normalize `root` to forward slashes before passing it to the expected-message callback, for example with `root.replaceAll('\\', '/')`, so these assertions remain portable.

internal/plugins/node/nodeutil/resolution.go
```diff
@@ -51,26 +62,60 @@ func ResolveModule(p *program.Program, name, containingFile string, options Reso
...
+		if tspath.IsExternalModuleNameRelative(name) {
+			// Module search directories govern bare packages, not local files.
+			folders = []string{"node_modules"}
+		}
+		var resolveError string
 		for _, base := range append(slices.Clone(options.Paths), tspath.GetDirectoryPath(containingFile)) {
-			for _, folder := range options.Modules {
+			base = tspath.ResolvePath(p.CurrentDirectory(), base)
+			resolveError = "Can't resolve '" + originalName + "' in '" + base + "'"
+			for _, folder := range folders {
```
elecmonkey, line 82: **[P2] Preserve imports maps when module-directory lookup is disabled**

With `resolverConfig.modules: []`, this loop never invokes the resolver, so package self-references and `package.json#imports` aliases are incorrectly disabled along with external package lookup. In a package containing `"imports":{"#ok":"./local.js"}` and `"exports":{".":"./local.js"}`, upstream accepts both `import '#ok'` and `import 'self-name'`, while rslint reports both missing. Please resolve the current package scope independently of the configured external module-directory list.

internal/plugins/node/nodeutil/resolution.go
```diff
@@ -51,26 +62,60 @@ func ResolveModule(p *program.Program, name, containingFile string, options Reso
...
-				base = tspath.ResolvePath(p.CurrentDirectory(), base)
 				result, _ := resolver.ResolveModuleName(name, tspath.ResolvePath(base, "__import__.js"), core.ResolutionModeCommonJS, nil)
 				if !view.unresolved && result != nil && result.IsResolved() {
-					return view.Realpath(result.ResolvedFileName)
+					if view.builtin {
+						return nodeResolution{}
+					}
+					return nodeResolution{path: view.Realpath(result.ResolvedFileName)}
+				}
+				if view.exportsFile != "" {
```
elecmonkey, line 99: **[P2] Stop after the first module directory finds the package**

After finding a package whose `exports` map rejects the request or points to a missing target, the resolver continues through later `resolverConfig.modules` entries and can incorrectly resolve a different package with the same name. With `modules: ['first_modules', 'second_modules']`, where the first `shadow` package blocks `./private` and the second exports it, upstream reports the first package's exports error while rslint emits no diagnostic. Please treat an exports failure from the first located package as terminal instead of continuing to later module directories.

internal/plugins/node/nodeutil/imports.go
```diff
@@ -75,10 +134,42 @@ func HasTypeScriptAlias(p *program.Program, fileName, name string) bool {
...
+// their upstream contract exempts an alias even when its target is missing.
+func ImportResolveError(p *program.Program, name, fileName string, typeOnly bool, options ResolutionOptions) string {
+	if isNodeBuiltin(name) || !typeOnly && isImportURL(name) {
+		return ""
+	}
+	moduleName, _ := ImportModuleName(name)
+	options.NoDirectory = moduleName == ""
+	if tspath.HasTSFileExtension(fileName) {
+		if config := nearestCompilerOptions(p, fileName); config != nil && config.Paths != nil {
+			for alias, targets := range config.Paths.Entries() {
```
elecmonkey, line 148: **[P2] Support non-trailing wildcards in TypeScript path aliases**

The alias matching only removes a trailing `/*` and then performs exact or slash-prefix matching, so valid single-wildcard mappings elsewhere in the key are ignored. For `paths: {"@pre*-post":["./generated/*/index.ts"]}` with `generated/value/index.ts` present, upstream accepts `import '@prevalue-post'`, but rslint reports it missing. Please match and substitute the wildcard in both the alias key and target rather than treating it as a literal character.

elecmonkey (commented):
The failing Windows npm CI job is not flaky; see the first inline comment for details.

elecmonkey (approved):

PR #2155, opened 2026-09-13
feat(rstest): add unbound-method rule and align @typescript-eslint/unbound-method with upstream

elecmonkey:
## Motivation

Passing a method without its receiver can give it an unintended `this` value when it is called later.

The existing `@typescript-eslint/unbound-method` rule detects these references, but it cannot distinguish Rstest APIs that only inspect a function from APIs and matchers that invoke it.

## Changes

Add the type-aware `rstest/unbound-method` rule as an Rstest adapter over the existing TypeScript rule.

Extract the existing TypeScript rule listeners behind an injectable exemption callback. The original `@typescript-eslint/unbound-method` rule continues to use the same implementation without exemptions, preserving its options, messages, ranges, and destructuring behavior.

Before applying Rstest-specific exemptions, align the TypeScript implementation with computed member access and union/intersection constituent handling. Computed string, template, numeric, and literal-union keys now resolve the corresponding method, while dynamic string keys remain unreported.

Allow an unbound method reference when it is:

- the subject of an Rstest `expect()` or `expect.soft()` assertion whose built-in matcher inspects the value without invoking it, including equality, mock, ordinary snapshot, and Chai-style value matchers;
- passed directly as the first argument to `mocked()` on a proven Rstest utilities object.

Continue reporting method references used with APIs that may invoke the subject, including:

- `toThrow` and `toThrowError`;
- Chai `throw` and `throws`;
- throwing snapshot matchers;
- `rejects`;
- `expect.poll()` and `expect.element()`;
- user-defined, overridden, or otherwise unknown matchers.

Handle same-file `expect.extend()` and `expect.addEqualityTesters()` customization conservatively. An overridden built-in matcher is no longer considered safe, and matchers that pass their received value through custom equality testers remain diagnostic when equality testers are registered.

Resolve Rstest provenance across globals, named and renamed imports, namespace imports, CommonJS bindings, `rstack/test`, `import.meta.rstest`, and Test Context `expect`. Foreign APIs, local shadows, reassigned bindings, and method references outside the direct subject or `mocked()` argument remain subject to the TypeScript rule.

Add focused Go tests covering matcher behavior, matcher and equality customization, `mocked()` provenance, local shadowing, computed members, union and intersection constituents, `this: void`, static methods, and exact diagnostic ranges. Add source-only filtering and compiler-backed capability proofs covering cross-file types, ambient declarations, default-library symbols, narrowing, type-parameter constraints, Promise and thenable detection, awaited types, and compiler options.

Add API and JavaScript integration coverage and user documentation. Register the rule in the Rstest plugin without adding it to the `recommended` preset.

## Credits

Ported from [`eslint-plugin-jest@29.16.0`](https://github.com/jest-community/eslint-plugin-jest/blob/v29.16.0/src/rules/unbound-method.ts).

- **Rstest utilities are resolved by provenance.** The `mocked()` exemption recognizes `rs` and `rstest` globals, renamed imports, namespace and CommonJS forms, `rstack/test`, and `import.meta.rstest`. A local or foreign object with the same name is not treated as Rstest.
- **Rstest matcher semantics determine the exemption.** Ordinary snapshot and Chai value matchers are exempt because they inspect the subject without invoking it. Chai throwing aliases and Rstest throwing snapshot matchers continue to report.
- **Unknown and customized matchers remain diagnostic.** A matcher registered or replaced through `expect.extend()` may invoke the received function. Matchers that use custom equality testers may also execute user code, so affected assertions are not assumed to be safe.
- **Promise modifiers follow Rstest runtime behavior.** `resolves` unwraps the promise without invoking a function subject and therefore follows the safety of the final matcher. `rejects` may invoke a function subject and remains diagnostic.
- **Matcher customization analysis is file-scoped.** Customizations declared in the current source file are detected, while extensions loaded exclusively through another setup file are not visible to the rule.
- **Only direct API arguments are exempt.** A method nested inside another expression or passed in another argument position remains diagnostic.

## Related Links

Related #935.

## Checklist

- [x] Tests updated.
- [x] Documentation updated.


internal/plugins/rstest/rules/unbound_method/unbound_method.go
```diff
@@ -0,0 +1,136 @@
...
+				if accessor.Parent != nil && ast.IsElementAccessExpression(accessor.Parent) {
+					if _, ok := utils.AccessExpressionStaticName(accessor.Parent); !ok {
+						return false
+					}
+				}
+			}
+			for _, modifier := range parsed.Modifiers {
+				if modifier != "not" && modifier != "resolves" {
+					return false
+				}
```
fansenze: [P2] Check `rejects` throughout the assertion chain

`parsed.Modifiers` only contains modifiers before the first matcher, so this Chai chain is incorrectly exempted:

```ts
expect(service.method).to.be.a('function').and.rejects.toBeDefined();
```

Rstest's `rejects` getter calls the subject immediately, so it executes `service.method` without its receiver. The focused rule test reports zero diagnostics. Please inspect the full chain for invoking modifiers, including those after a matcher.

internal/plugins/rstest/rules/unbound_method/unbound_method.go
```diff
@@ -0,0 +1,136 @@
...
+			}
+			for _, matcher := range parsed.Matchers {
+				if analysis.IsExpectMatcherOverridden(matcher.Name) ||
+					(analysis.HasCustomEqualityTesters() && usesCustomEqualityTesters(matcher.Name)) ||
+					!isNonInvokingMatcher(matcher.Name) {
```
fansenze, line 95: [P2] Account for custom asymmetric matchers in equality assertions

`toEqual` can invoke a custom `asymmetricMatch` passed as its expected value, without any `expect.extend()` or `expect.addEqualityTesters()` call:

```ts
expect(service.method).toEqual({
  asymmetricMatch(received) {
    received();
    return true;
  },
});
```

This currently produces no diagnostic even though Rstest executes the unbound method. Please account for custom asymmetric matchers when deciding whether an equality assertion is safe; checking the outer matcher name and file-scoped customization calls misses this path.

internal/plugins/rstest/rules/unbound_method/unbound_method.go
```diff
@@ -0,0 +1,136 @@
...
+			}
+			if ast.IsIdentifier(root) {
+				symbol := ctx.Refs.Resolve(root)
+				stable, checked := stableRoots[symbol]
+				if !checked {
+					stable = true
+					for _, reference := range ctx.Refs.References(symbol) {
+						for reference.Parent != nil && (utils.SkipAssertionsAndParens(reference.Parent) == reference ||
+							(ast.IsAccessExpression(reference.Parent) && reference.Parent.Expression() == reference)) {
+							reference = reference.Parent
```
fansenze: [P2] Unwrap both sides when walking API write references

A normal type cast hides a write from this loop:

```ts
import { rs } from '@rstest/core';
(rs as any).mocked = (fn: () => void) => fn();
rs.mocked(service.method);
```

After advancing to the `as` expression, `SkipAssertionsAndParens(reference.Parent)` still returns the inner `rs` node, which no longer equals `reference`. The walk stops before the `.mocked` assignment target, so the root is cached as stable and the final call is incorrectly exempted (zero diagnostics). Compare the unwrapped parent with the unwrapped reference, as the import-meta write analysis already does.

fansenze (commented):
Found three reproducible false negatives in the Rstest-specific exemptions. The existing tests in the five affected Go packages pass, but each focused reproducer below produces zero diagnostics. The first two cases were also verified against the Rstest runtime.

elecmonkey:
Fixed `rejects` handling across full assertion chains, custom asymmetric matcher detection in equality assertions, and Rstest API write detection through TypeScript wrappers, with regression tests for all three cases.

fansenze:
@elecmonkey Please resolve the merge conflicts. Thanks!

elecmonkey:
Resolved.

fansenze (commented):
The three original examples are now diagnosed correctly. I found three remaining false-negative variants and one new false positive, verified with @rstest/core's actual types and runtime. The existing tests in all five affected Go packages pass.

internal/plugins/rstest/rules/unbound_method/unbound_method.go
```diff
@@ -0,0 +1,221 @@
...
+	node = utils.SkipAssertionsAndParens(node)
+	if node == nil || isSafeBuiltinAsymmetricMatcher(analysis, node) {
+		return false
+	}
+	t := ctx.TypeChecker.GetTypeAtLocation(node)
+	for _, part := range utils.UnionTypeParts(checker.Checker_getApparentType(ctx.TypeChecker, t)) {
+		member := checker.Checker_getPropertyOfType(ctx.TypeChecker, part, "asymmetricMatch")
+		if member != nil {
+			memberType := ctx.TypeChecker.GetTypeOfSymbolAtLocation(member, node)
+			if len(utils.GetCallSignatures(ctx.TypeChecker, memberType)) != 0 {
```
fansenze: [P2] Recognize invoking asymmetric factories even when they return `any`

Rstest declares `expect.toSatisfy(...)` as returning `any`, so this type-based lookup cannot find its `asymmetricMatch` member:

```ts
expect(service.method).toEqual(
  expect.toSatisfy(received => { received(); return true; }),
);
```

This still produces zero diagnostics, while Rstest calls `service.method` without its receiver. Excluding `toSatisfy` from the safe-factory list is insufficient because the subsequent type check falls through. Please treat known invoking factories as unsafe directly, independently of their declared return type.

internal/plugins/rstest/rules/unbound_method/unbound_method.go
```diff
@@ -0,0 +1,221 @@
...
+}
+
+func isSafeBuiltinAsymmetricMatcher(analysis *rstestUtils.RstestCallAnalysis, node *ast.Node) bool {
+	if node.Kind != ast.KindCallExpression {
+		return false
+	}
+	parsed := analysis.ParseExpectCall(node)
+	return parsed != nil && parsed.Entry == rstestUtils.RstestExpectEntryStatic &&
+		rstestUtils.RSTEST_ASYMMETRIC_MATCHERS[parsed.Matcher] &&
+		parsed.Matcher != "toSatisfy" && parsed.Matcher != "schemaMatching"
```
fansenze: [P2] Inspect the samples of `expect.toBeOneOf` before considering it safe

`expect.toBeOneOf` delegates to equality comparisons, which can invoke a custom asymmetric matcher in its sample list:

```ts
const matcher = {
  asymmetricMatch(received: () => void) { received(); return true; },
};
expect(service.method).toEqual(expect.toBeOneOf([matcher]));
```

The name-only allowlist makes `expressionContainsAsymmetricMatcher` return early, so this reports nothing even though the matcher calls the unbound method at runtime. Please inspect the samples of this composite matcher instead of exempting it unconditionally.

internal/plugins/rstest/rules/unbound_method/unbound_method.go
```diff
@@ -0,0 +1,221 @@
...
+				return true
+			}
+		}
+	}
+	switch node.Kind {
+	case ast.KindArrayLiteralExpression:
+		for _, element := range node.AsArrayLiteralExpression().Elements.Nodes {
+			if expressionContainsAsymmetricMatcher(ctx, analysis, element) {
+				return true
+			}
```
fansenze, line 234: [P2] Detect asymmetric matchers in collections stored in variables

The array traversal only handles an inline array literal. Moving the same collection into a variable bypasses it:

```ts
const matchers = [{
  asymmetricMatch(received: () => void) { received(); return true; },
}];
expect(service.method).toBeOneOf(matchers);
```

The argument is now an identifier whose array type has no `asymmetricMatch` property, so the helper returns false and the rule emits zero diagnostics. Rstest still passes the unbound method to the custom matcher. Please follow the collection's value or element type so extracting the array into a variable does not change the result.

internal/plugins/rstest/rules/unbound_method/unbound_method.go
```diff
@@ -0,0 +1,221 @@
...
+			case ast.KindPropertyAssignment:
+				value = property.AsPropertyAssignment().Initializer
+			case ast.KindShorthandPropertyAssignment:
+				value = property.Name()
+			case ast.KindSpreadAssignment:
+				value = property.AsSpreadAssignment().Expression
+			}
+			if expressionContainsAsymmetricMatcher(ctx, analysis, value) {
+				return true
+			}
```
fansenze: [P2] Avoid reporting nested matchers that equality will never execute

Recursively scanning every object-literal value introduces a false positive:

```ts
const matcher = {
  asymmetricMatch(received: () => void) { received(); return true; },
};
expect(service.method).not.toEqual({ nested: matcher });
```

For an ordinary method and object, Rstest detects the Function/Object type mismatch before traversing properties. The assertion passes without invoking either the nested matcher or the method, but the new code reports `service.method`. This case passes before this commit and fails afterward. Please distinguish the values each matcher actually compares: `toEqual` does not descend into this object, whereas `toBeOneOf` compares its sample elements directly.

internal/plugins/rstest/rules/unbound_method/unbound_method.go
elecmonkey: Fixed. Invoking factories are now handled by runtime semantics rather than return types. Added tests for both `toSatisfy` and the same issue in `schemaMatching`.

internal/plugins/rstest/rules/unbound_method/unbound_method.go
elecmonkey: Fixed. `toBeOneOf` now inspects its direct array or `Set` samples while preserving exemptions for safe built-ins such as `expect.any`.

internal/plugins/rstest/rules/unbound_method/unbound_method.go
elecmonkey: Fixed. The analysis now follows stable aliases and spreads, checks array/`Set` element types, and conservatively handles mutated or opaque collections.

internal/plugins/rstest/rules/unbound_method/unbound_method.go
elecmonkey: Fixed. Ordinary equality now checks only values directly compared with the received method, avoiding false positives for nested matchers in objects and arrays.

fansenze (commented):
The four previously reported reproductions now pass on f753d977. The two inline findings below cover one new false positive and one remaining false negative through alias mutations. Both were confirmed with Rstest's actual types and runtime; the existing tests for both unbound-method rule packages pass.

internal/plugins/rstest/rules/unbound_method/unbound_method.go
```diff
@@ -0,0 +1,418 @@
...
+		}
+	}
+	t := analysis.ctx.TypeChecker.GetTypeAtLocation(node)
+	if !utils.IsTypeFlagSet(t, checker.TypeFlagsAny|checker.TypeFlagsUnknown) &&
+		analysis.typeHasCallableAsymmetricMatcher(t, node) {
+		return true
+	}
+	if initializer, resolved, mutated := analysis.localInitializer(node); mutated {
+		return true
```
fansenze: [P2] Preserve exemptions after harmless property writes

Any property write makes `localInitializer` return `mutated`, so this now reports an unbound method for a comparison that cannot call it:

```ts
class Service { method() { return this; } }
const service = new Service();
const expected = { name: 'first' };
expected.name = 'second';
expect(service.method).not.toEqual(expected);
```

Rstest stops at the Function/Object type mismatch, so this assertion passes without invoking the method. This case passes on `3c930086` but fails on this commit. Please preserve the exemption for mutations that cannot introduce an invoking matcher. The equivalent fallback in the collection analysis also incorrectly reports `toBeOneOf` with a `number[]` or `Set<number>` after `push(1)` or `add(1)`.

internal/plugins/rstest/rules/unbound_method/unbound_method.go
```diff
@@ -0,0 +1,418 @@
...
+		}
+	case ast.KindConditionalExpression:
+		conditional := node.AsConditionalExpression()
+		return analysis.collectionMayContainInvokingMatcher(conditional.WhenTrue) ||
+			analysis.collectionMayContainInvokingMatcher(conditional.WhenFalse)
+	}
+	if initializer, resolved, mutated := analysis.localInitializer(node); mutated {
+		return true
+	} else if resolved {
+		return analysis.collectionMayContainInvokingMatcher(initializer)
```
fansenze: [P2] Account for alias mutations before trusting a collection initializer

The mutation check only examines references to the original binding. A write through an alias is missed, so this branch analyzes the original empty array and treats the assertion as safe:

```ts
class Service { method() { return this; } }
const service = new Service();
const matcher = {
  asymmetricMatch(received: () => void) { received(); return true; },
};
const values: (typeof matcher)[] = [];
const alias = values;
alias.push(matcher);
expect(service.method).toBeOneOf(values);
```

This produces zero diagnostics, but Rstest invokes the matcher, which calls `service.method` with `this === undefined`. An aliased `Set.add(matcher)` has the same issue. Please account for writes through aliases before trusting the initializer, and retain the element-type fallback when the initializer no longer proves the collection is safe.

internal/plugins/rstest/rules/unbound_method/unbound_method.go
elecmonkey: Fixed. Harmless property and collection mutations now fall back to static type analysis instead of being treated as unconditionally unsafe. Added regression coverage for object properties, arrays, Sets, and aliases.

internal/plugins/rstest/rules/unbound_method/unbound_method.go
elecmonkey: Fixed. Collection mutation tracking now follows aliases and falls back to element types after the initializer becomes unreliable. Added coverage for arrays, Sets, alias chains, indexed writes, and widened aliases.

fansenze (approved):

PR #2160, opened 2026-09-14
feat(rstest): add prefer-to-have-been-called rule

elecmonkey:
## Motivation

Zero-call assertions are clearer when written with `toHaveBeenCalled()` and `.not.toHaveBeenCalled()`.

## Changes

Add `rstest/prefer-to-have-been-called` with Rstest-aware detection, conservative autofixes, documentation, and tests. Also exclude reassigned or reinitialized `expect` bindings from shared Rstest call analysis.

The rule is not included in the `recommended` preset.

## Credits

Adapted from [`eslint-plugin-jest@29.16.6`'s `prefer-to-have-been-called` rule](https://github.com/jest-community/eslint-plugin-jest/blob/v29.16.6/src/rules/prefer-to-have-been-called.ts).

- Recognizes Rstest globals, imports, CommonJS bindings, `rstack/test`, `import.meta.rstest`, Playwright imports, Test Context `expect`, `expect.soft()`, `expect.poll()`, promise modifiers, and Chai assertion chains.
- Reports `expect.poll()` assertions without autofixes because their callbacks can observe the assertion's negation state through `this`.
- Withholds autofixes when assertion results are consumed or reused, when chains continue after the matcher, or when removing arguments or type arguments would discard comments or observable evaluation.
- Preserves accessor delimiters and optional chaining, removes parenthesized zero arguments completely, and removes explicit matcher type arguments.
- Excludes reassigned and reinitialized Rstest bindings, including destructuring and `for-in` or `for-of` declarations, while preserving `var` redeclarations without initializers.

## Related Links

Related #935.

## Checklist

- [x] Tests updated.
- [x] Documentation updated.


internal/plugins/rstest/rules/prefer_to_have_been_called/prefer_to_have_been_called.md
```diff
@@ -0,0 +1,49 @@
+# rstest/prefer-to-have-been-called
```
elecmonkey:
```suggestion
# prefer-to-have-been-called
```

internal/plugins/rstest/utils/parse_rstest_expect.go
```diff
@@ -742,6 +743,27 @@ func classifyRstestExpectRoot(
...
+		}
+		if initializations > 1 {
+			return rstestExpectRoot{Kind: rstestExpectRootNone}
+		}
+	}
+	if ctx.Refs != nil {
+		for _, reference := range ctx.Refs.References(symbol) {
+			if utils.IsWriteReference(reference) {
+				return rstestExpectRoot{Kind: rstestExpectRootNone}
+			}
```
fansenze, line 766: **[P2] Skip reassigned global `expect` before offering a fix**

This check only runs for resolved symbols. In source-only linting (for example, a file without a tsconfig), a global `expect` has no symbol, so `rstestExpectRootMatch` accepts its name before reaching this check.

With an existing global `expect`, the rule still reports and autofixes this code:

```js
expect = () => ({ toHaveBeenCalledTimes() {} });
expect(null).toHaveBeenCalledTimes(0);
// Autofix: expect(null).not.toHaveBeenCalled();
```

The original call succeeds, but the fix throws `TypeError` because the replacement has no `.not`. I reproduced the generated fix in a source-only Go test and confirmed the runtime failure.

Please also check writes to unresolved global `expect` before accepting it as an Rstest root, and add a source-only regression test.

elecmonkey: Fixed. The 24 additional rules using the shared expect parser also benefit automatically.

elecmonkey: Fixed. The 24 additional rules using the shared expect parser also benefit automatically.

fansenze (approved):

PR #2162, opened 2026-09-14
fix(react): align jsx-no-target-blank with upstream

elecmonkey:
## Motivation

`react/jsx-no-target-blank` reads raw JSX attribute text instead of the value produced by the JSX runtime. Character references can therefore hide `_blank` targets and external URLs, while an encoded safe `noreferrer` token can be reported incorrectly.

The rule also rejects template literals that contain a statically guaranteed security token and may autofix an identifier named `undefined` even when it refers to a shadowed binding. Some string fixes can additionally change the runtime value or produce invalid JSX.

## Changes

- Decode direct JSX string attributes using TypeScript's JSX runtime semantics when checking targets, URLs, and `rel` tokens.
- Recognize complete, statically bounded security tokens across template literal quasis without trusting tokens that interpolations can extend.
- Stop autofixing identifiers named `undefined`, since the binding may be shadowed or otherwise dynamic.
- Serialize direct JSX attributes and JavaScript string expressions safely when adding `rel` tokens, preserving character references, quotes, backslashes, Unicode values, and parseability.
- Split upstream and rslint-specific Go coverage and migrate all 113 canonical cases from `eslint-plugin-react@7.37.5`.
- Add regression coverage for entity decoding, template boundaries, shadowed bindings, and autofix safety.


internal/plugins/react/rules/jsx_no_target_blank/jsx_no_target_blank.go
```diff
@@ -109,6 +122,28 @@ func templateOrStringText(node *ast.Node) (string, bool) {
...
+				tokens = tokens[:len(tokens)-1]
+			}
+			return strings.Join(tokens, " ")
+		}
+		parts := []string{completeTokens(template.Head.Text(), true, false)}
+		for _, span := range template.TemplateSpans.Nodes {
+			literal := span.AsTemplateSpan().Literal
+			parts = append(parts, completeTokens(literal.Text(), false, literal.Kind == ast.KindTemplateTail))
+		}
+		return strings.Join(parts, " "), true
```
fansenze, line 146: **[P2] Check both `_blank` branches before accepting the template**

The new template support exposes a false negative in `relStrings`' matched-condition shortcut:

```jsx
<a
  href="https://example.com"
  target={ok ? "_blank" : "_blank"}
  rel={ok ? `noreferrer ${extra}` : "nofollow"}
/>
```

When `ok` is false, this opens a new tab without `noreferrer`, but the rule now reports nothing. The shortcut stops at the first `_blank` branch, accepts its template, and never checks the unsafe alternate. I verified that this example reports on the base revision and produces no diagnostic on this PR.

Please validate every literal target branch equal to `_blank` (case-insensitively), and add a regression test for this case.

elecmonkey: Fixed in `6f7d6b993`. Both `_blank` branches are now checked case-insensitively, with regression tests added.

fansenze (approved):

PR #2167, opened 2026-09-15
feat(rstest): add max-nested-describe rule

elecmonkey:
## Motivation

Deeply nested suites make test names, reports, and inherited suite configuration harder to understand.

Rstest supports suite forms beyond basic inline `describe` calls, including parameterized and conditional suites, Playwright `test.describe`, and callbacks passed by reference. These forms need Rstest-aware call provenance and callback ownership rather than name-only matching.

## Changes

Add `rstest/max-nested-describe` with a configurable maximum nesting depth. The default maximum is `5`.

The rule recognizes Rstest globals, named and renamed imports, namespace and CommonJS imports, `rstack/test`, `import.meta.rstest`, Playwright `test.describe`, chained modifiers, `describe.each`, `describe.for`, `describe.runIf`, and `describe.skipIf`.

TypeScript call wrappers are treated as transparent. Calls such as `(it as any)(...)`, `it!(...)`, and `(it satisfies Function)(...)` are recognized because TypeScript erases the wrappers and Rstest executes the same registration. Existing tests that treated these forms as unrelated calls are reversed accordingly.

Named callbacks are resolved through the shared Rstest call analysis in both typed and source-only programs. Function-scoped, block-scoped, and class-static-block callbacks participate in suite depth when their binding is stable. Mutable or reassigned bindings are conservatively excluded because their value at the registration point cannot be proven without control-flow analysis.

The framework-independent option handling, depth tracking, diagnostics, and schema are shared with the existing Jest rule while preserving its diagnostic contract.

The rule is registered in the Rstest plugin but is not included in the `recommended` preset because the preferred maximum depth is project-specific.

## Credits

- Adapted from [`eslint-plugin-jest@29.16.0`'s `max-nested-describe` documentation](https://github.com/jest-community/eslint-plugin-jest/blob/v29.16.0/docs/rules/max-nested-describe.md), [source](https://github.com/jest-community/eslint-plugin-jest/blob/v29.16.0/src/rules/max-nested-describe.ts), and tests.

- **Rstest call provenance is resolved.** The rule recognizes Rstest globals, imports, aliases, namespace and CommonJS forms, `rstack/test`, `import.meta.rstest`, and Playwright `test.describe`, while ignoring foreign APIs and locally shadowed functions.

- **Rstest-specific suite APIs are recognized.** Parameterized suites through `describe.each` and `describe.for`, conditional suites through `runIf` and `skipIf`, and Rstest suite modifiers all participate in nesting depth.

- **TypeScript call wrappers preserve registration identity.** Type assertions, non-null assertions, and `satisfies` expressions around a test or suite callee are erased by TypeScript and therefore remain Rstest registrations.

- **Stable named callbacks participate in nesting depth.** Inline callbacks and same-file callbacks referenced by identifier produce the same suite hierarchy at runtime, so both are counted.

- **Mutable callback bindings are excluded.** `let`, `var`, reassigned function declarations, and bindings with assignment or update references do not establish callback ownership. This avoids attributing a registration to a stale initializer without introducing registration-point control-flow analysis.

- **Rstest does not recognize Jest aliases or Vitest's `suite` alias.** `fdescribe`, `xdescribe`, and `suite` are ignored because they are not Rstest APIs.

- **Source-only and typed behavior are aligned.** Same-file callback bindings resolve through the binder-backed reference store even when no TypeChecker is available.

## Related Links

Related #935

## Checklist

- [x] Tests updated.
- [x] Documentation updated.


internal/plugins/rstest/utils/test_callback.go
```diff
@@ -191,14 +206,48 @@ func resolveRstestCallbackArgument(ctx rule.RuleContext, argument *ast.Node) rst
...
+	}
 	switch declaration.Kind {
 	case ast.KindFunctionDeclaration:
 		return rstestCallbackInfo{functionNode: declaration, name: name}
 	case ast.KindVariableDeclaration:
+		// Const is the only variable declaration whose initializer is guaranteed
+		// to remain the value observed by every registration. Mutable bindings
+		// would require registration-point dataflow to resolve safely.
+		if !ast.IsVarConst(declaration) {
+			return rstestCallbackInfo{}
```
fansenze: **[P2] Preserve assertion detection for callbacks that are never reassigned**

This guard also affects `rstest/expect-expect`, which consumes `analysis.TestCallback()`. Returning an empty result makes it report `Test has no assertions` for this valid test:

```ts
let run = () => { expect(true).toBeDefined(); };
test('case', run);
```

The same happens with `var`. Both examples pass on the PR base and fail on this revision, even without enabling `max-nested-describe`.

Excluding mutable bindings from ownership must not turn an unknown callback into an assertion-free test. Please preserve detection for bindings with no writes, or adapt the consumer to handle unresolved callbacks without this false positive.

internal/utils/test_framework/rules/max_nested_describe/max_nested_describe.go
```diff
@@ -0,0 +1,97 @@
...
+				ast.KindCallExpression: func(node *ast.Node) {
+					if !runtime.IsDescribeCall(node) {
+						return
+					}
+					describes = append(describes, node)
+					depth := len(describes)
+					if runtime.DescribeDepth != nil {
+						if semanticDepth := runtime.DescribeDepth(node); semanticDepth > depth {
+							depth = semanticDepth
+						}
```
fansenze: **[P2] Carry the named callback's outer depth into nested syntax**

Taking the maximum of the two depth counts misses a level when a named suite callback contains a wrapped callback:

```ts
function wrap(fn) { return fn; }
function body() {
  describe('middle', wrap(() => {
    describe('inner', () => {});
  }));
}
describe('outer', body);
```

With `{ max: 2 }`, `inner` should report depth **3**, but this currently produces **no diagnostic**. The syntax stack counts `middle → inner` as 2, while ownership skips the wrapped callback and counts `outer → inner` as 2; taking their maximum still yields 2.

Please carry the resolved outer depth through the syntax fallback so this combination is counted correctly, and add the combined case to the regression tests.

fansenze (commented):
Found two reproducible issues. The inline comments include minimal examples and the expected behavior.

fansenze (approved):

PR #2172, opened 2026-09-16
feat: add node/global-require rule

fansenze:
## Motivation

Port `global-require` from [eslint-plugin-n v18.3.0](https://github.com/eslint-community/eslint-plugin-n/blob/v18.3.0/lib/rules/global-require.js) as `node/global-require`. The rule reports dependency loads outside the allowed top-level positions while ignoring locally declared `require` functions.

Refs #475.

## Changes

- Add the rule, catalog entry, documentation and JS integration test using existing rslint reference resolution and tsgo AST helpers.
- Preserve upstream diagnostics across parentheses, optional chains, dynamic imports, shadowed bindings and TypeScript namespace/import declarations.
- Cover all 21 upstream tests and 4 documentation examples, plus 65 Go regression and edge cases. Skip binding resolution for calls already in allowed positions.

Validation: 90 Go rule cases, the catalog registration test, 25 JS integration cases, scoped Go lint, formatting and spell checks. Differential checks against the pinned upstream rule cover 5 real CommonJS files with 8 diagnostics and 73 additional inputs; native disable-directive behavior is also verified. A temporary Go benchmark covered unrelated, top-level, nested, shadowed and mixed calls; its source is excluded from this PR.


elecmonkey (approved):
For `type require = unknown; function f() { require("x"); }`, Rslint reports the call while upstream does not. But Rslint’s behavior is more consistent with TypeScript’s separate type and value namespaces: a type-only declaration does not create a runtime binding, so the call still refers to the global `require`

PR #2173, opened 2026-09-16
feat: add node/exports-style rule

fansenze:
## Motivation

Add CommonJS export-style enforcement as part of the eslint-plugin-n rule migration. Refs #475.

## Changes

- Add `node/exports-style`, ported from eslint-plugin-n v18.3.0, with both export styles, `allowBatchAssign`, and matching diagnostics and ranges.
- Intentionally provide no automatic fixes or suggestions. Expanding an exported object can change its identity, compound assignment behavior and method semantics. Even replacing a member receiver can target the wrong object when `exports` is shadowed or either export reference has changed, including through indirect module access.
- Reuse the shared scope-analysis cache and tsgo's identifier index. Keep the export-style decisions within the rule package.
- Preserve all 44 upstream tests and 6 documentation examples in Go and JS, with fix expectations adjusted for the documented safety difference. Add 112 Go edge cases plus edit-demand and schema checks. Document the actual behavior difference from upstream.

## Validation

- `go test -json ./internal/plugins/node/rules/exports_style`: 50 upstream/documentation cases and 112 extras passed, together with edit-demand and schema checks.
- `golangci-lint run --new-from-merge-base=origin/main ./internal/plugins/node ./internal/plugins/node/rules/exports_style`: 0 issues.
- Rebuilt the native binary, rule schemas and JS artifacts. Targeted JS tests for exports-style, global-require and no-missing-import: 154 passed, with 3 existing no-missing-import skips (two synthetic `<input>` filenames and filesystem case sensitivity). All 50 exports-style cases passed without skips.
- Compared diagnostics and ranges with the pinned upstream rule on 116 edge/option combinations and 15 real files from graceful-fs, enhanced-resolve and picomatch, plus one known-positive sample. All four style/batch-option combinations matched. Also asserted that no fixes or suggestions are emitted and `--fix` leaves every source unchanged.
- Executed 23 CommonJS examples before and after `--fix`; all retained their source and runtime results, including object identity, `super`, shadowing, reassignment, `require.cache`, `eval` and wrapper-argument cases.
- Reran temporary Go benchmarks for eight workloads. Benchmark files are not committed.
- Changed-file spell check, `pnpm run format:check`, Go formatting and `git diff --check` passed.


internal/plugins/node/rules/exports_style/exports_style.go
```diff
@@ -0,0 +1,338 @@
...
+		node = parent
+	}
+	return node
+}
+
+func fixModuleExports(ctx rule.RuleContext, node *ast.Node) []rule.RuleFix {
+	if memberParent(node) != nil {
+		return []rule.RuleFix{rule.RuleFixReplace(ctx.SourceFile, node, "exports")}
+	}
+	assignment := utils.ESTreeParent(node)
```
elecmonkey: **[P1] Avoid a semantics-changing object-literal autofix**

Upstream `eslint-plugin-n@18.3.0` has the same behavior, but this autofix is not semantics-preserving. For example, `const old = exports; module.exports = { a: 1 };` makes `old !== module.exports`, whereas the generated `exports.a = 1;` mutates the old object and leaves `old === module.exports`. The empty-object case is fixed from `module.exports = {};` to just `;`, and compound assignments such as `module.exports += { a: 1 }` are also incorrectly rewritten as property assignments. Please suppress object-literal expansion unless equivalence can be guaranteed; at minimum, non-`=` assignments must not enter this path. Fixing this would intentionally and safely diverge from upstream.

elecmonkey (approved):

PR #2174, opened 2026-09-16
feat: add node/handle-callback-err

fansenze:
## Motivation

Add callback error-parameter checks to the Node.js rules. Part of #475.

## Changes

- Port `handle-callback-err` from eslint-plugin-n v18.3.0 as `node/handle-callback-err`, including its literal/regular-expression option and upstream diagnostic messages and ranges.
- Reuse binding utilities, reference tracking, ECMAScript regular expressions and tsgo scanning. Keep initialization-write and AST-range adaptations within the rule, with coverage for destructuring, shadowing, decorators, TypeScript and JSX.
- Add all upstream tests and documentation examples, separate Go regressions, JS integration coverage and user-facing documentation of regular-expression differences. Validate caret-prefixed patterns with the existing Unicode regexp schema format, so invalid patterns fail during configuration loading while literal names remain unrestricted strings.
- Use temporary Go benchmarks to remove unnecessary reference collection and repeated method scanning. Benchmark files are not included.

## Validation

- `go test ./internal/plugins/node/rules/handle_callback_err` — 112 rule cases and 16 option-validation cases.
- `golangci-lint run --new-from-merge-base=origin/main ./internal/plugins/node ./internal/plugins/node/rules/handle_callback_err` — 0 issues.
- Rebuilt the native binary, regenerated rule schemas and built `@rslint/core` JS/public option types.
- `CI=true pnpm --dir packages/rslint-test-tools exec rs test run tests/eslint-plugin-node/rules/handle-callback-err.test.ts` — 52 passed, none skipped.
- Compared diagnostics with ESLint on 12 real source files under two configurations, positive controls and 36 additional semantic inputs; no differences in those comparisons.
- Verified the rebuilt public API rejects invalid patterns before checking callbacks and accepts literal names and supported patterns (9 configuration checks).
- Temporary `go test -bench` measurements across seven workloads, changed-file spell check, `pnpm run format:check` and `git diff --check` passed.


elecmonkey:
No issues in this rule's implementation. There's a shared regex validation issue: `^]` is invalid in JavaScript’s `u` mode, so upstream throws an error, but Rslint’s `regex-u` validator accepts it and the rule silently reports nothing.

This comes from Rslint’s shared regex infrastructure and also affects other rules. If you think it is worth addressing, I suggest handling it in a separate PR.

elecmonkey (approved):

PR #2177, opened 2026-09-16
feat: add node/no-mixed-requires rule

fansenze:
## Motivation

Add `node/no-mixed-requires` to keep CommonJS module loading separate from ordinary variable declarations, with optional grouping by module kind.

Refs #475.

## Changes

- Port `eslint-plugin-n` v18.3.0 behavior, including `grouping`, `allowCall`, legacy boolean options, both messages, and complete declaration ranges.
- Reuse existing AST and reporting helpers. Keep syntax classification local to the rule, skip single-variable declarations, and stop scanning once a mixed declaration is established.
- Add the options schema, rule documentation, catalog registration, and JS integration coverage. Preserve all 23 upstream tests and 6 documentation examples, plus 58 Go edge cases for optional chains, parentheses, decoded strings, JSDoc, TypeScript, and diagnostic ranges.

Validation passed:

- `go test ./internal/plugins/node/rules/no_mixed_requires` — 87 cases.
- `CI=true pnpm --dir packages/rslint-test-tools exec rs test run tests/eslint-plugin-node/rules/no-mixed-requires.test.ts` — 29 tests, no skips.
- `golangci-lint run --new-from-merge-base=origin/main ./internal/plugins/node ./internal/plugins/node/rules/no_mixed_requires`.
- Native and JS builds, schema generation, `pnpm run format:check`, and spell-checking all changed text files.
- Differential validation against the pinned rule: 18 real CommonJS files plus a known positive input, and 253 generated edge cases, each under all four option combinations. Paths, message IDs/text, severity, full ranges, and absence of fixes/suggestions match.

Temporary listener-only Go benchmarks on Apple M5 Max, Go 1.26.5 (`-cpu=1 -benchmem -benchtime=300ms -count=7`): median single-variable handling is 1.68 ns/op with zero allocations; a grouped pair is 20.34 ns/op with zero allocations; a mixed pair reports in 67.18 ns/op, 160 B/op, 1 allocation. Parsing and setup are excluded. Benchmark source files are not included.


elecmonkey (approved):
no issues found

PR #2179, opened 2026-09-16
feat: add node/no-top-level-await rule

fansenze:
## Summary

Add `node/no-top-level-await` from eslint-plugin-n v18.3.0 to report top-level `await`, `for await`, and `await using` in published modules. Support `ignoreBin`, `convertPath`, and shared settings using existing package and path helpers.

- Reuse `nodeutil.ConvertPath`, including main's shared `CompileGlob` implementation.
- Keep publication relative to the source package after path conversion, including nested package metadata, scoped ignore files, and filesystem case sensitivity. `node/hashbang` shares the corrected publication policy.
- Port all 36 upstream cases and two documentation examples. Add Go regressions for function boundaries, diagnostic ranges, publication paths, and documented rewrites.
- Document remaining differences with concrete code and configuration examples. Some await expressions in `.js` files without import/export can be missed; parenthesized awaits in computed property names can also be missed; `await !value` in JavaScript can produce a syntax error. The documentation gives working alternatives. Two additional cases remain explicitly skipped, and no upstream cases are skipped.

## Validation

- `go test ./internal/utils/gitignore ./internal/plugins/node/nodeutil ./internal/plugins/node/rules/no_top_level_await ./internal/plugins/node/rules/hashbang ./internal/plugins/node/rules/no_restricted_import`
- `go test ./internal/plugins/node/rules/no_top_level_await -run '^TestNoTopLevelAwaitASTBoundaries$'`
- `go test -race ./internal/plugins/node/nodeutil -run '^TestPublicationGenerationAndConfiguration$'`
- `pnpm --filter @rslint/core build:bin`
- `CI=true pnpm --dir packages/rslint-test-tools exec rs test run tests/eslint-plugin-node/rules/no-top-level-await.test.ts tests/eslint-plugin-node/rules/hashbang.test.ts tests/eslint-plugin-node/rules/no-restricted-import.test.ts` — 148 passed.
- `golangci-lint run --new-from-merge-base=origin/main ./internal/utils/gitignore ./internal/plugins/node ./internal/plugins/node/nodeutil ./internal/plugins/node/rules/hashbang ./internal/plugins/node/rules/no_top_level_await` — no issues.
- `pnpm run format:check` and `pnpm run check-spell` with changed text files explicitly selected — passed.
- Publication fixtures exercise POSIX, Windows drive, and UNC paths under both casing modes, mixed separators, scoped ignores, CRLF, Unicode directories, and package-boundary escapes. Native macOS smoke checks and seven `npm pack --dry-run` comparisons verified publication behavior.
- Native API comparisons confirm the documented JavaScript limitations and the `export {}` / `.mjs` alternatives.
- Temporary Go benchmarks checked ordinary, converted, and ignored files. Benchmark and profiling helpers are not included.

Rebased onto the latest fetched `origin/main` (`eaa9cc7db`); retained main's new Node utilities and rule registrations.


internal/plugins/node/rules/no_top_level_await/no_top_level_await.go
```diff
@@ -0,0 +1,82 @@
...
+				}
+			}
+			start := scanner.GetTokenPosOfNode(node, ctx.SourceFile, false)
+			ctx.ReportRange(node.Loc.WithPos(start), rule.RuleMessage{
+				Id:          "forbidden",
+				Description: "Top-level `await` is forbidden in published modules.",
+			})
+		}
+		return rule.RuleListeners{
+			ast.KindAwaitExpression: report,
```
elecmonkey, line 64: [P2] This rule misses valid top-level `await` forms in source-only `.js` modules. Because it only listens for `AwaitExpression`, inputs like `await (value);`, `await [value];`, `await ({ value });`, `await !value;`, and ``await `value`;`` produce no diagnostic when `input.js` is configured with `languageOptions.sourceType: "module"`. Upstream `eslint-plugin-n@18.3.0` reports them, and Rslint starts reporting once the file has `export {}` or uses `.mjs`, so the gap is that the source-only parser path is not honoring the effective module source type. Please align source-only parsing with `sourceType: "module"` so these awaits are visible to the rule.

elecmonkey (approved):
Confirmed: source-only `.js` issue are explicitly ignored in this PR. However, this may not be reasonable: the long-term direction should be for source-only Program creation to respect the effective source type.

PR #2180, opened 2026-09-16
feat: add node/no-process-env rule

fansenze:
## Motivation

Add `node/no-process-env` to discourage direct environment-variable access throughout a project and support keeping configuration in one place.

Part of #475.

## Changes

- Port `eslint-plugin-n` v18.3.0 behavior, including literal bracket access, `allowedVariables`, and the original diagnostic message and ranges.
- Register the rule and option schema, add documentation, and mirror every upstream test and documentation example in the Go and JS suites.
- Reuse existing AST utilities for parentheses, optional chains, JSX, private properties and TypeScript syntax.

## Validation

- Rule package: 51 cases passed.
- Rule catalog registration test passed.
- JS integration: 20 cases passed with `CI=true`.
- Differential checks against the pinned upstream rule: 447 parseable syntax cases and three real source files, each with default and allowlist configurations; messages, IDs, severity, ranges and edit artifacts matched.
- Temporary Go benchmarks covered eight workloads with six samples each. For 64 repeated code fragments per operation, current medians are 2.08 µs/op for an allowed access and 5.39 µs/op for a reported dot access (rule setup and listeners, parsing excluded). Benchmark source was removed.
- Scoped Go lint, spelling and formatting checks passed.


elecmonkey (approved):
The only known difference is this case:

```js
class C {
  #env;
  f(process) {
    process?.#env;
  }
}
```

ts-go treats it as a syntax error (`TS18030`), so the rule cannot run on it. This is not an Rslint rule issue. LGTM.

PR #2181, opened 2026-09-16
feat: add node/no-unpublished-bin rule

fansenze:
## Summary

Add `node/no-unpublished-bin` to report package executables excluded by `files`, `.npmignore`, or the applicable `.gitignore`. Supports `convertPath` options and shared settings. Part of #475.

Reuse the existing package and publication helpers. Converted executables follow the declaring package's publication settings, including subdirectory ignore files. Match executable aliases and `main` using the filesystem's case rules; cover the shared `hashbang` and `no-top-level-await` behavior with regressions.

Reject invalid conversion regexes in rule options as configuration errors, while accepting JavaScript lookarounds, named captures and literal replacement text. Invalid conversion regexes in shared settings skip the check, as documented in Differences.

## Validation

- Targeted Go helper/rule tests passed on macOS and Linux. All 147 JS cases across the three affected rules passed with no skips.
- Preserved all 39 upstream cases plus the documentation example; compared fixture metadata and verified that the only schema changes are the two regex checks.
- Passed 96 complete-rule path attacks on each of macOS and Linux: POSIX, drive and UNC paths; both case modes; Unicode, aliases, package boundaries and CRLF ignore files. Eight additional attacks passed on the actual macOS filesystem.
- Passed 36 public API scenarios and 17 pinned-upstream comparisons. All six new schema mutations were detected by test assertions; the previous eight runtime mutation checks remain applicable.
- Native/schema/JS builds, rule registration, scoped Go lint, explicit-path spelling and formatting checks passed. Windows amd64 test binaries compile; native Windows execution runs in CI.

Temporary Go benchmarks exercised 12 scenarios with diagnostic assertions. No benchmark files are included and no performance improvement is claimed.


internal/plugins/node/rules/no_unpublished_bin/no_unpublished_bin.go
```diff
@@ -0,0 +1,55 @@
...
+//go:embed no_unpublished_bin.schema.json
+var schemaJSON []byte
+
+// https://github.com/eslint-community/eslint-plugin-n/blob/v18.3.0/lib/rules/no-unpublished-bin.js
+var NoUnpublishedBinRule = rule.Rule{
+	Name:   "node/no-unpublished-bin",
+	Schema: rule.NewSchema(schemaJSON),
+	Run: func(ctx rule.RuleContext, options []any) rule.RuleListeners {
+		program := ctx.Program()
+		if program == nil || ctx.SourceFile.FileName() == "<input>" {
```
elecmonkey, line 20: This only skips the literal `"<input>"`, but Rslint's public `lintText(code)` path uses an absolute synthetic filename like `<cwd>/__text__.ts` when the caller does not provide `filePath`. That synthetic filename can still find the caller's `package.json`, pass through `convertPath`, match a `bin` entry, and be reported as unpublished even though the user is linting unnamed text. Upstream receives `"<input>"` for this case and skips the rule, which is the safer behavior for a rule that depends on real package publication metadata. Please make the unnamed `lintText` path skip this rule as well, rather than treating `__text__.ts` as a real executable.

internal/plugins/node/nodeutil/paths.go
```diff
@@ -66,19 +66,22 @@ func (pkg *PackageJSON) Directory() string { return pkg.directory }
...
-func isBinFile(fileName string, bin any, directory string) bool {
+func isBinFile(fileName string, bin any, directory string, caseSensitive bool) bool {
+	comparison := tspath.ComparePathsOptions{UseCaseSensitiveFileNames: caseSensitive}
 	match := func(value any) bool {
 		name, ok := value.(string)
 		if !ok {
 			return false
 		}
 		resolved := tspath.ResolvePath(directory, name)
 		withoutJS := strings.TrimSuffix(fileName, ".js")
```
elecmonkey: The case-insensitive comparison here is still incomplete because the alias normalization happens with case-sensitive suffix stripping before `ComparePaths` runs. On a case-insensitive filesystem, `bin: "bin/cli"` should match both `BIN/CLI.JS` and `BIN/CLI/INDEX.JS`, but `strings.TrimSuffix(fileName, ".js")` and `strings.TrimSuffix(withoutJS, "/index")` do not remove uppercase `.JS` or `/INDEX`, so those executable aliases are missed. This affects both `node/no-unpublished-bin` and `node/hashbang`; upstream has the same bug, but this PR is already making filesystem-case behavior more correct, so the suffix alias handling should be made case-aware too.

elecmonkey:
Seems unnamed lintText inputs issue are fixed by propagating `UnnamedInput`, but later removed. Currently `lintText(code)` without `filePath` can still report, unlike upstream’s behavior. Could you please clarify whether this was an intentional trade-off?

fansenze:
Regarding [the unnamed `lintText` comment](https://github.com/web-infra-dev/rslint/pull/2181#issuecomment-5698730124):

The removal of `UnnamedInput` was intentional to keep this PR focused on the rule.

I rechecked with ESLint 10.2.1 and eslint-plugin-n 18.3.0: **both rslint’s and ESLint’s `lintText()` can report unnamed input**. For example, when `convertPath` maps that input to an ignored `bin` file, both report it. The automatic skip applies to `Linter.verify()` without a filename, which supplies `<input>`, not to `ESLint.lintText()`.

So this does not demonstrate a missing check in the rule’s implementation. Making unnamed text always skip this check would be a separate API behavior change, which I’ve left outside this PR.

elecmonkey (approved):

PR #2182, opened 2026-09-16
feat: add node/no-sync rule

fansenze:
## Motivation

Add `node/no-sync` so projects can flag synchronous calls and configure exceptions for initialization code or selected APIs.

Refs https://github.com/web-infra-dev/rslint/issues/475.

## Changes

- Port eslint-plugin-n v18.3.0 behavior, including `allowAtRootLevel`, string ignores, and declaration-based file, library, and package ignores.
- Reuse existing AST, type, path, glob, and JavaScript regexp helpers. Share function-range handling with `node/handle-callback-err` and cache declaration matches within each file's rule run.
- Mirror upstream tests and documentation examples in Go and JS. Add Go coverage for optional chains, computed keys, destructuring, JSX, decorators, anonymous/private symbols, aliases, and cache isolation.
- Preserve whitespace in file glob patterns, use tsgo path containment to distinguish sibling directories, and cover Windows drive letters, UNC paths, filesystem case sensitivity, type roots, and linked packages.
- Reject invalid package regular expressions during option validation using the existing Unicode regexp format.
- Document configuration and user-visible compatibility differences, including file glob examples, directory boundaries, and working file-pattern ignores on Windows and under working directories containing uppercase letters.

Validation:

- Focused Go rule, declaration-matching, and rule-catalog tests passed.
- Rebuilt native, schema, and JS artifacts; the no-sync JS integration suite passed with 36 tests passing and 3 explained skips for upstream JavaScript dependency injection. Four invalid package-regexp configurations were also rejected through the native JS API.
- Compared three real repository scripts against the pinned upstream rule under default, root-level, and string-ignore configurations; all diagnostic fields matched. Additional AST and declaration-type probes matched in 1,246 comparisons. A 1,680-case file-glob comparison identified the advanced syntax differences documented with regression examples.
- Scoped Go lint, formatting, and explicit-path spell checks passed. The 48-case path/glob matrix, 12 regexp-option cases, and linked-package regressions passed. Temporary Go benchmarks were run and removed from the change.


internal/utils/type_matches_specifier.go
```diff
@@ -376,6 +380,67 @@ func typeMatchesSpecifier(
...
+	canonical := func(name string) string {
+		return tspath.GetCanonicalFileName(tspath.NormalizePath(name), program.FS().UseCaseSensitiveFileNames())
+	}
+	cwd := canonical(program.CurrentDirectory())
+	typeRoots, _ := program.Options().GetEffectiveTypeRoots(program.CurrentDirectory())
+	// Picomatch treats whitespace in a configured pattern literally.
+	globOptions := minimatch3.Options{PreserveWhitespace: true}
+	for _, file := range files {
+		name := canonical(file.FileName())
+		if Some(typeRoots, func(root string) bool { return strings.HasPrefix(name, canonical(root)) }) {
```
elecmonkey: [P2] Use path containment instead of prefix checks for declaration files

This path check should use a real path-containment predicate instead of `strings.HasPrefix`. With `typeRoots: ["/repo/project/types"]`, a local declaration in `/repo/project/types-extra/file.ts` is currently treated as if it were inside the configured type root, so `{ from: "file" }` fails to ignore a normal project declaration. The same issue also applies to the `cwd` check below: `/repo/project-extra/file.ts` is considered inside `/repo/project`. Upstream `ts-declaration-location@1.0.6` has the same bug, but Rslint already has `tspath.ContainsPath`, so repairing this would be an intentional but more correct difference.

internal/plugins/node/rules/no_sync/no_sync.schema.json
```diff
@@ -0,0 +1,79 @@
...
+                },
+                "additionalProperties": false
+              },
+              {
+                "type": "object",
+                "properties": {
+                  "from": {
+                    "enum": ["package"]
+                  },
+                  "package": {
```
elecmonkey, line 57: [P2] Reject invalid package ignore regexps during option validation

`package` is later compiled as a JavaScript Unicode regexp, so the schema should reject invalid regexp strings here, for example with the existing `regex-u` format. With `ignores: [{ from: "package", package: "(" }]`, upstream `eslint-plugin-n@18.3.0` throws `Invalid regular expression: /(|(/u`, but Rslint accepts the config, caches a nil matcher, and silently behaves as if the ignore did not match. That makes a configuration typo look like an ordinary lint report instead of a configuration error.

elecmonkey (dismissed):

elecmonkey (dismissed):

PR #2183, opened 2026-09-16
feat: add node/no-deprecated-api rule

fansenze:
## Motivation

Add Node.js deprecation diagnostics for APIs such as `fs.exists()` and `new Buffer(10)`.

Related to [#475](https://github.com/web-infra-dev/rslint/issues/475).

## Changes

- Port `eslint-plugin-n` v18.3.0 as `node/no-deprecated-api`, covering configured globals, CommonJS, ESM, re-exports, `process.getBuiltinModule()`, aliases and destructuring. Support version-dependent replacement advice and separate global/module ignore lists.
- Extract and extend the existing require collector's API value traversal so require collection, module-property checks and deprecation checks share one implementation. Reuse `RuleContext.Refs`, its per-file cache, rslint's binding/reference/static-evaluation helpers and tsgo's destructuring and semver helpers. Keep API metadata and diagnostic policy in the rule; prepare messages and replacement checks only for reported APIs.
- Mirror all 138 upstream tests and 8 documentation examples in Go and JS, with additional Go regressions for language/AST boundaries and version configuration. Match upstream's alphanumeric prereleases and wildcard-order validation, including its caret, tilde and hyphen-range exceptions.
- Scope schema spelling exclusions to core, plugin and shared test-framework rule directories, retaining the existing generated catalog exclusion. Keep other schemas spell-checked, and exclude Go dependency manifests through the shared configuration without adding API identifiers to the dictionary or spelling directives to `go.mod`.
- Make exits explicit after five nil-failure assertions in existing tests flagged by CI's staticcheck.

## Validation

- Scoped Go tests passed for `nodeutil`, `no_deprecated_api`, `no_extraneous_require` and `no_path_concat`.
- Scoped tests passed for the affected LSP, reference-store, compiler-compatibility and rstest checks.
- Rebuilt core artifacts; all three affected JS rule suites passed: 208 tests, no skips.
- Compared against the pinned reference on 4 real source files (8 diagnostics) and 31 additional JS/TS scenarios (45 diagnostics), including complete diagnostic ranges. Real-source checks also compared edits and suggestions.
- Matched all 478 version-range cases against npm semver 7.8.5, with zero skips, plus 8 full rule-diagnostic comparisons for version options.
- Scoped Go lint, changed-file spelling and repository formatting checks passed.
- Ran temporary Go benchmarks for unrelated code, supported APIs, repeated deprecated calls and real `safe-buffer` source. Benchmark files are excluded from this PR.

Rebased onto `origin/main` at `289a759d` (`node/no-top-level-await`).


elecmonkey:
Not a rule impl issue: could you re-consider ignoring spell-check for `*schema*`? Many of them contain natural-language descriptions, so a narrower ignore may better.

fansenze:
@elecmonkey Agreed, `**/*schema*.json` was too broad. I've narrowed it to `.schema.json` files under the core, plugin, and shared test-framework rule directories, keeping the existing exclusion for the generated `packages/rslint/rule-schemas.json`. Other schemas, including `rslint-schema.json`, remain spell-checked.

Many rule schemas come from upstream and contain API names and option values that are not ordinary words. I think ignoring rule schemas is the more appropriate scope: it keeps those identifiers out of the shared dictionary and reduces repeated edits to that file, which can otherwise cause merge conflicts as more rules are ported.

elecmonkey (approved):

PR #2187, opened 2026-09-16
fix(playground): allow import cdn scripts in worker-src CSP

swwind:
## Motivation

Firefox seems have more strict security guards, anything import from a worker follows `worker-src` rules instead of `script-src`. So currently our playground is not working on firefox.

### Refs

- https://developer.mozilla.org/en-US/docs/Web/API/Worker/Worker
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/worker-src

## Changes

Changes are trivial.


elecmonkey:
@codex review

fansenze (approved):

PR #2188, opened 2026-09-16
feat: add node/file-extension-in-import rule

fansenze:
## Motivation

Add `node/file-extension-in-import` so projects can enforce consistent extensions in imports and re-exports.

## Changes

- Port eslint-plugin-n v18.3.0 behavior for `always` / `never`, per-extension overrides, directory indexes, and TypeScript extension mappings.
- Reuse the Node import visitor/resolver and tsgo configuration and extension utilities. Share forward and reverse extension mappings in `nodeutil`; keep style decisions and fixes in the rule.
- Cache directory entries per Program generation and retain ambiguity checks for broken symlinks.
- Preserve escape spelling when removing extensions by reusing the existing JavaScript string source mapping. With `never`, `import './\u0061.\x6as'` safely becomes `import './\u0061'`. Omit edits for incomplete strings, non-string arguments, unsafe custom extensions, and Unix backslash paths that cannot be checked reliably.
- Keep query, fragment, and loader suffixes outside extension checks and edits: `./a?raw` becomes `./a.js?raw`, while `never` changes `./a.js?raw` to `./a?raw`. Fixes preserve suffix spelling and stop after one round.
- Include all 62 upstream tests and six documentation examples in Go and JS, plus Go regressions for escaped paths, resource suffixes, AST forms, filesystem behavior, Windows separators, and edit demand. Document observable differences with user-facing examples.

## Validation

- Scoped Go tests for the new rule, Node utilities, affected import/require consumers, and rule registration.
- `golangci-lint run --new-from-merge-base=origin/main ./internal/plugins/node/rules/file_extension_in_import`: no issues.
- Rebuilt the Go binary and generated JS artifacts. The new JS suite passes all 68 cases; affected existing wrapper suites also passed.
- Pinned-upstream comparison on 38 real source files across three configurations: all diagnostics, ranges, fixes, and suggestions match. An additional 155-case boundary comparison has only the 23 explicitly asserted, documented differences, including corrected handling of resource suffixes.
- 81 resource-suffix regression cases check exact fixes and clean subsequent linting. Adversarial checks cover 13,824 import/export variants: every fix stays inside the literal, preserves the intended decoded path, produces parseable code, and is idempotent under a second lint pass.
- Temporary Go benchmarks cover seven scenarios and an additional escaped-fix workload; benchmark files are excluded from this PR.
- Rule tests pass on macOS arm64 and Linux arm64 (Go 1.26.0 in Docker). Windows amd64 test compilation passes; the new Windows-specific cases have not been run locally.
- Formatting, changed-file spelling, and `git diff --check` pass.


internal/plugins/node/rules/file_extension_in_import/file_extension_in_import.go
```diff
@@ -0,0 +1,198 @@
...
+						return nil
+					}
+					text := expectedExt
+					if isDirectory {
+						text = "index" + text
+						if !strings.HasSuffix(name, "/") && !strings.HasSuffix(name, `\`) {
+							text = "/" + text
+						}
+					}
+					position := source.End() - 1
```
elecmonkey, line 168: Rslint already separates resource suffixes from the file path during import resolution, but this rule does not use the same boundary when generating fixes.

For `import './a!loader'`, resolution checks `./a` and correctly determines that `.js` is missing. The fixer, however, appends `.js` after the loader suffix, producing `./a!loader.js` instead of `./a.js!loader`. On every subsequent fix round, resolution still checks `./a`, so another `.js` is appended. After the ten-round limit, the result is `./a!loader.js.js.js.js.js.js.js.js.js.js`.

Query suffixes expose the same inconsistency. `import './a?raw'` is fixed to `import './a?raw.js?raw'` instead of `import './a.js?raw'`. With `"never"`, `import './a.js?raw'` is fixed to `import './a'` instead of `import './a?raw'`, dropping the query entirely.

The rule should split the specifier into its file path and resource suffix once, then use that split consistently for resolution, extension checks, and fixes.

elecmonkey: Simply put, upstream does not consider loaders or query strings at al. Rslint considers them during detection but not during fixing. This makes Rslint’s own behavior internally inconsistent.

fansenze: Fixed

elecmonkey (dismissed):

elecmonkey (approved):

PR #2189, opened 2026-09-16
feat: add node/no-restricted-require rule

fansenze:
## Motivation

Add `node/no-restricted-require` so projects can prohibit CommonJS modules or resolved file paths, with custom messages and ordered exclusions.

## Changes

- Port eslint-plugin-n 18.3.0 behavior for `require()`, `require.resolve()`, aliases and constant arguments, with schema, public option types and documentation.
- Share restriction matching with `node/no-restricted-import` and reuse the existing require tracker, RuleContext services, static evaluator, tsgo identifier table and cached runtime resolver.
- Support resolver aliases, `mainFields`, `mainFiles`, `aliasFields`, and explicit extension/export-condition overrides in the shared Node resolver. Preserve TypeScript defaults unless overridden, resolve file aliases before symlinks, and retain query/fragment suffixes for path restrictions.
- Evaluate constant module names with JavaScript coercion, arithmetic and comparison semantics, reusing tsgo number operations. Handle huge BigInt exponents with bases `0n`, `1n` and `-1n` while bounding calculations that grow the result.
- Preserve 30 upstream cases and 10 documentation examples/configurations; add 97 Go edge cases plus schema validation. Document remaining user-visible limits and the intentional correction for false wildcard aliases that upstream can apply to unrelated modules.

## Validation

- Targeted Go tests, catalog registration and scoped Go lint pass. The review follow-up passed tests in all 27 traced root-module packages plus the new number shim; scoped lint reported zero issues. The branch is rebased onto `origin/main` (`3229314c`).
- Native binary, schema and JS builds pass. The review follow-up rebuilt the native binary and passed 364 tests across 23 affected JS integration files. Earlier resolver integration coverage passed 213 tests across five files, with three existing `no-missing-import` skips.
- Entry-option review: 1,296 resolver combinations and 61 rule-level cases (35 diagnostics) match the pinned reference. Another 92 fixture cases cover entry ordering, nested fields, false mappings, cycles, ESM overrides, exports, extension aliases and POSIX/drive/UNC paths. Real symlink cases pass. Cycles remain errors within an entry/alias chain; independent resolvePaths directories still get their own lookup after an error or ignored target.
- Differential review: 3,036 adversarial expressions (1,960 diagnostics), 86 generated cases (97 diagnostics), and five real repository scripts under two configurations (10 diagnostics) match the pinned upstream reference, including complete diagnostic ranges and edit fields.
- Review follow-up: 21,459 generated binary expressions match the pinned constant evaluator; eight rule-level files produce the same 16 complete diagnostics. Aggregate identity comparisons and fractional-exponent approximation cases remain outside that expression matrix. Extreme BigInt decimal conversion was benchmarked; the existing resource bound remains intentional.
- POSIX, drive-letter, UNC, query/fragment, case-insensitive filesystem and real symlink cases pass. Linux and Windows Go/npm CI pass on the updated revision. See the [CI run](https://github.com/web-infra-dev/rslint/actions/runs/35174431222) for the remaining Windows VS Code extension result.
- Formatting and explicit-file spelling checks pass. Temporary Go benchmarks covered nine scenarios; benchmark sources are excluded. Current warm-Program medians for 40 calls using the new settings are 37.2 µs (`mainFields`), 33.6 µs (`mainFiles`) and 34.2 µs (`aliasFields`); parsing is excluded. No extra optimization was justified.


internal/plugins/node/nodeutil/require_calls.go
```diff
@@ -19,6 +34,40 @@ func CollectRequireCalls(ctx rule.RuleContext) []*ast.Node {
...
+		if len(args) == 0 {
+			continue
+		}
+		source := utils.ESTreeRuntimeExpression(args[0])
+		name, ok := utils.GetStaticExpressionValue(utils.SkipAssertionsAndParens(source))
+		if !ok {
+			if evaluator == nil {
+				evaluator = utils.NewStaticStringEvaluatorWithoutScope()
+			}
+			name, ok = evaluator.EvalToString(source)
```
elecmonkey, line 61: `getStringIfConstant` also evaluates some constant binary expressions. For example, upstream resolves `require(0 < 1 ? 'fs' : 'path')`, `require(0 + true)`, and `require(0 - '1')` to `"fs"`, `"1"`, and `"-1"` respectively, while Rslint currently skips them. Would you like to support these evaluation paths, or is this an intentional boundary?

internal/utils/static_string_evaluator_bigint.go
```diff
@@ -0,0 +1,151 @@
...
+import (
+	"math"
+	"math/big"
+
+	"github.com/microsoft/TypeScript/tsc/shim/ast"
+	"github.com/web-infra-dev/rslint/internal/utils/ecmascript"
+)
+
+// Bound constant folding independently of source-controlled exponents/shifts.
+const maxStaticBigIntBits = 1 << 20
```
elecmonkey, line 12: One resource-budget consideration: this limit still allows each expression to materialize and stringify a roughly one-megabit integer. In a synthetic test, 1,000 occurrences of `require(1n << 1048575n)` took about 14.8 seconds, compared with about 18 ms for 1,000 ordinary literal calls. I do not consider this necessarily a defect, but it may be worth confirming that this per-expression cost is acceptable.

internal/utils/static_string_evaluator_bigint.go
```diff
@@ -0,0 +1,151 @@
...
+		if r.Sign() == 0 {
+			return staticEvalResult{}, true
+		}
+		if operator == ast.KindSlashToken {
+			value.Quo(l, r) // JS truncates toward zero, not negative infinity.
+		} else {
+			value.Rem(l, r)
+		}
+	case ast.KindAsteriskAsteriskToken:
+		if r.Sign() < 0 || !r.IsUint64() || r.Uint64() > maxStaticBigIntBits ||
```
elecmonkey: This guard also rejects algebraically trivial cases such as `0n ** huge`, `1n ** huge`, and `(-1n) ** huge`, whose results can be determined without materializing a large integer. Upstream reports the corresponding constant module names, while Rslint skips them. Would you like to handle these special cases before applying the general exponent-size limit, or intentionally leave them outside the supported evaluation boundary?

elecmonkey (commented):
These observations are non-blocking. If the current evaluation and resource boundaries are intentional, I have no further issues with this PR.

internal/plugins/node/nodeutil/require_calls.go
fansenze: Fixed in 85f26872. Constant comparisons and numeric coercions now evaluate correctly, with regression tests for all three examples.

internal/utils/static_string_evaluator_bigint.go
fansenze: Fixed in 85f26872. Bases 0n, 1n and -1n now support huge exponents using math/big’s existing fast paths. Added regression tests, including parity and negative exponents.

internal/utils/static_string_evaluator_bigint.go
fansenze: Confirmed the cost (~15 ms for a BigInt at the current limit). Keeping the cap to avoid skipping more valid constants; this extreme case is an accepted tradeoff.

elecmonkey (approved):

PR #2190, opened 2026-09-16
feat: add node/prefer-node-protocol rule

fansenze:
## Motivation

Add `node/prefer-node-protocol` to enforce and automatically insert `node:` for Node.js built-in modules when the configured Node versions support it.

Part of https://github.com/web-infra-dev/rslint/issues/475.

## Changes

- Check imports, re-exports, dynamic imports, `require()` and `process.getBuiltinModule()`, preserving quotes, escapes and comments in fixes.
- Leave custom local bindings unchanged while recognizing direct Node `process` imports and CommonJS loads, plus `require` initialized with an imported `createRequire()`. Reuse the existing reference store and AST utilities, caching binding checks within each file.
- Reuse existing Node version configuration and tsgo range expansion. Reversibly encode large numeric components during expansion, then compare their original safe-integer values. For example, `>4294967295` and `>=16 <4294967296` enable prefix checks, while `<=4294967296` does not. The same range handling preserves `node/no-deprecated-api` replacement recommendations.
- Ignore contradictory union alternatives regardless of their order. Document only the intentional differences for custom local bindings and empty alternatives, with user-visible examples.
- Preserve all upstream tests and documentation examples, and add regressions for shadowing, reassignment, JavaScript/TypeScript syntax, version-source precedence, numeric boundaries and path-like module names.

## Validation

- `go test ./internal/plugins/node/nodeutil ./internal/plugins/node/rules/prefer_node_protocol ./internal/plugins/node/rules/no_deprecated_api`
- `pnpm --filter @rslint/core build:bin`, schema generation and `pnpm --filter @rslint/core build:js`
- `CI=true pnpm --dir packages/rslint-test-tools exec rs test run tests/eslint-plugin-node/rules/prefer-node-protocol.test.ts tests/eslint-plugin-node/rules/no-deprecated-api.test.ts tests/eslint-plugin-node/rules/no-unpublished-bin.test.ts` — 260 passed, no skips.
- Compared 692 numeric boundary ranges against npm semver 7.8.5, then compared the native rule's 1,240 diagnostics and fixes for those ranges with eslint-plugin-n v18.3.0; all matched.
- Compared 68 additional boundary cases: 66 match upstream, and two exercise the documented correction for empty union alternatives. Six real source files across three Node version ranges also match upstream.
- Targeted Go benchmarks used temporary files only; no benchmark files are included.
- Affected-package Go lint, explicit-file spell checks and formatting checks passed.


elecmonkey:
Most of the module-matching, diagnostic, and autofix logic duplicates the existing `unicorn/prefer-node-protocol` implementation; could we extract it into a shared helper?

> ref: https://github.com/web-infra-dev/rslint/pull/1483

fansenze:
> Most of the module-matching, diagnostic, and autofix logic duplicates the existing `unicorn/prefer-node-protocol` implementation; could we extract it into a shared helper?
> 
> > ref: #1483

Sure, I will open a new pr to do it.

elecmonkey (approved):

PR #2194, opened 2026-09-17
feat(rstest): add no-unneeded-async-expect-function rule

elecmonkey:
## Summary

Reports an `async` function wrapper passed to `expect()` when its whole body awaits a single call, so that call can be asserted on directly.

The wrapper analysis is now shared with `jest/no-unneeded-async-expect-function`. That rule keeps its autofix, but now withholds it in the same cases where this rule withholds its suggestion, since applying it there would break the test's syntax or change what it asserts.

## Credits

Port from [`@vitest/eslint-plugin@1.6.27`](https://github.com/vitest-dev/eslint-plugin-vitest/blob/v1.6.27/src/rules/no-unneeded-async-expect-function.ts)'s `no-unneeded-async-expect-function`.

- **Only chains that go through `resolves` or `rejects` are reported.** Every other chain asserts on the function object itself, so unwrapping `expect(async () => { await run() }).toBeInstanceOf(Function)` would assert on a promise instead.
- **The rewrite is offered as a suggestion instead of an autofix.** A call that throws synchronously rejects the promise the wrapper returns, but escapes the assertion once the wrapper is gone, turning a passing assertion into a failing test.
- **`expect.poll()` and `expect.element()` are not reported.** The first retries a callback and throws when combined with a promise modifier, and the second asserts on a browser locator.
- **The suggestion is withheld when the unwrapped call would not mean the same thing.** That covers a wrapper with parameters, type parameters, a function-expression name, `this`, `arguments` or `new.target`, an awaited call that contains another `await` of the wrapper's own, such as `run(await load())`, an explicit type argument on `expect()`, and a comment inside the wrapper but outside the awaited call.

## Related Links

Related #935.

## Checklist

- [x] Tests updated.
- [x] Documentation updated.


packages/rslint/src/config/presets/rstest.ts
```diff
@@ -14,6 +14,7 @@ const recommended: RslintConfigEntry = {
     'rstest/no-interpolation-in-snapshots': 'error',
     'rstest/no-mocks-import': 'error',
     'rstest/no-standalone-expect': 'error',
+    'rstest/no-unneeded-async-expect-function': 'error',
```
fansenze, line 17: **[P1] Update the preset test when adding this rule**

The existing [preset test](https://github.com/web-infra-dev/rslint/blob/ffe62ae0fa68ab8a943a9835d883b61bc5331a8d/packages/rslint-test-tools/tests/cli/js-config/presets.test.ts#L158) uses `toEqual` with the old 16-rule object. This addition makes it fail because the actual object now has 17 rules; I reproduced that failure locally.

Please add `'rstest/no-unneeded-async-expect-function': 'error'` to the expected object in the same PR.

internal/plugins/rstest/rules/no_unneeded_async_expect_function/no_unneeded_async_expect_function.go
```diff
@@ -0,0 +1,146 @@
...
+	if !keepsWrapperBindings(fn, match.Awaited) || !keepsComments(ctx, match.Wrapper, match.Awaited) {
+		return nil
+	}
+
+	return []rule.RuleSuggestion{{
+		Message: removeWrapperSuggestion,
+		FixesArr: []rule.RuleFix{rule.RuleFixReplace(
+			ctx.SourceFile,
+			match.Wrapper,
+			utils.TrimmedNodeText(ctx.SourceFile, match.Awaited),
```
fansenze: **[P2] Keep the wrapper when the call contains a nested `await`**

The replacement also moves any nested `await` into the surrounding scope:

```js
// Before: valid synchronous test callback
test('rejects', () =>
  expect(async () => { await run(await load()); }).rejects.toThrow(),
);

// After applying the suggestion: SyntaxError
test('rejects', () =>
  expect(run(await load())).rejects.toThrow(),
);
```

A targeted regression test confirms that this suggestion is currently offered. Please withhold it when the call contains an `await` that belongs to the wrapper being removed.

internal/plugins/rstest/rules/no_unneeded_async_expect_function/no_unneeded_async_expect_function.go
```diff
@@ -0,0 +1,146 @@
...
+// `arguments`, which a function expression binds and an expression in the
+// assertion's scope does not.
+func referencesCallerBindings(node *ast.Node) bool {
+	found := false
+	var walk func(*ast.Node) bool
+	walk = func(child *ast.Node) bool {
+		if found || child == nil {
+			return true
+		}
+		switch child.Kind {
```
fansenze: **[P2] Preserve the `new.target` binding**

`new.target` is also scoped to the enclosing ordinary function, but this walk only guards `this` and `arguments`. The rule therefore offers this rewrite at module scope:

```js
// Before: valid
expect(async function () { await run(new.target); }).rejects.toThrow();

// Suggested: SyntaxError outside a function
expect(run(new.target)).rejects.toThrow();
```

I reproduced both the offered suggestion and the syntax error. Please include `new.target` in the binding check and withhold the suggestion when it refers to the function being removed.

packages/rslint/src/config/presets/rstest.ts
elecmonkey: Fixed in 0fe834f.

internal/plugins/rstest/rules/no_unneeded_async_expect_function/no_unneeded_async_expect_function.go
elecmonkey: Fixed in 0fe834f: the suggestion is withheld when the awaited call contains the wrapper's own `await`. The jest rule's autofix now follows the same check.

internal/plugins/rstest/rules/no_unneeded_async_expect_function/no_unneeded_async_expect_function.go
elecmonkey: Fixed in 0fe834f: `new.target` is now checked alongside `this` and `arguments`.

fansenze (approved):

PR #2196, opened 2026-09-17
feat(rstest): add prefer-to-have-length rule

elecmonkey:
## Motivation

Use `toHaveLength()` to make length assertions clearer and improve failure output.

## Changes

Add `rstest/prefer-to-have-length`.

The rule reports `.length` assertions using `toBe()`, `toEqual()`, or `toStrictEqual()` and recommends `toHaveLength()`.

It recognizes Rstest globals, imports and aliases, namespace and CommonJS bindings, `rstack/test`, `import.meta.rstest`, Playwright value assertions, `expect.soft()`, and test-context `expect`.

The matcher-independent detection logic is shared with `jest/prefer-to-have-length`, while each plugin keeps its own parser and autofix policy.

## Intentional differences

- Rstest only autofixes safe `toBe()` assertions.
- `toEqual()` and `toStrictEqual()` are reported without autofix because custom equality testers can affect their result.
- The rule does not autofix negative zero, getter-capable receivers, continued assertion chains, custom messages, extra arguments, or used assertion results.
- `expect.poll()`, promise modifiers, and `expect.element()` are excluded.
- Chai matchers that change the current assertion value, such as `property()` and `throw()`, stop later matcher checks.

## Credits

Adapted from [`eslint-plugin-jest@29.16.1`](https://github.com/jest-community/eslint-plugin-jest/blob/v29.16.1/src/rules/prefer-to-have-length.ts).

## Related Links

Related #935.

## Checklist

- [x] Tests updated.
- [x] Documentation updated.


internal/plugins/rstest/rules/prefer_to_have_length/prefer_to_have_length.go
```diff
@@ -0,0 +1,189 @@
...
+				}
+			}
+
+			onlyTerminalMatcher := endsAtOnlyMatcher(parsed)
+			matches := make([]*shared.ExpectCall, 0, len(parsed.Matchers))
+			for index := range parsed.Matchers {
+				matcher := &parsed.Matchers[index]
+				if subjectMutatingChaiMatchers[matcher.Name] {
+					break
+				}
```
fansenze, line 172: **[P2] Stop checking equality matchers after the assertion value changes**

This guard misses `toContain`, `toThrow`, and `toThrowError`, which can also replace the value used by later assertions. For example, both of these passing chains currently produce a false-positive `useToHaveLength` diagnostic:

```ts
expect([].length).not.toContain(1).and.toEqual(0);
expect({ length: () => { throw 2; } }.length).toThrow().and.toBe(2);
```

In the first chain, `toContain` converts the assertion value from `0` to `[]`, so `toEqual` compares an array with `0`. In the second, `toThrow()` delegates to Chai's `throws()` and changes the assertion value to the thrown `2`. Neither final matcher compares the original `.length` value.

Please stop inspecting later matchers after these operations and add regression coverage for all three names. I reproduced the diagnostics for `toContain`, `toThrow`, and `toThrowError`; these chains already withhold autofixes, so the issue is the incorrect diagnostic.

elecmonkey: Fixed in 1d066c3e957ab07a69dc4303c269354441b67e1b. Added the missing matcher boundaries and regression tests.

fansenze: @elecmonkey I checked the latest `main` and this PR, but the fix isn't present, and GitHub can't find `1d066c3e`. The PR still points to `00c958f8`. Did you forget to push the fix commit?

elecmonkey: sorry sorry 😭

fansenze (approved):

PR #2197, opened 2026-09-17
feat(jest,rstest): add prefer-lowercase-title rule

elecmonkey:
## Motivation

Keep test and suite titles consistently lowercase without changing the rest of the title.

## Changes

Add `rstest/prefer-lowercase-title` and `jest/prefer-lowercase-title`.

The rules report static test and suite titles that begin with an uppercase character and provide deferred autofixes.

Title checks, options, and autofix logic are shared between both plugins, while each plugin keeps its own framework parser.

Supported options:
- `ignore`
- `allowedPrefixes`
- `ignoreTopLevelDescribe`
- `ignoreTodos`

## Intentional differences

- Rstest uses its own call analysis rather than inheriting Jest-specific API aliases.
- Top-level `bench()` calls are not checked.
- Neither rule is added to the recommended preset.
- The describe-block nesting counter is incremented on every `describe` enter and decremented on exit, unlike `eslint-plugin-jest@29.16.1` which skips incrementing when the first argument is not a string literal. This avoids an upstream enter/exit counter imbalance that causes false positives with `ignoreTopLevelDescribe: true` when a title-less `describe()` precedes a describe with an uppercase title.

## Credits

Adapted from `eslint-plugin-jest@29.16.1`.

## Related Links

Related #935, #476.

## Checklist

- [x] Tests updated.
- [x] Documentation updated.


internal/utils/test_framework/rules/prefer_lowercase_title/prefer_lowercase_title.go
```diff
@@ -0,0 +1,201 @@
...
+
+					ctx.ReportNodeWithDeferredFixes(inner, rule.RuleMessage{
+						Id:          "unexpectedCase",
+						Description: fmt.Sprintf("`%s`s should begin with lowercase", parsed.Name),
+						Data:        map[string]string{"method": parsed.Name},
+					}, func() []rule.RuleFix {
+						r := internalUtils.TrimNodeTextRange(ctx.SourceFile, inner)
+						fixRange := r.WithPos(r.Pos() + 1).WithEnd(r.End() - 1)
+						return []rule.RuleFix{
+							rule.RuleFixReplaceRange(fixRange, newDescription),
```
fansenze: **[P1] Preserve escapes when rewriting the title**

`newDescription` contains the decoded string value, so inserting it between the original quotes removes escapes. Both rules turn valid code into invalid code:

```js
// Before
test('Doesn\'t mutate', () => {});

// After --fix: syntax error
test('doesn't mutate', () => {});
```

Escaped newlines also become literal newlines, and `\${name}` inside a template becomes an interpolation. Preserve the original escape sequences, or re-escape the replacement for the existing quote/template delimiter before applying the fix.

internal/utils/test_framework/rules/prefer_lowercase_title/prefer_lowercase_title.go
```diff
@@ -0,0 +1,201 @@
...
+					if parsed.Kind == testFramework.FnKindDescribe {
+						numberOfDescribeBlocks++
+						if opts.ignoreTopLevelDescribe && numberOfDescribeBlocks == 1 {
+							return
+						}
+					} else if parsed.Kind != testFramework.FnKindTest {
+						return
+					}
+
+					if opts.ignoreTodos && slices.Contains(parsed.Members, "todo") {
```
fansenze: **[P2] Honor `ignoreTodos` for Rstest aliases**

With `{ ignoreTodos: true }`, this still reports `unexpectedCase`:

```js
const todoTest = test.todo;
todoTest('Should work');
```

The equivalent `test.todo('Should work')` is correctly ignored. Rstest's `Members` only contains members written at the call site; `.todo` resolved through an alias is represented by `parsed.Todo`. The adapter returns only `parsed.ParsedCall`, dropping that flag. Pass the semantic todo flag into the shared rule so the option also applies to aliases.

internal/utils/test_framework/rules/prefer_lowercase_title/prefer_lowercase_title.go
```diff
@@ -0,0 +1,201 @@
...
+			return rule.RuleListeners{
+				ast.KindCallExpression: func(node *ast.Node) {
+					parsed := runtime.Parse(node)
+					if parsed == nil {
+						return
+					}
+
+					if parsed.Kind == testFramework.FnKindDescribe {
+						numberOfDescribeBlocks++
+						if opts.ignoreTopLevelDescribe && numberOfDescribeBlocks == 1 {
```
fansenze: **[P2] Keep nested Rstest suites checked when the callback is extracted**

With `{ ignoreTopLevelDescribe: true }`, this reports no diagnostics:

```js
describe('Outer', suiteBody);

function suiteBody() {
  describe('Inner', () => {}); // Should report and fix to 'inner'.
}
```

The equivalent inline callback correctly reports `Inner`. When traversal reaches the separate function declaration, this counter has returned to zero, so the nested suite is mistaken for a top-level suite. Have the Rstest adapter supply the depth from the existing `GetRstestDescribeDepth` helper, which accounts for callback ownership.

fansenze (commented):
Three issues reproduced on `44c1a9c`; details and examples are attached inline.

internal/utils/test_framework/rules/prefer_lowercase_title/prefer_lowercase_title.go
elecmonkey: Fixed in db718978. The autofix now replaces only the source span for the first title character, so the rest of the literal keeps its original escape spelling.

internal/utils/test_framework/rules/prefer_lowercase_title/prefer_lowercase_title.go
elecmonkey: Fixed in db718978. The Rstest adapter now passes the semantic todo flag from the call analysis, so `ignoreTodos` also applies to `test.todo` aliases.

internal/utils/test_framework/rules/prefer_lowercase_title/prefer_lowercase_title.go
elecmonkey: Fixed in db718978. The Rstest adapter now supplies describe depth from `GetRstestDescribeDepth`, so describes inside extracted suite callbacks are not treated as top-level describes.

fansenze (approved):

PR #2199, opened 2026-09-17
feat(unicorn): add no-array-sort rule

HashemKhalifa:
## Motivation

Refs #1309.

Port `unicorn/no-array-sort` from [eslint-plugin-unicorn v75.0.0](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/80939f9de76d4544bae132800f26bc7f248bee27/rules/no-array-sort.js). Verified that the rule was absent from main and exact-name PR searches returned no matches, including immediately before publication.

## Changes

- Add suggestion-only `Array#toSorted()` replacements and the upstream `allowExpressionStatement` option (default `true`).
- Preserve spread-copy alternatives, parentheses and optional member access; exclude optional calls, computed names and known non-array receivers consistently with the upstream cases.
- Share suggestion construction through a reusable array-mutation helper. This PR is self-contained against main.
- Register the rule, option schema and JavaScript mirror, with focused documentation.
- Extract the existing callback classifier from `prefer-array-some` instead of duplicating it. Fix optional calls being mistaken for ordinary call expressions; add two regressions to that existing consumer as well as the new rule.

## Verification completed locally

- 43 upstream/documentation cases and 65 additional AST/semantic cases. Go assertions check message IDs/text, full ranges, applied autofixes and suggestion IDs/outputs against the pinned upstream results.
- Representative edit-demand tests cover diagnostics-only, autofix-only, suggestion-only and all-edit consumers, with stable diagnostic identity and no edits produced for unrequested categories.
- Focused Go 1.26.0 tests passed: `go test ./internal/plugins/unicorn/rules/no_array_sort ./internal/plugins/unicorn/unicornutil ./internal/plugins/unicorn/rules/prefer_array_some`.
- Diff-aware golangci-lint 2.12.2 passed for the changed Go packages, including the catalog and affected helper consumers.
- Changed-file spelling, root `pnpm run format:check`, and `git diff --check` passed. Normal repository commit hooks remain enabled.

## Validation not completed locally

The full catalog test reached linking, but the Mac encountered disk pressure and sustained linker memory/I/O pressure. Those full-build jobs were stopped; no catalog-test pass is claimed.

Fresh native/core JavaScript builds, the JavaScript integration mirror, and the additional native-API differential run have **not** completed locally. The JS mirror is included but is not claimed as passing. CI should validate the native/JavaScript integration before merge.

The focused Go run also passed the existing `prefer-array-some` suite, including its two optional-comparator regressions.


HashemKhalifa:
@codex review

elecmonkey (commented):
As with #2200, please add this rule to the recommended rules list. I found no other issues.

elecmonkey (dismissed):

HashemKhalifa:
> As with #2200, please add this rule to the recommended rules list. I found no other issues.

@elecmonkey thanks for reminding me, I added it already to the PR's, kindly review

elecmonkey:
I suggest that we stop making changes to `presets.test.ts` in the future. Honestly, I don’t think this test is useful. It seems to be something AI agents just like to do, and it has caused endless conflicts. We only need to enable the appropriate rules in `unicorn.ts`.

Another small tip: try to work on non-adjacent entries when possible. For example, if the rules are ordered A-B-C-D, and A and D have already been merged while B and C are still unported, avoid porting B and C in the same batch, since adjacent entries are more likely to cause conflicts.

elecmonkey (approved):

elecmonkey:
@fansenze @HashemKhalifa Based on my testing, we can merge:

1. #2199
2. #2202
3. #2205

these three should merge without conflicts. After that, update and merge one PR at a time.

Keeping the parallelism under control helps avoid duplicate work and makes things easier on reviewers. :-)

HashemKhalifa:
> @fansenze @HashemKhalifa Based on my testing, we can merge:
> 
> 
> 
> 1. #2199
> 
> 2. #2202
> 
> 3. #2205
> 
> 
> 
> these three should merge without conflicts. After that, update and merge one PR at a time.
> 
> 
> 
> Keeping the parallelism under control helps avoid duplicate work and makes things easier on reviewers. :-)

Thank you, I will keep that in mind and apologise for the inconvenience 🫡

HashemKhalifa:
> I suggest that we stop making changes to presets.test.ts in the future. Honestly, I don’t think this test is useful. It seems to be something AI agents just like to do, and it has caused endless conflicts. We only need to enable the appropriate rules in unicorn.ts.
> 
> Another small tip: try to work on non-adjacent entries when possible. For example, if the rules are ordered A-B-C-D, and A and D have already been merged while B and C are still unported, avoid porting B and C in the same batch, since adjacent entries are more likely to cause conflicts.

You are 100% right, the agent liked to add test to the presets.test.ts and its useless!
Will make sure this won't happen again, thanks for the heads up!
Live and learn 🫡

fansenze (approved):

PR #2200, opened 2026-09-17
feat(unicorn): add no-zero-fractions rule

HashemKhalifa:
## Motivation

Refs #1309.

Port `unicorn/no-zero-fractions` from [eslint-plugin-unicorn v75.0.0](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/80939f9de76d4544bae132800f26bc7f248bee27/rules/no-zero-fractions.js). Verified that the rule was absent from main and exact-name PR searches returned no matches, including immediately before publication.

## Changes

- Remove trailing zero fractions and dangling decimal points, preserving separators and exponent spelling.
- Reuse the existing keyword-spacing and semicolon helpers so autofixes preserve valid member receivers and statement boundaries.
- Register the native rule and JavaScript mirror, with focused documentation.

## Verification completed locally

- 201 upstream/documentation cases and 31 additional AST/semantic cases. Go assertions check message IDs/text, full ranges, applied autofixes and suggestion IDs/outputs against the pinned upstream results.
- Representative edit-demand tests cover diagnostics-only, autofix-only, suggestion-only and all-edit consumers, with stable diagnostic identity and no edits produced for unrequested categories.
- Focused Go 1.26.0 tests passed: `go test ./internal/plugins/unicorn/rules/no_zero_fractions`.
- Diff-aware golangci-lint 2.12.2 passed for the changed Go packages, including the catalog.
- Changed-file spelling, root `pnpm run format:check`, and `git diff --check` passed. Normal repository commit hooks remain enabled.

## Validation not completed locally

The full catalog test reached linking, but the Mac encountered disk pressure and sustained linker memory/I/O pressure. Those full-build jobs were stopped; no catalog-test pass is claimed.

Fresh native/core JavaScript builds, the JavaScript integration mirror, and the additional native-API differential run have **not** completed locally. The JS mirror is included but is not claimed as passing. CI should validate the native/JavaScript integration before merge.


HashemKhalifa:
@codex review

internal/plugins/unicorn/all.go
```diff
@@ -90,6 +91,7 @@ func GetAllRules() []rule.Rule {
 		no_unsafe_string_replacement.NoUnsafeStringReplacementRule,
 		no_useless_switch_case.NoUselessSwitchCaseRule,
 		no_xor_as_exponentiation.NoXorAsExponentiationRule,
+		no_zero_fractions.NoZeroFractionsRule,
```
elecmonkey, line 96: [P2] Please also enable `unicorn/no-zero-fractions` in `packages/rslint/src/config/presets/unicorn.ts`. This line makes the rule available for explicit configuration, but the recommended preset still leaves it commented as “not implemented”. Both Unicorn v73 (the preset’s stated baseline) and v75 enable this rule in `recommended` and `unopinionated`, so users of Rslint's Unicorn recommended preset silently miss the newly implemented rule.

internal/plugins/unicorn/rules/no_zero_fractions/no_zero_fractions_extras_test.go
```diff
@@ -0,0 +1,216 @@
...
+					Consumer: rule.DiagnosticConsumer{Demand: demand, Report: func(d rule.RuleDiagnostic) { got = append(got, d) }},
+				})
+				if len(got) != 1 {
+					t.Fatalf("demand %d: got %d diagnostics", demand, len(got))
+				}
+				diagnostics[demand] = got[0]
+			}
+			baseline := diagnostics[rule.EditDemandNone]
+			all := diagnostics[rule.EditDemandAll]
+			for demand, diagnostic := range diagnostics {
```
elecmonkey, line 143: [P2] Please avoid iterating this map while calling `ApplyRuleFixes`. `ApplyRuleFixes` sorts `diagnostic.Fixes()` in place, so if the `EditDemandAll` entry is visited first, `all.FixesPtr` is reordered before the later `DeepEqual` comparison with the autofix-only entry. Because Go map iteration order is nondeterministic, `go test -p=1 ./internal/plugins/unicorn/rules/no_zero_fractions -run TestNoZeroFractionsEditDemand -count=100` fails intermittently with `demand 1 changed autofix artifacts`. Iterate a fixed demand slice and compare before mutation, or normalize copies before comparing.

elecmonkey (changes requested):
Changes requested: please address the two inline findings.

internal/plugins/unicorn/all.go
HashemKhalifa: Thanks, fixed in 1dcffb1940e5bc6c329e8242c26c62f08340d34d. `unicorn/no-zero-fractions` is now enabled at `error` severity in `packages/rslint/src/config/presets/unicorn.ts`. I also added an assertion to the existing preset test so this registration is covered. The targeted source-backed preset test passes.

internal/plugins/unicorn/rules/no_zero_fractions/no_zero_fractions_extras_test.go
HashemKhalifa: Good catch, fixed in 1dcffb1940e5bc6c329e8242c26c62f08340d34d. The test now finishes all diagnostic and edit-artifact comparisons before applying any fixes in a separate pass. That prevents `ApplyRuleFixes` from reordering the baseline slice before another comparison, regardless of map iteration order. I also removed the unused suggestion-expectation branch; this rule only produces autofixes.

I reproduced the intermittent failure before the change. After the fix, `TestNoZeroFractionsEditDemand` passes with `-count=100`, and the full `no_zero_fractions` package tests and scoped Go lint pass too.

elecmonkey (dismissed):
LGTM. Thanks :-)

elecmonkey (approved):

fansenze (approved):

PR #2202, opened 2026-09-17
feat(unicorn): add no-negation-in-equality-check rule

HashemKhalifa:
## Motivation

Refs #1309.

Port `unicorn/no-negation-in-equality-check` from [eslint-plugin-unicorn v75.0.0](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/80939f9de76d4544bae132800f26bc7f248bee27/rules/no-negation-in-equality-check.js). The rule was absent from main and exact-name, all-state PR searches returned no matches before implementation and on the final eligibility check.

## Changes

Report single left-side negation in equality checks and offer an explicit operator-inversion suggestion. Reuse keyword-spacing and semicolon helpers; preserve return/throw newline boundaries without automatic semantic changes.

Add documentation, catalogue registration, a JavaScript upstream mirror, and separate upstream/extras Go suites. This branch is independent of the other new rule PRs and does not modify shared utility implementations.

## Validation

- 20 upstream/documentation source cases and 32 additional syntax and regression cases passed through the native Go RuleTester. Assertions cover diagnostic IDs/text, full ranges, absence of automatic fixes, and suggestion IDs/output.
- Two representative scenarios cover all four edit-demand modes; diagnostic identity stays unchanged and suggestions are produced only when requested.
- Direct Go-versus-ESLint comparison passed on copies of two real repository TypeScript files (`packages/rslint/src/service/service.ts` and `packages/rslint/src/eslint-plugin/plugin/plugin-loader.ts`), each with an appended positive control. Checked diagnostic IDs/text, full ranges, applied autofixes and suggestion IDs/outputs. Reference: Unicorn 75.0.0, ESLint 10.10.0, typescript-eslint parser 8.62.0. This exercises the native rule harness, not the IPC layer.
- Focused Go 1.26.0 tests passed: `go test ./internal/plugins/unicorn/rules/no_negation_in_equality_check`.
- golangci-lint 2.12.2 passed with `--new-from-merge-base=origin/main` on the new rule package and changed Unicorn catalogue package.
- Changed-file spelling, `pnpm run format:check`, `gofmt`, and `git diff --check` passed. Normal repository pre-commit hooks remained enabled.

The full catalogue test, schema dump, complete native/core-JavaScript builds, and JavaScript integration run were not executed in this batch. The Mac previously exhausted disk and experienced sustained full-catalogue linker pressure, so this batch reused the unchanged pinned compiler with task-local workspace files and bounded, focused verification. No repository toolchain or dependency versions were changed.

The JavaScript mirror is included and registered but is not claimed as passing locally. CI must validate full registration/build and native-JavaScript integration before merge.


internal/plugins/unicorn/rules/no_negation_in_equality_check/no_negation_in_equality_check.go
```diff
@@ -0,0 +1,81 @@
...
+				start := utils.TrimNodeTextRange(ctx.SourceFile, left).Pos()
+				bang := core.NewTextRange(start, start+1)
+				ctx.ReportRangeWithDeferredSuggestions(bang, rule.RuleMessage{
+					Id: "no-negation-in-equality-check/error", Description: "Negated expression is not allowed in equality check.",
+				}, func() []rule.RuleSuggestion {
+					fixes := unicornutil.SpaceAroundKeywordFixes(ctx.SourceFile, node)
+					afterBang, _ := utils.TokenAtOrAfter(ctx.SourceFile, bang.End())
+					fixes = append(fixes, returnOrThrowParentheses(ctx.SourceFile, node, afterBang.Start)...)
+					prefix := ""
+					if unicornutil.NeedsSemicolonBefore(ctx.SourceFile, node, afterBang.Text) {
```
elecmonkey, line 46: **[P2] Preserve TypeScript-specific statement boundaries**

`NeedsSemicolonBefore` does not cover all of the upstream helper's TypeScript cases. For example, applying this suggestion to `foo\n!<boolean>bar === baz`, `foo<string>\n!(bar) === baz`, or `foo!\n!(bar) === baz` omits the required leading semicolon. The resulting text is parsed as one expression instead of two statements. Unicorn 75.0.0 inserts `;` in all three cases. Please extend the shared helper to treat `<` as a hazardous replacement prefix and recognize preceding `TSInstantiationExpression` and `TSNonNullExpression` boundaries, with regression tests for these cases.

internal/plugins/unicorn/all.go
```diff
@@ -77,6 +78,7 @@ func GetAllRules() []rule.Rule {
 		no_invalid_remove_event_listener.NoInvalidRemoveEventListenerRule,
 		no_magic_array_flat_depth.NoMagicArrayFlatDepthRule,
 		no_new_buffer.NoNewBufferRule,
+		no_negation_in_equality_check.NoNegationInEqualityCheckRule,
```
elecmonkey, line 82: **[P2] Enable the rule in the Unicorn recommended preset**

This registers the rule for explicit use, but `packages/rslint/src/config/presets/unicorn.ts` still leaves `unicorn/no-negation-in-equality-check` commented out as “not implemented.” Unicorn marks this rule as `recommended: 'unopinionated'` in both v73 and v75, so its recommended and unopinionated presets enable it. As a result, `unicornPlugin.configs.recommended` currently produces no diagnostic for `!foo === bar`, while explicitly enabling the rule does. Please enable the rule in the preset and add it to the preset integration assertions.

elecmonkey (changes requested):
Two correctness issues remain: TypeScript-specific ASI boundaries can produce semantics-changing suggestions, and the newly registered rule is still omitted from the Unicorn recommended preset.

internal/plugins/unicorn/all.go
HashemKhalifa: Thanks, fixed in 611d9f77b25c42be59b4cf9908dc691d3f26f743. `unicorn/no-negation-in-equality-check` is now enabled at `error` severity in `packages/rslint/src/config/presets/unicorn.ts`, and the existing preset integration test asserts that entry. The targeted source-backed preset regression failed before the registration change and passed afterward.

internal/plugins/unicorn/rules/no_negation_in_equality_check/no_negation_in_equality_check.go
HashemKhalifa: Good catch—these TypeScript cases were not covered by the earlier boundary fix. Fixed in dda4e6fba20e00af8c55a0f842df67ce270b1fdd.

The shared `NeedsSemicolonBefore` helper now treats `<` as a hazardous prefix and recognizes preceding instantiation and non-null expressions. I reproduced the three examples before the change and added 15 upstream-verified suggestion cases, including nested generics and negative controls for explicit semicolons and declaration boundaries. The suggested outputs also preserve the parsed statement count.

The full `no_negation_in_equality_check`, `unicornutil`, and affected `prefer_array_flat` package tests pass, along with scoped Go lint, formatting, and spelling.

HashemKhalifa:
> Two correctness issues remain: TypeScript-specific ASI boundaries can produce semantics-changing suggestions, and the newly registered rule is still omitted from the Unicorn recommended preset.
Good catch, thanks!

Fixed, kindly review :)

internal/plugins/unicorn/unicornutil/semicolon.go
```diff
@@ -73,13 +78,19 @@ func NeedsSemicolonBefore(
...
-		)
+	case ast.KindCloseBraceToken, ast.KindGreaterThanToken, ast.KindExclamationToken:
+		// These tokens can end a declaration or a runtime value, including
+		// TypeScript instantiations and non-null assertions. Only values can
+		// absorb a following expression across the statement boundary.
+		for previousNode := ast.GetNodeAtPosition(sourceFile, previous.Start, false); previousNode != nil && previousNode.End() == previous.End; previousNode = previousNode.Parent {
+			switch previousNode.Kind {
+			case ast.KindObjectLiteralExpression, ast.KindFunctionExpression,
+				ast.KindArrowFunction, ast.KindClassExpression,
+				ast.KindExpressionWithTypeArguments, ast.KindNonNullExpression:
```
elecmonkey, line 89: **Consider covering generic `new` expressions**

This is admittedly an extreme edge case, but handling it would make the shared boundary helper more complete:

```ts
new Foo<string>
!(bar) === baz
```

The current suggestion produces:

```ts
new Foo<string>
(bar) !== baz
```

In the ts-go AST, the original source contains two expression statements, but the suggested output contains one binary expression: the second line becomes the argument list of `new Foo<string>(bar)`. Unicorn 75.0.0 inserts a leading semicolon here.

The `>` token's ancestor chain contains a `NewExpression` rather than an `ExpressionWithTypeArguments`, so the current check does not recognize this boundary. Although this case is unusual, could we specially account for a generic `NewExpression` and add a focused regression test?

HashemKhalifa: Fixed in `7aff78427b1d2ff4e484d059735e27d70fd4c73b` (latest head `4867bce7ac99a0c7d1c84f460e65a190d9c9663` after merging current main). `NeedsSemicolonBefore` now recognizes a generic `NewExpression` ending at its explicit type arguments, so `new Foo<string>\n!(bar) === baz` suggests `new Foo<string>\n;(bar) !== baz`. I added both a shared-helper regression and the exact rule-level case. The full `no_negation_in_equality_check`, `unicornutil`, `prefer_array_flat`, and newly merged `no_useless_error_capture_stack_trace` suites pass, along with formatting, spelling, TypeScript syntax, diff checks, and catalog-aware scoped golangci-lint (0 issues).

elecmonkey (dismissed):

elecmonkey (approved):

fansenze (approved):

PR #2204, opened 2026-09-17
feat(unicorn): add no-unnecessary-slice-end rule

HashemKhalifa:
## Motivation

Refs #1309.

Port the recommended `unicorn/no-unnecessary-slice-end` rule from [Unicorn v75.0.0](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/80939f9de76d4544bae132800f26bc7f248bee27/rules/no-unnecessary-slice-end.js). The implementation was absent from main and exact-name, all-state PR searches returned no matches immediately before publication.

## Changes

Remove redundant end arguments from array, string and typed-array slice calls. Reuse the canonical method matcher and keep the argument-removal helper identical to the independent splice-rule port. Preserve comments, optional member syntax and trailing commas.

Enable the rule at `error` severity in the recommended preset via `packages/rslint/src/config/presets/unicorn.ts`, and include catalogue registration, documentation, complete upstream/documentation Go cases, separate regression tests, and a registered JavaScript upstream mirror.

Locally shadowed or explicitly disabled `Infinity` and `Number` are not treated as built-in constants. This documented, regression-tested safety difference avoids upstream autofixes that could remove a finite argument.

## Verification

- 74 source cases: 40 upstream, 6 documentation and 28 additional regressions. Go assertions cover exact IDs/text, full ranges and applied edits.
- Four-mode edit-demand checks finish artifact comparisons before applying any fixes.
- `go test ./internal/plugins/unicorn/rules/no_unnecessary_slice_end ./internal/plugins/unicorn/unicornutil` passed on Go 1.26.0.
- A fresh focused native build passed 76 cases through production JavaScript/service/Go IPC, including two real repository-source copies with appended positive controls. Checked file/rule execution, IDs/text, severity, complete UTF-16 ranges and edit outputs against the pinned reference, with the documented safety differences applied explicitly.

The focused IPC build uses a temporary six-rule catalogue to bound memory and disk use. It is not a full-catalogue build. The normal full-catalogue test, schema/core distribution build and standard full-catalogue JavaScript integration were not run locally; CI must verify those before merge. No production build scripts, dependency versions or compiler pins were changed. This branch targets main independently.


internal/plugins/unicorn/rules/no_unnecessary_slice_end/no_unnecessary_slice_end.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/web-infra-dev/rslint/internal/plugins/unicorn/unicornutil"
+	"github.com/web-infra-dev/rslint/internal/rule"
+)
+
+var NoUnnecessarySliceEndRule = rule.Rule{
+	Name: "unicorn/no-unnecessary-slice-end", Schema: rule.EmptyArraySchema,
+	Run: func(ctx rule.RuleContext, options []any) rule.RuleListeners {
+		argumentCount := 2
+		return rule.RuleListeners{ast.KindCallExpression: func(node *ast.Node) {
+			call, ok := unicornutil.MatchDotMethodCall(node, unicornutil.DotMethodCallOptions{Methods: []string{"slice"}, ArgumentsLength: &argumentCount, RejectSpreadElement: true, AllowOptionalMember: true})
```
elecmonkey, line 15: [P2] This reports on any `.slice(...)` member call, but the rule docs say the autofix is only safe for `Array`, `String`, and `TypedArray` receivers. A known custom receiver can change behavior: `class Box { length = 2; slice(start: number, end = 99) { return end; } } const value = new Box(); value.slice(1, value.length);` currently reports and fixes to `value.slice(1)`, changing the returned value from `2` to `99`. It is fine for source-only runs to keep following upstream's broad syntax-only behavior, but when Rslint has a checker it should use that type information to skip receivers known not to be `Array`, `String`, or `TypedArray`.

internal/plugins/unicorn/unicornutil/unnecessary_length_argument.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+	if !ast.IsPropertyAccessExpression(argument) {
+		return ""
+	}
+	member := argument.AsPropertyAccessExpression()
+	receiver := utils.ESTreeRuntimeExpression(member.Expression)
+	if !ast.IsIdentifier(member.Name()) {
+		return ""
+	}
+	if member.Name().Text() == "POSITIVE_INFINITY" && !ast.IsOptionalChain(argument) && ast.IsIdentifier(receiver) && receiver.Text() == "Number" && ctx.Globals.Access("Number").IsDeclared() && ctx.Refs.IsGlobalReference(receiver) {
```
elecmonkey: [P2] This only verifies that `Number` resolves as a global, so it still fixes after the global binding has been overwritten earlier in the file. For example, `Number = {POSITIVE_INFINITY: 2}; [0, 1, 2].slice(1, Number.POSITIVE_INFINITY);` is fixed to `.slice(1)`, changing the result from `[1]` to `[1, 2]`. Please also reject `Number.POSITIVE_INFINITY` when the global `Number` binding has been written before this reference, similar to the existing pristine-global checks used by other rules.

internal/plugins/unicorn/unicornutil/unnecessary_length_argument.go
```diff
@@ -0,0 +1,56 @@
...
+		return "Number.POSITIVE_INFINITY"
+	}
+	if member.Name().Text() != "length" || !utils.IsSameReference(object, receiver, false) {
```
elecmonkey: [P3] `utils.IsSameReference` only compares literal element keys, so this misses upstream-valid static-property cases such as `foo[-0].slice(1, foo[0].length)` and `foo["a" + "b"].slice(1, foo.ab.length)`. Unicorn v75 reports and fixes both because `getStaticPropertyName` folds the computed key to the same property name; Rslint has a static evaluator that can do this coercion, so the receiver comparison should use it for static element-access names.

elecmonkey (commented):
I found a few correctness issues; details inline.

internal/plugins/unicorn/rules/no_unnecessary_slice_end/no_unnecessary_slice_end.go
HashemKhalifa: Good catch. Fixed in `bb4206f34bee2dafe97331501c2cb92cd7f6860d`. When a TypeChecker is available, `no-unnecessary-slice-end` now skips receivers proven to be neither string-like nor indexed collections, so the custom `Box#slice` example is left unchanged. The source-only path still keeps Unicorn's broad syntax behavior. I added the exact typed custom-receiver regression; the full rule + shared-helper suites and scoped Go lint pass.

internal/plugins/unicorn/unicornutil/unnecessary_length_argument.go
HashemKhalifa: Fixed in `bb4206f34bee2dafe97331501c2cb92cd7f6860d`. The shared length/count helper now requires `Infinity`/`Number` to still be pristine at the reference site: an earlier write to the global prevents the diagnostic/autofix. This uses the existing reference index and global-resolution boundary rather than a rule-local text check. The reassigned-`Number` example is covered by a regression and remains unchanged.

internal/plugins/unicorn/unicornutil/unnecessary_length_argument.go
HashemKhalifa: Fixed in `bb4206f34bee2dafe97331501c2cb92cd7f6860d`. Receiver equivalence now uses the existing static-string evaluator recursively for member keys instead of relying only on literal element keys. Regression cases cover both `foo[-0]` vs `foo[0]` and `foo["a" + "b"]` vs `foo.ab`; both diagnose and fix as expected. Focused rule/helper tests and scoped lint pass.

elecmonkey (dismissed):

elecmonkey (dismissed):

elecmonkey (approved):

PR #2205, opened 2026-09-17
feat(unicorn): add no-unnecessary-array-splice-count rule

HashemKhalifa:
## Motivation

Refs #1309.

Port the recommended `unicorn/no-unnecessary-array-splice-count` rule from [Unicorn v75.0.0](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/80939f9de76d4544bae132800f26bc7f248bee27/rules/no-unnecessary-array-splice-count.js). The native implementation was absent from main and exact-name, all-state PR searches found no existing PR immediately before publication.

## Changes

Remove redundant deleteCount/skipCount arguments from two-argument splice/toSpliced calls. Calls with inserted items remain unchanged. Reuse canonical receiver classification and the same shared argument-removal helper as the independent slice-rule port, preserving comments and trailing commas.

Enable the rule at `error` severity in the recommended preset via `packages/rslint/src/config/presets/unicorn.ts`, and include catalogue registration, documentation, complete upstream/documentation Go cases, separate regressions, and a registered JavaScript upstream mirror.

Locally shadowed or explicitly disabled `Infinity` and `Number` are not assumed to be built-in constants. This documented, tested difference prevents removing a potentially finite count, which upstream v75.0.0 checks only by spelling.

## Verification

- 150 source cases: 82 upstream, 12 documentation and 56 additional regressions, with exact diagnostic IDs/text, complete ranges and edit outputs.
- Four-mode edit-demand checks compare artifacts before applying fixes.
- `go test ./internal/plugins/unicorn/rules/no_unnecessary_array_splice_count ./internal/plugins/unicorn/unicornutil` passed on Go 1.26.0.
- A fresh focused native build passed 152 cases through production JavaScript/service/Go IPC, including two real repository-source copies with appended positive controls. Verified file/rule execution, IDs/text, severity, full UTF-16 ranges and edits against the pinned reference, with documented safety differences applied explicitly.

The focused build uses a temporary six-rule catalogue, not the full catalogue. The normal full-catalogue test, schema/core distribution build and standard full-catalogue JavaScript integration were not run locally; CI must verify these before merge. Production build scripts, dependency versions and compiler pins are unchanged. This branch targets main independently.


internal/plugins/unicorn/unicornutil/unnecessary_length_argument.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+	if !ast.IsPropertyAccessExpression(argument) {
+		return ""
+	}
+	member := argument.AsPropertyAccessExpression()
+	receiver := utils.ESTreeRuntimeExpression(member.Expression)
+	if !ast.IsIdentifier(member.Name()) {
+		return ""
+	}
+	if member.Name().Text() == "POSITIVE_INFINITY" && !ast.IsOptionalChain(argument) && ast.IsIdentifier(receiver) && receiver.Text() == "Number" && ctx.Globals.Access("Number").IsDeclared() && ctx.Refs.IsGlobalReference(receiver) {
```
elecmonkey: [P2] This still reports when the global `Number` binding has been reassigned. For example, `Number = {POSITIVE_INFINITY: 1}; const array = [0, 1, 2]; array.splice(1, Number.POSITIVE_INFINITY);` removes one element, while the autofix changes it to `array.splice(1)` and removes everything after index 1. Upstream v75.0.0 has the same bug, but this port can avoid introducing the unsafe fix by only accepting `Number.POSITIVE_INFINITY` when the global binding is unmodified.

internal/plugins/unicorn/unicornutil/unnecessary_length_argument.go
```diff
@@ -0,0 +1,56 @@
...
+		return "Number.POSITIVE_INFINITY"
+	}
+	if member.Name().Text() != "length" || !utils.IsSameReference(object, receiver, false) {
```
elecmonkey: [P2] `IsSameReference` proves the two expressions have the same shape, but not that evaluating them twice is equivalent to evaluating them once. A getter-backed receiver such as `let i = 0; const first = [0,1,2], second = [0]; const obj = { get a() { return i++ ? second : first; } }; obj.a.splice(1, obj.a.length);` leaves `first` as `[0, 2]`, while the autofix changes it to `obj.a.splice(1)` and leaves `first` as `[0]`. Upstream v75.0.0 also reports this, but we should gate this autofix/report on a receiver path that is safe to re-evaluate.

elecmonkey (changes requested):
Two autofix correctness issues need to be fixed before this lands.

internal/plugins/unicorn/unicornutil/unnecessary_length_argument.go
HashemKhalifa: Fixed in `3497758131938623958f4322d51f688f86b296e8`. The shared helper now rejects `Number.POSITIVE_INFINITY` when the global `Number` binding was written earlier, so the reassigned-`Number` splice example is left unchanged. The same pristine-global guard is shared with `no-unnecessary-slice-end`; focused rule/helper tests and scoped lint pass.

internal/plugins/unicorn/unicornutil/unnecessary_length_argument.go
HashemKhalifa: Fixed in `3497758131938623958f4322d51f688f86b296e8`. `.length` rewrites now reject known getter-backed receiver paths, and they also reject a first argument that can mutate/re-evaluate the receiver before the length read. I added regressions for the getter example and for `a.splice((a = b, 1), a.length)`. The safety policy lives in the shared length/count helper rather than being duplicated in this rule. Full focused tests and scoped lint pass.

internal/plugins/unicorn/unicornutil/unnecessary_length_argument.go
```diff
@@ -0,0 +1,190 @@
...
+			return false
+		}
+		return isRepeatableReference(ctx, utils.AccessExpressionObject(node))
+	default:
+		return false
+	}
+}
+
+func accessHasGetter(ctx rule.RuleContext, node *ast.Node) bool {
+	if ctx.TypeChecker == nil || node == nil {
```
elecmonkey, line 147: [P2] This still lets the unsafe getter autofix through when the lint run has no TypeChecker. The rule does not require type info, so source-only / `project: false` JavaScript can reach this branch with `ctx.TypeChecker == nil`; `accessHasGetter` then returns false, `isRepeatableReference` treats `obj.a` as repeatable, and `fix: true` still rewrites `obj.a.splice(1, obj.a.length)` to `obj.a.splice(1)`. I reproduced this on the latest head with `let i = 0; const first = [0,1,2], second = [0]; const obj = { get a() { return i++ ? second : first; } }; obj.a.splice(1, obj.a.length);` under `project: false`, and the output still drops the second getter evaluation. The typed regression passes, but the checker-less path needs to conservatively skip member-chain receivers it cannot prove are getter-free.

elecmonkey (changes requested):
One source-only path still needs tightening before this is safe to approve.

internal/plugins/unicorn/unicornutil/unnecessary_length_argument.go
HashemKhalifa: Fixed in `b514372a21c1a8fc5827c677e2406db7189d91f5`. In checker-less/source-only runs, member-chain receivers are now treated conservatively as non-repeatable because we cannot prove their properties are getter-free. I added an explicit `project:false` regression using your exact getter example and assert that it produces no diagnostic/autofix. The full rule + shared-helper suites, formatting, spelling, diff checks, and scoped golangci-lint all pass.

elecmonkey (dismissed):

elecmonkey (approved):

fansenze (approved):

PR #2206, opened 2026-09-17
feat(unicorn): add prefer-dom-node-append rule

HashemKhalifa:
## Motivation

Refs #1309.

Port the recommended `unicorn/prefer-dom-node-append` rule from [Unicorn v75.0.0](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/80939f9de76d4544bae132800f26bc7f248bee27/rules/prefer-dom-node-append.js). The implementation was missing from main and exact-name, all-state PR searches found no competing PR immediately before publication.

## Changes

Prefer append to appendChild, using the existing dot-method matcher. Only autofix standalone expression statements: a used appendChild return value must not silently become undefined. Preserve optional member access and exclude syntactic non-node receivers/arguments, computed methods, optional calls and spread arguments.

Enable the rule at `error` severity in the recommended preset via `packages/rslint/src/config/presets/unicorn.ts`. Include native catalogue registration, documentation, complete upstream/documentation Go cases, separate regressions, and a registered JavaScript upstream mirror. This branch targets main independently.

## Verification

- 102 source cases: 69 upstream, 6 documentation and 27 additional syntax/regression cases. Assertions include exact diagnostic IDs/text, complete ranges, applied fixes and absence of fixes when return values are used.
- Representative tests cover all four edit-demand modes, comparing artifacts before fix application.
- `go test ./internal/plugins/unicorn/rules/prefer_dom_node_append` passed on Go 1.26.0.
- A fresh focused native build passed 104 cases through production JavaScript/service/Go IPC, including two real repository-source copies with appended positive controls. Checked file/rule execution, IDs/text, severity, complete UTF-16 ranges and edit outputs against the pinned reference.

The focused IPC build uses a temporary six-rule catalogue to bound memory/disk use. It does not establish a full-catalogue build. The normal full-catalogue test, schema/core distribution build and standard full-catalogue JavaScript integration were not run locally; CI must verify those before merge. No production build scripts, dependencies or compiler pins were changed.


internal/plugins/unicorn/rules/prefer_dom_node_append/prefer_dom_node_append.go
```diff
@@ -0,0 +1,41 @@
...
+					return []rule.RuleFix{rule.RuleFixReplace(ctx.SourceFile, call.Property, "append")}
+				}
+				return nil
+			})
+		}}
+	},
+}
+
+func isNotDOMNode(node *ast.Node) bool {
+	node = utils.ESTreeRuntimeExpression(node)
```
elecmonkey: This classifier is too shallow for fix safety. Inputs like `node.appendChild(void 0)`, `node.appendChild((sideEffect(), null))`, and `node.appendChild(null as unknown as Node)` still pass this check, so the rule reports and autofixes them to `append(...)`. That changes runtime behavior: `appendChild(null/undefined)` throws because the argument is not a `Node`, while `append(null/undefined)` succeeds and inserts the strings `"null"` / `"undefined"`. Upstream has the same bug, but Rslint should not preserve it because the rule already tries to exclude values that cannot be DOM nodes. Please unwrap/evaluate the relevant expression forms, or at least suppress the autofix for statically impossible Node values.

internal/plugins/unicorn/rules/prefer_dom_node_append/prefer_dom_node_append.go
```diff
@@ -0,0 +1,41 @@
...
+	switch node.Kind {
+	case ast.KindArrayLiteralExpression, ast.KindArrowFunction, ast.KindClassExpression, ast.KindFunctionExpression,
+		ast.KindObjectLiteralExpression, ast.KindTemplateExpression, ast.KindNoSubstitutionTemplateLiteral,
+		ast.KindStringLiteral, ast.KindNumericLiteral, ast.KindBigIntLiteral, ast.KindRegularExpressionLiteral,
+		ast.KindNullKeyword, ast.KindTrueKeyword, ast.KindFalseKeyword:
+		return true
+	}
+	return utils.IsUndefinedIdentifier(node)
```
elecmonkey: This treats every identifier named `undefined` as the global undefined value, which creates false negatives when the name is locally bound. For example, `function f(undefined) { node.appendChild(undefined); }` currently produces no diagnostic even though the argument may be a DOM node. Upstream v75.0.0 also misses this, but Rslint has binding information available in source-only mode through the existing shadowing helpers, so this should distinguish the global `undefined` from a local binding.

internal/plugins/unicorn/rules/prefer_dom_node_append/prefer_dom_node_append.go
HashemKhalifa: Fixed in `e2b299f9a3f4735880b262539eff7cd0f9842268`. The non-DOM classifier now unwraps parentheses/TS assertions, follows the final value of comma expressions, and treats `void` plus the existing literal/object/function shapes as statically impossible DOM nodes. Regressions cover `void 0`, `(sideEffect(), null)`, and `null as unknown as Node`; none is reported or autofixed now. The full rule suite and scoped lint pass.

internal/plugins/unicorn/rules/prefer_dom_node_append/prefer_dom_node_append.go
HashemKhalifa: Fixed in `e2b299f9a3f4735880b262539eff7cd0f9842268`. An identifier named `undefined` is now considered the undefined value only when it resolves to the configured global. A locally bound parameter named `undefined` therefore remains a possible DOM node and is diagnosed/fixed. The exact `function f(undefined) { ... }` case is covered by a regression.

elecmonkey (dismissed):

elecmonkey (approved):

fansenze (approved):

PR #2207, opened 2026-09-17
feat(unicorn): add prefer-identifier-import-export-specifiers rule

HashemKhalifa:
## Motivation

Refs #1309.

Port the recommended `unicorn/prefer-identifier-import-export-specifiers` rule from [Unicorn v75.0.0](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/80939f9de76d4544bae132800f26bc7f248bee27/rules/prefer-identifier-import-export-specifiers.js). The native rule was absent from main and exact-name, all-state PR searches found no existing PR immediately before publication.

## Changes

Prefer identifier syntax for import/export specifiers and import-attribute keys when the decoded string is a valid identifier name. Reuse the TypeScript identifier scanner and existing token-safe replacement helper. Cover Unicode names, aliases, re-exports, namespace exports, type-only forms and attributes without duplicate diagnostics. Ordinary object keys and module source strings remain untouched.

Enable the rule at `error` severity in the recommended preset via `packages/rslint/src/config/presets/unicorn.ts`, and add catalogue registration, documentation, complete upstream/documentation Go cases, separate regressions, and a registered JavaScript upstream mirror. This branch targets main independently.

## Verification

- 88 source cases: 56 upstream, 9 documentation and 23 additional cases, including supplementary Unicode and identifier-format characters. Assertions check exact IDs/text, complete UTF-16 ranges and autofix outputs.
- Representative tests cover all four edit-demand modes, with comparisons completed before fix application.
- `go test ./internal/plugins/unicorn/rules/prefer_identifier_import_export_specifiers` passed on Go 1.26.0.
- A fresh focused native build passed 90 cases through production JavaScript/service/Go IPC, including two real repository-source copies with appended positive controls. Checked file/rule execution, IDs/text, severity, complete ranges and edits against the pinned reference.

The focused IPC build uses a temporary six-rule catalogue, not the full catalogue. The normal full-catalogue test, schema/core distribution build and standard full-catalogue JavaScript integration were not run locally; CI must verify those before merge. Production build scripts, dependency versions and compiler pins are unchanged.


elecmonkey (dismissed):
Known difference: Rslint currently uses typescript-go's Unicode 15.1 identifier tables, while Unicorn v75 on Node 22+ uses Unicode 17. Some newer valid IdentifierName characters, such as `𑎀`, are therefore not reported or autofixed by this rule in Rslint. This is intentionally left to the parser/table layer rather than handled with a rule-local heuristic.

elecmonkey (dismissed):

elecmonkey (dismissed):

fansenze (approved):

PR #2214, opened 2026-09-20
feat: add node/no-missing-require rule

fansenze:
## Motivation

Add `node/no-missing-require` so unresolved `require()` and `require.resolve()` targets can be caught before runtime. Part of [#475](https://github.com/web-infra-dev/rslint/issues/475).

## Changes

- Port eslint-plugin-n 18.3.0 behavior and options, reusing the existing require reference tracker and shared module resolution utilities.
- Cache results per file, including allowed targets, and skip package-name parsing when no allow list is configured.
- Preserve all 79 upstream cases and three documentation examples, add 68 Go edge cases, and document user-visible resolver differences.

Validation on macOS arm64:

- `go test ./internal/plugins/node/rules/no_missing_require ./internal/plugins/node/nodeutil`
- `golangci-lint run --new-from-merge-base=origin/main ./internal/plugins/node/nodeutil ./internal/plugins/node/rules/no_missing_require ./internal/plugins/node`
- `pnpm --filter @rslint/core build:bin` and `pnpm --filter @rslint/core build:js`
- `CI=true pnpm --dir packages/rslint-test-tools exec rs test run tests/eslint-plugin-node/rules/no-missing-require.test.ts` — 79 passed, 3 explained skips.
- Differential check against the pinned upstream rule on seven real repository files plus one positive probe; compare messages, IDs, severity, full ranges, fixes and suggestions.
- Temporary Go benchmarks covering builtin, allowed, existing, missing, repeated and TypeScript alias targets.
- `pnpm run format:check` and `pnpm run check-spell` with all changed text paths.

Two upstream `<input>` cases are retained as explained skips because the native lint API needs a filename. The JS case-sensitive filesystem case is also skipped on macOS, as upstream does; the Go virtual-filesystem case runs. Linux and Windows execution was not performed. Temporary Go benchmark files are excluded from this PR.


elecmonkey (commented):
Withdrawn.

elecmonkey (approved):
all previous findings are documented diffs.

fansenze:
#2219 adds `fallback` and `fullySpecified` support in the shared resolver. Object alias ordering remains a documented limitation; use arrays for explicit priority. The fix does not change the config layer or RuleContext.

PR #2217, opened 2026-09-20
feat(rstest): add prefer-ending-with-an-expect rule

elecmonkey:
## Motivation

A test that ends by calling into the code under test, changing state, or awaiting a promise has exercised something without checking the result, which usually means the test was left unfinished.

## Changes

Add `rstest/prefer-ending-with-an-expect`.

The rule reports a test case whose last statement is not an assertion. A block body contributes its last statement, a concise arrow body contributes its expression, and a leading `await` is looked through. `expect`, `expect.soft`, `expect.poll` and Chai's `assert` all count, including when they reach the call site through the test context, a destructured fixture, a namespace import or a renamed import. Further asserting functions can be named through `assertFunctionNames`, and further test wrappers through `additionalTestBlockFunctions`.

Share the framework-neutral engine with `jest/prefer-ending-with-an-expect`, which becomes a thin adapter over it, while each plugin keeps its own call parser, callback position and assertion resolution. The jest rule's behavior is unchanged.

Add Rstest-specific tests, a source-only test, a JS integration test, and documentation. The rule is not part of the `recommended` preset.

Fix the shared Rstest call analysis so a destructured key is resolved through the static property-name helper. Reading `binding.PropertyName` with `Text()` panicked on a computed key, and the `import.meta.rstest` binding resolver silently fell through to the local alias when it could not read the key, so `const { ['expect']: check } = import.meta.rstest` resolved to `check` and matched no API. Both paths now fold a constant key and give up on a dynamic one. The same unguarded `Text()` in `@typescript-eslint/no-deprecated` is fixed alongside it.

Resolve Chai's `assert` through the shared Rstest call analysis rather than by matching callee text, so `assert as check`, `rstest.assert` and `const { assert: check } = import.meta.rstest` are recognized the way the same shapes of `expect` already were. The test context is excluded because Rstest provides a callback its own `expect`, not its own `assert`. The identifier that gates the resolving hook now accepts either name, since a file asserting only through `assert` previously switched the hook off entirely.

## Intentional differences

- The `(name, options, fn)` overload is checked. Rstest takes the callback as either the second or the third argument, and `@rstest/core` 0.11 runs the third one even when the second is not an options object, so a case that passes its options second is no longer skipped.
- `test.todo` is exempt even when it is handed a function, because a todo registration never runs that function. This matches `rstest/expect-expect`.
- Assertions reached through the test context, a destructured fixture, a namespace import or a renamed import count, resolved through the same analysis `rstest/expect-expect` consumes. Matching callee text alone reports every one of them.
- `assertFunctionNames` defaults to `["expect", "assert"]`, because Chai's `assert` is an Rstest global, again matching `rstest/expect-expect`.
- A test may end in a Chai property assertion such as `expect(cart).to.be.empty`, which asserts through a getter and is not a call. The shared engine gained an optional hook for a non-call statement, consulted only after the call shape misses; the jest adapter leaves it nil and keeps the call-only behavior. A chain that resolves no matcher, such as `expect(cart).not` or `expect(cart).toBe`, still reports.

## Known limitations

- A callback passed by reference, such as `test('places the order', run)`, is not inspected. Deciding whether the referenced function's last statement is the one the test ends with costs a false positive on working code whenever the answer is wrong.
- Only the syntactic last statement is examined, so `return expect(cart).toBe('ok')` and `test('x', () => promise.then(() => expect(cart).toBe('ok')))` are reported although they do assert. Recognizing them needs control-flow analysis this rule does not do.
- A destructuring key computed from a runtime value, such as `const { [key]: check } = import.meta.rstest`, names no API the rule can resolve, so a test that asserts only through that binding is reported. Any other answer would have to guess which API the key selects, and guessing `expect` misreports the case where it selects something else.

## Credits

Adapted from [`eslint-plugin-jest@29.16.1`](https://github.com/jest-community/eslint-plugin-jest/blob/v29.16.1/src/rules/prefer-ending-with-an-expect.ts).

## Related Links

Related #935.

## Checklist

- [x] Tests updated.
- [x] Documentation updated.


internal/plugins/rstest/rules/prefer_ending_with_an_expect/prefer_ending_with_an_expect.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+var PreferEndingWithAnExpectRule = shared.NewRule(shared.Config{
+	Name: "rstest/prefer-ending-with-an-expect",
+	// Rstest ships both `expect` and `assert` (chai) as globals
+	// (packages/core/src/utils/constants.ts globalApiList), so the default
+	// asserting functions match rstest/expect-expect rather than jest.
+	DefaultAssertFunctionNames: []string{"expect", "assert"},
+	Prepare: func(ctx rule.RuleContext) shared.Runtime {
+		analysis := rstestUtils.GetRstestCallAnalysis(ctx)
```
elecmonkey, line 141: This eagerly initializes the shared Rstest call analysis, which can panic for valid computed destructuring from `import.meta.rstest`, for example:

```ts
const { ['expect']: check } = import.meta.rstest;

test('places the order', () => {
  check(checkout()).toBe('ok');
});
```

`collectVariableCandidates` reaches the destructuring property with `binding.PropertyName` as a `ComputedPropertyName` and calls `Text()` on it, which panics (`Unhandled case in Node.Text: *ast.ComputedPropertyName`) before any diagnostic can be produced. This is especially visible for this new rule because its docs say destructured `import.meta.rstest` assertion sources are recognized.

Please resolve the property name in the shared analyzer through the existing static-string helper, and skip unsupported dynamic computed keys instead of calling `Text()` unguarded.

internal/plugins/rstest/rules/prefer_ending_with_an_expect/prefer_ending_with_an_expect.go
```diff
@@ -0,0 +1,94 @@
...
+		// Rstest reaches `expect` through the test context, namespace imports and
+		// import aliases, none of which the callee-text patterns can match. Reuse
+		// the same resolution rstest/expect-expect consumes so both rules agree on
+		// what an assertion is.
+		var isAssertion func(node *ast.Node) bool
+		if sourceMayContainRstestExpect(ctx.SourceFile) {
+			isAssertion = func(node *ast.Node) bool {
+				return analysis.IsExpectCall(node)
```
fansenze: [P2] Recognize aliases and namespace imports of `assert`

With the default rule options, this valid test reports `Tests should end with an assertion`:

```ts
import { test, assert as check } from '@rstest/core';

test('x', () => {
  check.equal(1, 1);
});
```

The same false positive occurs with `rstest.assert.equal(...)` after a namespace import, and with `const { assert: check } = import.meta.rstest`.

The default `assert` pattern only matches the literal callee name, while this fallback resolves only `expect`. Please also resolve native Rstest `assert` bindings, including in files without an `expect` identifier, and add regression cases for these forms.

internal/utils/test_framework/rules/prefer_ending_with_an_expect/prefer_ending_with_an_expect.go
```diff
@@ -0,0 +1,171 @@
...
+	patterns []*esregexp.RegExp,
+) bool {
+	if node == nil {
+		return false
+	}
+	if node.Kind == ast.KindAwaitExpression {
+		node = ast.SkipParentheses(node.AsAwaitExpression().Expression)
+	}
+	if node == nil || node.Kind != ast.KindCallExpression {
+		return false
```
fansenze: [P2] Accept Chai property assertions supported by Rstest

This test ends with a valid Rstest assertion, but the rule reports `Tests should end with an assertion`:

```ts
import { test, expect } from '@rstest/core';

test('x', () => {
  expect(true).to.be.true;
});
```

The final expression is a property access, so this guard rejects it before the Rstest assertion resolver can run. `context.expect(true).to.be.true` has the same problem, and configuring `assertFunctionNames` cannot bypass this guard.

Please let the Rstest adapter recognize property-style assertions while preserving Jest's existing behavior, and add regression coverage for both direct and context assertions.

fansenze (commented):
Found two reproducible false positives with native Rstest assertions. The existing tests in the six affected Go packages pass, but the additional cases below are reported incorrectly.

internal/plugins/rstest/rules/prefer_ending_with_an_expect/prefer_ending_with_an_expect.go
elecmonkey: Fixed in 26956c07e. `assert` is resolved through the shared analysis like `expect`, minus the test context. The `HasIdentifier("expect")` gate was a second layer and now accepts either name. `rstest/expect-expect` had the same bug and is fixed with it.

internal/utils/test_framework/rules/prefer_ending_with_an_expect/prefer_ending_with_an_expect.go
elecmonkey: Fixed in 26956c07e. The shared engine gained an optional non-call hook; jest leaves it nil and is unchanged. Rstest hands the chain to the existing expect parser, so `expect(x).not` and `expect(x).toBe` still report.

internal/plugins/rstest/utils/assert_call.go
```diff
@@ -0,0 +1,118 @@
...
+	if isImportMetaRstestObjectAlias(symbol) {
+		return rstestExpectRootReceiver
+	}
+	if testFramework.IsModuleNamespaceSymbolModules(symbol, RstestAllImportModules) {
+		return rstestExpectRootReceiver
+	}
+	name, _, _ := testFramework.ResolveFunctionIdentifierReferenceFromSymbolModules(
+		rstestAssertAPIName,
+		nil,
+		symbol,
```
fansenze, line 124: [P2] Resolve the actual identifier before accepting an assert binding

This test contains no assertion, but both `rstest/prefer-ending-with-an-expect` and `rstest/expect-expect` now accept it:

```ts
import { test, assert as check } from '@rstest/core';

test('x', () => {
  const check = () => {};
  check();
});
```

It was correctly reported at `6d860bf`. Passing `nil` as the identifier makes `ResolveFunctionIdentifierReferenceFromSymbolModules` immediately return the supplied name (`assert`), without inspecting the symbol. As a result, the local function shadowing the imported alias is mistaken for a native assertion.

Please pass the actual root identifier and its local name, and add regression cases for local-variable and parameter shadowing.

internal/plugins/rstest/utils/assert_call.go
```diff
@@ -0,0 +1,118 @@
...
+func (analysis *RstestCallAnalysis) isAssertCandidate(node *ast.Node) bool {
+	root := testFramework.ResolveFirstIdentifier(node.AsCallExpression().Expression)
+	return root == nil ||
+		root.Kind != ast.KindIdentifier ||
+		analysis.candidates[root.AsIdentifier().Text]&rstestCandidateAssert != 0
+}
+
+func (analysis *RstestCallAnalysis) resolveAssertCall(node *ast.Node) bool {
+	entries := firstRstestExpectEntries(node.AsCallExpression().Expression)
+	if entries.count == 0 || entries.first == nil || entries.first.Kind != ast.KindIdentifier {
```
fansenze, line 60: [P2] Recognize direct import.meta.rstest.assert calls

This valid test still reports `Tests should end with an assertion`:

```ts
test('x', () => {
  import.meta.rstest.assert.equal(1, 1);
});
```

The destructured form now passes, but this direct form does not. This resolution path expects an identifier root and does not recognize the `import.meta.rstest` chain as the assertion source.

Please handle direct `import.meta.rstest.assert` chains before identifier-based resolution, as the expect resolver already does for `import.meta.rstest.expect`, and add a regression case.

fansenze (commented):
The original eight reproduction cases now pass. Follow-up checks found one new regression affecting both assertion rules and two remaining false positives; details are attached inline.

internal/plugins/rstest/rules/prefer_ending_with_an_expect/prefer_ending_with_an_expect.go
```diff
@@ -0,0 +1,162 @@
...
+		return false
+	}
+	return parsed.Matchers[len(parsed.Matchers)-1].Kind == rstestUtils.RstestExpectMatcherProperty
+}
+
+// chainHeadCall returns the call a member chain starts from, so
+// `expect(cart).to.be.empty` yields `expect(cart)`.
+func chainHeadCall(node *ast.Node) *ast.Node {
+	for node != nil {
+		switch node.Kind {
```
fansenze, line 120: [P2] Unwrap parentheses while finding the assertion head

This valid test still reports `Tests should end with an assertion`; removing the parentheses around `expect(true)` makes it pass:

```ts
import { test, expect } from '@rstest/core';

test('x', () => {
  (expect(true)).to.be.true;
});
```

`chainHeadCall` reaches a `ParenthesizedExpression` inside the member chain and returns `nil` through the default branch. Unwrapping only the outer expression in `rstestPropertyAssertion` does not cover this case.

Please skip parentheses at each step of this traversal and add regression cases for both this form and `(expect(true).to.be).true`.

internal/plugins/rstest/utils/assert_call.go
elecmonkey: Fixed in 4c3ef9cbd. The resolver now gets the identifier written at the call site and its local name, so it sees the local declaration and refuses it. Regression cases cover const, function and parameter shadowing, each keeping the import that makes the local name a candidate — without it the candidate gate short-circuits before the resolver, which is why my earlier negative case passed.

internal/plugins/rstest/utils/assert_call.go
elecmonkey: Fixed in 4c3ef9cbd. Handled before the identifier path, mirroring the expect resolver. Covered for both the member and callable forms, plus a negative for another `import.meta.rstest` member.

internal/plugins/rstest/rules/prefer_ending_with_an_expect/prefer_ending_with_an_expect.go
elecmonkey: Fixed in 4c3ef9cbd. `chainHeadCall` skips parentheses at every step now. The parser already resolved all three forms with `reason=none`, so this only brings the traversal up to what the parser accepts. Both forms you named are covered.

fansenze (approved):

PR #2219, opened 2026-09-20
fix(node): honor fallback and fullySpecified resolver options

fansenze:
## Summary

Honor `resolverConfig.fallback` and `fullySpecified` across Node rules. Unresolved requests can use fallback targets, and fully specified requests no longer guess extensions or directory entries.

- Let path restrictions inspect builtin fallback targets, including directory shims.
- Preserve malformed package metadata, package exports/imports errors, and alias cycles instead of hiding them with fallbacks.
- Reuse alias matching and tsgo lookup in the shared resolver. Package entry points and alias targets retain their normal lookup behavior.
- Document unsupported resolver options and alphabetical priority for overlapping object aliases. Arrays support explicit priority.

Follow-up to #2214; related to [#475](https://github.com/web-infra-dev/rslint/issues/475).

## Validation

- Targeted Go tests passed for `moduleresolver`, `nodeutil`, and all eight affected Node rules.
- Rebuilt the native binary and JS packages. Integration tests for both missing-module and both restriction rules: 239 passed, 6 existing filename/filesystem-related skips.
- 69 resolver-case expectations verified against enhanced-resolve 5.24.3; 84 diagnostic comparisons passed against eslint-plugin-n 18.3.0 across six rule families. Eight malformed-metadata diagnostics retain the documented message-wording difference; all other diagnostic fields match.
- Filtered Go lint, formatting, and spelling checks passed.
- Temporary Go benchmarks covered nine require-rule scenarios; benchmark files are excluded from the PR.

Verified locally on macOS arm64. POSIX, drive-letter, and UNC paths are covered by virtual-filesystem tests; native Linux and Windows execution remains for CI.


internal/utils/moduleresolver/resolution_alias.go
```diff
@@ -69,7 +69,14 @@ func (resolver *nodeResolver) resolve(name string) (result Result) {
 		request = request[:index]
 	}
 	if !resolver.mainTarget {
-		if result, matched := resolver.alias(request); matched {
+		defer func() {
+			if result.Error != "" && !result.recursive && !result.terminal {
```
elecmonkey, line 73: **[P2] Allow configured fallbacks to redirect Node builtins**

`enhanced-resolve@5.24.3` applies a user-provided fallback to builtins: with `fallback: { fs: './shim.js' }`, both `fs` and `node:fs` resolve to the shim. Here, builtin resolution returns `Result{}` before this deferred fallback runs because the fallback is attempted only when `result.Error != ""`. As a result, path-based consumers such as `node/no-restricted-import` and `node/no-restricted-require` cannot observe or restrict the configured shim. Please let an explicitly configured fallback run for builtin requests.

elecmonkey, line 73: **[P2] Do not apply fallback after malformed package metadata**

When `node_modules/bad/package.json` is malformed and `fallback: { bad: './shim.js' }` is configured, `eslint-plugin-n@18.3.0` with `enhanced-resolve@5.24.3` reports the package JSON parse failure. Rslint currently marks the read as unresolved in `nodeResolutionFS.ReadFile`, returns a normal resolution error, and this condition replaces it with the fallback result, so `node/no-missing-require` emits no diagnostic. Please preserve malformed package metadata as a terminal resolution failure so fallback applies only to an ordinary missing request.

fansenze: Fixed in [32be1b83](https://github.com/web-infra-dev/rslint/commit/32be1b838250d0eb64e854f6fef6bf35e61aedf5). Both restriction rules now match builtin fallback targets, including directories. `fs` and `node:fs` require separate fallback keys, matching upstream.

fansenze: Fixed in [32be1b83](https://github.com/web-infra-dev/rslint/commit/32be1b838250d0eb64e854f6fef6bf35e61aedf5). Malformed `package.json` files now stop resolution, so fallbacks cannot hide the error. Added regression tests for direct requests, alias targets, and fallback arrays.

PR #2222, opened 2026-09-20
feat: add node/prefer-global/console rule

fansenze:
## Motivation

Add native support for `node/prefer-global/console` so projects can consistently use either the global `console` or the Node.js module. This ports the rule from eslint-plugin-n v18.3.0 and contributes to #475.

## Changes

- Support `always` (default) and `never`, with upstream messages and ranges for globals, CommonJS, `process.getBuiltinModule()`, and static ESM imports/re-exports.
- Reuse `nodeutil.ReferenceTracker` and the existing context/reference utilities. Stop following aliases when a shared reference trace has no remaining property, call, or constructor events; reads are still reported at their original locations.
- Preserve upstream checks for type-only imports and re-exports. Recognize constant string `.toString()` module names through the existing static string evaluator, retaining its guards for unknown inputs and custom methods.
- Cover all 15 upstream tests and 5 documentation examples, plus regressions for shadowing, aliases, optional chains, TypeScript, JSX, JSDoc, diagnostic ranges, and suppression directives. Add shared evaluator coverage for `.toString()`, including empty strings, optional/computed calls, extra arguments, and UTF-16 strings. Configure Node globals in the documentation example.

Validation passed:

- Targeted Go tests across 30 affected packages: shared evaluators and reference/module/array helpers, plus their consuming Node, Unicorn, TypeScript, and core rules. The rule catalog registration check also passed.
- 39 JS integration tests for `node/prefer-global/console` and `node/prefer-global/process`, using a rebuilt native binary in an isolated worktree.
- Eleven CLI checks matched eslint-plugin-n v18.3.0, covering the copied documentation configuration, both options, type-only declarations, constant string methods, and unknown/custom calls. Initial differential validation also covered six real source files with 44 matching diagnostics.
- Scoped Go lint, changed-file spelling checks, and `pnpm run format:check`.

Temporary Go benchmarks were used during development and are excluded from this PR.


internal/plugins/node/rules/prefer_global/console/console_extras_test.go
```diff
@@ -0,0 +1,463 @@
...
+		// Type query without local binding.
+		{Code: "type Logger = typeof console; interface Options { logger: typeof console }",
+			Options:         []any{"never"},
+			LanguageOptions: rule.LanguageOptions{SourceType: "module"},
+			FileName:        "input.ts",
+			Errors: []rule_tester.InvalidTestCaseError{
+				consoleError("preferModule", 1, 22, 1, 29),
+				consoleError("preferModule", 1, 66, 1, 73),
+			}},
+		// Type-only imports still load the module for this rule.
```
elecmonkey: [P2] Type-only imports and exports are erased at runtime, so they should not be reported as module usage. Could we skip cases such as `import type { Console } from "console"` and `export type { Console } from "console"`? Upstream has the same false positive, but Rslint has the AST information needed to avoid it.

internal/plugins/node/rules/prefer_global/console/console.md
```diff
@@ -0,0 +1,53 @@
...
+Enforce consistent use of the global `console` variable or the `console` module.
+
+## Options
+
+- `"always"` (default): use the global `console` variable.
+- `"never"`: load `console` from a module instead of using the global variable.
+
+```javascript
+export default [{
+  plugins: ['node'],
```
elecmonkey, line 14: [P2] This example does not configure the Node globals, so `console` and `process` are not tracked when users copy it directly. Could we add `languageOptions: { globals: globals.node }`, including the corresponding import, or state that an existing Node environment configuration is required?

internal/plugins/node/rules/prefer_global/console/console.go
```diff
@@ -0,0 +1,55 @@
...
+					if !utils.IsJSDocSyntaxNode(node) {
+						references = append(references, node)
+					}
+				}},
+			}
+			tracker := nodeutil.NewReferenceTracker(ctx)
+			if preferModule {
+				tracker.TrackGlobals(trace)
+			} else {
+				tracker.TrackModules(trace)
```
elecmonkey, line 47: [non-blocking] Upstream also recognizes uncommon constant expressions such as `require("console".toString())`, while Rslint currently misses them. This seems reasonable to leave unsupported if it is too much of an edge case, though documenting or testing the boundary would be helpful.

internal/plugins/node/rules/prefer_global/console/console_extras_test.go
fansenze: Restored upstream behavior after reviewing the rule's scope: it standardizes where console references come from, including type-only imports and exports. Regression tests now enforce that behavior.

internal/plugins/node/rules/prefer_global/console/console.md
fansenze: Added `import { globals } from '@rslint/core'` and `languageOptions: { globals: globals.node }`. Verified the copied config against `require`, `process.getBuiltinModule`, and global `console`.

internal/plugins/node/rules/prefer_global/console/console.go
fansenze: Now supported through the existing static string evaluator. Added coverage for both module loaders, computed/optional calls, unknown inputs, custom methods, and UTF-16 strings. Removed the documented limitation.

elecmonkey (approved):

PR #2226, opened 2026-09-20
feat: add node/prefer-global/url rule

fansenze:
## Motivation

Add `node/prefer-global/url` so projects can consistently use either the global `URL` constructor or the `URL` export from Node's `url` module.

## Changes

- Port the `always` (default) and `never` behavior from eslint-plugin-n v18.3.0, including CommonJS, `process.getBuiltinModule()`, ESM imports/re-exports, aliases and configured globals.
- Reuse the existing Node reference tracker and rule-context services, and register the rule with its option schema and documentation.
- Preserve upstream messages and diagnostic ranges without adding fixes or suggestions. Cover all upstream cases and documentation examples, plus scope, JavaScript and TypeScript boundaries.

## Validation

- Targeted Go rule tests: 84 cases; catalog registration check passed.
- CI-mode JS integration: 20 cases, no skips.
- Differential comparison with eslint-plugin-n v18.3.0: 80 file/rule runs, including seven real source files and positive controls; 66 diagnostics agree on normalized paths, rule names, message IDs/text, severity, complete ranges and edit artifacts.
- Scoped Go lint, changed-file spelling, formatting and whitespace checks.


internal/plugins/node/rules/prefer_global/url/url.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Options
+
+- `"always"` (default): use the global `URL`.
+- `"never"`: use `URL` from `url` or `node:url`.
+
+```javascript
+export default [
+  {
+    plugins: ["node"],
```
elecmonkey, line 18: ### [P2] Document the required Node globals in the configuration example

The example enables the rule without configuring `globals.node`. With that exact configuration, Rslint reports neither of the documentation's primary violations:

- `"always"` with `const { URL } = require("node:url");` produces 0 diagnostics because `require` is undeclared.
- `"never"` with `new URL("https://example.com");` produces 0 diagnostics because `URL` is undeclared.

Adding `languageOptions: { globals: globals.node }` produces the expected `preferGlobal` and `preferModule` diagnostics. The adjacent `prefer-global/process` documentation already demonstrates this prerequisite.

The example should import `globals` from `@rslint/core` and configure `languageOptions.globals`.

fansenze: Fixed in ef748be4. The example now imports `globals` and configures `globals.node`. Verified that both modes report the expected diagnostics.

elecmonkey (dismissed):

elecmonkey (approved):

PR #2228, opened 2026-09-20
feat(rstest): add prefer-hooks-on-top rule

elecmonkey:
## Motivation

Rstest attaches a lifecycle hook to the whole suite wherever it is written, so a hook declared after a test case still wraps every case above it. Reading the file top to bottom suggests the opposite.

## Changes

Add `rstest/prefer-hooks-on-top`.

The rule reports `beforeAll`, `beforeEach`, `afterEach`, and `afterAll` declared after the first test case of the same suite body. Each suite body is judged on its own, and the order of the hooks among themselves is not checked. A function body is a scope of its own, whether it is the callback written inline at the `describe` call or a function passed to it by name.

It recognizes Rstest globals, imports and aliases, namespace and CommonJS bindings, `rstack/test`, `import.meta.rstest`, and the hooks `@rstest/playwright` exposes as members of its test object. A test case counts when it is registered plainly, through a modifier such as `.only`, through `.each` or `.for`, and through a fixture-extended test API.

The framework-neutral scope tracking and reporting are shared with `jest/prefer-hooks-on-top`, which is refactored onto the shared engine, while each plugin keeps its own call parser.

Add Rstest-specific tests and documentation.

## Intentional differences

- Calling `test.extend({ ... })` on its own builds another test function and does not open the scope, while a registration made through the extended API does. The rule therefore needs no exempt-modifier list.
- `onTestFinished` and `onTestFailed` are registered from inside a running test and are not treated as suite hooks.
- A hook written in a function that is never called registers nothing at run time, and a function passed to `describe` by name is that suite's body rather than a continuation of the code around it, so neither inherits the test cases registered outside it. Upstream opens a new scope only at a call expression and reports both. The cost is that a hook registered through a helper the suite does call is judged against the helper and goes unreported. `jest/prefer-hooks-on-top` picks this up from the shared engine and documents it.
- Only hooks written directly in a suite body are checked; one reached indirectly, through a function the suite calls, is outside what the rule reads.
- The rule is not added to the recommended preset.

## Credits

Adapted from [`eslint-plugin-jest@29.16.1`](https://github.com/jest-community/eslint-plugin-jest/blob/v29.16.1/src/rules/prefer-hooks-on-top.ts).

## Related Links

Related #935.

## Checklist

- [x] Tests updated.
- [x] Documentation updated.


internal/utils/test_framework/rules/prefer_hooks_on_top/prefer_hooks_on_top.go
```diff
@@ -0,0 +1,104 @@
...
+				rule.ListenerOnExit(ast.KindCallExpression): popScope,
+
+				ast.KindFunctionDeclaration:                      pushScope,
+				rule.ListenerOnExit(ast.KindFunctionDeclaration): popScope,
+				ast.KindFunctionExpression:                       pushScope,
+				rule.ListenerOnExit(ast.KindFunctionExpression):  popScope,
+				ast.KindArrowFunction:                            pushScope,
+				rule.ListenerOnExit(ast.KindArrowFunction):       popScope,
+				ast.KindMethodDeclaration:                        pushScope,
+				rule.ListenerOnExit(ast.KindMethodDeclaration):   popScope,
```
fansenze, line 106: [P2] Isolate getters, setters, and constructors as function scopes

This list omits `KindGetAccessor`, `KindSetAccessor`, and `KindConstructor`, so calls inside those bodies incorrectly share the enclosing scope:

```ts
class Helper {
  constructor() {
    test('unused', () => {});
  }
}
beforeEach(() => {}); // incorrectly reported
```

`Helper` is never instantiated, so no test is registered and this hook should be allowed. Getter/setter bodies have the same problem. The reverse also happens: an outer test makes hooks inside these uncalled bodies report.

Please add matching enter/exit listeners for all three node kinds and cover both directions in regression tests. I reproduced all six cases in both the Rstest and Jest suites; adding the missing listeners made them pass.

elecmonkey: Fixed in `7ac8561d4`: `KindConstructor`, `KindGetAccessor` and `KindSetAccessor` now push and pop a scope frame too, and all six cases pass in both suites.

`KindClassStaticBlockDeclaration` is left out on purpose — a static block runs when the class declaration is evaluated, so what it registers does belong to the surrounding scope.

fansenze (approved):

PR #2229, opened 2026-09-20
feat(rstest): add require-hook rule

elecmonkey:
## Motivation

Rstest collects a test file by executing it. Module-level statements and every `describe` callback run during collection, before any `beforeAll` or `beforeEach`, and they run even for a suite that is skipped. Setup written there happens whether or not the tests that need it run, and `afterAll` cannot undo it.

## Changes

Add `rstest/require-hook`. It reports a statement standing at the top level of a file or of a `describe` callback when it is a call that is not part of Rstest's own API, or a `let` / `var` declaration whose initializer is not `null` or `undefined`.

Registrations, hooks, `expect` chains, and calls on the `rs` / `rstest` utilities object stay exempt. A `describe` callback is inspected only when the function is written at the call site; a callback passed by name is not, since the same function may be registered more than once.

The `allowedFunctionCalls` option matches the whole callee chain, so `helper.setup` allows `helper.setup()` but not `setup()`.

`onTestFinished` and `onTestFailed` are reported with a separate message, because Rstest rejects either call during collection with `onTestFinished() can only be called inside a test`. Moving such a call into a test callback is the repair; moving it into a hook is not.

Register the rule in the Rstest plugin without adding it to the `recommended` preset, because whether top-level initialization is acceptable depends on how a project organizes its test files.

## Credits

Port from [`eslint-plugin-jest@29.16.0`](https://github.com/jest-community/eslint-plugin-jest/blob/v29.16.0/src/rules/require-hook.ts)'s `require-hook`.

- **The utilities object is resolved instead of name-prefixed.** Upstream exempts any callee chain spelled `jest.*`. This port resolves the receiver, so it also covers a renamed import, a namespace import and `import.meta.rstest`, while an unrelated local object named `rs` is still reported. A plugin-managed member such as `rs.mock` is additionally matched by the receiver as written, because Rstest's build rewrites it that way even in a file that declares its own `rs`.

- **Rstest's execution-time APIs get their own diagnostic.** Upstream tells the user to move every reported call into a hook. `onTestFinished()` and `onTestFailed()` register work against the test that is currently running, so that repair fails too. They are reported under a separate message id that names the test callback instead, and the API is resolved through Rstest provenance so a local function of the same name keeps the ordinary message.

- **Diagnostics are emitted in source order.** Upstream checks the module body and each suite body as separate passes, and ESLint sorts its messages afterwards. This port checks each statement where the traversal reaches it, producing the same order without a sorting step.

## Related Links

Related #935

## Checklist

- [x] Tests updated.
- [x] Documentation updated.


elecmonkey (commented):
Reviewed against `@rstest/core@0.11.11`, `@vitest/eslint-plugin@1.6.27`, and `eslint-plugin-jest@29.16.0`. The focused Go suites pass, but the inline findings below reproduce correctness gaps against Rstest’s actual collection semantics.

internal/plugins/rstest/rules/require_hook/require_hook.go
```diff
@@ -0,0 +1,277 @@
...
+	if call == nil || call.Kind != ast.KindCallExpression {
+		return false
+	}
+	// A `describe` nested in another call is reported as that outer statement
+	// instead, so its body is left alone.
+	if hasCallExpressionParent(call) {
+		return false
+	}
+	args := call.AsCallExpression().Arguments
+	if args == nil || len(args.Nodes) < 2 || args.Nodes[1] != callback {
```
elecmonkey: [P2] Please use the existing overload-aware callback resolver here. Rstest 0.11.11 supports `describe('suite', { timeout: 100 }, () => { setup(); })`, and the callback executes during collection, but this identity check only accepts an unwrapped second argument, so PR head reports nothing. It also misses parenthesized or TypeScript-wrapped inline callbacks such as `describe('suite', (() => { setup(); }))`. `RstestCallAnalysis.callbackInfo` already resolves the second/third-argument overload and strips transparent wrappers, so this rule should reuse that logic.

internal/plugins/rstest/rules/require_hook/require_hook.go
```diff
@@ -0,0 +1,277 @@
...
+	node *ast.Node,
+	ctx rule.RuleContext,
+	analysis *rstestUtils.RstestCallAnalysis,
+) bool {
+	if analysis.ParseFnCall(node) != nil {
+		return true
+	}
+	// A plugin-managed member is matched by the receiver as written, because
+	// the build rewrites it that way even in a file that declares its own `rs`.
+	if rstestUtils.ParseRstestPluginManagedCall(node) != nil {
```
elecmonkey, line 197: [P2] Respect `managed.API.ResolvesReceiver` before exempting this call. For `importActual` and `requireActual`, Rstest deliberately resolves the receiver, so a local binding such as `const rs = { importActual() { return setup() } }; rs.importActual()` invokes user code during collection. PR head reports nothing because every syntactically plugin-managed call is accepted here. The shared metadata and `ReceiverIsLocallyDeclared` helper already encode this distinction.

internal/plugins/rstest/rules/require_hook/require_hook.go
```diff
@@ -0,0 +1,277 @@
...
+			ctx,
+			analysis,
+			allowedFunctionCalls,
+		)
+	case ast.KindCallExpression:
+		call := node.AsCallExpression()
+		// A dynamic import is an ImportExpression upstream rather than a
+		// CallExpression, and an optional call is wrapped in a ChainExpression,
+		// so neither reaches the CallExpression branch there.
+		if call.Expression.Kind == ast.KindImportKeyword ||
```
elecmonkey, line 229: [P2] These syntax-based exclusions leave real Rstest collection work unreported. With `@rstest/core@0.11.11`, `setup?.()`, `await setup()`, `void setup()`, and `export let value = setup()` all execute while the file is collected, but PR head reports none of them. The exact upstream linters also miss these forms because of their ESTree visitor shapes, but inheriting that limitation conflicts with this rule’s Rstest purpose and with the documentation’s unconditional description of calls and initialized mutable declarations.

internal/plugins/rstest/rules/require_hook/require_hook.go
```diff
@@ -0,0 +1,277 @@
...
+			call.QuestionDotToken != nil ||
+			ast.IsOptionalChain(node) {
+			return false
+		}
+		if isRstestFnCall(node, ctx, analysis) {
+			return false
+		}
+		return !slices.Contains(allowedFunctionCalls, testFramework.CalleeChainName(node))
```
elecmonkey: [P3] Guard the allowlist lookup when `CalleeChainName` returns an empty string. With `{ allowedFunctionCalls: [''] }`, both `(condition ? setup : teardown)()` and `helper[key.value]()` are silently exempt because their callee cannot be named. Both reference linters report these inputs. An empty name should remain reportable, or the schema should reject empty entries.

internal/plugins/rstest/rules/require_hook/require_hook.go
```diff
@@ -0,0 +1,277 @@
...
+	if callback == nil || !testFramework.IsFunction(callback) {
+		return false
+	}
+	call := callback.Parent
+	if call == nil || call.Kind != ast.KindCallExpression {
+		return false
+	}
+	// A `describe` nested in another call is reported as that outer statement
+	// instead, so its body is left alone.
+	if hasCallExpressionParent(call) {
```
elecmonkey, line 334: [P3] This suppression depends on the outer call being reported, but `allowedFunctionCalls` can exempt that outer call. For `register(describe('suite', () => { setup(); }))` with `allowedFunctionCalls: ['register']`, PR head reports neither `register(...)` nor `setup()`, even though the suite callback executes during Rstest collection. Please continue inspecting the recognized `describe` callback when the enclosing call produces no diagnostic.

internal/plugins/rstest/rules/require_hook/require_hook.go
```diff
@@ -0,0 +1,277 @@
...
+		// by scanning each block, so diagnostics come out in source order even
+		// when a suite body sits between two module-level statements.
+		check := func(statement *ast.Node) {
+			if !runsDuringCollection(statement, analysis) {
+				return
+			}
+			if !shouldBeInHook(statement, ctx, analysis, opts.AllowedFunctionCalls) {
+				return
+			}
+			ctx.ReportNode(statement, buildUseHookMessage())
```
elecmonkey: [P3] `onTestFinished()` and `onTestFailed()` need a different diagnostic direction. Rstest 0.11.11 rejects either call during collection with “can only be called inside a test,” while this message tells users to move it into a lifecycle hook. The rule intentionally reports these APIs, so it should tell users to place them inside a test callback rather than suggesting an invalid repair.

internal/plugins/rstest/rules/require_hook/require_hook.go
elecmonkey: Partly addressed in 9ea3c2bbe. `runsDuringCollection` now unwraps symmetrically: `outermostTransparentWrapper` walks up from the callback through `ParenthesizedExpression`, `AsExpression`, `SatisfiesExpression`, `NonNullExpression` and `TypeAssertionExpression` to reach the call, and the argument side compares through `SkipAssertionsAndParens`. `describe('suite', (() => { setup(); }))` and the `as` / `!` wrapped forms are reported now, with regression cases for all four.

The `(description, options, fn)` overload is deliberately out of scope here. `eslint-plugin-jest` does not descend into it either, so this widens the ported behaviour rather than closing a parity gap, and it needs an exported registration-callback accessor — `callbackInfo` is unexported and `TestCallback` is gated on `ParseTestCall`. Left to separate work.

internal/plugins/rstest/rules/require_hook/require_hook.go
elecmonkey: Acknowledged. `importActual` and `requireActual` are the only entries carrying `ResolvesReceiver` in `plugin_managed_api.go`, and `isRstestFnCall` currently accepts every syntactically plugin-managed call. Not changing it in this PR: the reachable shape needs a file-local `const rs = { importActual() { … } }`, and honouring the flag means threading `ReceiverIsLocallyDeclared` through the exemption, which deserves its own change and its own regression cases.

internal/plugins/rstest/rules/require_hook/require_hook.go
elecmonkey: Acknowledged, and the runtime claim is correct — all four execute during collection. Keeping the ported behaviour in this PR: `eslint-plugin-jest` reports none of them, so widening here changes what the rule reports rather than fixing a porting defect, and that decision should be reviewable on its own.

internal/plugins/rstest/rules/require_hook/require_hook.go
elecmonkey: Fixed in 9ea3c2bbe. `shouldBeInHook` now returns early when `CalleeChainName` yields an empty string, so an unnameable callee is never matched against the allowlist. `(condition ? setup : teardown)()` and `helper[key.value]()` with `allowedFunctionCalls: [""]` are reported again, both covered by regression cases.

internal/plugins/rstest/rules/require_hook/require_hook.go
elecmonkey: Acknowledged. `eslint-plugin-jest` behaves the same way — a `describe` nested in another call is not descended into, regardless of `allowedFunctionCalls` — so this is a behaviour extension rather than a porting defect. It also touches the argument-position logic that the callback-resolver change above would replace, so the two are better designed together in a follow-up.

internal/plugins/rstest/rules/require_hook/require_hook.go
elecmonkey: Fixed in 9ea3c2bbe. Added a `useTest` message id — `onTestFinished() can only be called inside a test`, matching the runtime error in `packages/core/src/runtime/runner/runner.ts` — with the API name in the diagnostic data. `executionTimeAPIName` resolves the API through Rstest provenance, so globals, renamed imports, namespace imports and `import.meta.rstest` all get the new message, while a file-local function of the same name keeps `useHook`. Rule documentation and the integration test are updated.

internal/plugins/rstest/rules/require_hook/require_hook.go
fansenze: **Non-blocking follow-up: support `describe(name, options, callback)`.**

Thanks for clarifying the scope. Keeping this port aligned with Jest's upstream behavior is reasonable. The parenthesized and TypeScript-wrapped callback cases are fixed in `9ea3c2b`, and I'm happy for this PR to merge with the options overload left for a separate change.

For that follow-up, this supported Rstest form currently produces no diagnostic:

```ts
describe('suite', { timeout: 5000 }, () => {
  setup();
});
```

The same gap applies to `describe.skip` and `describe.each`. A follow-up can recognize the inline callback in either supported argument position and add regression tests. It would also be useful to document the current limitation.

fansenze (approved):

PR #2230, opened 2026-09-20
feat(rstest): add prefer-to-have-been-called-times rule

elecmonkey:
## Summary

Add `rstest/prefer-to-have-been-called-times` to report `toHaveLength()` assertions on a mock's `mock.calls` array and rewrite them to `toHaveBeenCalledTimes()`. It is registered in the Rstest plugin without being added to the `recommended` preset, matching upstream.

The table of Chai matchers that replace the assertion subject and the predicate for a numeric literal written in source move to `internal/utils/test_framework`, behind `IsSubjectMutatingChaiMatcher` and `StaticNumericLiteral`, shared with `rstest/prefer-to-have-length` so the two rules cannot drift apart.

## Credits

Port from [`eslint-plugin-jest`](https://github.com/jest-community/eslint-plugin-jest)'s `prefer-to-have-been-called-times`.

- **Identifier bracket keys are not treated as member names.** `expect(handler[mock].calls)` reads a variable rather than the `mock` property, so reporting and rewriting it would be a false positive.
- **Dynamic matcher names are not reported.** `expect(handler.mock.calls)[matcher](2)` names its matcher through a variable, which the assertion parser does not resolve to `toHaveLength`.
- **Parenthesized links inside the subject are rewritten correctly.** Upstream deletes one accessor chain as a single span and turns `expect((handler.mock).calls).toHaveLength(2)` into unparseable code.
- **Optional chaining on the matcher is preserved.** Upstream rewrites the whole matcher member, which drops the `?.` in `expect(handler.mock.calls)?.toHaveLength(2)`.
- **Computed accessor delimiters are preserved.** Upstream rewrites `expect(handler.mock.calls)["toHaveLength"](2)` to a dotted accessor; this port changes only the literal value.
- **`expect.soft` assertions are checked.** Soft assertions record the same call count, so they are treated like ordinary assertions.
- **Matchers after a subject-changing Chai matcher are not reported.** `expect(handler.mock.calls).property(0).toHaveLength(2)` measures the arguments of one recorded call rather than the call count, so the chain ends at `property`.
- **`expect.poll` assertions are not checked.** `expect.poll` takes a callback, so the only shape this rule could match there is an argument that throws before any matcher runs.
- **Only a count written as a number is fixed.** `toHaveLength` compares loosely and `toHaveBeenCalledTimes` strictly, so rewriting `toHaveLength('1')` fails an assertion that passed and rewriting `not.toHaveLength('1')` passes one that failed.
- **Arguments that can reset the mock are reported without a fix.** `expect()` captures the calls array while `toHaveBeenCalledTimes` reads `mock.calls` when the matcher runs, so `expect(handler.mock.calls, (handler.mockClear(), 'called once'))` would change result after the rewrite.
- **A `super` receiver is reported without a fix.** A bare `super` is not a value, so `expect(super.mock.calls)` cannot become `expect(super)` without producing code that does not parse.
- **Assertions whose result is reused are reported without a fix.** Rstest matchers return a Chai assertion that carries the rewritten subject into every later matcher, so only standalone assertion statements are fixed.

## Known gaps

- **`expect.extend` overrides are not consulted.** A file-local override of `toHaveLength` or `toHaveBeenCalledTimes` changes what either matcher means, and no matcher-rewrite rule in the plugin checks `IsExpectMatcherOverridden` today; making this one the exception would be inconsistent, so it is left for a separate change covering the whole family.

## Related Links

Related #935

## Checklist


internal/plugins/rstest/rules/prefer_to_have_been_called_times/prefer_to_have_been_called_times.go
```diff
@@ -0,0 +1,142 @@
...
+				}
+				arguments := parsed.Head.AsCallExpression().Arguments
+				if arguments == nil || len(arguments.Nodes) == 0 {
+					return
+				}
+				mockCalls := parseMockCallsAccess(arguments.Nodes[0])
+				if mockCalls == nil {
+					return
+				}
+				for _, matcher := range parsed.Matchers {
```
elecmonkey, line 38: [P2] Stop after a matcher that changes the Chai assertion subject

`parsed.Matchers` may contain matchers after `.property(...)`, but Chai's `property` matcher changes the current assertion subject. For example, `expect(fn.mock.calls).property(0).toHaveLength(1)` checks the argument count of the first recorded call, not the mock's call count. This assertion passes under Rstest 0.11.11, while the rule reports the later `toHaveLength`. Please stop at subject-changing matchers, as `prefer-to-have-length` already does, so only matchers still applied to the original `mock.calls` value are reported.

internal/plugins/rstest/rules/prefer_to_have_been_called_times/prefer_to_have_been_called_times.go
```diff
@@ -0,0 +1,142 @@
...
+					if matcher.Kind != rstestUtils.RstestExpectMatcherCall || matcher.Name != "toHaveLength" {
```
elecmonkey, line 42: [P2] Respect `expect.extend` overrides before reporting or fixing

Rstest allows `expect.extend` to override either matcher involved in this rewrite. For example, a custom `toHaveLength` can make `expect(fn.mock.calls).toHaveLength(999)` pass even when the mock was never called, but the current autofix changes it to `expect(fn).toHaveBeenCalledTimes(999)`. Overriding `toHaveBeenCalledTimes` can similarly make the replacement behave differently. Both cases were reproduced with Rstest 0.11.11. Please consult `analysis.IsExpectMatcherOverridden` for both matcher names: an overridden source matcher should not be treated as the built-in call-count assertion, and an overridden replacement matcher must not be introduced by an autofix.

internal/plugins/rstest/rules/prefer_to_have_been_called_times/prefer_to_have_been_called_times.go
```diff
@@ -0,0 +1,142 @@
...
+						continue
+					}
+					ctx.ReportNodeWithDeferredFixes(matcher.Entry.Node, rule.RuleMessage{
+						Id: "preferMatcher", Description: "Prefer `toHaveBeenCalledTimes`",
+					}, func() []rule.RuleFix {
+						// Rewriting the factory argument changes the subject that the
+						// returned Chai assertion carries, so it must not be reused.
+						if parsed.Entry == rstestUtils.RstestExpectEntryPoll ||
```
elecmonkey: [P2] Do not apply this direct-subject rule to `expect.poll`

Rstest 0.11.11 requires the first `expect.poll` argument to be a callback. The tested form `expect.poll(fn.mock.calls)` passes an array and throws immediately when the runtime calls `.bind`; nevertheless, this rule reports it and recommends a call-count matcher. The valid form is `expect.poll(() => fn.mock.calls).toHaveLength(1)`, which this implementation does not recognize and which has different polling semantics. Please exclude `RstestExpectEntryPoll` from this rule and remove the current poll documentation/tests unless callback-return analysis and a semantics-preserving replacement are implemented separately.

internal/plugins/rstest/rules/prefer_to_have_been_called_times/prefer_to_have_been_called_times.go
```diff
@@ -0,0 +1,142 @@
...
+							len(parsed.Matchers) != 1 ||
+							parsed.Expression != testFramework.InvokedAccessorCall(&matcher.Entry) ||
+							!isDiscardedAssertion(parsed.Expression) {
+							return nil
+						}
+						nameRange, name, ok := testFramework.AccessorReplacement(ctx.SourceFile, matcher.Entry.Node, "toHaveBeenCalledTimes")
+						if !ok {
+							return nil
+						}
+						fixes := []rule.RuleFix{rule.RuleFixReplaceRange(nameRange, name)}
```
elecmonkey, line 64: [P2] Remove or reject explicit `expect<T>` type arguments when changing the subject

The autofix retains head-call type arguments even though it replaces the subject. `expect<unknown[]>(fn.mock.calls).toHaveLength(0)` type-checks, but the generated `expect<unknown[]>(fn).toHaveBeenCalledTimes(0)` fails with TS2345 because the mock function is not an `unknown[]`. Please remove uncommented head-call type arguments when applying this rewrite, as `rstest/prefer-to-have-length` does, and withhold the fix when that range contains comments.

elecmonkey (commented):
I found four correctness issues in the Rstest-specific behavior and autofix boundaries. Inline comments include minimal reproductions and suggested directions.

internal/plugins/rstest/rules/prefer_to_have_been_called_times/prefer_to_have_been_called_times.go
```diff
@@ -0,0 +1,164 @@
...
+					ctx.ReportNodeWithDeferredFixes(matcher.Entry.Node, rule.RuleMessage{
+						Id: "preferMatcher", Description: "Prefer `toHaveBeenCalledTimes`",
+					}, func() []rule.RuleFix {
+						// Rewriting the factory argument changes the subject that the
+						// returned Chai assertion carries, so it must not be reused.
+						if len(parsed.Matchers) != 1 ||
+							matcherCall == nil ||
+							parsed.Expression != matcherCall ||
+							!isDiscardedAssertion(parsed.Expression) {
+							return nil
```
fansenze, line 55: [P2] Withhold the autofix for counts that are not known to be numbers

In Rstest 0.11.11, `toHaveLength` delegates to Chai's loose comparison (`==`), while `toHaveBeenCalledTimes` uses strict comparison (`===`). For a mock called once, this JavaScript assertion passes:

```js
expect(fn.mock.calls).toHaveLength('1');
```

The autofix produces `expect(fn).toHaveBeenCalledTimes('1')`, which fails. With `.not`, the same rewrite turns a failing assertion into a passing one. Boolean counts also differ. Please verify that the expected count is a number before offering this fix; otherwise, retain the diagnostic without an autofix.

internal/plugins/rstest/rules/prefer_to_have_been_called_times/prefer_to_have_been_called_times.go
```diff
@@ -0,0 +1,164 @@
...
+						}
+						nameRange, name, ok := testFramework.AccessorReplacement(ctx.SourceFile, matcher.Entry.Node, "toHaveBeenCalledTimes")
+						if !ok {
+							return nil
+						}
+						fixes := []rule.RuleFix{rule.RuleFixReplaceRange(nameRange, name)}
+						for _, key := range mockCalls {
+							ranges, ok := testFramework.RemoveAccessorEntryRanges(ctx.SourceFile, ctx.Comments.All(), &testFramework.MemberEntry{Node: key})
+							if !ok {
```
fansenze, line 67: [P2] Preserve the captured calls array when arguments reset the mock

This assertion passes in Rstest 0.11.11 after one call to `fn`:

```js
fn();
expect(fn.mock.calls).toHaveLength((fn.mockClear(), 1));
```

The original captures the calls array before evaluating the matcher argument. `mockClear()` replaces that array, so the captured array still contains one call. After the fix, `expect(fn).toHaveBeenCalledTimes((fn.mockClear(), 1))` reads the new array and fails with a count of zero. Please withhold the fix when later `expect` arguments or matcher arguments can replace the mock state, and cover this evaluation-order difference with a regression test.

internal/plugins/rstest/rules/prefer_to_have_been_called_times/prefer_to_have_been_called_times.go
```diff
@@ -0,0 +1,164 @@
...
+
+// parseMockCallsAccess returns the `mock` and `calls` key nodes of a
+// `<value>.mock.calls` access, or nil when node is any other expression.
+func parseMockCallsAccess(node *ast.Node) []*ast.Node {
+	receiver, calls := unwrapStaticMember(node, "calls")
+	if calls == nil {
+		return nil
+	}
+	if _, mock := unwrapStaticMember(receiver, "mock"); mock != nil {
+		return []*ast.Node{mock, calls}
```
fansenze: [P2] Reject a bare super receiver before removing the accessors

When a base class exposes the mock context through a static `mock` getter, this is a valid assertion:

```js
class Derived extends Base {
  static verify() {
    expect(super.mock.calls).toHaveLength(1);
  }
}
```

The autofix removes `.mock.calls` and produces `expect(super).toHaveBeenCalledTimes(1)`. A bare `super` cannot be used as a value, so the generated code fails to parse with TS1034. Please inspect the receiver returned by the second `unwrapStaticMember` call and skip this case instead of generating an invalid fix.

fansenze (commented):
Three remaining autofix issues, reproduced on this revision with Rstest 0.11.11. The existing targeted Go tests and rule registration check pass; additional regression probes confirm the issues below.

internal/plugins/rstest/rules/prefer_to_have_been_called_times/prefer_to_have_been_called_times.go
elecmonkey: Fixed in a22ebf01a. The fix is now offered only when the expected count is a number written in source; a variable, string, boolean, missing count or extra matcher argument keeps the diagnostic and loses the fix. Confirmed both directions on Rstest 0.11.11, including `not.toHaveLength('1')` turning a failing assertion into a passing one. Regression cases added for 11 fixable and 11 non-fixable spellings.

internal/plugins/rstest/rules/prefer_to_have_been_called_times/prefer_to_have_been_called_times.go
elecmonkey: Fixed in a22ebf01a. The remaining `expect()` arguments must now be literals, so the `mockClear()` form keeps its diagnostic without a fix. The second `expect()` argument is affected the same way — `expect(fn.mock.calls, (fn.mockClear(), 'm'))` also passes before the rewrite and fails after — so both positions are covered, with regression tests.

internal/plugins/rstest/rules/prefer_to_have_been_called_times/prefer_to_have_been_called_times.go
elecmonkey: Fixed in a22ebf01a. A bare `super` receiver is now rejected before the fix is built, so the assertion is reported without one. Confirmed the previous output failed with TS1034, and added a regression case.

fansenze (approved):

PR #2237, opened 2026-09-21
chore: upgrade to Go 1.27 and golangci-lint 2.13.2

swwind:
## Summary

Supersedes #1830 with the Go 1.27 and golangci-lint update on current `main`.

- Update the root Go module, workspace, and workflow pins to Go 1.27.0; leave the shim and typescript-go modules on their existing directives.
- Update CI to golangci-lint v2.13.2 and adjust the lint rules for Go 1.27 Unicode categories.
- Remove the Unicode 17 compatibility package, use Go 1.27's Unicode tables, and retain coverage for JavaScript case behavior.
- Remove obsolete `t.Fatal` nil-guard returns and annotate deprecated compiler-option test inputs.
- Intentionally omit the gofmt migration and bulk Go formatting from #1830.

## Verification

- `go test -mod=readonly` passed for seven directly affected Unicode and rule packages.
- `golangci-lint config verify` passed.
- Full `golangci-lint run --timeout=15m ./cmd/... ./internal/...` passed with 0 issues after suppressing two deprecated test-option uses. Scoped branch-diff lint also passed.
- `pnpm run format:check` passed. The CI command `pnpm check-spell` passed (3,947 files, zero issues); an explicit-path check of the changed Go and Markdown files also passed.
- A compile-only check of the remaining changed test packages could not finish because `/tmp` hit its temporary-space quota.


elecmonkey:
Glad we can say goodbye to our hand-rolled Unicode 17 implementation

fansenze (approved):

PR #2238, opened 2026-09-21
feat(rstest): add prefer-expect-resolves rule

elecmonkey:
## Motivation

When a promise is awaited inside `expect()`, an unexpected rejection throws before the matcher runs. That makes the failure look like a plain thrown error instead of an assertion failure, and it loses the clearer `.resolves` failure message.

Using `await expect(promise).resolves.<matcher>()` keeps the promise state inside the assertion chain and mirrors the `.rejects` form used for rejected promises.

## Changes

Add `rstest/prefer-expect-resolves`.

The rule reports `await` used as the first argument to an Rstest `expect()` call and prefers moving the `await` outside the assertion with a `.resolves` modifier.

It recognizes Rstest globals, imports and aliases, namespace and CommonJS bindings, `rstack/test`, `import.meta.rstest`, Playwright imports, and test-context `expect`, including renamed destructuring and `test.for` context parameters.

The framework-neutral await detection and edit construction are shared with `jest/prefer-expect-resolves`, while each plugin keeps its own parser and edit policy. The Jest rule keeps its existing autofix behavior. The Rstest rule uses type information when available: known non-promise values and mixed promise/value unions are not reported, unknown promise values get a suggestion, and known promise values may be autofixed when the rewrite is safe.

Add upstream, Rstest-specific, source-only, JS integration, and runtime-sensitive coverage. Register the rule in the Rstest plugin without adding it to the `recommended` preset.

## Intentional differences

- Rstest skips `expect.soft()`, `expect.poll()`, `expect.element()`, browser locator/page matchers, Chai property assertions, chains with multiple assertions, and chains already using `resolves` or `rejects`.
- Rstest only autofixes known thenables with built-in value matchers and literal message / matcher arguments. Other reported cases are left without edits, or receive a suggestion when the promise type is unknown.
- Optional chains, explicit type arguments on `expect`, custom or locally overridden matchers, and assertion results used by another expression are reported without edits.
- Known non-promise values, `never`, and unions with non-promise alternatives are not reported because JavaScript `await` accepts those values but `.resolves` requires a thenable.
- Copying `import.meta.rstest.expect` to a variable is not recognized; direct `import.meta.rstest.expect(...)` and destructured bindings from `import.meta.rstest` are recognized.

## Credits

Adapted from [`eslint-plugin-jest@29.16.1`](https://github.com/jest-community/eslint-plugin-jest/blob/v29.16.1/src/rules/prefer-expect-resolves.ts).

## Related Links

Related #935.

## Checklist

- [x] Tests updated.
- [x] Documentation updated.


internal/utils/test_framework/rules/prefer_expect_resolves/prefer_expect_resolves.go
```diff
@@ -0,0 +1,214 @@
...
+	for _, part := range utils.UnionTypeParts(typ) {
+		if part.Flags()&checker.TypeFlagsNever != 0 {
+			return true, false
+		}
+		if part.Flags()&(checker.TypeFlagsAny|checker.TypeFlagsUnknown) != 0 {
+			return false, false
+		}
+	}
+	for _, part := range utils.UnionTypeParts(typ) {
+		if !utils.IsThenableType(ctx.TypeChecker, subject, part) {
```
fansenze, line 117: [P2] Avoid autofixing objects with an optional `then`

`IsThenableType` accepts a callable branch of `then` even when that property may be missing. This can turn a passing assertion into a runtime error. For example, in a JavaScript file with `allowJs: true`:

```js
/** @param {{ then?: (resolve: (value: number) => void) => void }} value */
async function check(value) {
  expect(await value).toBeDefined();
}
await check({});
```

The original assertion passes. The rule automatically rewrites it to `await expect(value).resolves.toBeDefined()`, which throws a `TypeError` because `{}` is not a promise. I reproduced both the autofix and the runtime failure with `@rstest/core@0.11.11`.

Please require `then` to be definitely present and callable before offering an autofix, and add a regression test for this case. The same issue applies to a required `then` property whose type includes `undefined`.

fansenze (approved):

PR #2239, opened 2026-09-21
perf(node): compact builtin API metadata

swwind:
## Summary

- Restore `generate.mjs` from the initial version of #2234 and keep the upstream metadata pinned to `eslint-plugin-n` v18.3.0.
- Encode built-in API paths as indexed trie nodes, with interned path segments and release-version strings. Preserve feature order and the three recursive aliases.
- Keep the existing reference tracking and diagnostics behavior; add no dependencies.

## Results

- Stripped CLI binary: 40,694,023 → 40,542,471 bytes (**148 KiB smaller**) using the same `go build -p 2 -ldflags='-s -w'` command for both revisions.
- Dogfood with only `node/no-unsupported-features/node-builtins` enabled across `packages/`, `scripts/`, and `website/`: 1,292 files and 86 byte-identical diagnostics. Default parallel CLI wall-time median was 274.3 → 272.2 ms (40 runs per revision); single-thread rule-time mean was 91.8 → 91.6 ms (30 runs per revision).
- Regeneration is deterministic, and all 3,271 generated API records match the prior table in content and order.

## Verification

- `go test ./internal/plugins/node/rules/no_unsupported_features/node_builtins`
- `CI=true pnpm --dir packages/rslint-test-tools exec rs test run tests/eslint-plugin-node/rules/no-unsupported-features/node-builtins.test.ts` (704 passed)
- `GOFLAGS=-p=2 pnpm --filter @rslint/core build:bin`
- `golangci-lint run --new-from-merge-base=origin/main ./internal/plugins/node/rules/no_unsupported_features/node_builtins` (0 issues)
- `pnpm run check-spell` on the three changed files, and `pnpm run format:check`


elecmonkey:
I agree with restore generate.mjs but seems that @fansenze delete it intentionally. Perhaps he could weigh in on this again. Anyway, do you think it's better split it into a seperate PR. This PR should focus on compact data

fansenze:
I think we can leave the generator out and document the generation conventions in the generated file. Agents already handle upstream syncs and can follow those conventions to regenerate the data, while maintaining a separate script adds overhead.

If we keep the script, it should live under `scripts/`.

elecmonkey:
> I think we can leave the generator out and document the generation conventions in the generated file. Agents already handle upstream syncs and can follow those conventions to regenerate the data, while maintaining a separate script adds overhead.
> 
> If we keep the script, it should live under `scripts/`.

@fansenze I had not carefully read through the PR code earlier and assumed by reflex that it was purely a generator. This script contains the trie indexed compaction codes introduced by this PR, so it should be committed together with the PR. It is the key the binary size is reduced by 156 KB.

fansenze:
@swwind @elecmonkey

The binary size reduction is a great improvement. My concern is keeping the JS generator in the Go rule directory. It loads third-party dependencies from `node_modules` through `load()`, and mixing JS tooling into the Go source tree makes the structure harder to follow.

I still prefer documenting the generation conventions and letting agents handle the compression and regeneration. I think this would be easier to maintain as upstream changes.

If we need to keep the script, it should live under `scripts/`.

swwind:
> If we need to keep the script, it should live under `scripts/`.

Indeed, I have moved the gen script into `scripts/`.

swwind:
> I still prefer documenting the generation conventions and letting agents handle the compression and regeneration. I think this would be easier to maintain as upstream changes.

The current apis_generate.go was not intended to be able for manual editing. Every string now deduplicated and referenced by array index now. That would be hard for human to read and understand it directly. So I think keep a generate script would be better.

fansenze (approved):

PR #2240, opened 2026-09-21
feat(jest,rstest): add prefer-mock-return-shorthand rule

elecmonkey:
## Motivation

A mock implementation that only returns a value says what the mock produces with a function wrapped around it.

## Changes

Add `rstest/prefer-mock-return-shorthand` and `jest/prefer-mock-return-shorthand`.

The rules report `mockImplementation()` and `mockImplementationOnce()` given a callback that does nothing but return a value, and rewrite them to `mockReturnValue()` and `mockReturnValueOnce()`. Dot, quoted and template accessors are recognized, as is optional chaining; a bracketed key that is a variable is not, since the call site does not name a method.

The whole rule is shared between both plugins. It matches on the method name alone and never resolves the receiver, so nothing about it is framework-specific and each plugin only supplies its rule name.

Register the rules without adding them to either `recommended` preset, because collapsing a callback into a value shared by every call is a style choice a project makes for itself.

## Credits

Port from [`eslint-plugin-jest@29.16.1`](https://github.com/jest-community/eslint-plugin-jest/blob/v29.16.1/src/rules/prefer-mock-return-shorthand.ts)'s and [`@vitest/eslint-plugin@1.6.27`](https://github.com/vitest-dev/eslint-plugin-vitest/blob/v1.6.27/src/rules/prefer-mock-return-shorthand.ts)'s `prefer-mock-return-shorthand`.

- **A mutable binding is looked for in the whole returned expression.** `@vitest/eslint-plugin` only inspects its outermost node, so `() => ({ value })` is rewritten and freezes whatever `value` held at configuration time, and both plugins miss the binding inside a template, a tagged template, a class expression, `as`, `satisfies` and `!`.

- **A write anywhere in the returned expression withholds the report.** Upstream guards only a top-level update expression, which lets `() => (n += 1)` through and turns a mock that counted 1, 2, 3 across three calls into one that returns 1 every time.

- **A callback returning `Promise.reject()` is not reported.** The shorthand builds the rejected promise when the mock is configured, so a mock that is never called leaves an unhandled rejection the runner reports as an error even though every test passed, and the safe rewrite is `mockRejectedValue()` rather than this rule's.

- **A generator callback is not reported.** Calling it produces an iterator, so rewriting `function* () { return 1; }` to `mockReturnValue(1)` changes what every caller of the mock receives.

- **A callback declaring type parameters is not reported.** The rewrite drops the `<T>` declaration and leaves `T` dangling in the argument.

- **A `function` callback reading `this`, `arguments` or `new.target` is not reported.** All three are bound by the call, and `new.target` lifted out of every enclosing function does not parse at all, so the rewrite either changes the value or produces invalid code.

- **A quoted or template accessor keeps its delimiters.** Upstream replaces the whole literal, turning `mock['mockImplementation'](() => 1)` into `mock[mockReturnValue](1)`, which reads an undeclared variable.

- **A comma expression keeps its parentheses.** Upstream emits `mockReturnValue(0, value)`, which passes two arguments and returns the wrong operand.

- **A comment inside the callback withholds the fix.** The comment is not part of the returned expression's text, so the rewrite would delete it; the diagnostic still stands.

## Related Links

Related #935.

## Checklist

- [x] Tests updated.
- [x] Documentation updated.


internal/utils/test_framework/rules/prefer_mock_return_shorthand/prefer_mock_return_shorthand.go
```diff
@@ -0,0 +1,444 @@
...
+
+					callback := ast.SkipParentheses(arguments[0])
+					if !isCollapsibleCallback(callback) {
+						return
+					}
+
+					expression := singleReturnExpression(callback)
+					if expression == nil || !isEvaluatedOnceSafe(ctx, expression) ||
+						buildsRejectedPromise(ctx, expression) ||
+						readsCallSiteBinding(callback, expression) {
```
fansenze, line 344: **[P2] Preserve bindings defined inside the callback**

The fix removes the callback's scope, but these checks allow references to bindings declared there:

```ts
mock.mockImplementation(function impl() { return impl; });
// Autofix:
mock.mockReturnValue(impl); // ReferenceError: impl is not defined
```

A callback containing `return getValue(); function getValue() { return 42; }` has the same problem: the hoisted local declaration is deleted. Please skip the rewrite when the returned expression references a binding that would disappear with the callback.

internal/utils/test_framework/rules/prefer_mock_return_shorthand/prefer_mock_return_shorthand.go
```diff
@@ -0,0 +1,444 @@
...
+// Both reference plugins rewrite this shape anyway, and one of them documents
+// `mockReturnValue(Promise.reject(...))` as an acceptable result. It is not
+// reported here at all, rather than reported without a fix,
+// because the advice itself is the problem: the safe rewrite of a rejecting mock
+// is `mockRejectedValue`, which builds the promise per call, and that belongs to
+// the promise-shorthand rule rather than this one.
+func buildsRejectedPromise(ctx rule.RuleContext, expression *ast.Node) bool {
+	call := ast.SkipParentheses(expression)
+	if call == nil || call.Kind != ast.KindCallExpression {
+		return false
```
fansenze: **[P2] Unwrap TypeScript expressions before checking `Promise.reject`**

`SkipParentheses` leaves `as` and non-null assertions intact, so this guard misses:

```ts
mock.mockImplementation(
  () => Promise.reject(new Error('nope')) as Promise<never>
);
// Autofix:
mock.mockReturnValue(Promise.reject(new Error('nope')) as Promise<never>);
```

The rejected promise is now created during setup. Even if the mock is never called, it causes an unhandled rejection. Please unwrap these expressions before checking the call and cover both `as` and `!` in regression tests.

internal/utils/test_framework/rules/prefer_mock_return_shorthand/prefer_mock_return_shorthand.go
```diff
@@ -0,0 +1,444 @@
...
+			return true
+		case ast.KindIdentifier:
+			return node.Text() == "arguments" && scope.IsReferenceIdentifier(node)
+		case ast.KindMetaProperty:
+			// `import.meta` is the other meta property, and it names the module
+			// rather than the call, so it travels with the expression.
+			return node.AsMetaProperty().KeywordToken == ast.KindNewKeyword
+		case ast.KindFunctionExpression, ast.KindFunctionDeclaration,
+			ast.KindMethodDeclaration, ast.KindGetAccessor, ast.KindSetAccessor:
+			return false
```
fansenze: **[P2] Check computed keys and class heritage for callback bindings**

Method bodies have their own `this`, but computed keys use the enclosing callback's binding:

```ts
mock.mockImplementation(function () {
  return { [this.key]() {} };
});
// Autofix:
mock.mockReturnValue({ [this.key]() {} });
```

The fixed code reads `this` at the configuration site instead of using the mock call's receiver, and can throw. Returning `class extends this.Base {}` has the same problem because classes are skipped below. Please inspect computed names and `extends` expressions instead of skipping the entire method/class node.

internal/utils/test_framework/rules/prefer_mock_return_shorthand/prefer_mock_return_shorthand.go
```diff
@@ -0,0 +1,444 @@
...
+		if ast.IsAssignmentOperator(node.AsBinaryExpression().OperatorToken.Kind) {
+			return true
+		}
+	case ast.KindDeleteExpression:
+		return true
+	}
+
+	if testFramework.IsFunction(node) || ast.IsClassLike(node) {
+		return false
+	}
```
fansenze, line 204: **[P2] Check writes in class initialization and computed method names**

Skipping the whole class or method also skips expressions that execute immediately:

```ts
const state = { count: 0 };
const mock = rs.fn().mockImplementation(
  () => class { static value = ++state.count; }
);
```

Two calls to `mock().value` return `1`, then `2`. After the autofix, both return `1`: the increment only runs during setup. Computed method keys such as `{ [++state.count]() {} }` bypass this guard too. Please inspect the parts evaluated immediately, including static initializers, static blocks, computed names, and heritage expressions.

fansenze (commented):
Found four autofix safety issues in the shared implementation, affecting both Jest and Rstest. The existing Go tests for both rules and the rule registration check pass; eight additional cases reproduce the unsafe rewrites described below.

internal/utils/test_framework/rules/prefer_mock_return_shorthand/prefer_mock_return_shorthand.go
elecmonkey: Fixed in c0b71fa: the fix is withheld when statements follow the `return` or a named function expression references its own name.

internal/utils/test_framework/rules/prefer_mock_return_shorthand/prefer_mock_return_shorthand.go
elecmonkey: Fixed in c0b71fa: the `Promise.reject` check now unwraps `as`, `!`, `satisfies` and `<T>`, on the call, callee and `Promise` object.

internal/utils/test_framework/rules/prefer_mock_return_shorthand/prefer_mock_return_shorthand.go
elecmonkey: Fixed in c0b71fa: computed names, decorators and `extends` of nested methods/classes are now checked for `this`/`arguments`/`new.target`.

internal/utils/test_framework/rules/prefer_mock_return_shorthand/prefer_mock_return_shorthand.go
elecmonkey: Fixed in c0b71fa: static initializers, static blocks, computed names and `extends` are now checked for writes.

fansenze (commented):
The eight original reproductions are fixed in c0b71fa. One remaining TypeScript case still produces an unsafe autofix; I reproduced it on both the previous and current implementations, so this is an uncovered case rather than a regression from the latest changes.

internal/utils/test_framework/rules/prefer_mock_return_shorthand/prefer_mock_return_shorthand.go
```diff
@@ -0,0 +1,551 @@
...
+	// Promise<never>` and `(Promise as any).reject(e)` build the same promise.
+	call := utils.SkipAssertionsAndParens(expression)
+	if call == nil || call.Kind != ast.KindCallExpression {
+		return false
+	}
+
+	access := utils.SkipAssertionsAndParens(call.AsCallExpression().Expression)
+	if access == nil || !ast.IsAccessExpression(access) {
+		return false
+	}
```
fansenze: **[P2] Unwrap generic instantiation before checking `Promise.reject`**

`SkipAssertionsAndParens` does not unwrap `ExpressionWithTypeArguments`, so this valid TypeScript expression still bypasses the guard:

```ts
mock.mockImplementation(() => (Promise.reject<never>)(new Error('nope')));
// Current autofix:
mock.mockReturnValue((Promise.reject<never>)(new Error('nope')));
```

The type argument is erased at runtime. The fixed code creates the rejected promise during setup, so even an unused mock now causes an unhandled rejection. Please also unwrap generic instantiation expressions before checking the callee and add a regression case for this form.

elecmonkey: Fixed in c177897: the `Promise.reject` check now uses `ast.SkipOuterExpressions(…, ast.OEKAll)`, which also unwraps instantiation expressions.

fansenze (approved):

PR #2241, opened 2026-09-21
feat: add Node.js recommended presets

fansenze:
## Motivation

The native `node/*` rules currently require manual configuration. Add the three recommended Node.js presets to `@rslint/core`, including the language settings needed for their rules to work in ES modules and CommonJS.

## Changes

- Export `nodePlugin.configs.recommended`, `nodePlugin.configs.recommendedModule`, and `nodePlugin.configs.recommendedScript`, using camelCase keys for the corresponding presets in [eslint-plugin-n@18.3.0](https://github.com/eslint-community/eslint-plugin-n/tree/v18.3.0/lib/configs).
- Enable all 13 available recommended rules under the native `node/*` prefix. Preserve upstream globals access levels, `sourceType`, and the syntax rule's `ignores` options. The unimplemented `process-exit-as-throw` rule is explicitly documented and omitted.
- Make `recommended` select the nearest valid working-directory package's type and apply `.cjs`/`.mjs` overrides, including dotfiles. Keep package lookup and globals loading lazy so importing the core root does not load environment assets.
- Add integration coverage that triggers every enabled native rule and checks package lookup, globals, scopes, and option merging. Test that built presets lazily load only their two bundled globals assets. Correct the existing imported-globals test so config-load failures cannot pass as an empty diagnostic result.
- Register the presets in the shared website preset table and import snippet. Add the preset API reference and source comments with links mapping `recommended`, `recommendedModule`, and `recommendedScript` to upstream `recommended`, `recommended-module`, and `recommended-script`. Fix three broken links in the Node publication-rule documentation exposed by the docs build.

The presets use core's bundled globals catalog: all 134 globals supplied by upstream retain their access levels, with seven additional globals from core's newer catalog. Auto selection uses `process.cwd()`, as upstream does, but runs when `recommended` is accessed instead of at import time. API `cwd` does not affect it; callers whose process directory differs from the lint project, or nested packages with different types, should use explicit presets. `sourceType` affects lint scopes, not TypeScript parsing.

## Usage

```ts
import { defineConfig, nodePlugin } from '@rslint/core';

export default defineConfig([
  nodePlugin.configs.recommended,
  // Use one of these instead to select a module system explicitly:
  // nodePlugin.configs.recommendedModule,
  // nodePlugin.configs.recommendedScript,
]);
```

## Validation

On macOS arm64 with Node.js 22.18.0:

- `pnpm --filter @rslint/core build:bin` — passed.
- `pnpm --filter @rslint/core build:js` — passed, including declarations.
- `CI=true pnpm --dir packages/rslint exec rs test run tests/define-config.test.ts tests/config-loader.test.ts tests/globals.test.ts` — 99 passed.
- `CI=true pnpm --dir packages/rslint-test-tools exec rs test run tests/cli/js-config/presets.test.ts tests/cli/js-config/language-options-globals.test.ts` — 50 passed.
- With `TMPDIR` beneath an ESM package, reproduced the missing-package fixture failure before the fix. `TMPDIR=<isolated-dir> CI=true pnpm --dir packages/rslint-test-tools exec rs test run tests/cli/js-config/presets.test.ts` — all 37 passed after adding the CommonJS fixture boundary.
- Packed core and the host native package with `pnpm pack`, installed the tarballs into an isolated temporary project without `globals`, and ran the packaged CLI: all three presets passed in both module and CommonJS packages across `.js`, `.mjs`, and `.cjs` (6 runs, 18 files). Verified root import loads no globals assets and preset normalization loads only `node.json` and `es2021.json`.
- Reproduced the API/process-directory mismatch against both rslint and the published upstream preset to confirm the auto-selection limitation described above.
- Compared the built presets against the published `eslint-plugin-n@18.3.0` package: all 13 rule configurations and 134 upstream globals match after prefix mapping; confirmed the upstream duplicate `Object.fromEntries` diagnostics from the syntax and builtins rules.
- `pnpm --filter @rslint/website build` — passed; verified the rendered preset names, API mapping links, and the repaired anchor.
- `pnpm run format:check`, `pnpm run check-spell` with all changed text paths explicitly supplied, and `git diff --check` — passed.

No Go implementation changed. Full-repository tests and other platforms were not run.


website/docs/en/api/presets/node-plugin.md
```diff
@@ -0,0 +1,23 @@
+# nodePlugin
+
+`nodePlugin` exposes Rslint's built-in implementation of supported rules from [`eslint-plugin-n` 18.3.0](https://github.com/eslint-community/eslint-plugin-n/tree/v18.3.0). Rules use Rslint's native `node/*` prefix instead of upstream's `n/*`.
```
chenjiahan: Should we pin the version to 18.3.0?

packages/rslint-test-tools/tests/cli/js-config/presets.test.ts
```diff
@@ -236,3 +243,311 @@ describe('defineConfig and config presets', () => {
...
+      child: 'null',
+      module: true,
+    },
+    {
+      name: 'array package',
+      parent: '{"type":"module"}',
+      child: '[]',
+      module: true,
+    },
+    { name: 'no package', parent: undefined, child: undefined, module: false },
```
elecmonkey: In `presets.test.ts`, the `no package` case lets `usesModules()` walk from the temp directory all the way up to the filesystem root. On a host where some ancestor of `os.tmpdir()` happens to declare `"type": "module"`, the case would fail for reasons that have nothing to do with the code under test. Would adding a sentinel `package.json` at the top of the fixture tree be worth it, to make the case hermetic?

packages/rslint/src/config/presets/index.ts
```diff
@@ -45,6 +50,28 @@ export const importPlugin = {
   configs: { recommended: importRecommended },
 };
 
+export const nodePlugin = {
+  configs: {
+    /**
+     * Maps to eslint-plugin-n's `recommended` preset.
+     * @see https://github.com/eslint-community/eslint-plugin-n/blob/v18.3.0/lib/configs/recommended.js
+     */
+    get recommended() {
```
elecmonkey, line 59: `configs.recommended` is a getter, so every read walks the filesystem again from `process.cwd()`. Would it be worth memoizing that locally?

website/docs/en/api/presets/node-plugin.md
fansenze: Changed the label to `18.x`, consistent with other presets. Kept the source links pinned to `v18.3.0`.

elecmonkey: The versions declarations for these plugins are a bit messy now. A lot of rules got pinned to different versions as ported them, while the docs just declare one version so docs don’t match what’s actually impled. This will need some cleanup later and we need to discuss Rslint's built-in plugins' versioning policy.

packages/rslint-test-tools/tests/cli/js-config/presets.test.ts
fansenze: Added a CommonJS package.json above the fixture project to stop lookup at the fixture boundary. Verified with an ESM ancestor above TMPDIR; all 37 preset tests pass.

packages/rslint/src/config/presets/index.ts
fansenze: Keeping this uncached so reloaded configs pick up changes to cwd or package.json. The getter runs during config evaluation, not per linted file.

chenjiahan (approved):
LGTM 👍

PR #2242, opened 2026-09-21
feat(unicorn): add prefer-type-error rule

HashemKhalifa:
## Motivation

Refs #1309.

Port the recommended `unicorn/prefer-type-error` rule from [eslint-plugin-unicorn v75.0.0](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v75.0.0/rules/prefer-type-error.js). Exact-name, all-state PR searches found no existing `prefer-type-error` PR immediately before publication.

## Changes

Enforce `TypeError` instead of the generic `Error` when an `if` branch consists only of a throw and its condition is exclusively a recognized type check. The port covers `typeof`, non-error `instanceof` checks, global `isNaN` / `isFinite`, and Unicorn's full `.isType()` identifier set.

The autofix replaces only the `Error` constructor identifier. Rslint adds conservative binding safety: locally shadowed `Error` is not reported; bare `isNaN` / `isFinite` count only when they resolve to environment globals; and a shadowed or disabled `TypeError` still reports but does not autofix.

Register the rule in the native Unicorn catalog, enable it at `error` severity in `packages/rslint/src/config/presets/unicorn.ts`, add documentation, and add the JavaScript upstream mirror. `presets.test.ts` is intentionally untouched.

## Test-first coverage

The first branch commit, `b3e394e`, contains the complete behavior suite with an intentionally inert rule. The invalid suite fails at that checkpoint, establishing the red baseline before implementation.

A deterministic upstream-test extractor accounted for:
- 48 upstream valid cases
- 50 upstream invalid cases, including all 37 generated type-check identifiers
- all 4 documentation invalid/valid example pairs

The Go suite therefore carries 52 upstream/documentation valid cases and 54 upstream/documentation invalid cases, plus Rslint-specific binding, AST-shape, disabled-global and edit-demand regressions.

After implementation, `go test -coverprofile=...` on the rule package reports **100.0% statement coverage**.

## Verification

- `go test ./internal/plugins/unicorn/rules/prefer_type_error` — passed.
- Rule-package `golangci-lint --new-from-merge-base=upstream/main` — 0 issues.
- `pnpm run format:check` — passed.
- Changed-file spelling — 0 issues.
- `git diff --check` — passed.
- A focused production native build using a temporary one-rule catalog completed successfully. The temporary catalog files were restored byte-for-byte before delivery.
- `CI=true pnpm --dir packages/rslint-test-tools exec rs test run tests/eslint-plugin-unicorn/rules/prefer-type-error.test.ts` — 2/2 groups passed through the production service wrapper.

A normal full-catalog native build was attempted, but Go build temp/cache growth exhausted the available local disk. The focused build above proves this rule through the real production binary/service path without changing production build scripts. Full-catalog registration/build remains for CI.

The catalog-only Go test and full-catalog lint were not claimed as passing locally; both spend most of their time loading/relinking the entire rule catalog and were stopped after the focused rule and production-path proofs were already green.


internal/plugins/unicorn/rules/prefer_type_error/prefer_type_error.go
```diff
@@ -0,0 +1,248 @@
...
+			isTypecheckingExpression(ctx, binary.Right, callExpression)
+
+	case ast.KindInstanceOfKeyword:
+		return !isErrorConstructor(binary.Right)
+	}
+
+	if isExistenceCheck(binary) {
+		return false
+	}
+	return isTypecheckingExpression(ctx, binary.Left, callExpression) ||
```
elecmonkey, line 215: [P2] Exclude sequence and assignment expressions before recursing

TypeScript represents sequence and assignment expressions as `BinaryExpression` nodes, so this fallback currently accepts a condition whenever either operand contains a recognized type check. For example, Rslint reports and autofixes both `if (Array.isArray(value), flag) { throw new Error(); }` and `if (flag = Array.isArray(value)) { throw new Error(); }`, while `eslint-plugin-unicorn@75.0.0` reports neither. The first condition is ultimately controlled by `flag`, and assignment expressions are likewise not exclusively type checks. Compound assignments such as `+=`, `&&=`, `||=`, and `??=` are affected too. Please restrict this fallback to operators corresponding to ESTree `BinaryExpression`, or explicitly reject comma and all assignment operators.

internal/plugins/unicorn/rules/prefer_type_error/prefer_type_error.go
```diff
@@ -0,0 +1,248 @@
...
+}
+
+func isTypecheckingExpression(ctx rule.RuleContext, node *ast.Node, callExpression *ast.CallExpression) bool {
+	node = ast.SkipParentheses(node)
+
+	switch node.Kind {
+	case ast.KindIdentifier:
+		return isTypecheckingIdentifier(ctx, node, callExpression, false)
+
+	case ast.KindPropertyAccessExpression, ast.KindElementAccessExpression:
```
elecmonkey, line 139: [P2] Preserve the upstream optional-chain boundary

Optional member access and optional calls are currently treated as ordinary type-checking calls. For example, Rslint reports and autofixes both `utils?.isArray(value)` and `utils.isArray?.(value)`, whereas `eslint-plugin-unicorn@75.0.0` reports neither because these expressions are exposed through an unsupported `ChainExpression`. That behavior is appropriate here: the condition also checks whether the receiver or function exists, so it is not exclusively a type check. Please reject optional chains before handling member and call expressions.

elecmonkey (commented):
Really appreciate your work! I found two false-positive cases that need addressing before merge.

HashemKhalifa:
> Really appreciate your work! I found two false-positive cases that need addressing before merge.

Thanks for your kind words! I’m really happy to be able to contribute. I genuinely love Rspack and its ecosystem.

Will do a follow up fix

internal/plugins/unicorn/rules/prefer_type_error/prefer_type_error.go
HashemKhalifa: Fixed in `a0da49dd4a9917e6c8a6d1114d78107683189f3d`. The tsgo-only `BinaryExpression` fallback now rejects comma expressions and every assignment operator before recursing, so sequence, plain assignment, arithmetic compound assignment, and `&&=`/`||=`/`??=` no longer trigger. Added exact/regression coverage; the focused rule suite passes and scoped golangci-lint reports 0 issues.

internal/plugins/unicorn/rules/prefer_type_error/prefer_type_error.go
HashemKhalifa: Fixed in `a0da49dd4a9917e6c8a6d1114d78107683189f3d`. `isTypecheckingExpression` now reuses `unicornutil.HasOptionalChainElement` to preserve the upstream ChainExpression boundary, so optional member access, optional call, and nested optional receiver paths are excluded. Added explicit regressions; the focused rule suite passes and scoped golangci-lint reports 0 issues.

elecmonkey (approved):

PR #2243, opened 2026-09-21
refactor(jest): share call analysis across jest recommended rules

elecmonkey:
## Motivation

Bring Rstest's mature per-file shared call-analysis architecture to Jest so recommended rules can reuse parsing and callback analysis.

## Changes

- add a file-scoped `JestCallAnalysis`, organized consistently with `RstestCallAnalysis`
- cache successful parses, non-Jest calls, and invalid expect-chain reasons
- lazily index calls and named functions for callback analysis
- share callback discovery and named callback resolution across rules
- migrate all 15 recommended rules that consume Jest call analysis
- remove the callback collector's private parse cache
- remove the one-entry cache from `no-standalone-expect`
- remove duplicate expect-chain parsing from `valid-expect`
- retain the pure parser entry point for rules that have not been migrated yet

## Performance

Benchmarks used `eslint-plugin-jest@29.16.1`, the same 19 recommended rules and options, single-threaded execution, and the median of seven runs.

| Fixture | Files | Upstream | main baseline | this PR | Improvement |
| --- | ---: | ---: | ---: | ---: | ---: |
| [Stencil](https://github.com/stenciljs/core/tree/1c81131f659dd0beaf9bdadd59d5f2a5f895dd38) | 258 | 152.1 ms | 138.3 ms | 52.2 ms | 62.3% |
| [Immutable.js](https://github.com/immutable-js/immutable-js/tree/686b8baf1dc825419f5ded2968b48e3432644aff) | 56 | 73.3 ms | 45.1 ms | 14.6 ms | 67.6% |
| [Carbon](https://github.com/Sage/carbon/tree/e10647e972d55c696aebc94168c4be8676b57f6e) | 279 | 250.5 ms | 185.8 ms | 60.2 ms | 67.6% |
| [AntV L7](https://github.com/antvis/L7/tree/6e3dbfe392b687231447088333da9b0efbfcae53) | 125 | 74.0 ms | 55.2 ms | 18.8 ms | 65.9% |
| [SubQuery](https://github.com/subquery/subql/tree/51e2a2c985c3c869510a36e8c37a3b3557af4a3c) | 117 | 54.1 ms | 44.8 ms | 17.3 ms | 61.4% |

Across the five fixtures, this PR is 2.91x to 5.02x faster than upstream and reduces Rslint's recommended-rule time by 61.4% to 67.6%. Diagnostics are unchanged from the main baseline on all five fixtures.

## Related Links

Related #476.

## Checklist


internal/plugins/jest/utils/parse_jest_fn.go
```diff
@@ -58,25 +63,25 @@ func ParseJestFnCall(node *ast.Node, ctx rule.RuleContext) *ParsedJestFnCall {
 
 	callExpr := node.AsCallExpression()
 	if isEachFactoryCall(callExpr, members) || isInvalidTaggedTemplateCall(callExpr, members) || isInnerExpectCall(node, localName, members, ctx.Settings) {
-		return nil
+		return jestCallParseResult{}
```
fansenze: **[P2] Preserve diagnostics for rejected `expect` chains**

This early return loses the error reason that `valid-expect` previously computed after `ParseJestFnCall` returned `nil`. These invalid expressions now produce no diagnostics:

```js
expect(1).not.not.each(); // Previously: modifierUnknown
expect`value`();         // Previously: matcherNotFound
```

I verified both cases against the base commit (`c4d047da`) and this PR (`855318d0`): the base reports the expected errors, while the PR reports zero errors. Please preserve the expect-chain error analysis when `isEachFactoryCall` or `isInvalidTaggedTemplateCall` rejects a call, and add regression coverage.

fansenze (approved):

PR #2244, opened 2026-09-21
feat(unicorn): add no-async-promise-finally rule

HashemKhalifa:
## Motivation

Refs #1309.

Port the recommended `unicorn/no-async-promise-finally` rule from eslint-plugin-unicorn v75.0.0.

## Changes

Disallow async non-generator functions as `Promise#finally()` callbacks. The port recognizes direct async callbacks, local `const` aliases, and local async function declarations, including TypeScript assertion wrappers and statically computed `finally` property names.

When type information is available, Rslint suppresses the diagnostic only when the receiver is proven non-Promise-like. Unknown/source-only receivers, `any`/`unknown`, Promise-like receivers, error types, and mixed Promise/non-Promise unions retain Unicorn's best-effort behavior.

The rule is registered in the native Unicorn catalog and enabled in `packages/rslint/src/config/presets/unicorn.ts`. `presets.test.ts` is intentionally untouched.

## Test-first coverage

The first branch commit, `444feed`, contains the complete behavior suite with an intentionally inert rule; the invalid cases fail at that checkpoint before implementation.

Coverage accounting for Unicorn v75.0.0:
- 21 upstream valid cases
- 17 upstream invalid cases
- 4 documentation code blocks, yielding 3 additional valid and 3 additional invalid examples

Rslint extras cover TypeScript transparent wrappers, PromiseLike/any/error/mixed receiver types, dynamic computed properties, and an explicit `project:false` path for both `const` aliases and async function declarations.

The final focused rule package reports **100.0% statement coverage**.

## Verification

- `go test -coverprofile=... ./internal/plugins/unicorn/rules/no_async_promise_finally` — passed, 100.0% statements
- scoped `golangci-lint --new-from-merge-base=upstream/main` — 0 issues
- `pnpm run format:check` — passed
- changed-file spelling — passed
- `git diff --check` — passed
- focused production native build with a temporary one-rule catalog — passed; catalog files restored byte-for-byte
- `CI=true pnpm --dir packages/rslint-test-tools exec rs test run tests/eslint-plugin-unicorn/rules/no-async-promise-finally.test.ts` — 2/2 groups passed


internal/plugins/unicorn/rules/no_async_promise_finally/no_async_promise_finally.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	symbol := ctx.Refs.Resolve(node)
+	if symbol == nil || len(symbol.Declarations) != 1 {
+		return false
+	}
+
+	declaration := symbol.Declarations[0]
+	return declaration != nil &&
+		declaration.Kind == ast.KindFunctionDeclaration &&
```
fansenze, line 92: **[P2] Also recognize references to named async function expressions**

This condition skips an async function passed through its own name:

```js
const run = async function cleanup() {
  promise.finally(cleanup);
};
```

Here, `cleanup` resolves to a `FunctionExpression`, so the `FunctionDeclaration` check returns false. [Unicorn v75.0.0](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v75.0.0/rules/no-async-promise-finally.js#L33-L44) reports this callback because its `FunctionName` definition also covers named function expressions. I reproduced **0 diagnostics instead of 1** on this PR in both the type-aware and `project: false` paths.

Please also accept `ast.KindFunctionExpression` here, keeping the async/non-generator check, and add regression coverage for both paths.

HashemKhalifa: Fixed in `afbf6b34276043325cc3e08a6931e7a68e60ac6a`. The local-reference resolver now accepts exactly the two upstream `FunctionName` declaration kinds—`FunctionDeclaration` and named `FunctionExpression`—and reuses the existing async/non-generator classifier. Added the reviewer-exact named async function-expression regression in both the type-aware RuleTester path and `project:false`. Focused tests pass, rule-package coverage remains 100.0%, scoped golangci-lint reports 0 issues, and `git diff --check` is clean.

internal/plugins/unicorn/rules/no_async_promise_finally/no_async_promise_finally.go
```diff
@@ -0,0 +1,142 @@
...
+					arguments[0].Kind == ast.KindSpreadElement {
+					return
+				}
+
+				callback := arguments[0]
+				if !isAsyncFinallyCallback(ctx, callback) {
+					return
+				}
+
+				ctx.ReportNode(callback, message)
```
elecmonkey: **[P2] Strip parentheses before reporting the callback**

Reporting the raw ts-go `callback` node makes source parentheses part of the diagnostic range, unlike Unicorn’s ESTree range. This also breaks line-based suppression:

```js
promise.finally(
  (
    // eslint-disable-next-line unicorn/no-async-promise-finally
    async () => {}
  )
);
```

With `eslint-plugin-unicorn@75.0.0`, the directive suppresses the diagnostic. At this revision, Rslint still reports one diagnostic spanning lines 2–5 because `DisableManager` checks the outer parenthesis line. The same range mismatch occurs for ordinary parenthesized direct callbacks and `const` references.

Please report `ast.SkipParentheses(callback)` instead. This preserves authored TypeScript assertion wrappers while matching the ESTree-visible callback range. I verified that change against the upstream suite and a 56-case differential corpus.

PR #2246, opened 2026-09-23
feat(jest,rstest): add prefer-expect-assertions rule

elecmonkey:
## Motivation

Assertions inside callbacks, loops and promise handlers may never run, and a test whose assertions are all skipped still passes. Declaring the expected count with `expect.assertions()` or `expect.hasAssertions()` makes the runner fail such a test.

## Changes

- Add `rstest/prefer-expect-assertions` and `jest/prefer-expect-assertions`, sharing one rule engine with a per-plugin call parser.
- Require every test callback to start with `expect.assertions(n)` or `expect.hasAssertions()`, validate their arguments, and suggest inserting or correcting the call.
- Count `beforeEach` and `afterEach` declarations for Jest, which checks the count after `afterEach`, but only `beforeEach` for Rstest, which checks it before `afterEach` runs.
- Accept `disallowHasAssertions` in the Rstest rule.
- Leave both rules out of the `recommended` presets.

## Credits

Port from [`eslint-plugin-jest@29.16.0`](https://github.com/jest-community/eslint-plugin-jest/blob/v29.16.0/src/rules/prefer-expect-assertions.ts)'s `prefer-expect-assertions`.

- **A hook covers every test of its `describe` block wherever it is declared.** Upstream tracks hooks in source order and reports a test written above the covering `beforeEach`, although the runner collects all hooks of the block first.

- **Only a declaration made directly by the hook callback counts.** Upstream also counts one made in a timer the hook starts, which runs after the test has been checked.

- **`expect(value).assertions(1)` and a declaration inside an `if` branch do not count.** The first is a matcher, and the second only declares a count on some paths.

- **A satisfied test no longer exempts the next one.** With an `only*` option, upstream leaves its flag set after a skipped test and accepts the next test the option selects.

## Related Links

Related #935.

## Checklist

- [x] Tests updated.
- [x] Documentation updated.


internal/utils/test_framework/rules/prefer_expect_assertions/prefer_expect_assertions.go
```diff
@@ -0,0 +1,727 @@
...
+					case RegistrationHook:
+						if !slices.Contains(config.CoveringHooks, registration.HookName) || registration.Callback == nil {
+							return
+						}
+						suite := currentSuite()
+						// Type assertions do not change what the hook runs.
+						callback := utils.SkipAssertionsAndParens(registration.Callback)
+						switch {
+						case ast.IsFunctionExpressionOrArrowFunction(callback):
+							hookCallbacks[callback] = append(hookCallbacks[callback], suite)
```
fansenze: **[P2] Keep hooks registered by helpers scoped to their actual suite**

This hook only applies to suite A, but both rules also consider the test in B covered:

```js
function installAssertions() {
  beforeEach(() => expect.hasAssertions());
}
describe('A', () => {
  installAssertions();
  test('a', () => {});
});
describe('B', () => {
  test('b', () => {}); // Should report missing assertion declaration.
});
```

`currentSuite()` is `nil` while traversing the helper's definition, so the hook eventually sets `fileCovered` and suppresses diagnostics for unrelated suites. Please resolve the hook's registration scope, or conservatively avoid granting coverage when the registration is inside an ordinary helper.

internal/utils/test_framework/rules/prefer_expect_assertions/prefer_expect_assertions.go
```diff
@@ -0,0 +1,727 @@
...
+	case ast.KindIfStatement:
+		return child == parent.AsIfStatement().Expression
+	case ast.KindSwitchStatement:
+		return child == parent.AsSwitchStatement().Expression
+	case ast.KindExpressionStatement, ast.KindVariableStatement, ast.KindReturnStatement, ast.KindThrowStatement:
+		return true
+	}
+	// Any other statement, such as a loop, may run its expressions zero times
+	// or only after its body.
+	return !ast.IsStatement(parent)
```
fansenze: **[P2] Account for optional calls and destructuring defaults**

Both rules accept this test even though the assertion declaration never executes:

```js
const maybe = undefined;
beforeEach(() => maybe?.(expect.hasAssertions()));
test('t', () => {});
```

The same happens with `const { x = expect.hasAssertions() } = { x: 1 }` inside the hook. The fallback here treats every non-statement parent as evaluating its child unconditionally, but optional-call arguments and destructuring defaults can be skipped. Please handle these conditional positions before counting the declaration as satisfying a test or hook.

internal/utils/test_framework/rules/prefer_expect_assertions/prefer_expect_assertions.go
```diff
@@ -0,0 +1,727 @@
...
+// insertionPoint is just inside the opening brace, or after the directive
+// prologue so an inserted statement cannot turn `'use strict'` into an
+// ordinary expression.
+func insertionPoint(sourceFile *ast.SourceFile, body *ast.Node) int {
+	insertAt := utils.TrimNodeTextRange(sourceFile, body).Pos() + 1
+	for _, statement := range body.AsBlock().Statements.Nodes {
+		if !ast.IsPrologueDirective(statement) {
+			break
+		}
+		insertAt = statement.End()
```
fansenze: **[P2] Preserve a separator after directives without semicolons**

For this valid input:

```js
test('t', () => {
  'use strict'
  run();
});
```

Applying either suggestion inserts the new call directly after the string literal, producing `'use strict'expect.hasAssertions();` (or the `assertions()` variant). This is a `SyntaxError`, reproduced for both rules. Please ensure a statement separator is present when the preceding directive has no explicit semicolon.

fansenze (commented):
Reviewed `4b082159`. The existing targeted Go rule tests pass, but additional regression cases reproduce the four issues below. JS integration tests were not run locally.

internal/plugins/rstest/rules/prefer_expect_assertions/prefer_expect_assertions.go
```diff
@@ -0,0 +1,441 @@
...
+				if rstest.RequireBindingImportedName(element) == "expect" && result.expect == "" && !isRebound(ctx, element) {
+					result.expect = local.Text()
+				}
+			}
+		}
+		return
+	}
+	if name.Kind == ast.KindIdentifier && name.Text() == "expect" {
+		result.declaresExpect = true
+	}
```
fansenze: **[P2] Check destructured bindings before suggesting global expect**

With Rstest globals enabled:

```js
const { expect } = require('chai');
test('t', () => {});
```

The rule reports the missing declaration, but also suggests inserting `expect.hasAssertions()` and `expect.assertions()`. Both suggestions resolve to Chai's local binding instead of Rstest's `expect`. This check only handles a plain identifier, so a top-level destructured `expect` is missed. Please inspect the names bound by destructuring as well and withhold the global suggestions when `expect` is shadowed.

internal/utils/test_framework/rules/prefer_expect_assertions/prefer_expect_assertions.go
elecmonkey: Fixed in 246ea072. Hooks are now attributed to the suites that actually run them: a function passed to `describe` is that suite's body, and a helper, IIFE or callback belongs to the suites its call sites run in, so `installAssertions()` in `A` no longer covers `B`. When not every call site can be found (the helper is exported, stored elsewhere, or never referenced), the hook still counts as possibly covering any test, which can only miss a report.

internal/utils/test_framework/rules/prefer_expect_assertions/prefer_expect_assertions.go
elecmonkey: Fixed in b5c25028. Arguments after `?.` in an optional chain and defaults in binding patterns, destructuring assignments and shorthand properties are no longer treated as always evaluated, for both test declarations and hooks.

internal/utils/test_framework/rules/prefer_expect_assertions/prefer_expect_assertions.go
elecmonkey: Fixed in 47ed59d7. When the last directive has no semicolon, the suggestion now inserts one before the declaration, producing `'use strict';expect.hasAssertions();`.

internal/plugins/rstest/rules/prefer_expect_assertions/prefer_expect_assertions.go
elecmonkey: Fixed in abdacd6c. Module-level destructuring is now checked as well, including nested and array patterns and `const { assert: expect } = require('@rstest/core')`, so the global `expect` is not suggested when the file binds its own.

internal/utils/test_framework/rules/prefer_expect_assertions/prefer_expect_assertions.go
```diff
@@ -0,0 +1,752 @@
...
+					return false
+				}
+				isCovered := func(test *testEntry) bool {
+					if covered[unknownSuite] {
+						return true
+					}
+					for _, suite := range resolver.suitesOf(enclosingFunction(test.call)) {
+						if isSuiteCovered(suite) {
+							return true
+						}
```
fansenze: **[P2] Require coverage in every suite that registers a shared test**

`commonTests()` creates one test in A and another in B:

```js
function commonTests() {
  test('case', () => {});
}

describe('A', () => {
  beforeEach(() => expect.hasAssertions()); // A requires assertions.
  commonTests();
});

describe('B', () => {
  commonTests(); // B has no assertion requirement.
});
```

The loop finds that A is covered and immediately returns `true`, so neither rule reports the unprotected test in B. Please require coverage in **all known registration scopes**, not just one. Reusing a named `describe` callback under covered and uncovered parents has the same problem. Both scenarios reported a missing declaration before this change.

internal/utils/test_framework/rules/prefer_expect_assertions/suites.go
```diff
@@ -0,0 +1,197 @@
...
+		// A recursive helper adds no suite beyond those its other callers reach.
+		return nil
+	}
+	r.inProgress[scope] = true
+	var suites []*ast.Node
+	uses := r.functionUses(scope)
+	if len(uses) == 0 {
+		// Never referenced in this file: dead code, or reached some way the
+		// resolver does not see.
+		suites = []*ast.Node{unknownSuite}
```
fansenze: **[P2] Do not let an unused setup helper exempt unrelated tests**

No hook is registered in this example because `setup()` is never called:

```js
describe('A', () => {
  function setup() {
    beforeEach(() => expect.hasAssertions());
  }
  // setup() is never called.
});

describe('B', () => {
  test('b', () => {}); // Should report a missing assertion declaration.
});
```

Both rules now emit no diagnostic for B. This branch maps the unused helper to `unknownSuite`, and `covered[unknownSuite]` then exempts every test in the file. Please keep unused local helpers from contributing coverage, and avoid treating uncertain hook ownership as proof that unrelated suites are covered. A helper called through a local alias in A can cause the same suppression in B; both cases reported B before this change.

fansenze (commented):
The original 10 reproduction cases now pass. I found two additional regressions in the suite resolver, affecting both Jest and Rstest. Four new scenarios per framework pass with the shared engine from `abdacd6c` and fail with `246ea072`.

internal/utils/test_framework/rules/prefer_expect_assertions/prefer_expect_assertions.go
elecmonkey: Fixed in 7984b74d: a test is now covered only when every suite that registers it is covered.

internal/utils/test_framework/rules/prefer_expect_assertions/suites.go
elecmonkey: Fixed in 7984b74d: helpers that are never called in this file no longer contribute coverage, and calls through `const` aliases are followed.

fansenze (approved):

Issue #181, opened 2025-08-06
[Feature]: Port eslint-plugin-import


colinaaa:
### What problem does this feature solve?

Port https://www.npmjs.com/package/eslint-plugin-import

### What does the proposed API of configuration look like?

```jsonc
// rslint.json
[
  {
    "language": "javascript",
    "rules": {
      "import/no-commonjs": "error",
      "import/no-cycle": "error",
      "import/first": "error",
      "import/newline-after-import": "error",
      // ...
    },
    "plugins": ["import"]
  }
]
```

PR #2260, opened 2026-09-23
feat(import): add consistent-type-specifier-style

fansenze:
## Motivation

Add `import/consistent-type-specifier-style` so users can enforce inline or declaration-level markers for named type imports. Part of https://github.com/web-infra-dev/rslint/issues/181.

## Changes

- Port the rule from eslint-plugin-import v2.32.0, including both options, diagnostics, autofixes, schema, registration, and documentation. Omitted options follow upstream's actual `prefer-top-level` behavior.
- Reuse existing tsgo and rslint utilities. Preserve quoted import names and withhold autofixes that would lose import attributes, invalidate `resolution-mode` imports, or delete comments.
- Mirror upstream tests and documentation examples, with explicit skips for unsupported Flow forms. Add Go regressions for import attributes, comments, keyword identifiers, Unicode, BOM/CRLF, ambient modules, quoted names, and edit demand.

Validation passed: the rule's Go tests, the selected JS integration file with `CI=true`, native JS interface regression checks for diagnostics and fix safety, scoped Go lint, formatting, and spelling. User-visible differences from upstream are documented.


internal/plugins/import/rules/consistent_type_specifier_style/consistent_type_specifier_style.go
```diff
@@ -0,0 +1,154 @@
...
+		// Preserve quoted export names rather than changing the imported binding.
+		if imported.Kind == ast.KindStringLiteral {
+			name = utils.TrimmedNodeText(sourceFile, imported)
+		}
+		if imported.Kind == ast.KindStringLiteral || name != specifier.Name().Text() {
+			name += " as " + specifier.Name().Text()
+		}
+		names = append(names, name)
+	}
+	newImport := "import type {" + strings.Join(names, ", ") + "} from " + utils.TrimmedNodeText(sourceFile, declaration.ModuleSpecifier) + ";"
```
elecmonkey: [P2] Preserve import attributes when creating the type-only import

This reconstruction keeps only the module specifier, so `import { type Style } from "plain.style" with { type: "css" };` is fixed to `import type {Style} from "plain.style";`. The dropped attributes can affect module loading or resolution; the same loss occurs when splitting a mixed import. TypeScript supports attributes on type-only imports, so the generated import needs to preserve the original attribute clause, or the rule should withhold the fix when it cannot do so safely.

internal/plugins/import/rules/consistent_type_specifier_style/consistent_type_specifier_style.go
```diff
@@ -0,0 +1,154 @@
...
+					ctx.ReportNodeWithDeferredFixes(specifier, preferTopLevelMessage(), buildFixes)
+				}
+			},
+		}
+	},
+}
+
+func inlineFixes(sourceFile *ast.SourceFile, clause *ast.ImportClause, specifiers []*ast.Node) []rule.RuleFix {
+	fixes := make([]rule.RuleFix, 0, len(specifiers)+2)
+	if token, ok := utils.TokenAtOrAfter(sourceFile, clause.Pos()); ok && token.Kind == ast.KindTypeKeyword {
```
elecmonkey: [P2] Do not convert `resolution-mode` imports to inline form

For `import type { Foo } from "pkg" with { "resolution-mode": "import" };` under `prefer-inline`, these edits produce `import { type Foo } from "pkg" with { "resolution-mode": "import" };`. TypeScript rejects that result with TS1454 because `resolution-mode` is only allowed on a top-level type-only import. Please avoid offering this autofix for that attribute, or otherwise preserve a valid equivalent.

internal/plugins/import/rules/consistent_type_specifier_style/consistent_type_specifier_style.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+func topLevelFixes(sourceFile *ast.SourceFile, node *ast.Node, typeSpecifiers []*ast.Node, lastValueSpecifier *ast.Node) []rule.RuleFix {
+	declaration := node.AsImportDeclaration()
+	clause := declaration.ImportClause.AsImportClause()
+	names := make([]string, 0, len(typeSpecifiers))
+	for _, node := range typeSpecifiers {
+		specifier := node.AsImportSpecifier()
+		imported := node.PropertyNameOrName()
+		name := imported.Text()
```
elecmonkey: [P2] Avoid deleting comments while rebuilding type specifiers

Reconstructing each specifier from its parsed names silently drops comments. For example, `import { type /*A*/ Foo /*B*/ as /*C*/ Bar } from "m";` becomes `import type {Foo as Bar} from "m";`. eslint-plugin-import v2.32.0 has the same behavior, but deleting user comments is still unsafe for an autofix. Please preserve the original specifier text while relocating the `type` marker, or withhold the fix when comments cannot be retained.

internal/plugins/import/rules/consistent_type_specifier_style/consistent_type_specifier_style.go
fansenze: Fixed in e5b36657: imports with attributes keep their diagnostics but no longer receive unsafe autofixes. Added regressions.

internal/plugins/import/rules/consistent_type_specifier_style/consistent_type_specifier_style.go
fansenze: Fixed in e5b36657: `resolution-mode` imports no longer receive an inline autofix; both modes are covered.

internal/plugins/import/rules/consistent_type_specifier_style/consistent_type_specifier_style.go
fansenze: Fixed in e5b36657: fixes that would delete comments are withheld; safe fixes remain available. Added regressions.

elecmonkey (approved):

PR #2261, opened 2026-09-23
feat: port import/export rule

fansenze:
## Motivation

Add `import/export` from eslint-plugin-import v2.32.0 to detect duplicate named/default exports and conflicts introduced by star re-exports.

Refs https://github.com/web-infra-dev/rslint/issues/181.

## Changes

- Check exports within each namespace/module scope, preserving TypeScript overload and declaration-merging behavior, and report star re-exports without named exports.
- Reuse the existing binding/range helpers and import export maps. Shared utilities expose known export names and namespace-assignment members; duplicate detection stays in the rule. Occurrence lists are allocated only for repeated names.
- Register the rule in the native catalog and import recommended preset, and add documentation, the upstream Go/JS cases, and Go regressions for TypeScript AST shapes, string names, re-export graphs, diagnostic ranges, and suppression directives.

## Validation

- Targeted Go tests for `export`, import utils, `default`, and `namespace`, plus the rule catalog check.
- Branch-filtered golangci-lint v2.13.2; rebuilt native binary; targeted JS integration tests for `export`, `default`, `namespace`, and the import recommended preset.
- Differential checks against the pinned upstream on 64 real TypeScript files plus a positive re-export probe, and 12 additional AST/directive cases. Messages and complete diagnostic ranges match.
- Temporary Go benchmarks for small/large modules, duplicates, namespace merging, and star re-exports; benchmark files are excluded from this PR.
- `pnpm run format:check`, explicit-path spell checking, and `git diff --check`.

Two malformed-dependency upstream cases remain documented skips because the test Program rejects their syntax before rule execution. The rule documentation explains the user-visible diagnostic difference for invalid dependency syntax.


internal/plugins/import/rules/export/export.go
```diff
@@ -0,0 +1,223 @@
...
+		if body == nil {
+			return
+		}
+		switch body.Kind {
+		case ast.KindModuleBlock:
+			for _, statement := range body.AsModuleBlock().Statements.Nodes {
+				c.collect(statement)
+			}
+		case ast.KindModuleDeclaration:
+			c.collect(body)
```
elecmonkey: [P2] Keep nested dotted namespaces out of the root export scope

Recursing into the inner `ModuleDeclaration` makes its name fall back to the root scope because `add` only recognizes a direct `ModuleBlock` parent. As a result, `export namespace A.B {} export const B = 1;` reports both `B` occurrences as duplicate exports, while eslint-plugin-import v2.32.0 reports nothing. The same false positive occurs for `namespace A.B {}` and for the innermost name in longer dotted namespaces. Please preserve the enclosing dotted namespace as the scope when visiting this nested declaration.

elecmonkey (commented):
I found one additional wiring issue that cannot be attached inline because the affected file is outside this PR's diff:

[P1] Enable `import/export` in the public import recommended preset

The rule is registered in the native rule catalog, but `packages/rslint/src/config/presets/import.ts` still leaves `'import/export': 'error'` commented out as “not implemented.” I verified this against the built package: `importPlugin.configs.recommended.rules` does not contain `import/export`, and linting `export const duplicated = 1; export { duplicated };` with that preset produces no diagnostic. eslint-plugin-import v2.32.0 enables this rule in its recommended preset, so users of Rslint's corresponding preset will not receive the newly implemented check.

internal/plugins/import/rules/export/export.go
fansenze: Fixed in 4c006021. Dotted namespace segments no longer enter the root export scope; added regression tests.

fansenze:
Enabled `import/export` in the recommended preset and added an integration test in 4c006021.

elecmonkey (approved):

PR #2263, opened 2026-09-23
feat(import): add no-unresolved rule

fansenze:
## Motivation

Add `import/no-unresolved` so rslint can report imports whose target module cannot be found, with behavior based on `eslint-plugin-import` v2.32.0.

Refs https://github.com/web-infra-dev/rslint/issues/181.

## Changes

- Check ES imports, re-exports and dynamic imports, plus opt-in CommonJS and AMD dependencies. Support ignore patterns, explicit type-only exclusions, core-module exemptions and path casing checks.
- Adapt import resolver settings to the existing shared runtime resolver and TypeScript project resolution. Keep filesystem lookup in those existing services and import-specific configuration in the import plugin.
- Cover parenthesized and optional calls, JavaScript JSDoc casts versus TypeScript assertions, escaped source ranges, repeated resolver configurations and symlinked importers. Reuse the rule context's Program cache for casing checks and initialize the resolver only when needed.
- Add upstream Go and JS cases, focused Go regressions and user-facing documentation. Webpack/third-party resolvers and Babel's experimental export-default-from syntax remain documented limitations.

Validation: affected Go packages and the targeted rule catalog check pass; Go lint reports no new issues. The JS rule suite passes 22 tests with 7 explained skips for Webpack/Babel cases. Differential checks against v2.32.0 match all diagnostic fields on 20 real repository files plus a positive control, and on 11 additional boundary cases. Temporary Go benchmarks, the JS test typecheck, spell check and format check also pass; benchmark source files are excluded from this PR.


elecmonkey:
Please enable it in recommend preset

fansenze:
@elecmonkey Done in b414fef5: enabled as `error` in `importPlugin.configs.recommended`, with an integration test.

elecmonkey (approved):

